PR #40, opened 2014-11-25
Set a default MEDIA_URL value.

amjith:
An empty MEDIA_URL causes an exception at server start.

ImproperlyConfigured at /
Empty static prefix not permitted
Request Method:	GET
Request URL:	http://localhost:8000/
Django Version:	1.7.1
Exception Type:	ImproperlyConfigured
Exception Value:	Empty static prefix not permitted


berggren:
Hi Amjith!

Sorry for the delay. I have been busy with a rather big re-factor of the project, and this will be fixed in that process. Thanks for contributing!

Issue #44, opened 2015-02-03
Move over js/html/css


berggren:
## Move Javascript and HTML to the refactor branch
* Clean up HTML
* Clean up CSS
* Move Javascript over (as is for now, refactor later)
* Move over third_party libs

PR #45, opened 2015-02-03
Move javascript html

berggren:
## Move over Javascript and HTML
Move over the Javascript controllers/directives and HTML templates to the refactor branch in order to get them merged.
#### Note
* Moving over in it's current state with the goal of refactor it in a later PR.
* We will move the API calls to services and add tests in the near future. 

Reviewer: @mbushkov
Related issue: #44


timesketch/ui/js/controllers.js
```diff
@@ -0,0 +1,176 @@
...
+
+timesketch.controller('ExploreCtrl', function($scope, $http) {
+        $scope.init = function(sketch_id, view_id, timelines) {
+            $scope.sketch_id = sketch_id;
+            $scope.view_id = view_id;
+            $scope.timelines = timelines;
+            $scope.star = false;
+            $scope.filter = {};
+            $scope.filter.indices = $scope.timelines.split(",");
+            $http.get("/api/v1/sketches/" + $scope.sketch_id + "/views/" + $scope.view_id + "/").success(function(data) {
```
mbushkov: I'd suggest implementing a service to work with Timesketch API. Then you can pas the service as a dependency with controller's arguments (i.e. something like function($scope, $http, timesketchApiService)). Having a dedicated service provides much better documentation and makes the code more testable. I.e. here it would look like:
timesketchService.getView($scope.sketch_id, $scope.view_id)).success(...)

Then in the tests you can just mock getView function and not care about the details on how the URL is built.

timesketch/ui/templates/home/home.html
```diff
@@ -0,0 +1,73 @@
...
+limitations under the License.
+-->
+
+{% extends "base.html" %}
+
+{% block sidenav %}
+<div id="navigation">
+  <div class="leftnav-title">
+    <form action="{{ url_for('home_views.home') }}" method="post">
+      {{ form.name }}
```
mbushkov: Mentioning this just in case. It's really confusing to mix server-side templating with client-side templating. IIRC this is a temporary thing and you're going to migrate all the server-side templates to be client-side.

timesketch/ui/js/directives.js
```diff
@@ -0,0 +1,161 @@
...
+      restrict: 'A',
+      scope: {
+        filter: '=',
+        search: '=',
+        meta: '='
+      },
+      link: function (scope, elem, attrs) {
+        scope.$watch("filter", function(value) {
+            var i = scope.filter.indices.indexOf(attrs.index);
+            if (i == -1) {
```
mbushkov: There's quite a loot of "manual" DOM manipulation here. Any reason not to do the same via traditional directive template?

mbushkov: You can have this directive as a div-container if you use ng-transclude.

timesketch/ui/js/directives.js
```diff
@@ -0,0 +1,161 @@
...
+      link: function (scope, elem, attrs) {
+          var timeline_names = scope.meta.timeline_names;
+          elem.addClass("label");
+          elem.css("color", "#999");
+          elem.append(timeline_names[scope.event._index])
+      }
+    }
+});
+
+directives.directive('indexChooser', function() {
```
mbushkov: I'd suggest having an app-specific prefix for all timesketch controllers/services/directives registered in global AngularJs namespace. Something like "ts" (i.e. tsIndexChooser).

timesketch/ui/js/directives.js
```diff
@@ -0,0 +1,161 @@
...
+                elem.find("input").prop("checked", true);
+
+            }
+            scope.search()
+        })
+      }
+    }
+});
+
+directives.directive('indexDisabler', function() {
```
mbushkov: Same here - why not use a directive template?

timesketch/ui/js/directives.js
```diff
@@ -0,0 +1,161 @@
...
+                e.find("input").prop("checked", false);
+            }
+            scope.filter.indices = [];
+            scope.search()
+        })
+      }
+    }
+});
+
+directives.directive('indexEnabler', function() {
```
mbushkov: And the same template question here.

timesketch/ui/templates/sketch/explore.html
```diff
@@ -0,0 +1,154 @@
...
+
+    <!-- Result card -->
+    <div class="row">
+      <div class="col-md-12">
+        <div class="card">
+          {% raw %}
+          <div ng-show="meta" style="padding:10px;">
+            {{ meta.es_total_count }} events ({{ meta.es_time/1000 }}s)
+          </div>
+          <div ng-repeat="event in events" ng-controller="EventCtrl">
```
mbushkov: Does it make sense to make this a directive with its own template?

timesketch/ui/js/directives.js
berggren: Done.

timesketch/ui/js/controllers.js
berggren: Agree, and as discussed this will be done in a near future PR, the great Angular refactor.

timesketch/ui/templates/home/home.html
berggren: Agree, and yes there is a WIP to move over to pure client side templates, but there are some things that needs to be done before this can happen. Moving this to a future PR.

timesketch/ui/js/directives.js
berggren: Agree. I have added TODOs for this and will file an issue to follow up with a bigger refactor in a future PR.

timesketch/ui/js/directives.js
berggren: Ack. See earlier comment. Will be done in another PR as a larger refactor.

timesketch/ui/js/directives.js
berggren: Ack. See earlier comment. Will be done in another PR as a larger refactor.

timesketch/ui/templates/sketch/explore.html
berggren: Absolutely. I have added a TODO to fix this. In this PR I want to move the stuff in and get comments and implement the changes in a future PR.

berggren:
As discussed off list, The current code will be merged and then later refactored. The following issue has been created to track this: https://github.com/google/timesketch/issues/46

mbushkov:
LGTM. Changes look good to me and refactoring after the merge is done sounds like a right way to go.

PR #47, opened 2015-02-25
Example config file and version change

berggren:
* Add example configuration
* Change version
* Some config checks
* Add missing dependencies

Reviewer: @onager


timesketch.conf
```diff
@@ -0,0 +1,29 @@
+# Timesketch configuration
+#
+# The default location for this configuration file in /etc/timesketch.conf
```
Onager: * is

timesketch.conf
```diff
@@ -0,0 +1,29 @@
...
+# If you put it somewhere else you can pass the path to tsctl.py
+# Example:
+#
+# $ tsctl.py -c /path/to/this/timesketch.conf runserver
+
+# Turn off debugging.
+# It is a security risk to have this enabled in production.
+DEBUG = False
+
+# This should be a unique random string. Don't share this with anyone.
```
Onager: What does this actually secure? Is this the seed for XSRF protection, or something else?

timesketch.conf
berggren: Done

timesketch.conf
berggren: Added more infor

Onager:
As discussed:
*  It'd be nice to generate the secret key on install
*  It'd be nice at some point to move the test dependencies somewhere nicer

These are for later though.

LGTM

Issue #48, opened 2015-02-26
Redirect after login


berggren:
Honor the ?next= argument when login.

PR #49, opened 2015-02-26
Redirect to next= if present

berggren:
The "next" argument is set when login_required is in place. This redirects the user back to the requested page after successful authn.
Issue: #48 

Reviewer: @onager


Onager:
LGTM

Issue #51, opened 2015-03-15
REMOTE_USER authentication


berggren:
To support common SSO solutions Timesketch should be able to setup a session when the REMOTE_USER environment variable is set.

berggren:
This is fixed and merged.

PR #56, opened 2015-03-24
55 remote user authn

berggren:
This implements support for REMOTE_USER authentication. This is used in many Single Sign On (SSO) solutions.
Ref: Issue #51 

Reviewer: @onager

### DIFF
--- a/timesketch/ui/views/user_test.py
+++ b/timesketch/ui/views/user_test.py
@@ -13,10 +13,10 @@
 # limitations under the License.
 """Tests for the user views."""
 
+from flask_login import current_app
+
 from timesketch.lib.definitions import HTTP_STATUS_CODE_REDIRECT
 from timesketch.lib.testlib import BaseTest
-
-from flask_login import current_app
 
 
 class UserViewTest(BaseTest):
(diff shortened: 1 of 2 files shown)


timesketch/ui/views/user_test.py
```diff
@@ -13,19 +13,33 @@
...
+        self.login()
+        response = self.client.get('/login/')
+        self.assertEquals(response.status_code, HTTP_STATUS_CODE_REDIRECT)
+
+    def test_login_view_sso_authenticated(self):
+        """Test the login view handler with an SSO authenticated session."""
+        current_app.config['SSO_ENABLED'] = True
+        response = self.client.get(
+            '/login/', environ_base={'REMOTE_USER': 'test1'})
+        self.assertEquals(response.status_code, HTTP_STATUS_CODE_REDIRECT)
```
Onager: Check also that the correct user was set.

Onager:
Just add one asset, then LGTM.

timesketch/ui/views/user_test.py
berggren: Done.

PR #57, opened 2015-03-26
Support virtualenv in the wsgi application

berggren:
Add sane defaults.

* Commented out support for virtualenv.
* Add Apache configuration example.
* Apache mod_wsgi expects the wsgi application be named "application", not "app".

Reviewer: @onager


Onager:
LGTM

PR #58, opened 2015-03-27
Unicode everywhere

berggren:
## Unicode everywhere
In order to be as consistent as possible we need to clean up the codebase. This PR moves all strings to unicode and all SQLAlchemy database column type from String/Text -> Unicode/UnicodeText.

Issue: #59 
Reviewer: @onager


Onager:
Yay for unicode!

LGTM

PR #72, opened 2015-04-20
Dump dict to string json

berggren:
When saving a view, the API handler doesn't convert the filter dict to json string before saving to database, which raises an exception.

Reviewer: @darrenbilby


darrenbilby:
LGTM

Issue #61, opened 2015-03-30
Refactor JS - Services


berggren:
Move REST API calls to services.

berggren:
Done.

Issue #46, opened 2015-02-18
Refactor the JS frontend code


berggren:
## Refactor the JS frontend code
The current structure of the frontend JS code is hard to test and is somewhat confusing, e.g mixing server-side templates with client-side templates. In order to make this better we need to refactor the code.

#### Sub tasks
- [x] #61 - Move REST API calls to services
- [x] #66 - Move controllers to directives
- [x] #67 - Utilize templateURL for directives

See this old PR and comments as starting points: https://github.com/google/timesketch/pull/45

berggren:
This is done.

PR #75, opened 2015-04-21
61 js refactor services

berggren:
# First round of JS refactoring - Services

This is the first round on refactoring the Angular JS code. See issue #46 for tracking of the different steps.

Move all code that talks to the Timesketch REST API to services. Test will be added at a later time but with this refactor it will at least be possible. :)

Ref issue: #61 
Reviewer: @mbushkov


timesketch/ui/static/js/controllers.js
```diff
@@ -104,21 +98,17 @@ timesketch.controller('TsExploreCtrl', function($scope, $http) {
...
-                var id = data.objects[0].id;
-                var view_url = '/sketch/' + $scope.sketch_id + '/explore/view/' + id + '/';
-                window.location.href = view_url;
+            timesketchApi.saveView(
+                $scope.sketch_id, $scope.view_name, $scope.query, $scope.filter)
+                .success(function(data) {
+                    var id = data.objects[0].id;
+                    var view_url = '/sketch/' + $scope.sketch_id + '/explore/view/' + id + '/';
+                    window.location.href = view_url;
             });
```
mbushkov: It probably makes sense to handle requests failures in some generic way. For example, timesketchApi may expose a variable with the last error and then you can a piece of UI watching on it. Then, when error happens - no matter when or during which API call - you'll see it in the UI. This way you'll also won't have to manually handle failures everywhere.

mbushkov:
LGTM! Looks really good, and sorry for taking so long to review.

timesketch/ui/static/js/controllers.js
berggren: Great idea! I added a interceptor for response failures that broadcasts http-error. This is picked up by a directive that shows a butterbar telling something went wrong.

I will revisit this in the future to make this more elegant.

timesketch/ui/static/js/controllers.js
```diff
@@ -104,21 +102,17 @@ timesketch.controller('TsExploreCtrl', function($scope, $http) {
...
-            $http.post('/api/v1/sketches/' + $scope.sketch_id + '/views/', params).success(function(data) {
-                var id = data.objects[0].id;
-                var view_url = '/sketch/' + $scope.sketch_id + '/explore/view/' + id + '/';
-                window.location.href = view_url;
+            timesketchApi.saveView(
+                $scope.sketch_id, $scope.view_name, $scope.query, $scope.filter)
+                .success(function(data) {
+                    var id = data.objects[0].id;
+                    var view_url = '/sketch/' + $scope.sketch_id + '/explore/view/' + id + '/';
+                    window.location.href = view_url;
```
mbushkov: Nit: you can use $window Angular alias: https://docs.angularjs.org/api/ng/service/$window
Then you'll have to inject it via dependency injection mechanism and will be able to mock it out in tests.

Issue #79, opened 2015-05-22
Create timeline from json file


berggren:
Add functionality to create a timeline from a json file with events.

PR #80, opened 2015-05-22
Add JSON import suppport

berggren:
### Import timeline from JSON document
This PR adds support for ingesting timeline events from JSON. It is not the most optimal solution due to the way the built in JSON module works (reading the whole document into memory) but for reasonable sized files it should be ok. To address this I plan to implement a CSV importer later.

When this is submitted we will be able to do:
$ tsctl json2ts --name foo --file /path/to/file.json

Ref issue: #79 
Reviewer: @liamjm


liamjm:
LGTM

PR #82, opened 2015-05-29
81 fix query bug with pgsql

### LINKED ISSUE #81: Unable to get event and add event annotation when using PGSQL
When using postgresql some queries fails with DataError.

This is due to the way SQLalchemy constructs the query, and the way we pass in the searchindex name to the filter.

berggren:
Fixes #81 

Reviewer: @onager


Onager:
LGTM

Issue #97, opened 2015-07-06
Unable to get xsrf token in Safari


berggren:
This doesnt work in safari:
var csrftoken = document.getElementsByTagName('meta')['csrf-token'].getAttribute('content');

But this does:
var csrftoken = document.getElementsByTagName('meta')[0]['content'];

PR #98, opened 2015-07-06
Fix Safari bug with xsrf token

berggren:
Fixes a bug where Safari couldnt get the XSRF token from the meta tag.
Ref issue: #97 

Reviewer: @onager


Onager:
LGTM

Issue #104, opened 2015-07-15
UI polish and more intuitive settings


berggren:
To polish the release we need to clean up the settings file and make some minor UI polish changes.

Add UPLOAD_ENABLED to settings and make the checks look for this.

PR #105, opened 2015-07-15
104 release polish

berggren:
Prep for the release.
Small UI polish and more intuitive settings.

Reviewer: @onager
Ref issue: #104

### DIFF
--- a/timesketch.conf
+++ b/timesketch.conf
@@ -6,30 +6,37 @@
 #
 # $ tsctl -c /path/to/this/timesketch.conf runserver
 
-# Turn off debugging.
-# It is a security risk to have this enabled in production.
+# Show debug information.
+#
+# Note: It is a security risk to have this enabled in production.
 DEBUG = False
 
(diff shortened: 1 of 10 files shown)


timesketch.conf
```diff
@@ -6,30 +6,37 @@
...
 DEBUG = False
 
+# Key for signing cookies and for CSRF protection.
+#
 # This should be a unique random string. Don't share this with anyone.
-# The key is used for signing cookies and for CSRF protection.
 # To generate a key, you can for example use openssl:
 # $ openssl rand -base64 32
-# If the key is not set the server will not start.
 SECRET_KEY = u''
```
Onager: You don't want to tell people this is required?

Onager:
LGTM - one minor comment

timesketch.conf
berggren: Yep, this is redundant because the server will complain in any case if the key is missing and refuse to start.

PR #106, opened 2015-07-23
2015.7.post1 Release with fixes for pypi

berggren:
Release fixes because I don't know pypi good enough. Makes sure that the configuration file is included in the packaged file and stop depending on 0.9.8 of SQLAlchemy.


Onager:
LGTM

Issue #111, opened 2015-09-20
Plaso file loading doesn't set data_location for frontend


Onager:
No data_location being set breaks event log message string expansion, among other things.

Onager:
At present the logic for setting a data location default in Plaso looks to be tied to CLI processing: https://github.com/log2timeline/plaso/blob/master/plaso/cli/tools.py#L101 so that will likely need to be moved too.

PR #115, opened 2015-09-23
111 data location for frontend

berggren:
This PR enables message string support in the Plaso frontend task.
Resolves: #111 
Reviewer: @onager

### DIFF
--- a/timesketch.conf
+++ b/timesketch.conf
@@ -62,6 +62,6 @@ UPLOAD_FOLDER = u'/tmp'
 CELERY_BROKER_URL='redis://ip:port',
 CELERY_RESULT_BACKEND='redis://ip:port'
 
-# Path to Plaso WinEvtx message string database.
+# Path to plaso data directory.
 # If not set, defaults to system prefix + share/plaso
-#WINEVT_DB = u'/path/to/dir/with/plaso/data/files'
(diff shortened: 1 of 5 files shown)


timesketch/lib/tasks.py
```diff
@@ -32,11 +61,15 @@ def run_plaso(source_file_path, timeline_name, index_name):
     Returns:
         Dictionary with count of processed events.
     """
+    # Try to read the winevt-rc database path from the config file.
+    data_location = get_data_location(
```
Onager: This is a bit ugly - maybe move the config retrieval to get_data_location

timesketch.conf
```diff
@@ -61,3 +61,7 @@ UPLOAD_FOLDER = u'/tmp'
 # server is running.
 CELERY_BROKER_URL='redis://ip:port',
 CELERY_RESULT_BACKEND='redis://ip:port'
+
+# Path to Plaso WinEvtx message string database.
```
Onager: I think its better to refer to this as just the "plaso data directory" - there could be more things added in the future.

timesketch/lib/tasks.py
berggren: This was mostly for testing convenience. I have refactored it now.

timesketch.conf
berggren: Done.

Onager:
LGTM

Issue #85, opened 2015-05-29
Ability to sort dates in explore view


hiddenillusion:
Currently, once a batch of events is returned they are listed by oldest date/time first by default. While this is most likely the way most users will view their timelines, it'd be convenient for one to have the ability to easily change the sort on the same page the events are displayed.

Two use cases this would be beneficial in are
1. when there is an excessive amount of events to page through as this would be easier than continuously applying date/time filters or (if implemented) clicking through search pages
2. a situation where you want to backtrace a sequence of events

berggren:
Yes, this would be very handy, and I have been thinking of implementing this. When the current JS refactoring is done this should be easier to add.

berggren:
This is implemented.

PR #119, opened 2015-09-30
Sort event list asc or desc

berggren:
Sort result event list in either ascending order or descending order. This PR adds this to the filter object so that we can render correctly and also make it possible to save it in the view.

Reviewer: @onager
Ref issue: #85


Onager:
LGTM

PR #131, opened 2015-10-13
Adds explore/<query>/ ability

Splinks:
Quickly wrote this together, figured that if someone else wanted something like this before the winter release they could do this aswell


berggren:
Thanks for doing this! I have another branch implementing this in the pipeline. I will align these two.

berggren:
This functionality was implemented in #143 

Thanks again!

Issue #132, opened 2015-10-14
Filter on data_type


berggren:
Aggregate number of event of data_type attribute and show in the filter UI with the ability to enable/disable certain data_types.

Issue #30, opened 2014-11-11
Heatmap visualization 


berggren:
## Experimental UI
Add events per day/hour heatmap to query results by using elasticsearch aggregation ability.

Something like this:
![screen shot 2014-11-11 at 09 09 07](https://cloud.githubusercontent.com/assets/316362/4990038/7db6d204-6982-11e4-9e92-c65ff0483428.png)

berggren:
This also need backend code for general aggregation support.

PR #133, opened 2015-10-14
Heatmap aggregation

berggren:
* Implements basic aggregation support in the datastore client.
* Adds heatmap aggregation, both backend and new UI.
* Use data_type aggregation for filtering.

Ref issues: #30 #132
Reviewer: @liamjm


timesketch/ui/static/components/explore/explore-heatmap-directive.js
```diff
@@ -0,0 +1,152 @@
...
+            scope: {
+                sketchId: '=',
+                filter: '=',
+                query: '=',
+                meta: '=',
+                showCharts: '='
+            },
+            require: '^tsSearch',
+            link: function(scope, element, attrs, ctrl) {
+                var AGGREGATION_TYPE = 'heatmap';
```
liamjm: Not sure what is the best style here for this constant. Can it be outside of this function? 
Should it have trailing underscore to indicate private?

timesketch/api/v1/resources.py
```diff
@@ -339,11 +342,25 @@ def post(self, sketch_id):
             for timeline in sketch.timelines:
                 tl_colors[timeline.searchindex.index_name] = timeline.color
                 tl_names[timeline.searchindex.index_name] = timeline.name
+
+            histogram = None
+            es_total_count_unfiltered = 0
+            try:
+                buckets = result[u'aggregations'][u'data_type'][u'buckets']
```
liamjm: I'm not a huge fan of strings literals used like this due to the increased chance of typos causing hidden bugs (especially with the catching of all KeyErrors), but I"m not familiar enough with the code base to have a good feel for how much of an issue this is.
Ideas:
* define constants for these strings.
* add accessor methods/properties to the result type to allow defined programmatic access to these internal components.

timesketch/api/v1/resources.py
```diff
@@ -339,11 +342,25 @@ def post(self, sketch_id):
...
+                histogram = buckets
+                for bucket in buckets:
+                    es_total_count_unfiltered += bucket[u'doc_count']
+            except KeyError:
```
liamjm: See above comment, but the KeyError catch should be the line under the buckets = ... line to reduce scope that is being caught.
The rest of the lines in the try block could be moved into an else block.

liamjm:
Looks good. A couple of comments.

timesketch/ui/static/components/explore/explore-heatmap-directive.js
berggren: We actually don't need a constant here. I just removed it.

timesketch/api/v1/resources.py
berggren: Done.

timesketch/api/v1/resources.py
berggren: I agree, I will go for 2 and add accessor methods but this require a refactor of the result backend. This needs to be done anyways because I want the result object to be more than it is today. I'll do this in a later PR though as this involves some thinking before implementing.

berggren:
Updated, PTAL.

liamjm:
LGTM

Issue #136, opened 2015-11-01
Cluster output based on time


berggren:
It can sometimes be difficult to browse a large timelines when the events are far apart. It would be nice to have an indication on how much time is between events if it exceeds a certain threshhold, e.g. more than a day.

PR #138, opened 2015-11-02
Calculate time clusters

berggren:
It can sometimes be difficult to browse a large timelines when the events are far apart. It would be nice to have an indication on how much time is between events if it exceeds a certain threshhold, e.g. more than a day.

Ref issue: #136 
Reviewer: @onager


Onager:
LGTM

Issue #137, opened 2015-11-01
Show tags as UI labels


berggren:
If there are tags in the tag attribute for event, show them in the UI as labels.

PR #140, opened 2015-11-02
Show tags in event result list

berggren:
If there is a tag array in the event we should display it in the UI.

Ref issue: #137 
Reviewer: @onager


Onager:
LGTM

Issue #94, opened 2015-06-23
Be able to render a view page based on URL query string


liamjm:
It would be great to be able to dynamically generate a view based on the query args, allowing this URL to be created by another system. eg view events for IP 1.1.1.1 with a filter blah:

timesketch.example.com/explore?sketch=1&q=ip:1.1.1.1&filter=blah

PR #143, opened 2015-11-11
Handle URL parameters for search and filter

berggren:
Initial support for setting query string and some limited filtering from GET request parameters. This gives you this e.g:

http://127.0.0.1:5000/sketch/1/explore/?q=foobar&time_start=2015-01-01&time_end=2015-12-13

Ref issue: #94 
Reviewer: @onager


Onager:
LGTM

Issue #86, opened 2015-05-29
Delete a view


hiddenillusion:
Once a view is created, I don't see a way that the view can be deleted without deleting the entire sketch.

berggren:
Yes, this is definitely on the agenda. If not before the upcoming release this will be added shortly after that. Also ability to delete timelines and other entities in the system will be added.

pettai:
I was also about to ask about the possibility to delete timelines.
I would be happy if tsctl could get a (admin) command for that.

berggren:
@pettai Do you mean deleteing as in purging the timeline along with the Elastic index? Or do you mean deleting a timeline from a sketch? A timeline can be part of multiple sketches, so purging it might get difficult without breaking potential other sketches.

If the latter, this is much easier to do, and I'm working on that. You will be able to delete the timeline from the current sketch you are working on.

pettai:
@berggren Delete from sketch is a good start. 
But I actually wanted to purge timelines too (but that could be an "admin" operation, only available via tsctl for instance)

(But showing import date and possible other metadata like number of events for the timeline would help to be able to avoid the less useful imported timelines. (For instance, I imported several copies of the same host from different plaso runs (using different versions of plaso) to check if / how much the timeline events would differentiate...)

berggren:
Good idea, makes sense. I'll add the purge option to be implemented.
Regards to getting stats from the timelines, that is a WIP.

PR #146, opened 2015-11-24
86 delete view

berggren:
This enables any user with write permission on the sketch to delete a view.

Ref issue: #86 
Reviewer: @onager


Onager:
LGTM

Issue #54, opened 2015-03-19
Hide events in view


berggren:
It would be nice to manually be able to hide events in a view. Possible implementation could be a special label that is applied to the event that will prevent the UI to render it.

This is meant for manual hiding events on a small dataset, not bulk hiding.
This will not limit the amount of events sent to the client, only what is visually displayed.

An indicator that there are hidden events in the view need to be in place and a way to toggle visibility of them.

PR #147, opened 2015-11-25
54 hide event

berggren:
This add the feature to hide events in the resulting view.

Ref issue: #54 
Reviewer: @onager


Onager:
LGTM

PR #148, opened 2015-11-25
Initial docker support

pettai:
Initial docker support
docker-compose up will build and start timesketch container +
dependencies.


docker/ts_config/timesketch.conf
```diff
@@ -0,0 +1,12 @@
+DEBUG = False
+SECRET_KEY = u'dockerdockerdocker'
```
berggren: This needs to be generated later otherwise the installation is open for XSRF attackes etc. Maybe in the install script?

berggren:
Thanks for doing this! I wil review this asap.

docker/ts_config/timesketch.conf
pettai: Right, I had that there for testing.
The entrypoint also adds a demo user, which you also might want to remove...
But better push something so we get somewhere...
(not working in this commit: celery worker)

pettai: Ok, now I remembered why I did like this.
Generating a secret during installation is ok, but then it will be statically configured for the built docker container (which then gets pushed to docker hub). That's equally bad I would say.
My original intention was to mount the configuration so you could customise it easily...

pettai: For now, just remove dockerdockerdocker with nothing, so you manually have to put in a secret.

berggren:
Closing this one out as there is another PR for docker floating around.
Thanks for working on this though!

Issue #150, opened 2015-11-27
Timeline picker bug


berggren:
There is a race condition on how the timeline picker gets rendered in the UI. In some cases the View XHR request returns to the client before the Sketch XHR request, making the list of selected timelines to reset to all selected.

PR #151, opened 2015-11-27
Timeline picker bug

berggren:
There is a race condition on how the timeline picker gets rendered in the UI. In some cases the View XHR request returns to the client before the Sketch XHR request, making the list of selected timelines to reset to all selected.

This change will not fetch the view before the sketch XHR is done, as intended.

Ref issue: #150 
Reviewer: @onager


Onager:
LGTM

Issue #120, opened 2015-09-30
Handle whitespace (tabs, newlines etc.) more nicely in events


Onager:
These are generated by some event types, notably the EVT/EVTX message string expansion.

PR #153, opened 2015-12-03
Preserve newlines etc in event info and comments

berggren:
Fix formatting in events info and comments. This fixes newlines etc.

Ref issue: #120 
Reviewer: @onager


Onager:
Looks great! LGTM

Issue #103, opened 2015-07-08
Move wsgi.py file inside the timesketch package?


gtback:
Would it be possible to move the `wsgi.py` file inside the `timesketch` package, so that if timesketch is installed with pip, so it's not necessary to have an additional copy of the timesketch code?

I'd like to run Timesketch with Gunicorn using the wsgi.py file (`gunicorn timesketch.wsgi ...`), rather than using the tsctl command.

berggren:
This sounds like a good idea. I also usually run timesketch with gunicorn by first moving the wsgi file somewhere gunicorn can read it and modify it to use virtualenv etc, then I:
$ gunicorn -b 127.0.0.1:4000 --log-file - wsgi:application

Moving it into the package might be another solution. I will try this out.

Thanks for the idea!

gtback:
Thanks, @berggren .

I realize there are other ways I can accomplish the same thing (I'll have to look into your approach), but I currently do the following:

- Clone repo onto machine
- Install timesketch from checked out code into a new virtualenv using pip (in order to automatically install all of Timesketch's dependencies). 
- Point gunicorn to the wsgi.py file in the cloned repo (via a `gunicorn_start` script, run by supervisord, inspired by https://gist.github.com/postrational/5747293). 

It would be easy enough to move the wsgi.py file into the virtualenv as well, just like the gunicorn_start script, but thought putting it in the timesketch package could be another alternative.

berggren:
@gtback This is now merged.

gtback:
Thanks, @berggren!

PR #154, opened 2015-12-03
Move wsgi into the package

berggren:
This makes more sense. The wsgi file is on the installed timesketch package and can now be run as follows:

$ gunicorn -b 127.0.0.1:80 --log-file - --timeout 120 timesketch.wsgi:application
Without moving around files etc.

Ref issue: #103 
Reviewer: @onager


Onager:
LGTM

Issue #110, opened 2015-09-07
Change/reset password capability for tsctl


pettai:
A few days ago I created a couple of test user with random passwords I forgot about... (oh, the shame)
And I can't just overwrite the users passwords with tsctl:

$ tsctl add_user -u demo123 -p <newpassword>
The username (demo123) is already taken, try another one.

So for instance, adding a -f option (force) so one could overwrite the password would be handy for short minded persons like me. (Or a new subcommand "ch_user" or something else would be just as fine...)

berggren:
Good point. I'll fix this asap.

PR #155, opened 2015-12-03
Update password if user exists

berggren:
When running tsctl add_user with a existing user, update the password.

Ref issue: #110 
Reviewer: @onager


Onager:
LGTM

Issue #217, opened 2016-09-01
Search Templates


berggren:
There should be a way to have generic saved views that could be applied to any sketch.

Example:
Show me all login events over RDP

berggren:
This has been i implemented.

PR #233, opened 2016-09-30
217 search templates

berggren:
This is Search Templates

Reviewer: @onager
Ref issue: #217


timesketch/api/v1/resources.py
```diff
@@ -338,12 +373,66 @@ def post(self, sketch_id):
         form = SaveViewForm.build(request)
         if form.validate_on_submit():
             sketch = Sketch.query.get_with_acl(sketch_id)
+            new_searchtemplate = form.new_searchtemplate.data
```
Onager: Can you factor this out into a a separate method? The logic here is hard to follow with everything in a conditional like this.

timesketch/api/v1/resources.py
```diff
@@ -861,3 +1023,30 @@ def post(self, sketch_id, story_id):
...
+
+        Args:
+            sketch_id: Integer primary key for a sketch database model
+            story_id: Integer primary key for a story database model
+
+        Returns:
+            A story in JSON (instance of flask.wrappers.Response)
+        """
+        form = ExploreForm.build(request)
+        if form.validate_on_submit():
```
Onager: This would be easier to follow as:

if not form.validate_on_submit():
 return abort(....)

sketch = ...

timesketch/api/v1/resources.py
berggren: Done. I'm not 100% sure this is optimal for readability. I'm gonna revisit this in a later CL.

timesketch/api/v1/resources.py
berggren: I agree (I think..) This requires a big refactor though (all API endpoints and views needs to be changed to be consistent, not just here). I'm gonna follow up on it in a separate CL.

Issue #243, opened 2016-10-27
Issue using tsctl


jamdunnDFW:
I'm getting the following error when trying to install timesketch using the instructions from the wiki (https://github.com/google/timesketch/wiki/Installation). I'm attempting to install on Ubuntu 14.04.

Anyone run into this issue before?

Traceback (most recent call last):
  File "/usr/local/bin/tsctl", line 34, in <module>
    from timesketch import create_app
  File "/usr/local/lib/python2.7/dist-packages/timesketch/__init__.py", line 26, in <module>
    from timesketch.api.v1.resources import AggregationResource
  File "/usr/local/lib/python2.7/dist-packages/timesketch/api/v1/resources.py", line 39, in <module>
    from flask_login import current_app
ImportError: cannot import name current_app

MajinLogan:
I have had the exact same issue when I attempt to add a user.

csash:
Flask-Login did some refactoring between v0.3.2 and v0.4.0 that changed where this import lives. As a temporary workaround, installing Flask-Login v0.3.2 instead will fix this issue. 

I think all the current_app imports can just be changed to importing from flask instead of flask_login though.

PR #244, opened 2016-10-29
Fix for #243 (update current_app imports) / update one FlaskWTF import in forms

csash:
Change all imports of current_app to import from flask instead of
flask_login. Refactoring in Flask-Login between v.0.3.2 and v0.4.0
moved their import of current_app into another module and broke
Timesketch stuff that relied on it (#243). Also updating forms to use
flask_wtf.FlaskForm instead of flask_wtf.Form, since flask_wtf.Form has
been re-named and is apparently going away soonish.

### DIFF
--- a/timesketch/api/v1/resources.py
+++ b/timesketch/api/v1/resources.py
@@ -33,10 +33,10 @@ import os
 import uuid
 
 from flask import abort
+from flask import current_app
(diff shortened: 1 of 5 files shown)


berggren:
Great, thanks for fixing this!

PR #296, opened 2017-03-06
API endpoint for tsctl's add_index

csash:
I think this has general utility for folks doing automation and integration work with Timesketch, so I figured I'd offer it up.

We wrote this to make some integrations possible between Timesketch and other parts of our stack. It's useful when you have things pushing data to ES (Timesketch) not using psort or tsctl and would like them to advise Timesketch of the existence of the data, without needing to have Timesketch installed, command line access to the Timesketch server, or access to Timesketch's database. 

To do this, we added a tsctl API under /api/v1/tsctl/ and added a searchindices endpoint that mirrors exactly what tsctl's add_index command does, with a couple minor changes for context. Requests to the tsctl API are authenticated with a token set in timesketch.conf; the tsctl API endpoints only get loaded if the token is set.

Related thoughts that came up while doing this:
- Would it make sense to enforce a uniqueness constraint on the ES index name on the SearchIndex. I can't think of a use case for multiple, differently named timelines using the same ES index. It just seems to add opportunities for things to go wrong or get confusing.
- Should all the things that create a SearchIndex check for the timesketch_label mapping's presence and add it if it's not present?
- The error handler for ApiHTTPError doesn't seem to be called when DEBUG is false, which precludes things using it from providing intelligent errors to users. I'll open a separate issue on that.

### DIFF
--- a/timesketch.conf
+++ b/timesketch.conf
@@ -73,7 +73,9 @@ UPLOAD_ENABLED = False
 
 # Folder for temporarily storage of Plaso dump files before being processed and
 # inserted into the datastore.
-UPLOAD_FOLDER = u'/tmp'o enable the tsctl API endpointss
+UPLOAD_FOLDER = u'/tmp'
+
+# Celery broker configuration. You need to change ip/port to where your Redis
 # server is running.
 CELERY_BROKER_URL='redis://127.0.0.1:6379',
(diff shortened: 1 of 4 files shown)


csash:
Need to fix some stuff to make the googlebot happy. Will re-submit shortly.

PR #297, opened 2017-03-06
API endpoint for tsctl's add_index

csash:
I think this has general utility for folks doing automation and integration work with Timesketch, so I figured I'd offer it up.

This was written to make integrations possible between Timesketch and other things. It's useful when you have things pushing data to ES/Timesketch not using psort or tsctl and would like them to advise Timesketch of the existence of the data, without needing to have Timesketch or Plaso installed, command line access to something running Timesketch, or access to Timesketch's database.

To do this, we added a tsctl API under /api/v1/tsctl/ and a searchindices endpoint that mirrors exactly what tsctl's add_index command does, with a couple minor changes for context. Requests to the tsctl API are authenticated with a token set in timesketch.conf; the tsctl API endpoints only get loaded if the token is set.

Related thoughts that came up while doing this:

- Would it make sense to enforce a uniqueness constraint on the ES index name on the SearchIndex. I can't think of a use case for multiple, differently named timelines using the same ES index. It just seems to add opportunities for things to go wrong or get confusing.
- Should all the things that create a SearchIndex check for the timesketch_label mapping's presence and add it if it's not present?
- The error handler for ApiHTTPError doesn't seem to be called when DEBUG is false, which precludes things using it from providing intelligent errors to users. I'll open a separate issue on that.


timesketch/__init__.py
```diff
@@ -170,7 +171,13 @@ def load_user(user_id):
         return User.query.get(user_id)
 
     # Setup CSRF protection for the whole application
-    CSRFProtect(app)
+    csrf = CSRFProtect(app)
+
+    # Load the tsctl API endpoints only if the TSCTL API key is set
+    if u'TSCTL_API_KEY' in app.config and app.config[u'TSCTL_API_KEY']:
```
berggren: I would rather implement a general API KEY feature instead of this special case. You can achieve the same with a service account and wrap the POSTs with scraping the CSRF token. I know this is not elegant, but it works before we have proper API KEYs implemented. Take a look at this example:

https://gist.github.com/berggren/365d4e6a74eb245f6f6a019c43a8103c

timesketch/api/v1/resources.py
```diff
@@ -25,6 +25,8 @@
 POST /sketches/:sketch_id/event/
 POST /sketches/:sketch_id/event/annotate/
 POST /sketches/:sketch_id/views/
+
+POST /tsctl/searchindices/
```
berggren: this can be just /searchindices/

Thinking about this, we should create 2 exnpoints here:
/index/ - Create or get ES index (create using es.create_index)
/searchindices/ - Create TS index mapping (the one you implemented here)

timesketch/api/v1/resources.py
```diff
@@ -1166,3 +1169,55 @@ def delete(self, sketch_id, timeline_id):
         sketch.timelines.remove(timeline)
         db_session.commit()
         return HTTP_STATUS_CODE_OK
+
+class SearchIndexResource(ResourceMixin, Resource):
```
berggren: This is great, just drop the API key. We should also create a IndexResource using es.create_index to create elastic indes via the API

timesketch/api/v1/resources.py
```diff
@@ -1166,3 +1169,55 @@ def delete(self, sketch_id, timeline_id):
...
+                raise ApiHTTPError(
+                    message=u'User {0} does not exist' \
+                        .format(form.username.data),
+                    status_code=HTTP_STATUS_CODE_BAD_REQUEST)
+
+            # Check that the index exists
+            es = self.datastore
+            if not es.client.indices.exists(index=form.index.data):
+                raise ApiHTTPError(
+                    message=u'Index {0} does not exist in the datastore' \
```
berggren: -in the datastore

timesketch/api/v1/resources.py
```diff
@@ -1166,3 +1169,55 @@ def delete(self, sketch_id, timeline_id):
...
+
+        # Check that the index exists
+        if form.validate_on_submit() or \
+                (len(form.errors) == 1 and u'csrf_token' in form.errors):
+
+            # Check that the user exists
+            user = User.query.filter_by(username=form.username.data).first()
+            if not user:
+                raise ApiHTTPError(
+                    message=u'User {0} does not exist' \
```
berggren: Please refactor so we can skip then "\"

csash: Sorry... don't understand your comment. What would you like me to change?

csash:
I think I've addressed your comments, with the exception of the "refactor" one. Let me know how it looks.

berggren:
Hey @csash ! Thanks for contributing this PR. I have added this functionality into the Python API client instead. This should cover your needs. Let me know if there is anything missing.

PR #300, opened 2017-03-16
Create timestamp from datetime string when importing CSV

tomchop:
It is now possible to build timelines from CSVs that do not necessarily have a timestamp (they will fall-back to the datetime field). Particularly handy since now it can ingest the output from psort.

### DIFF
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -16,7 +16,9 @@
 import colorsys
 import csv
 import random
+import time
 
+from dateutil import parser
 
 def random_color():
     """Generates a random color.
@@ -40,10 +42,11 @@ def read_and_validate_csv(path):
     """
     # Columns that must be present in the CSV file
     mandatory_fields = [
-        u'message', u'timestamp', u'datetime', u'timestamp_desc']
+        u'message', u'datetime', u'timestamp_desc']
 
     with open(path, u'rb') as fh:
-        reader = csv.DictReader(fh)
+
+        reader = csv.DictReader((x.replace('\x00', '') for x in fh))
         csv_header = reader.fieldnames
         missing_fields = []
         # Validate the CSV header
@@ -53,9 +56,16 @@ def read_and_validate_csv(path):
         if missing_fields:
             raise RuntimeError(
                 u'Missing fields in CSV header: {0:s}'.format(missing_fields))
-
         for row in reader:
+            if u'timestamp' not in csv_header and u'datetime' in csv_header:
+                try:
+                    row['timestamp'] = str(int(time.mktime(
+                        parser.parse(row["datetime"]).timetuple())))
+                except ValueError:
+                    continue
+
             yield row
+
 
 
 def get_validated_indices(indices, sketch_indices):
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -46,7 +46,7 @@ def read_and_validate_csv(path):
 
     with open(path, u'rb') as fh:
 
-        reader = csv.DictReader((x.replace('\x00', '') for x in fh))
+        reader = csv.DictReader(fh)
         csv_header = reader.fieldnames
         missing_fields = []
         # Validate the CSV header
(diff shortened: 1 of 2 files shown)


timesketch/lib/utils.py
```diff
@@ -40,10 +42,11 @@ def read_and_validate_csv(path):
     """
     # Columns that must be present in the CSV file
     mandatory_fields = [
-        u'message', u'timestamp', u'datetime', u'timestamp_desc']
+        u'message', u'datetime', u'timestamp_desc']
 
     with open(path, u'rb') as fh:
-        reader = csv.DictReader(fh)
+
+        reader = csv.DictReader((x.replace('\x00', '') for x in fh))
```
berggren: Why remove null bytes here instead of having the user provide clean data? Bail out and have the user clean up the input before maybe.

timesketch/lib/utils.py
```diff
@@ -53,11 +56,18 @@ def read_and_validate_csv(path):
         if missing_fields:
             raise RuntimeError(
                 u'Missing fields in CSV header: {0:s}'.format(missing_fields))
-
         for row in reader:
+            if u'timestamp' not in csv_header and u'datetime' in csv_header:
+                try:
+                    row['timestamp'] = str(int(time.mktime(
+                        parser.parse(row["datetime"]).timetuple())))
```
berggren: u'datetime' (instead of "")

timesketch/lib/utils.py
```diff
@@ -53,11 +56,18 @@ def read_and_validate_csv(path):
         if missing_fields:
             raise RuntimeError(
                 u'Missing fields in CSV header: {0:s}'.format(missing_fields))
-
         for row in reader:
+            if u'timestamp' not in csv_header and u'datetime' in csv_header:
+                try:
+                    row['timestamp'] = str(int(time.mktime(
```
berggren: u'timestamp'

berggren: Why cast to string? Elasticsearch is happy with an int.

timesketch/lib/utils.py
```diff
@@ -16,7 +16,9 @@
 import colorsys
 import csv
 import random
+import time
 
+from dateutil import parser
```
berggren: This adds a new dependency to the project, right?

timesketch/lib/utils.py
berggren: There are a lot of parentheses here, it is pretty hard to read. Can we break it apart?

Also, we need to make this microseconds since 1970-01-01

timesketch/lib/utils.py
tomchop: Yes, I'll add it to `setup.py`

timesketch/lib/utils.py
tomchop: The output provided by plaso includes the occasional null bytes; this is mainly to make the user's life easier and since we're talking about text I don't think it will change much to what's ingested.

timesketch/lib/utils.py
tomchop: We can make them microseconds by `*1000`, but `psort`'s output is still datetime-seconds: `1970-01-01T00:00:00+00:00`

tomchop: Ack

timesketch/lib/utils.py
tomchop: Ack

timesketch/lib/utils.py
```diff
@@ -59,8 +59,8 @@ def read_and_validate_csv(path):
         for row in reader:
             if u'timestamp' not in csv_header and u'datetime' in csv_header:
                 try:
-                    row['timestamp'] = str(int(time.mktime(
-                        parser.parse(row["datetime"]).timetuple())))
+                    parsed = parser.parse(row[u'datetime'])
```
berggren: Looks better, thx. Can we make this var name a bit more descriptive. Maybe parsed_datetime or similar?

berggren:
Hey @tomchop any updates on this PR? I would love to merge it :)

tomchop:
Done, sorry for the delay!

Issue #306, opened 2017-03-31
Enable streaming support for query results


berggren:
Elasticsearch supports scrolling through their scrolling API. This makes it possible to stream events and to get all event results for a query. The main motivator for this is for external programs that need all the results should be able to do so in an efficient way.

PR #307, opened 2017-03-31
Streaming query results

berggren:
Ref issue: #306 
Reviewer: @rgayon


timesketch/lib/datastores/elastic.py
```diff
@@ -221,6 +224,19 @@ def search(
         DEFAULT_LIMIT = 500  # Maximum events to return
         LIMIT_RESULTS = query_filter.get(u'limit', DEFAULT_LIMIT)
 
+        # Default timeout for the scroll API
+        default_scroll_timeout = u'1m'  # 1 minute
+        if scroll:
```
rgayon: I would make the distinction between the boolean "enable_scroll" parameter and the string "scroll_timeout" to pass down to the client below.
Otherwise LGTM =)

berggren: I agree, fixed now.

Issue #311, opened 2017-04-18
Purge timeline


berggren:
Ability to purge timelines, both from TImesketch and Elasticsearch.

PR #312, opened 2017-04-18
Delete timeline permanently

berggren:
Deletes a timeline in both Timesketch and Elasticsearch

Ref issue: #311 
Reviewer: @tomchop, @aarontp

### DIFF
--- a/tsctl
+++ b/tsctl
@@ -37,6 +37,7 @@ from timesketch.models import drop_all
 from timesketch.models.user import Group
 from timesketch.models.user import User
 from timesketch.models.sketch import SearchIndex
+from timesketch.models.sketch import Timeline
 
 
 class DropDataBaseTables(Command):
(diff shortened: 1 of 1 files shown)


tsctl
```diff
@@ -364,6 +365,49 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+    def __init__(self):
+        super(PurgeTimeline, self).__init__()
+
+    def run(self, index_name):
+        """Delete timeline in both Timesketch snd Elasticsearch.
+
+        Args:
+            index_name: The name of the index in Elasticsearch
+        """
+        index_name = unicode(index_name.decode(encoding=u'utf-8'))
```
tomchop: I think this is redudant, you can drop the unicode casting.

tsctl
```diff
@@ -364,6 +365,49 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+        searchindex = SearchIndex.query.filter_by(index_name=index_name).first()
+
+        es = ElasticsearchDataStore(
+            host=current_app.config[u'ELASTIC_HOST'],
+            port=current_app.config[u'ELASTIC_PORT'])
+
+        if not searchindex:
```
tomchop: Swap with block in Line 386?

tsctl
```diff
@@ -364,6 +365,49 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+            sys.stdout.write(u'No such index\n')
+            sys.exit()
+
+        timelines = [t for t in Timeline.query.filter_by(searchindex=searchindex).all()]
+        sketches = [t.sketch for t in timelines if t.sketch]
+        if sketches:
+            sys.stdout.write(u'WARNING: This timeline is in use by the following sketches:\n')
+            for sketch in sketches:
+                if not sketch.get_status.status == u'deleted':
```
tomchop: Integrate this test in the loop on l. 395. Otherwise you will get the warning but no sketch names if all sketches are marked as "deleted"

tsctl
berggren: It is needed because otherwise SQLAlchemy get's unhappy because the SQL schema enforces unicode only.

tomchop: The return value of `.decode()` is already unicode.

berggren: ahh, you mean that, yeah good point, I know this but I'm trying to stay consistent with the code base (which for whatever reason turned out this way, I blame myself). The plan is to clean this up at a later stage. :)

tsctl
berggren: Yeah, good catch. This is now fixed, ignoring deleted sketches all together.

tsctl
berggren: Done.

berggren:
Yeah, I'm aware that IDs are not ideal, but this is the first iteration of this and this feature will be available in the UI later on. The crux here is that multiple "Sketch timelines" can have the same name, so which one do you choose? etc.

The tsctl tool is meant for admins only.

berggren:
@aarontp

tsctl
```diff
@@ -364,6 +365,51 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+    """Delete timeline permanently from Timesketch and Elasticsearch."""
+    option_list = (
+        Option(u'--index', u'-i', dest=u'index_name', required=True),
+    )
+
+    def __init__(self):
+        super(PurgeTimeline, self).__init__()
+
+    def run(self, index_name):
+        """Delete timeline in both Timesketch snd Elasticsearch.
```
aarontp: nit: s/snd/and/

tsctl
```diff
@@ -364,6 +365,51 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+            host=current_app.config[u'ELASTIC_HOST'],
+            port=current_app.config[u'ELASTIC_PORT'])
+
+        timelines = Timeline.query.filter_by(searchindex=searchindex).all()
+        sketches = [t.sketch for t in timelines if
+                    t.sketch and t.sketch.get_status.status != u'deleted']
+        if sketches:
+            sys.stdout.write(u'WARNING: This timeline is in use by:\n')
+            for sketch in sketches:
+                sys.stdout.write(u' * {0:s}\n'.format(sketch.name))
```
aarontp: Just curious -- Why do this instead of just normal prints?

berggren: This is just a habit of mine when writing CLI tools. I want to control where to put the output (stdout, stderr etc).

tsctl
berggren: done.

PR #249, opened 2016-11-03
Docker Support

jessemillar:
Similar to [this previous pull request](https://github.com/google/timesketch/pull/148) but with more modern Docker practices and using the current Timesketch codebase. As an added bonus, I've commented a ton of things for educational purposes and added Docker container building and pushing to the Travis CI pipeline.

The following commands will build the Docker container from the cloned repository and start the container in conjunction with linked Postgres and Elastic containers.
```
docker-compose build && docker-compose up -d
```

[This line](https://github.com/jessemillar/timesketch/blob/master/Dockerfile#L15) in the `Dockerfile` will need to be changed once [this fix](https://github.com/google/timesketch/pull/244) is landed in the November release. I'm currently pulling a .zip archive of the master branch instead of installing from pip because the Flask fix isn't published yet which prevented builds from succeeding.

My updated `.travis.yml` will need to be updated on lines [12](https://github.com/jessemillar/timesketch/blob/master/.travis.yml#L12) and [25](https://github.com/jessemillar/timesketch/blob/master/.travis.yml#L25) to push to a different Docker Hub repository (potentially a Google repository if you choose to merge). As proof that the deploy pipeline works with the new Docker support, here is [my hosted version](https://hub.docker.com/r/jessemillar/timesketch/) of the Timesketch Docker container with Docker Hub documentation.

PR #323, opened 2017-06-02
Add Docker Support

witoff:
**Problem**
Timesketch is currently difficult to deploy in a dockerized environment, which is desirable to simplify deployments.

**Usage**
This adds docker support for simpler local development and test. Test with:

```shell 
docker-compose build
docker-compose up
# view at http://localhost:5000
``` 

**Notes**
Builds on @jessemillar's work in https://github.com/google/timesketch/pull/249.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
-# Pull and use the official Docker Hub Ubuntu 14.04 base image
-FROM ubuntu:14.04
+# Use the official Docker Hub Ubuntu 14.04 base image
+FROM ubuntu:16.04
 
 # Update the base image
 RUN apt-get update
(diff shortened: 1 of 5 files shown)


witoff:
I signed it!

Dockerfile
```diff
@@ -0,0 +1,36 @@
+# Use the official Docker Hub Ubuntu 14.04 base image
+FROM ubuntu:14.04
```
berggren: We wanna use 16.04 here, seeing 14.04 is getting rather old. Totally fine to do this in a later PR though.

timesketch/lib/utils.py
```diff
@@ -60,7 +60,8 @@ def read_and_validate_csv(path):
             if u'timestamp' not in csv_header and u'datetime' in csv_header:
                 try:
                     parsed_datetime = parser.parse(row[u'datetime'])
-                    row[u'timestamp'] = int(time.mktime(parsed_datetime.timetuple()))
+                    row[u'timestamp'] = str(int(
```
berggren: Was this meant to go into this PR?

berggren:
Looks good! Just a couple of small things, otherwise LGTM.

Dockerfile
witoff: fixed

timesketch/lib/utils.py
witoff: reverted.  this came in when rebasing and may have been something jesse added in testing.  so long as it still runs locally i'll leave this out.

PR #324, opened 2017-06-05
Add Docker Support

witoff:
**Problem**
Timesketch is currently difficult to deploy in a dockerized environment, which is desirable to simplify deployments.

**Usage**
This adds docker support for simpler local development and test. Test with:

```shell 
docker-compose build
docker-compose up
# view at http://localhost:5000
``` 

**Notes**
- Builds on @jessemillar's work in https://github.com/google/timesketch/pull/249.
- This replaces https://github.com/google/timesketch/pull/323 which had issues with git log emails and the google CLA.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,11 @@
+sudo: required
+services:
+  - docker
 language: python
 python:
   - "2.7"
+before_install:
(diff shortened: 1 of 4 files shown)


berggren:
Ref issue: #325

PR #348, opened 2017-07-07
Python API client library

### LINKED ISSUE #347: Python API client library
To integrate Timesketch it would be nice to have an easy to use library that takes care of the underlying HTTP REST calls and exposes python objects instead.

This library should not depend on the main Timesketch code but be stand alone.

berggren:
Fixes #347 
Ping @hiddenillusion


timesketch/api/v1/resources.py
```diff
@@ -952,8 +957,14 @@ def post(self):
...
 
-            return self.to_json(
-                searchindex, status_code=HTTP_STATUS_CODE_CREATED)
+            # Return Timeline if it was created.
+            if timeline:
+                return self.to_json(
+                    timeline, status_code=HTTP_STATUS_CODE_CREATED)
+            else:
+                return self.to_json(
+                    searchindex, status_code=HTTP_STATUS_CODE_CREATED)
```
rgayon: Should the status_code be different is timeline is not created?

api_client/python/timesketch_api_client/client.py
```diff
@@ -0,0 +1,460 @@
...
+            Dictionary with resource data.
+        """
+        if not self._data or refresh_cache:
+            resource_uri = u'sketches/{0:d}'.format(self.id)
+            self._data = self.api.fetch_resource_data(resource_uri)
+        return self._data
+
+    @property
+    def data(self):
+        """Property to fetch sketch data."""
```
berggren: Done.

api_client/python/timesketch_api_client/client.py
```diff
@@ -0,0 +1,460 @@
...
+
+    @property
+    def status(self):
+        """Property that returns sketch status.
+
+        Returns:
+            Sketch status as string.
+        """
+        sketch = self._lazyload_data()
+        return sketch[u'objects'][0][u'status'][0][u'status']
```
berggren: Heh, it is actually right :) It is a result of SQL object mapping magic.

timesketch/api/v1/resources.py
berggren: It will be a 500 if it failed, but in another part of the code.

berggren:
Re errors: The code raises HTTP errors where relevant. I'll add a TODO to see over this and raise more explicit with message string etc.

rgayon:
LGTM

Issue #262, opened 2017-01-19
Elasticsearch 5.x support


berggren:
There are some breaking changes in how Elasticsearch 5.x do queries. We need to update our query definitions.

berggren:
The code for this is written, and lives in a feature branch named "262-es5-support".
This will get merged before release as it brakes backward compatibility.

PR #350, opened 2017-07-07
Elasticsearch 5.x support

berggren:
Support for Elasticsearch 5.x. This will break backwards compatibility with Elasticsearch 2.x deployments. You will need to upgrade your cluster.
Fixes: #262


timesketch/lib/datastores/elastic.py
```diff
@@ -328,6 +335,7 @@ def set_label(
             script_name = u'toggle_label'
         script = {
             u'script': {
+                u'lang': u'groovy',
```
berggren: Groovy is still supported. I want to do some more testing first and update this in a later PR.

berggren: Re bootstrap.sh - missed that, thanks for the heads up!

tomchop: No idea how this will behave if you copy two scripts that have the same name. It will probably depend on how ES is configured (groovy or painless) but IIRC ES5 uses painless by default so you might want to tweak the config there.

berggren: ack, I will test this before release and update docs etc.

PR #354, opened 2017-07-17
upgrade to docker-compose v3 and fix docker support

witoff:
Notes:
* This fixes docker support now that docker files are in `/docker/`. 
* Upgrades docker-compose to v3 format
* Adds a redis container and sets .conf accordingly

Note that the docker [docs](https://github.com/google/timesketch/wiki/Docker) are now out of date.  What do you think about converting the wiki file into a new `docker/README.md` that's versioned for inline docs on using these properly?


berggren:
Thanks for updating! Re docs: SG moving it to docker/README.md - makes sense. Let's do that in a separeate PR.

Issue #377, opened 2017-08-16
API should warn if ES index_name already exists when creating a new timeline


hiddenillusion:
Using [`create_timeline()`](https://github.com/google/timesketch/pull/376/files#diff-cfc277cad8b430c4868c66b63ef6ded2R181), from #376, to manually create a timeline works as expected, if the `index_name` provided doesn't already exist within ES.

However, when a duplicate is used, there's no indication/warning. The only way to currently tell is to list all the timelines afterwards, and realize it was never created (the original will exist, the attempted new one silently fails).

One possible work around is to do a quick check (e.g.) -
```
def es_index_exists(self, index_name):
    response = self.fetch_resource_data(u'timelines/')
    for index in response[u'objects'][0]:
        if index_name == index.get(u'index_name'):
            return True
    return False
```

and then verify before creation:

```
def create_timeline(self, timeline_name, index_name=None, public=False):
...
    if not index_name:
        index_name = uuid4().hex

    if self.es_index_exists(index_name):
        <raise/return duplicate error>
```

berggren:
Good point. I have fixed this. When a duplicate is found the server will respond with a HTTP 409 (conflict) and the client code will catch this and raise an error.

berggren:
After thinking about this some more I believe that this is actually working as intended. We are confusing searchindex with timeline here and it was a mistake to call it create_timeline(), it should be named create_searchindex().

A timeline is always bound to a Sketch. A SearchIndex can be referenced in many timelines. This being said, I will change the name of the methods to make this clear.

PR #380, opened 2017-08-21
Create search indices and timelines

berggren:
Fixes: #377 
Reviewer: @hiddenillusion


hiddenillusion:
ack, will have a look

hiddenillusion:
This still doesn't seem to be working as expected. If the `es_index_name` supplied to `create_searchindex()` already exists in ES, `create_searchindex()` still returns a status code **201** even though a _new_ SearchIndex wasn't created.

The most reliable way I've tested existence prior to attempting to create a new searchindex was iterating & comparing index_names from the data returned in `fetch_resource_data(u'searchindices/')`, but that may be slow depending on the number.

berggren:
Hey, good point and I missed changing to the correct status code. I have made some small changes to make this a bit more intuitive.

The method is now named get_or_create_searchindex(), and it will return the searchindex object together with a boolean indicating if the index was created or if it already existed.

Example:
api_object.get_or_create_searchindex(es_index_name="123456", searchindex_name="foobar")
(<timesketch_api_client.client.SearchIndex at 0xdeadbeef>, False)

I was thinking of returning HTTP 409 (conflict) and raise an error if the index already existed, but I then this need to be handled in all client code, and I think that most people don't care. They just want the index object.

PTAL

hiddenillusion:
[one](https://github.com/google/timesketch/pull/380/commits/673e4f8623de9f5d8725035927687897ec4bd1a4#diff-cfc277cad8b430c4868c66b63ef6ded2R197) minor change needs to be done then it LGTM:

`if response.status_code != HTTP_STATUS_CODE_20X` needs to be `if response.status_code not in HTTP_STATUS_CODE_20X` since `HTTP_STATUS_CODE_20X` is a list.

berggren:
Done, thanks for the review @hiddenillusion !

PR #390, opened 2017-08-23
Add frontend build pipeline

franekp:


.gitignore
```diff
@@ -1,3 +1,10 @@
+# Ignore frontend build related files
+node_modules
+timesketch/ui/static/dist
+
+# Ignore data files
+*.plaso
```
berggren: You can move this to your own ignore:
.git/info/exclude

setup.py
```diff
@@ -29,6 +33,28 @@
     u'timelines and analyze them all at the same time.  Add meaning to '
     u'your raw data with rich annotations, comments, tags and stars.')
 
+def check_before_upload():
+    this_dir = os.path.dirname(__file__)
```
berggren: Please add docstring

timesketch/api/v1/resources.py
```diff
@@ -239,7 +239,7 @@ def to_json(self,
 
         schema = {u'meta': meta, u'objects': []}
 
-        if model:
+        if model:  # why is that?
```
berggren: Yeah this can probably be removed. Please test.

.gitignore
franekp: ".plaso" in .gitignore is useful for anyone using vagrant, so if you don't mind I will leave it here. Gitignore does not affect files already checked-in, so it will not break example timeline (vagrant/test.plaso)

berggren: Sound good.

setup.py
```diff
@@ -34,20 +34,23 @@
     u'your raw data with rich annotations, comments, tags and stars.')
 
 def check_before_upload():
+    """Raise UserWarning if frontend build is not present or is not recent,
```
berggren: Let's have the docstring consistent with the rest, i.e. Google style:

"""Warn user if frontend build is not present or is not recent.

Make sure that .js and .css bundles included in the PyPI package are up to date.

Raises:
    UserWarning
"""

PR #437, opened 2017-09-19
[don't merge] Add pycypher - bindings for libcypher-parser.

franekp:
It might be convenient to take a look here:
https://github.com/franekp/timesketch/tree/pycypher/contrib/pycypher
instead of navigating a long diff.


franekp:
moved to: https://github.com/cleishm/libcypher-parser/pull/3

Issue #446, opened 2017-10-04
Chart aggregations is inefficient


berggren:
The current aggregation for heatmap and histogram is inefficient and memory heavy and can cause Elastichsearch to get out of memory.

PR #447, opened 2017-10-04
Nested aggregations

berggren:
This PR changes how we do aggregations. It uses nested aggregations instead of computing all buckets server side, making it faster and less memory hungry.

Also some dafe guards in place to not render and calculate charts when there are many (>300k) events in the result.


berggren:
Fixes #446

aarontp:
Sorry for the delay, I'll take a look at this today.

timesketch/lib/aggregators.py
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 """Elasticsearch aggregations."""
 
+# Max number of search hits to chart
+MAX_RESULT_LIMIT = 300000
```
aarontp: Just curious; How did you come up with this limit?

timesketch/ui/explore/filter.directive.ts
```diff
@@ -68,6 +68,7 @@ export const tsFilter = function () {
                 scope.meta.es_total_count = 0
                 scope.meta.es_time = 0
                 scope.meta.noisy = false
+                ctrl.search(scope.query, scope.filter, scope.queryDsl)
```
aarontp: I'll have to ask you more about this later.  I don't really know what this does.

timesketch/lib/aggregators.py
berggren: Good question, I did some simple memory monitoring to see the sweetspot for most setups. This should be moved to config though. Will add todo.

timesketch/lib/aggregators.py
```diff
@@ -29,12 +32,52 @@ def heatmap(es_client, sketch_id, query_string, query_filter, query_dsl,
...
     aggregation = {
-        u'heatmap': {
-            u'date_histogram': {
-                u'field': u'datetime',
-                u'interval': u'hour',
-                u'format': u'e,H'
+        u'byDay': {
+            u'terms': {
+                u'script': {
+                    u'file': 'dateaggregation',
```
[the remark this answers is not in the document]
berggren: Correct.

timesketch/ui/explore/filter.directive.ts
berggren: This is how AngularJS 1.x works. Happy to explain more later.

Issue #448, opened 2017-10-04
Support for Plaso version 20170930


berggren:
Plaso has changed a lot and our upload feature is broken. Before we figure out how new Plaso works let's put a fix in place. Use subprocess to run psort.py in Celery.

berggren:
Done.

PR #449, opened 2017-10-04
Support new Plaso (20170930 aka Heimdall)

berggren:
Plaso support is currently broken. This is a quick fix while we figure out how to write a proper psort Celery task.


berggren:
Fixes #448

PR #453, opened 2017-10-11
Graph entities and registry

berggren:
Be flexible on how to render graph nodes and edges.
Fixes: #452


timesketch/lib/kb/graph/entities.py
```diff
@@ -0,0 +1,151 @@
...
+            setattr(self, key, properties[key])
+
+    @property
+    def human_readable(self):
+        """Format human readable string for the entity.
+
+        Returns:
+            Human readable string.
+        """
+        return NotImplemented
```
rgayon: raise?

timesketch/lib/kb/graph/entities.py
```diff
@@ -0,0 +1,151 @@
...
+    """Machine node entity."""
+
+    NEO4J_TYPE = u'Machine'
+
+    def __init__(self, node):
+        super(MachineNode, self).__init__(node)
+
+    @property
+    def human_readable(self):
+        return self.hostname
```
rgayon: You probably want to set self.hostname to None in __init__

timesketch/lib/kb/graph/entities.py
```diff
@@ -0,0 +1,151 @@
...
+    """User node entity."""
+
+    NEO4J_TYPE = u'User'
+
+    def __init__(self, node):
+        super(UserNode, self).__init__(node)
+
+    @property
+    def human_readable(self):
+        return self.username
```
rgayon: You probably want to set self.hostname to None in init

rgayon: Same with all the others

timesketch/lib/kb/graph/entities.py
```diff
@@ -0,0 +1,151 @@
...
+        Returns:
+            Dictionary with object attributes.
+        """
+        attributes = self.__dict__
+        attributes[u'human_readable'] = self.human_readable
+        return attributes
+
+
+class GenericEntity(BaseEntity):
+    """Generic entity."""
```
rgayon: What about setting a default NEO4J_TYPE here? Or in the BaseEntity class?

This might catch some errors in the registry below, and the line:
registry[klass.NEO4J_TYPE] = klass

timesketch/lib/kb/graph/entities.py
berggren: Nah, I use return for when it should be overridden.

timesketch/lib/kb/graph/entities.py
berggren: Done

berggren: Not at this moment, I might on Monday when some of this is probably going to be re-written :)

timesketch/lib/kb/graph/entities.py
berggren: Not at this moment, I might on Monday when some of this is probably going to be re-written :)

berggren:
Updated, PTAL.

PR #463, opened 2017-10-17
Graph generation - initial version without celery job.

franekp:
Version that does all importing during processing a request.


timesketch/api/experimental/resources.py
```diff
@@ -0,0 +1,64 @@
...
+                continue
+            result.append(self.graph_datastore.search(
+                statement, params=params))
+        return jsonify(result)
+
+
+class DeleteGraphResource(ResourceMixin, Resource):
+    @login_required
+    def post(self, sketch_id):
+        result = self.graph_datastore.search(delete_graph, params={
```
berggren: Seeing that we execute cypher queries that do more than search maybe we should rename the method from "search" to "query". What do you think?

timesketch/api/experimental/resources.py
```diff
@@ -0,0 +1,64 @@
...
+        params = {
+            u'sketch_id': sketch_id,
+            u'win_logins': win_logins(sketch_id),
+            u'win_services': win_services(sketch_id),
+        }
+        for statement in create_graph.split(';'):
+            statement = statement.strip()
+            if not statement or statement.startswith('//'):
+                continue
+            result.append(self.graph_datastore.search(
```
berggren: This will be expensive when creating large graphs. Could we batch the statements and send in bulk?

timesketch/api/experimental/resources.py
```diff
@@ -0,0 +1,64 @@
...
+
+from timesketch.api.v1.resources import ResourceMixin
+from timesketch.lib.experimental.win_logins import win_logins
+from timesketch.lib.experimental.win_services import win_services
+from timesketch.lib.experimental.create_graph import create_graph
+from timesketch.lib.experimental.delete_graph import delete_graph
+
+
+class WinLoginsResource(Resource):
+    @login_required
```
berggren: Please add a one-line docstring here and for the other methods in this file

timesketch/api/experimental/resources.py
franekp: It doesn't matter. Elasticsearch is the bottleneck here.
On greendale dataset:
 - gathering win_logins and win_services takes about 50s
 - putting this to neo4j takes 0.14s

timesketch/api/experimental/resources.py
franekp: Ok, I will add "Handles POST request to the resource". I'm sure this will be super-helpful.

franekp: Done.

timesketch/api/experimental/resources.py
franekp: Done.

timesketch/api/experimental/resources.py
berggren: Ack, makes sense.

timesketch/api/experimental/resources.py
berggren: Thanks, this is just to stay consistent with the rest of the code base.

franekp:
I approve

PR #465, opened 2017-10-18
Add ES queries and ES index to all edges

berggren:
Fixes: #455 and #454

### DIFF
--- a/timesketch/lib/experimental/create_graph.cql
+++ b/timesketch/lib/experimental/create_graph.cql
@@ -45,8 +45,7 @@ WHERE line.src = line.dst
 
   MERGE (user:WindowsADUser { username: line.user, sketch_id: {sketch_id} })
   MERGE (dst:WindowsMachine { hostname: line.dst, sketch_id: {sketch_id} })
-  SET user.es_index_name = line.es_index_name,
-      dst.es_index_name = line.es_index_name
+  SET dst.es_index_name = line.es_index_name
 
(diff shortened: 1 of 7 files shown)


timesketch/lib/experimental/create_graph.cql
```diff
@@ -32,6 +32,8 @@ WHERE line.src <> line.dst
   MERGE (src)-[r:ACCESS]->(dst)
```
franekp: What about adding es_index to nodes also?

timesketch/lib/experimental/win_services.py
```diff
@@ -69,18 +69,22 @@ def win_services(sketch_id):
             sketch_id=sketch_id, query=u'event_identifier:7045'):
         data = event[u'_source'][u'xml_string']
         res = parse_xml(data)
+        res.extend((event.get(u'_index'), event.get(u'_id')))
```
franekp: This will create multiple edges for one service started multiple times, right?

berggren: It will in the output of this method, but will be merged when cypher adds them to the graph.

franekp: Sounds good.

timesketch/lib/experimental/create_graph.cql
berggren: Done.

berggren:
PTAL

PR #467, opened 2017-10-18
Graphs: use label templates in frontend.

franekp:


berggren:
CLA bot is confused. This seems to happen as soon as I hit the update branch button.

PR #477, opened 2017-10-23
Fix memory leak - destroy Cytoscape instances when not used.

franekp:


berggren:
Can we also add runOutsideAngular() to this PR?

Example:
```
import {
  Component, ViewChild, ElementRef, Input, Output, OnChanges, SimpleChanges, EventEmitter,
  AfterViewInit, NgZone,
} from '@angular/core'
..
constructor(private zone: NgZone) {}
..
this._cy = this.zone.runOutsideAngular(() => cytoscape(options))

```

franekp:
Ok, I will add it.

PR #478, opened 2017-10-23
Minor ui changes

berggren:
Small changes to the UI.
Move the dropdowns in graph view to the right.


timesketch/ui/css/ts.scss
```diff
@@ -705,7 +705,7 @@ ts-graphs-graph-view {
   }
   .#{&}__dropdown-content {
     position: absolute;
-    left: 0px;
+    right: 0px;
```
franekp: is this really needed?

berggren: yep, to align the dropdown

PR #501, opened 2017-12-16
Update install instructions for ES 5.x

csash:
Installation instructions still say to use ES 2.x because ES 5.x is unsupported. Update ES 2.x references to ES 5.x.


berggren:
Thanks for updating!

PR #503, opened 2017-12-17
Add direct link to specific event (#414)

csash:
Resubmitting because my git skills suck... sorry

Adds a link above the comment section in the event detail view to a query that returns the event being viewed. I figure this probably won't get enough use to warrant its own icon in the event list and you'll probably already be looking at the detail view for any event you care about enough to want a direct link for it.


berggren:
CLA warning is wrong. This happens when you hit the update this branch button on GitHub. Was green before I hit the button.

PR #511, opened 2018-01-05
Addition of JSONL upload capability

adamjnichols:
Allows for the upload and ingestion of JSONL files via UI and tsctl, and provides some field correction for timestamp/datetime irregularity from plaso json_line output types that will be corrected in a later PR.

### DIFF
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -15,6 +15,7 @@
 
 import colorsys
 import csv
+import datetime
 import json
 import random
 import time
@@ -76,9 +77,7 @@ def read_and_validate_jsonl(path):
         path: Path to the JSONL file
     """
     # Fields that must be present in each entry of the JSONL file.
-    # We normally require datetime here, but plaso jsonl only has timestamp
-    # need to learn if this is necessary or can be changed
-    mandatory_fields = [u'message', u'timestamp', u'timestamp_desc']
+    mandatory_fields = [u'message', u'datetime', u'timestamp_desc']
 
     with open(path, 'rb') as fh:
 
@@ -87,6 +86,13 @@ def read_and_validate_jsonl(path):
             lineno += 1
             try:
                 linedict = json.loads(line)
+                if u'datetime' not in linedict.keys() and u'timestamp' in linedict.keys():
+                    epoch = int(str(linedict[u'timestamp'])[:10])
+                    dt = datetime.datetime.fromtimestamp(epoch)
+                    linedict[u'datetime'] = dt.isoformat()
+                if u'timestamp' not in linedict.keys() and u'datetime' in linedict.keys():
+                    linedict[u'timestamp'] = parser.parse(linedict[u'datetime'])
+
                 missing_fields = []
                 for field in mandatory_fields:
                     if field not in linedict.keys():
@@ -95,6 +101,7 @@ def read_and_validate_jsonl(path):
                     raise RuntimeError(
                         u"Missing fields in JSON at line {0:n}: {1:s}"
                         .format(lineno, missing_fields))
+
                 yield linedict
 
             except ValueError as e:
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -98,15 +98,13 @@ def read_and_validate_jsonl(path):
                     if field not in linedict.keys():
                         missing_fields.append(field)
                 if missing_fields:
-                    raise RuntimeError
-                        (u"Missing fields in JSON at line {0:n}: {1:s}"
+                    raise RuntimeError(u"Missing field(s) at line {0:n}: {1:s}"
                         .format(lineno, missing_fields))
 
                 yield linedict
 
             except ValueError as e:
-                raise RuntimeError
-                    (u"Error parsing JSON at line {0:n}: {1:s}"
+                raise RuntimeError(u"Error parsing JSON at line {0:n}: {1:s}"
                     .format(lineno, e))
 
 
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -68,6 +68,40 @@ def read_and_validate_csv(path):
             yield row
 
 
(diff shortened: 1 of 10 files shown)


tsctl
```diff
@@ -380,6 +381,46 @@ class CreateTimelineFromCsv(CreateTimelineBase):
...
+            flush_interval):
+        """Create the timeline from a JSONL file.
+
+        Args:
+            timeline_name: The name of the timeline in Timesketch
+            index_name: Name of the index in Elasticsearch
+            file_path: Path to the file to process
+            event_type: Type of event (e.g. plaso_event)
+            flush_interval: Number of events to queue up before bulk insert
+
```
berggren: -newline

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+    """Generator for reading a JSONL (json lines) file.
+
+    Args:
+        path: Path to the JSONL file
+    """
+    # Fields that must be present in each entry of the JSONL file.
+    mandatory_fields = [u'message', u'datetime', u'timestamp_desc']
+
+    with open(path, 'rb') as fh:
+
```
berggren: -newline

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+                    linedict[u'timestamp'] = parser.parse(linedict[u'datetime'])
+
+                missing_fields = []
+                for field in mandatory_fields:
+                    if field not in linedict.keys():
+                        missing_fields.append(field)
+                if missing_fields:
+                    raise RuntimeError(
+                        u"Missing fields in JSON at line {0:n}: {1:s}"
+                        .format(lineno, missing_fields))
```
berggren: line break at ( instead

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+
+                yield linedict
+
+            except ValueError as e:
+                raise RuntimeError(
+                    u"Error parsing JSON at line {0:n}: {1:s}"
+                    .format(lineno, e))
```
berggren: Line break at (

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+    mandatory_fields = [u'message', u'datetime', u'timestamp_desc']
+
+    with open(path, 'rb') as fh:
+
+        lineno = 0
+        for line in fh:
+            lineno += 1
+            try:
+                linedict = json.loads(line)
+                if u'datetime' not in linedict.keys()\
```
berggren: This logic is repeated in the read_and_vlidate_csv function. Can we refactor and break out the logic to it's own function and use that?

From read_and_vlidate_csv():

```
        for row in reader:
                try:
                    parsed_datetime = parser.parse(row[u'datetime'])
                    row[u'timestamp'] = str(
                        int(time.mktime(parsed_datetime.timetuple())))
                except ValueError:
                    continue
```

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+                  and u'timestamp' in linedict.keys():
+                    epoch = int(str(linedict[u'timestamp'])[:10])
```
berggren: If you use the imported dateutil parser you can keep the precision. Also, see the refactor suggestion above.

timesketch/lib/utils.py
```diff
@@ -68,6 +70,48 @@ def read_and_validate_csv(path):
...
+                    dt = datetime.datetime.fromtimestamp(epoch)
+                    linedict[u'datetime'] = dt.isoformat()
+                if u'timestamp' not in linedict.keys()\
```
berggren: I'm not a fan of "\" for line breaks. Can you refactor to get rid of them?

timesketch/lib/utils.py
adamjnichols: So, these similar, but actually different.  The current implementation of plaso json_lines output provides a timestamp, but not a datetime.  So we're working the other way here.  I've chatted with joachimmetz@ about adding it into that project, but for now we're turning a microsecond precision epoch timestamp into something ISO-formatted for use here.  Ultimately, because that's rather brittle, I opted to slice the first 10 intead of assume precision.  

Have I missed something?

timesketch/lib/utils.py
adamjnichols: Please see comment above.

timesketch/lib/utils.py
berggren: Ok, yeah it makes sense. Thanks for the explanation. You can proceed and discard this suggestion.

timesketch/lib/utils.py
berggren: Ack. SG

timesketch/lib/utils.py
```diff
@@ -68,6 +70,46 @@ def read_and_validate_csv(path):
...
+                if u'timestamp' not in ld_keys and u'datetime' in ld_keys:
+                    linedict[u'timestamp'] = parser.parse(linedict[u'datetime'])
+
+                missing_fields = []
+                for field in mandatory_fields:
+                    if field not in linedict.keys():
+                        missing_fields.append(field)
+                if missing_fields:
+                    raise RuntimeError
+                        (u"Missing fields in JSON at line {0:n}: {1:s}"
```
berggren: This errors. Move the paranthesis up and break after ".format(" instead

berggren: See https://travis-ci.org/google/timesketch/builds/326452079?utm_source=github_status&utm_medium=notification

adamjnichols: Yeah, didn't actually want it to run CI yet :P

berggren:
Approving this, the failing test is due to PyPi issues (timeout). PR travis run succeed.

PR #520, opened 2018-01-11
Docker overhaul/update

adamjnichols:
- All containers now run most current version of all components
- Frontend yarn build is now done automatically
- Upload capability is configured automatically
- Celery is started automatically
- chmod of /etc/timesketch.conf has been removed from entrypoint to allow celery to run as uid:nobody
- Plaso, plaso-tools are installed and functional
- Neo4j container added and integrated
- Redis container reconfigured and integrated
- pip is upgraded before use
- Groovy/painless scripts for annotations are mounted to ES container via Docker volume (less complex than a Dockerfile specific to that container)
- All datastores in all containers are mounted as Docker volumes from the host to persist data
- README.md has been updated

NOTE: Depends on #519


docker/Dockerfile
```diff
@@ -2,26 +2,38 @@
...
+RUN apt-get -y install apt-transport-https\
+                       curl\
+                       git\
+                       libffi-dev\
+                       lsb-release\
+                       python-dev\
+                       python-pip\
+                       python-psycopg2
+
+RUN curl -sS https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
```
berggren: Docker is for deployment, not development. We don't want/need yarn and nodeJS. THe bundle is part of the release to PyPi.

docker/Dockerfile
```diff
@@ -2,26 +2,38 @@
...
-RUN apt-get -y install python-plaso
+RUN add-apt-repository ppa:gift/stable && apt-get update
+RUN apt-get update && apt-get -y install python-plaso plaso-tools nodejs yarn
 
 # Use pip to install Timesketch
+RUN pip install --upgrade pip
 ADD . /tmp/timesketch
-RUN pip install /tmp/timesketch
+RUN cd /tmp/timesketch && yarn install && yarn run build
+RUN pip install /tmp/timesketch/
```
berggren: Seeing that docker is for deployment, we probably want to pip install timesketch, right?

docker/Dockerfile
berggren: Ok, after discussing this offline I can see the point of the docker stuff to run on HEAD. Future release changes (monthly) will address this and we can change then.

docker/docker-entrypoint.sh
```diff
@@ -2,49 +2,67 @@
...
+    TIMESKETCH_PASSWORD="$(openssl rand -base64 32)"
+    echo "TIMESKETCH_PASSWORD set randomly to: ${TIMESKETCH_PASSWORD}";
+  fi
+
+  # Sleep to allow the other processes to start
+  sleep 5
+  tsctl add_user -u "$TIMESKETCH_USER" -p "$TIMESKETCH_PASSWORD"
+
+  # Run the Timesketch server (without SSL)
+  exec `bash -c "/usr/local/bin/celery -A timesketch.lib.tasks worker --uid nobody --loglevel info & /usr/local/bin/tsctl runserver -h 0.0.0.0 -p 5000"`
```
berggren: Note: We should not run the development server here. Let's use gunicorn/wsgi instead. The development server is slow.

docker/docker-entrypoint.sh
```diff
@@ -2,49 +2,67 @@
...
+    TIMESKETCH_USER="admin"
+    echo "TIMESKETCH_USER set to default: ${TIMESKETCH_USER}";
+  fi
+  if [ -z ${TIMESKETCH_PASSWORD+x} ]; then
+    TIMESKETCH_PASSWORD="$(openssl rand -base64 32)"
+    echo "TIMESKETCH_PASSWORD set randomly to: ${TIMESKETCH_PASSWORD}";
+  fi
+
+  # Sleep to allow the other processes to start
+  sleep 5
```
berggren: This seems brittle. Any other way to check status of other processes?

docker/docker-entrypoint.sh
berggren: Example
```
gunicorn -b 127.0.0.1:80 --log-file - --timeout 120 timesketch.wsgi:application
```

docker/Dockerfile
berggren: See above comment. Fine to leave this as is.

docker/docker-entrypoint.sh
adamjnichols: There isn't, no.  Race conditions between containers are well known.  Some folks write very complex logic to determine the listening status of other containers in a compose, but that seems unnecessary here.  It's only cosmetic, and in place to ensure errors aren't displayed to users that are benign and would self-resolve once systems are online.  Functionally, it's unnecessary.

docker/docker-entrypoint.sh
```diff
@@ -62,7 +62,8 @@ if [ "$1" = 'timesketch' ]; then
   tsctl add_user -u "$TIMESKETCH_USER" -p "$TIMESKETCH_PASSWORD"
 
   # Run the Timesketch server (without SSL)
-  exec `bash -c "/usr/local/bin/celery -A timesketch.lib.tasks worker --uid nobody --loglevel info & /usr/local/bin/tsctl runserver -h 0.0.0.0 -p 5000"`
+  #exec `bash -c "/usr/local/bin/celery -A timesketch.lib.tasks worker --uid nobody --loglevel info & /usr/local/bin/tsctl runserver -h 0.0.0.0 -p 5000"`
+  exec `bash -c "gunicorn -b 127.0.0.1:80 --log-file - --timeout 120 timesketch.wsgi:application & /usr/local/bin/tsctl runserver -h 0.0.0.0 -p 5000"`
```
berggren: You are running 2 webservers here, this should be something like:
```
exec `bash -c "/usr/local/bin/celery -A timesketch.lib.tasks worker --uid nobody --loglevel info & gunicorn -b 127.0.0.1:80 --log-file - --timeout 120 timesketch.wsgi:application"`
```

```tsctl runsever``` is only for development

adamjnichols: Sorry, I should have mentioned this was not intended to be merged yet.  Playing with it a bit, and I tend to commit more often than I should.  Sorry. :(

berggren: Heh, no worries :) Let me know when you are ready.

berggren:
@adamjnichols are we ready to merge this PR?

adamjnichols:
This is ok to merge when you get a chance

berggren:
CLA warning is not true. Bug due to merge

PR #529, opened 2018-01-20
Mention the disksize plugin

deralexxx:
Without it, you will end up:

There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /Users/foo/timesketch/vagrant/Vagrantfile
Line number: 8
Message: RuntimeError: vagrant-disksize is not installed ("vagrant plugin install vagrant-disksize")

Not a big deal and it will tell you what to do anyway, but mentioning it lowers the bar to use it I hope.


deralexxx:
I signed it!

PR #530, opened 2018-01-22
Surface celery task errors

berggren:
Timesketch uses Celery to handle upload request and to ingest timelines. When a background task fails it is currently invisible to the user.

This PR fixes this. It will now:
* Set status on both searchindex and timeline to "fail" if anything goes wrong.
* Cleanup Elasticsearch when fail (i.e. remove broken index)
* Display error message and status in the UI
* Filter out broken timelines from explore view

screen shot 2018-01-22 at 15 07 25

### DIFF
--- a/timesketch/lib/datastores/elastic.py
+++ b/timesketch/lib/datastores/elastic.py
@@ -452,6 +452,18 @@ class ElasticsearchDataStore(datastore.DataStore):
         doc_type = unicode(doc_type.decode(encoding=u'utf-8'))
         return index_name, doc_type
 
+    def delete_index(self, index_name):
+        """Delete Elasticsearch index.
+
+        Args:
+            index_name: Name of the index to delete.
+        """
+        if self.client.indices.exists(index_name):
+            try:
+                self.client.indices.delete(index=index_name)
+            except ConnectionError:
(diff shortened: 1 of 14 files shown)


timesketch/lib/datastores/elastic.py
```diff
@@ -452,6 +452,18 @@ def create_index(self, index_name=uuid4().hex, doc_type=u'generic_event'):
...
+        """Delete Elasticsearch index.
+
+        Args:
+            index_name: Name of the index to delete.
+        """
+        if self.client.indices.exists(index_name):
+            try:
+                self.client.indices.delete(index=index_name)
+            except ConnectionError:
+                raise RuntimeError(u'Unable to connect to Timesketch backend.')
```
adamjnichols: Would it make sense to add specifics to the backend failing here?  With so much infrastructure, it might make it easier to troubleshoot.

berggren: Yeah, this needs some more work overall, but I fixed this one. The log will show more details anyways but I included the specific connection error here.

berggren:
CLA warning is not true, it got confused with recent rebasing,.

PR #534, opened 2018-01-26
Refactor CSV and JSONL importers and add error handling

berggren:
If uploaded file is malformed we report error back to the user and set the timeline and index as failed.

### DIFF
--- a/timesketch/lib/tasks.py
+++ b/timesketch/lib/tasks.py
@@ -15,6 +15,7 @@
 
 import logging
 import subprocess
+import traceback
 
 from flask import current_app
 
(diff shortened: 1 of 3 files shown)


timesketch/lib/tasks.py
```diff
@@ -96,76 +103,55 @@ def run_plaso(source_file_path, timeline_name, index_name, username=None):
...
-    total_events = es.import_event(index_name, event_type)
+    # Reason for the broad exception catch is that we want to capture
+    # all possible errors and exit the task.
+    try:
+        es.create_index(index_name=index_name, doc_type=event_type)
+        for event in read_and_validate(source_file_path):
+            es.import_event(index_name, event_type, event)
+        # Import the remaining events
+        total_events = es.import_event(index_name, event_type)
+    except Exception as e:
```
rbdebeer: todo (for future release): catch specific errors and handle.

rbdebeer:
LGTM.

I see some jenkins build errors around pylint and timesketch not being able to be called, assume this is a config issue of some sort?

timesketch/lib/tasks.py
berggren: Reason for the broad exception catch is that we want to capture
 -    for event in read_and_validate_csv(source_file_path):		 +    # all possible errors and exit the task.

berggren:
Lint error in tasks.py (trailing newline). Fixed now.

PR #539, opened 2018-02-01
Add support for CSV files using non-standard delimiters

### LINKED ISSUE #337: tsctl to support TSV files
`tsctl` currently supports reading CSV files, however, often times its easier/less likely data gets broken by using _tab_ delimited files.

This is trivial to implement, e.g - :
1.  add the option `tsv2ts` to [`shell_manager`](https://github.com/hiddenillusion/timesketch/blob/master/tsctl#L424)
2. in `CreateTimelineFromCsv()`have an option for _delimiter_ and if `tsv2ts` was provided, change the CSV [`reader`](https://github.com/hiddenillusion/timesketch/blob/master/tsctl#L412) delimiter to a tab.

adamjnichols:
Tested to work with tab, pipe, hash, and others.

One thing about this that I *really* don't love is that the current class structure prevents the use of Option() within only the CSV object, meaning that you are able to pass an (ignored) option for --delimiter into jsonl2ts.  It's not a huge issue but refactoring to correct it is.

Fixes #337


berggren:
CLA error is false. IT's a bug when you hit update branch button.

Issue #531, opened 2018-01-23
Vagrant install is broken due to polkit


adamjnichols:
Celery doesn't want to enable/start due to polkit policy issues.

berggren:
Can you add the error you get when starting celery?

adamjnichols:
Thinking about closing this, based on some of the things I've found.  Polkit isn't the issue as much as some of the permissions changes that removed the ubuntu user as the default for canonical cloud images earlier this month.  The polkit thing was just a symptom.

At any rate, I believe we fixed it with a merge some time ago, will dig that out, link, and close.

adamjnichols:
Should be corrected with last pagination [commit](https://github.com/google/timesketch/compare/ajn-paginate-the-things#diff-3a627674bc62724e5f508803f8ac3e2b)

PR #541, opened 2018-02-13
Event pagination

adamjnichols:
Provides pagination functionality for 10k events.  ES limitations prevent deep paging using native ES filter DSL, requiring us to explore alternatives longer term.  Some changes in this PR

- API changes to accept ES "size" and "from" parameters
- Suppression of these parameters for aggregation route as they are incompatible with groovy scripts
- Limit now defaults to match "size"
- UI changes to accommodate pager functionality
- Minor tweak to correct "Toggle All" button annoyance
- Minor vagrant changes


timesketch/lib/datastores/elastic.py
```diff
@@ -264,9 +271,9 @@ def search(self,
         # Default search type for elasticsearch is query_then_fetch.
         search_type = u'query_then_fetch'
 
-        # Set limit to 0 to not return any results
-        if not return_results:
-            limit_results = 0
+        # # Set limit to 0 to not return any results
```
berggren: Is this suppose to be commented out? In that case, please remove.

berggren:
When there is only one page of results, the bas still show up. Can we hide it when there is nothing to paginate?

screen shot 2018-02-14 at 19 24 26

adamjnichols:
So, I actually played with this before, and ultimately decided that it seemed more flaky than useful.  There seemed to be more value in having the expected component be there reassuring you that it's working correctly, compared to hiding it and making you wonder if it's broken or not.

It's easy enough to do, though.  Think about it and lmk if you still want the change and I'll shoot a quick commit.

berggren:
That makes sense. Let's keep it as is for now.

vagrant/bootstrap.sh
```diff
@@ -126,7 +126,7 @@ echo "dbms.shell.enabled=true" >> /etc/neo4j/neo4j.conf
 
 if [ "$VAGRANT" = true ]; then
   # Expose Neo4j to the host, for development purposes
-  echo "dbms.connectors.default_listen_address=0.0.0.0" >> /etc/neo4j/neo4j.conf
+  sudo echo "dbms.connectors.default_listen_address=0.0.0.0" >> /etc/neo4j/neo4j.conf
```
berggren: Why sudo here? Vagrant runs this script as root already.

adamjnichols: The script should run as the SSH user, which is now vagrant after some of the early January changes to the ubuntu cloud images.  This sudo makes everything consistent with other commands that might require it now.

berggren: No, the whole script runs with sudo, so this is redundant. See:
https://www.vagrantup.com/docs/provisioning/shell.html#privileged

adamjnichols: Aware, but this is the behavior that led me to file #531.  I don't seem to be able to recreate with newer versions of the cloud image, but the symlinks into systemd were inheriting permissions as vagrant:vagrant, leading to failures to init and run the services.  Prefacing those (and this) command with sudo corrected the issue.  I'm not sure why this PR has that remanent, though, so I've pulled it.

I was not able to repro this morning on 20180215.0.0, so perhaps it was transient, but I'm not making it up, I swear. :)

PR #545, opened 2018-02-17
fix: default Vagrant images are hosted on vagrantcloud

adulau:


adamjnichols:
Hi there!  What is it that you're hoping to fix with this PR?

adulau:
@adamjnichols As vagrant changed the distribution model of the default Vagrant boxes: https://github.com/hashicorp/vagrant/issues/9442 this allows to use the Vagrant installation.

adamjnichols:
I'm aware of that issue, but it doesn't stem from Hashi changing the way they distribute, so much as their acknowledgement of a bug affecting Vagrant installs < 1.9.6 when the metadata URLs were changed.  It seems they pulled redirects too early, and a fix will be forthcoming.  I'm not sure we want to make downstream changes for something we expect they will address in the near-term.  

@berggren thoughts?

adulau:
@adamjnichols I tested with various version of Vagrant including latest, it works with the new distribution path.  FYI, I wanted to be sure that I could build timesketch and MISP (misp-vagrant) together.

adamjnichols:
I'm sure it works great, they seem to list it as a work around.  My concerns have more to do with changing our Vagrantfile to address a problem that only affects a subset of Vagrant installs, and is something Hashi has already accepted as a bug (meaning they intend to fix).

berggren:
Thanks @adulau for filing this PR. I agree with Adam here in that we should not change out Vagrant file to add this workaround. Let's monitor the upstream bug and if they decide not to fix this we can open this PR again.

PR #551, opened 2018-03-06
Strip domain from username if present

berggren:
Quick fix to handle email based usernames in the ACL form. Strip on @ and extract the username.
Future: Move this to the API and present validation errors in the form.


berggren:
@aarontp friendly ping

PR #560, opened 2018-03-16
Update upstream image source in Vagrantfile

adricnet:
This line got me past a blocking error due to the relocation of the base images used by the Vagrant build. It's likely needed now to build on Vagrant less than 2.x. I did not discover the fix and all credit to @Val in this issue thread:
https://github.com/hashicorp/vagrant/issues/9442#issuecomment-363080565 

hth,
adric


adricnet:
CLA signed, "Thanks, your CLA has been submitted."

PR #572, opened 2018-03-23
Show labels for labeled events

tomchop:


berggren:
Looks good. Can you add a screenshot to the description? That way it is easier to see how the change looks like.

berggren:
The CLA error is wrong.

PR #576, opened 2018-03-27
Enable file import/export support for search templates.

aguilajesus:
First iteration to enable support for a search template repository. this change includes the following:

- export search templates into a given file.
   * exported as yaml file.
- import search templates from a given yaml file.
   * it skips the template if the name is already present in the database.

### DIFF
--- a/timesketch/models/__init__.py
+++ b/timesketch/models/__init__.py
@@ -107,15 +107,6 @@ class BaseModel(object):
     updated_at = Column(DateTime(), default=func.now(), onupdate=func.now())
 
     @classmethod
-    def get(cls, **kwargs):
-        """Get a database object.
-
-        Returns:
-            A model instance or None.
-        """
-        return cls.query.filter_by(**kwargs).first()
-
-    @classmethod
     def get_or_create(cls, **kwargs):
         """Get or create a database object.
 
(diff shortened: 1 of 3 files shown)


timesketch/models/__init__.py
```diff
@@ -106,6 +106,15 @@ def __tablename__(self):
     created_at = Column(DateTime(), default=func.now())
     updated_at = Column(DateTime(), default=func.now(), onupdate=func.now())
 
+    @classmethod
+    def get(cls, **kwargs):
```
berggren: I opt for removing this. This is a bit confusing with the build in Class.query.get(id). In the code below, just use the query syntax:
```
SearchTemplate.query.filter_by(name=u'foo').first()
```

tsctl
```diff
@@ -500,6 +503,52 @@ class SimilarityScore(Command):
...
+
+            with open(export_location, 'w') as fh:
+                yaml.safe_dump(search_templates, stream=fh)
+
+        if import_location:
+            try:
+                with open(import_location, 'rb') as fh:
+                    search_templates = yaml.load(fh)
+                    for search_template in search_templates:
+                        if not SearchTemplate.get(name=search_template[u'name']):
```
berggren: Change to:

```
name = search_template[u'name']
query_string = search_template[u'query_string']
query_dsl = search_template[u'query_dsl']

SearchTemplate.query.filter_by(name=name).first()
```

PR #577, opened 2018-03-28
add label support to search templates

aguilajesus:
with his change:
 - added "remote_template" label to differentiate  between local and imported templates.
 - added supported_os label (as seen in ForensicsArtifacts) to all imported templates.

### DIFF
--- a/tsctl
+++ b/tsctl
@@ -16,7 +16,6 @@
 
 import json
 import sys
-import os
 import yaml
 from uuid import uuid4
 
--- a/tsctl
+++ b/tsctl
@@ -521,10 +521,15 @@ class SearchTemplateManager(Command):
         if export_location:
             search_templates = []
             for search_template in SearchTemplate.query.all():
+                labels = []
+                for l in search_template.labels:
+                    if l.label.startswith(u'supported_os:'):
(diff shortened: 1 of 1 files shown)


tsctl
```diff
@@ -521,10 +520,15 @@ class SearchTemplateManager(Command):
         if export_location:
             search_templates = []
             for search_template in SearchTemplate.query.all():
+                labels = []
+                for l in search_template.labels:
```
berggren: use label instead of l, e.g:
```for label in ...```

File: tsctl [hunk 2]
```diff
@@ -534,23 +538,39 @@ class SearchTemplateManager(Command):
...
+                    continue
+
+                imported_template = SearchTemplate(
+                    name=name,
+                    user=User(None),
+                    query_string=query_string,
+                    query_dsl=query_dsl)
+
+                # Add supported_os labels.
+                for platform in search_template[u'supported_os']:
```
berggren: to follow the naming, use:
```for supported_os in ...```

tsctl
aguilajesus: Done.

tsctl
aguilajesus: Done.

PR #584, opened 2018-04-21
Replacing groovy with painless scripts for clean ES 6.0 support

mari0d:


mari0d:
Unintentional, disregard.

PR #588, opened 2018-04-27
Don't trample local OS PIP

adamjnichols:
https://github.com/pypa/pip/issues/5221

### DIFF
--- a/vagrant/bootstrap.sh
+++ b/vagrant/bootstrap.sh
@@ -52,7 +52,7 @@ apt-get install -y \
   python-pip python-dev libffi-dev redis-server python-plaso plaso-tools jq
 
 # Install python dependencies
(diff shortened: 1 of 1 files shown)


vagrant/bootstrap.sh
```diff
@@ -52,7 +52,7 @@ apt-get install -y \
   python-pip python-dev libffi-dev redis-server python-plaso plaso-tools jq
 
 # Install python dependencies
-pip install --upgrade pip
+# pip -v install --upgrade pip  # don't do this https://github.com/pypa/pip/issues/5221
```
berggren: Looks good. just remove the line altogether IMO.

pradyunsg: Passing `--user` as well to pip, is another option here. :)

PR #608, opened 2018-05-17
check status from elastic

deralexxx:
Had some troubles and by debugging turned out my elastic was not running, so I think it is good enough to mention it here to check.

(May 17 03:52:32 systemd[1]: elasticsearch.service: Main process exited, code=exited, status=1/FAILURE
May 17 03:52:32  systemd[1]: elasticsearch.service: Unit entered failed state.
May 17 03:52:32  systemd[1]: elasticsearch.service: Failed with result 'exit-code'.)

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -39,6 +39,9 @@ Make sure that Elasticsearch is started on boot:
     /bin/systemctl daemon-reload
     /bin/systemctl enable elasticsearch.service
     /bin/systemctl start elasticsearch.service
+    
+Make sure that Elasticsearch is running:
+
     /bin/systemctl status elasticsearch.service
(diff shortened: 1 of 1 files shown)


docs/Installation.md
```diff
@@ -39,6 +39,7 @@ Make sure that Elasticsearch is started on boot:
     /bin/systemctl daemon-reload
     /bin/systemctl enable elasticsearch.service
     /bin/systemctl start elasticsearch.service
+    /bin/systemctl status elasticsearch.service
```
berggren: Maybe add something like:

### Make sure Elasticsearch is running correctly
/bin/systemctl status elasticsearch.service

PR #612, opened 2018-05-28
Close DB session in tasks

berggren:
Make sure we close the db session after operating tasks.

### DIFF
--- a/timesketch/lib/tasks.py
+++ b/timesketch/lib/tasks.py
@@ -17,7 +17,6 @@ import logging
 import subprocess
 import traceback
 
-from celery import Task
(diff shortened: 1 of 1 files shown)


timesketch/lib/tasks.py
```diff
@@ -17,6 +17,7 @@
 import subprocess
 import traceback
 
+from celery import Task
```
tomchop: You don't seem to be using this anywhere in the code. Correct?

berggren: Good catch! Removed now.

berggren:
PTAL

PR #629, opened 2018-06-11
Google OpenID Connect authentication support

berggren:
This PR implements the full OAuth2 authentication support for Google OpenID Connect. It uses a normal Oauth2 flow to obtain a authentication code that is used to fetch a signed JWT.

Reference: https://developers.google.com/identity/protocols/OpenIDConnect


timesketch/lib/google_auth.py
```diff
@@ -34,6 +47,135 @@ class JwtKeyError(Exception):
...
+        'code': code,
+        'client_id': current_app.config.get('GOOGLE_OIDC_CLIENT_ID'),
+        'client_secret': current_app.config.get('GOOGLE_OIDC_CLIENT_SECRET'),
+        'redirect_uri': redirect_uri,
+        'grant_type': 'authorization_code'
+    }
+    token_url = discovery_document.get('token_endpoint')
+    try:
+        response = requests.post(token_url, data=post_data)
+        encoded_jwt = response.json().get('id_token')
```
tomchop: Maybe account for the situation where `id_token` is not in the response?

timesketch/lib/google_auth.py
```diff
@@ -164,5 +284,39 @@ def get_public_key_for_jwt(encoded_jwt, url):
...
+        }
+    }
+    get_oauth2_discovery_document.cache = cache
+    return new_discovery_document
+
+
 # Used to cache public keys.
 get_public_key_for_jwt.key_cache = {}
+
+# Used to cache public keys.
```
tomchop: `Used to cache discovery documents`?

timesketch/views/auth.py
```diff
@@ -52,6 +62,10 @@ def login():
         Redirect if authentication is successful or template with context
         otherwise.
     """
+    # Google OpenID Connect authentication.
+    if current_app.config.get(u'GOOGLE_OIDC_ENABLED', False):
```
tomchop: `from __future__ import unicode_literals`?

timesketch/views/auth.py
```diff
@@ -134,3 +148,71 @@ def logout():
...
+
+    validated_email = validated_jwt.get(u'email')
+    user_whitelist = current_app.config.get(u'GOOGLE_OIDC_USER_WHITELIST')
+
+    # Check if the authenticating user is on the whitelist.
+    if user_whitelist:
+        if validated_email not in user_whitelist:
+            return abort(HTTP_STATUS_CODE_UNAUTHORIZED)
+
+    user = User.get_or_create(username=validated_email, name=validated_email)
```
tomchop: One of `username` or `name` should be enough to have a unique user, right?

timesketch/lib/google_auth.py
berggren: Done.

timesketch/lib/google_auth.py
berggren: Done.

timesketch/views/auth.py
berggren: This is going into a bigger refactor later. Will be in a separate PR.

timesketch/views/auth.py
berggren: True, but the way the helper works I need both here so it will be created correctly. Might change this later, but this PR is not the place :)

berggren:
PTAL

timesketch/views/auth.py
tomchop: SG

timesketch/views/auth.py
tomchop: SG

PR #633, opened 2018-06-12
Searchtemplate source model

berggren:
Add source model to searchtemplates in order to track where the template came from and to support sharing of templates in the future.


berggren:
Closing this PR. We will revisit this with a different approach

Issue #640, opened 2018-06-24
Get some redline csv samples


jaegeral:
As discussed it would be handy to have different redline examples for redline export csv data

- a good and easy one
- a very broken one

kiddinn:
I think we can close this one,.

PR #641, opened 2018-06-24
Introduce test data

deralexxx:
https://github.com/google/timesketch/issues/640


berggren:
We are refactoring redline support in the near future. I will add test data later after this is done.

PR #642, opened 2018-06-24
Enable redline CSV support in upload

berggren:
Refactor the CSV importer to support Redline CSV as well.

### DIFF
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -36,87 +36,82 @@ def random_color():
     rgb = tuple(int(i * 256) for i in colorsys.hsv_to_rgb(hue, 0.5, 0.95))
     return u'{0:02X}{1:02X}{2:02X}'.format(rgb[0], rgb[1], rgb[2])
 
+
+def get_csv_dialect(csv_header):
+    """Get CSV dialect format.
+
+    Args:
+        csv_header: List of CSV column names
+
+    Returns:
+        Name of the dialect if known, otherwise None
+    """
+
+    # Check if redline format
+    redline_fields = {
+        u'Alert', u'Tag', u'Timestamp', u'Field', u'Summary'}
+    redline_intersection = set(
+        redline_fields).intersection(set(csv_header))
+    if len(redline_fields) == len(redline_intersection):
+        return u'redline'
+
+    # Check if Timesketch supported format
+    timesketch_fields = {u'message', u'datetime', u'timestamp_desc'}
+    timesketch_intersection = timesketch_fields.intersection(
+        set(csv_header))
+    if len(timesketch_fields) == len(timesketch_intersection):
+        return u'timesketch'
+
+    return None
+
+
 def read_and_validate_csv(path, delimiter):
     """Generator for reading a CSV or TSV file.
 
     Args:
         path: Path to the file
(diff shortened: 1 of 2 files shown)


timesketch/lib/utils.py
```diff
@@ -36,87 +36,82 @@ def random_color():
...
+        redline_fields).intersection(set(csv_header))
+    if len(redline_fields) == len(redline_intersection):
+        return u'redline'
+
+    # Check if Timesketch supported format
+    timesketch_fields = {u'message', u'datetime', u'timestamp_desc'}
+    timesketch_intersection = timesketch_fields.intersection(
+        set(csv_header))
+    if len(timesketch_fields) == len(timesketch_intersection):
+        return u'timesketch'
```
tomchop: Same comment as above

timesketch/lib/utils.py
```diff
@@ -36,87 +36,82 @@ def random_color():
...
     Args:
         path: Path to the file
         delimiter: character used as a field separator
-    """
-    # Columns that must be present in the CSV file
 
-    # check if it is the right redline format
-    mandatory_fields = [u'Alert', u'Tag', u'Timestamp', u'Field', u'Summary']
+    Returns:
+        Generator of event rows
```
tomchop: Would also be useful to know what type `event rows` are.

timesketch/lib/utils.py
```diff
@@ -36,87 +36,82 @@ def random_color():
...
-            row_to_yield["datetime"] = dt_iso_format
-            row_to_yield["timestamp_desc"] = timestamp_desc
-            row_to_yield["alert"] = alert #extra field
-            tags = [tag]
-            row_to_yield["tag"] = tags # extra field
-
-            yield row_to_yield
+        csv_dialect = get_csv_dialect(csv_header)
+
+        if u'redline' in csv_dialect:
```
tomchop: it seems that `get_csv_dialect` returns a string, not an array of strings. I think you want to use `csv_dialect == 'redline'`. Also, this will fail if `csv_dialect` is `None`.

timesketch/lib/utils.py
```diff
@@ -36,87 +36,82 @@ def random_color():
...
+                if u'timestamp' not in csv_header and u'datetime' in csv_header:
+                    try:
+                        parsed_datetime = parser.parse(row[u'datetime'])
+                        row[u'timestamp'] = int(
+                            time.mktime(parsed_datetime.timetuple())) * 1000000
+                    except ValueError:
+                        continue
+                yield row
+        else:
+            raise RuntimeError(u'Unknown CSV format')
```
tomchop: You can turn this into a guard statement before the other checks

```
if csv_dialect is None:
    raise RuntimeError()

if csv_dialect == 'timesketch':
    ...

if csv_dialect == 'redline':
    ...
```

deralexxx: I like the proposal

deralexxx:
@berggren any chance to review the changes from @tomchop

berggren:
This PR got lost in some other updates. This PR is partly obsolete and the validation of CSVs are being reworked. I'm dropping this one as thing are changing soon. Will enable redline uploads soon.

PR #693, opened 2018-11-05
API client methods for comment and label

deralexxx:
I found it way easier to use comment_event and label_event than the current available version of label_events.


api_client/python/timesketch_api_client/client.py
```diff
@@ -460,6 +460,44 @@ def explore(self,
...
+            {"annotation": comment_text,
+             "annotation_type": "comment",
+             "events": {"_id": event_id, "_index": index,
+                        "_type": "generic_event"}}
+        resource_url = u'{0:s}/sketches/{1:d}/event/annotate/'.format(
+            self.api.api_root, self.id)
+        response = self.api.session.post(resource_url, json=form_data)
+        return response.json()
+
+    def label_event(self, event_id, index, label_text):
```
berggren: This method is not necessary IMO, there is already a label_events() method that can add one or more labels.

deralexxx:
I will hope to work on that soon.

api_client/python/timesketch_api_client/client.py
```diff
@@ -460,6 +460,44 @@ def explore(self,
         response = self.api.session.post(resource_url, json=form_data)
         return response.json()
 
+    def comment_event(self, event_id, index, comment_text):
+        """
+        Adds a comment to a single event.
+
+        :param event_id:
```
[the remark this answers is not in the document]
deralexxx: done IMHO

api_client/python/timesketch_api_client/client.py
```diff
@@ -460,6 +460,44 @@ def explore(self,
...
+        :param index:
+        :param comment_text:
+        :return: a json data of the query.
+        """
+        form_data = \
```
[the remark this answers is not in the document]
deralexxx: done

api_client/python/timesketch_api_client/client.py
deralexxx: I see your point, however, the method label_events() needs an array of events, if you only want to label one, this one is way easier to use.

Maybe a way to address my point would be to give a better example of how to use the label_events function with multiple and only one label? Or I could make the label_event a wrapper function?

deralexxx: I removed the label_event() method now in order to be able to make that PR mergeable, what do you think @berggren  ?

kiddinn:
What is the status of this?

deralexxx:
imho should be mergeable, however I did not have success adding single label to an event.

api_client/python/timesketch_api_client/client.py
```diff
@@ -1245,8 +1269,8 @@ def __init__(self, timeline_id, sketch_id, api, name=None,
             timeline_id: The primary key ID of the timeline.
             sketch_id: ID of a sketch.
             api: Instance of a TimesketchApi object.
-            name: Name of the timeline (optional)
-            searchindex: The Elasticsearch index name (optional)
+            name: Name of the timelne (optional)
```
kiddinn: let's not introduce a spelling mistake ;)

api_client/python/timesketch_api_client/client.py
```diff
@@ -1245,8 +1269,8 @@ def __init__(self, timeline_id, sketch_id, api, name=None,
...
+            searchindex: The Elasticsearch index name (optional)i
```
kiddinn: same here

kiddinn: (seems like you cut the "i" from timeline and pasted it at the end of this line

api_client/python/timesketch_api_client/client.py
```diff
@@ -801,6 +801,30 @@ def store_aggregation(
...
+
+        Args:
+            event_id: id of the event
+            index: The Elasticsearch index name
+            comment_text: text to add as a comment
+        Returns:
+             a json data of the query.
+        """
+        form_data = {
+            u'annotation': comment_text,
```
kiddinn: don't have the u'' just use '' (this is py3 code and we've imported unicode_literals from future for py2 compatibility)

deralexxx: should be good now

api_client/python/timesketch_api_client/client.py
deralexxx: ouch, yes, fixed it

api_client/python/timesketch_api_client/client.py
deralexxx: seems like you added two comments to the same issue, both refer to https://github.com/deralexxx/timesketch/blob/patch-9/api_client/python/timesketch_api_client/client.py#L1272 --> fixed

kiddinn:
approved this, and merged... however I think we may possibly want to change this in the future... to make it easier for the end-user to add comments.... 

I'm thinking having an event object, similar to what is exposed in the analyzer interface, but one that uses the API... and it could have

```
__init__(self,  event_id=None, index=None):
```

That is having the event id and index optional, but then there would be few methods:

```
def from_dict
```

This would read the dict that comes back from the explore function, extracting the event id and index from those.

```
def from_data_frame
```

This would take in both a data frame and an index into it, that can be used to extract the index/event_id

```
def from_data_frame_row
```

this would just take in a single row from a data frame (a numpy array) to extract the data...

something like this... then this object could have

```
def add_comment(self, comment_text)
```
and
```
def add_label(self, label)
```

etc, etc.... to make it easier to change that particular event.

PR #704, opened 2018-11-13
Make sure sketch analyzers are run after indexing.

berggren:
Sketch analyzer is not run after indexing. This PR fixes that.

### DIFF
--- a/timesketch/api/v1/resources.py
+++ b/timesketch/api/v1/resources.py
@@ -1000,7 +1000,7 @@ class UploadFileResource(ResourceMixin, Resource):
 
         form = UploadFileForm()
         if form.validate_on_submit() and upload_enabled:
-            sketch_id = form.sketch_id.data
+            sketch_id = form.sketch_id.data or None
             file_storage = form.file.data
             _filename, _extension = os.path.splitext(file_storage.filename)
             file_extension = _extension.lstrip(u'.')
@@ -1046,11 +1046,11 @@ class UploadFileResource(ResourceMixin, Resource):
                 db_session.commit()
 
             # Start Celery pipeline for indexing and analysis.
-            if current_app.config.get(u'ENABLE_INDEX_ANALYZERS'):
-                from timesketch.lib import tasks
-                pipeline = tasks.build_index_pipeline(
-                    file_path, timeline_name, index_name, file_extension)
-                pipeline.apply_async(task_id=index_name)
+            from timesketch.lib import tasks
+            pipeline = tasks.build_index_pipeline(
+                file_path, timeline_name, index_name, file_extension, sketch_id)
+            print pipeline
+            pipeline.apply_async(task_id=index_name)
 
             # Return Timeline if it was created.
             # pylint: disable=no-else-return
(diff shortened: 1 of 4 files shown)


timesketch/api/v1/resources.py
```diff
@@ -1046,11 +1046,11 @@ def post(self):
                 db_session.commit()
 
             # Start Celery pipeline for indexing and analysis.
-            if current_app.config.get(u'ENABLE_INDEX_ANALYZERS'):
-                from timesketch.lib import tasks
-                pipeline = tasks.build_index_pipeline(
-                    file_path, timeline_name, index_name, file_extension)
-                pipeline.apply_async(task_id=index_name)
+            from timesketch.lib import tasks
```
kiddinn: shouldn't the import be at top?

kiddinn: or perhaps have a comment like you've got below to explain that this is to prevent circular imports

timesketch/api/v1/resources.py
```diff
@@ -1373,12 +1373,14 @@ def post(self, sketch_id):
...
 
             # If enabled, run sketch analyzers when timeline is added.
-            if current_app.config.get(u'ENABLE_SKETCH_ANALYZERS'):
-                from timesketch.lib import tasks
-                pipeline = tasks.build_sketch_analysis_pipeline(
-                    sketch_id, searchindex_id)
-                if pipeline:
-                    pipeline.apply_async(task_id=searchindex_id)
+            # Import here to avoid circular imports.
+            from timesketch.lib import tasks
```
kiddinn: so you removed both of the If sentences to check if current app config was set, so aren't you now always importing tasks? and doing that twice now, in line1049 and here

timesketch/lib/tasks.py
```diff
@@ -138,24 +139,36 @@ def build_index_pipeline(file_path, timeline_name, index_name, file_extension):
...
-    if not analyzer_task_group:
+    if not index_analyzer_group and not sketch_analyzer_group:
         return index_task
 
-    return chain(index_task, analyzer_task_group)
+    if sketch_analyzer_group:
+        if not index_analyzer_group:
+            return chain(
+                index_task, run_sketch_init.s(), sketch_analyzer_group)
+        else:
```
kiddinn: this else is not needed...

if not index_....:
  return chain(....

return chain(.....

is sufficient

timesketch/api/v1/resources.py
berggren: Yep, added comment. this is because how Celery works with the factory pattern I'm using.

timesketch/api/v1/resources.py
berggren: Yep, I'm doing the check in tasks.build_sketch_analysis_pipeline instead, returning an empty list of analyzers if they are missing.

timesketch/lib/tasks.py
berggren: Fixed.

berggren:
@kiddinn PTAL

timesketch/api/v1/resources.py
```diff
@@ -1373,12 +1373,15 @@ def post(self, sketch_id):
                 timeline = Timeline.query.get(timeline_id)
 
             # If enabled, run sketch analyzers when timeline is added.
+            # Import here to avoid circular imports.
```
kiddinn: perhaps remove this comment now, since you no longer have that import

PR #709, opened 2018-11-14
Enable similarity UI element

berggren:
the array isn't evaluated to True for the UI to show the similarity layer button. This PR fixes that.
Show experimental UI when ENABLE_EXPERIMENTAL_UI is True in config.

### DIFF
--- a/timesketch.conf
+++ b/timesketch.conf
@@ -7,7 +7,6 @@
 # $ tsctl -c /path/to/this/timesketch.conf runserver
 
 # Show debug information.
-#
 # Note: It is a security risk to have this enabled in production.
 DEBUG = False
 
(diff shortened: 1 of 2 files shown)


timesketch.conf
```diff
@@ -150,14 +144,17 @@ NEO4J_USERNAME = u'neo4j'
...
-# TODO: Add defaults.
+# Configuration.
+
+# What data_types to calculate a similarity score fore.
 SIMILARITY_DATA_TYPES = []
 
 #-------------------------------------------------------------------------------
+# Enable experimental UI features.
+
+ENABLE_EXPERIMENTAL_UI = False
```
berggren: Default is false here. This is WAI

Issue #700, opened 2018-11-07
Allow overriding the displayed "message" field with a custom one


obsidianforensics:
Now that Timesketch has custom analyzers that can add new fields, we should have a new field that when present would be displayed instead of the default message field.

For a simple example, currently a Google search URL is displayed like:

> [Last Visited Time] https://www.google.com/search?q=best+ketchup&ie=utf-8&oe=utf-8&client=firefox-ab (best ketchup - Google Search) [count: 1] Host: www.google.com (URL not typed directly) Transition: TYPED

We could instead use an analyzer to parse this entry into something more human-readable and display that instead in the Timesketch UI, like:

> Search for "best ketchup"

The original, more verbose message would still be available for searching and viewing when the event is expanded. This search example is a trivial one, but this same idea could be much more valuable with more complex analyzers or on events with verbose messages (like #371). 

The new field could be named something like "message_friendly", "event_description", "human_readable", or whatever, and Timesketch would display the contents of that field, if it exists for the event, and if not, just display the original message field.

berggren:
Great idea! Should be pretty straight forward to implement. 2 parts:
1. Backend analyzer code to format the message and update the event.
2. Frontend, extend the event-list component to display the field if it exist.

Also, searching for "best ketchup" should be everyones daily routine. ;)

deralexxx:
Uh I like that a lot. One other Use case could be a extract google maps GPS coordinates and do a reverse address lookup.

kiddinn:
I added a browser search "extraction" like the one that is described in the original FR (#733)

This is just one example of what can be done with extracting meaning from browser history, in this case I add a tag "browser_search" and an attribute which can be searched, as well as changing using the human_readable field to change what is displayed in the UI, adding an emoji for visual "spotting" of the search history and make the search query more visible

PR #733, opened 2018-12-13
Added a browser search sketch analyzer

kiddinn:
The analyzer inspects all browser history to extract search queries from the URL and making them more visible in the UI as well as adding tags to make it easier to search for browser history events.


kiddinn:
I signed it

kiddinn:
part of #700

timesketch/lib/analyzers/browser_search.py
```diff
@@ -0,0 +1,219 @@
...
+                    search_query = callback_method(url, parameter)
+                else:
+                    search_query = callback_method(url)
+
+                if not search_query:
+                    continue
+
+                event.add_attributes(
+                    {'search_string': search_query,
+                     'human_readable': '[{0:s}] Search: {1:s} - {2:s}'.format(
```
berggren: You can replace the human readable with:
`event.set_human_readable(<human_readable_string>)`

You don't need to have the original message in the string IMO, the attribute is still there if anyone want to see it.

Maybe something like:
`human_readable = '{0:s} search: {1:s}'.format(engine, search_query)`

timesketch/lib/analyzers/browser_search_test.py
```diff
@@ -0,0 +1,76 @@
...
+            'ret%20stuff&ln=en')
+        google_result = analyzer._extract_search_query_from_url(
+            google_search, 'q')
+        self.assertEqual(google_result, 'my secret stuff')
+        google_search = 'https://google.com/search?q=my%20secret%20stuff&ln=en'
+        google_result = analyzer._extract_search_query_from_url(
+            google_search, 'q')
+        self.assertEqual(google_result, 'my secret stuff')
+
+
```
berggren: remove 1 line

timesketch/lib/analyzers/browser_search.py
kiddinn: ack

timesketch/lib/analyzers/browser_search.py
```diff
@@ -202,9 +202,10 @@ def run(self):
                     continue
 
                 event.add_attributes(
-                    {'search_string': search_query,
-                     'human_readable': '[{0:s}] Search: {1:s} - {2:s}'.format(
-                         engine, search_query, message)})
+                    {'search_string': search_query})
```
berggren: No need for the newline now when the line is <80

timesketch/lib/analyzers/browser_search.py
berggren: As discussed offline, OK to keep the message in the human_readable

timesketch/lib/analyzers/browser_search.py
kiddinn: but of course, fixed

kiddinn:
pushed that single change you asked for

Issue #740, opened 2018-12-17
Move all emoji definitions to a separate library


kiddinn:
Instead of defining emojis inside each analyzer, with the Unicode code behind them, have them all in a central library that can be referenced in each analyzer. That way analyzer can more easily re-use already defined emojis

PR #741, opened 2018-12-17
Moved emoji definitions to a library using shared constants

kiddinn:
See issue #740


kiddinn:
already defiend in #742

PR #749, opened 2018-12-20
Sketch analyzers chained and human_readable set to a list.

kiddinn:
This PR has few changes:

- Celery now runs analyzers in a chain instead of a group.
   -  If more than a single sketch analyzer is working on the same data (eg browser related artifacts) attributes and other changes may be overwritten by one of the analyzers 
   - That is each analyzer buffers up changes to ES and therefore has a different state of the index than others, leading to analyzers overwriting one another
- Human readable is now a list instead of a string, and set_human_readable therefore changed into add_human_readable
   - Making it easier for multiple analyzers to add to a human_readable string for the same events
   - This also makes it easy to remove dedups from human_readable (e.g if the same timeline is added into more than one sketch, it will have dedups of human_readable)
- Minor visibility changes to similarity score to make it easier to run from other analyzers


kiddinn:
After this PR has been pushed, I'll make some changes to the UI to reflect human_readable now being a list instead of a string.

PR #762, opened 2019-01-02
Changed the emoji library to include helper text

kiddinn:
Adding more structure to the emoji lib, adding helper text.

This is the first step into making the emoji library more integrated into the UI. Next steps would be to add a helper text as soon as you hover over an emoji in the UI and adding an easier filter in the UI to filter for emojis.


kiddinn:
done, tests added.

PR #765, opened 2019-01-03
Only create one view

### LINKED ISSUE #747: add_view in analyzers interface needs to be modified to prevent duplicates
If an analyzer adds a view and it is run on more than a single timeline, it will create multiple views that are the exact copy of one another.

The add_view function in the analyzer interface needs to be changed so that it first checks to see if the view has already been added before it is added to the sketch.

berggren:
Only create one view per analyzer by reusing existing view if present.
Fixes #747

### DIFF
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -198,12 +198,13 @@ class Sketch(object):
         if not self.sql_sketch:
             raise RuntimeError('No such sketch')
 
-    def add_view(self, name, query_string=None, query_dsl=None,
-                 query_filter=None):
+    def add_view(self, view_name, analyzer_name, query_string=None,
+                 query_dsl=None, query_filter=None):
         """Add saved view to the Sketch.
 
         Args:
-            name: The name of the view.
+            view_name: The name of the view.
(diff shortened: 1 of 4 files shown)


timesketch/lib/analyzers/interface.py
```diff
@@ -219,11 +220,13 @@ def add_view(self, name, query_string=None, query_dsl=None,
...
-                    query_dsl=query_dsl, searchtemplate=None)
-
+        name = '[{0:s}] {1:s}'.format(analyzer_name, view_name)
+        view = View.get_or_create(name=name, sketch=self.sql_sketch, user=None)
+        view.query_string = query_string
         view.query_filter = view.validate_filter(query_filter)
+        view.query_dsl = query_dsl
+        view.searchtemplate = None
+
         db_session.add(view)
```
kiddinn: since you are using the get_or_create function of the BaseModel, which does the db_session.add/db_session.commit if there is no instance, is thsi needed?

berggren: Yeah, we still need it because the DB object (view) is modified after the get_or_create call. so we need to commit it to the DB with the updated data.

PR #766, opened 2019-01-03
Simple domain analyzer

berggren:
Simple analyzer that counts how often a domain occurs in a timeline.


kiddinn:
let's first review #767 and then we can make minor changes to this PR

timesketch/lib/analyzers/domain.py
```diff
@@ -0,0 +1,102 @@
...
+
+        Returns:
+            String with summary of the analyzer result
+        """
+        query = (
+            '{"query": { "bool": { "should": [ '
+            '{ "exists" : { "field" : "url" }}, '
+            '{ "exists" : { "field" : "domain" }} ] } } }')
+
+        return_fields = ['domain', 'url', 'message', 'human_readable']
```
kiddinn: in this case, no need for the message nor the human_readable

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -40,14 +40,15 @@ def __init__(self, index_name, sketch_id):
...
-        super(DomainsSketchPlugin, self).__init__(index_name, sketch_id)
+        super(PhishyDomainsSketchPlugin, self).__init__(index_name, sketch_id)
 
         self.domain_scoring_threshold = current_app.config.get(
             'DOMAIN_ANALYZER_WATCHED_DOMAINS_SCORE_THRESHOLD', 0.75)
         self.domain_scoring_whitelist = current_app.config.get(
             'DOMAIN_ANALYZER_WHITELISTED_DOMAINS', [])
 
-    def _get_minhash_from_domain(self, domain):
+    @staticmethod
```
kiddinn: why are you making this a staticmethod???

kiddinn: and not just have a self, domain there like it was.... don't see the need for that

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -144,7 +145,8 @@ def _get_similar_domains(self, domain, domain_dict):
 
         return similar
 
-    def _get_tld(self, domain):
+    @staticmethod
```
kiddinn: same here... why the static method?

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -19,10 +19,10 @@
 from timesketch.lib.analyzers import manager
 
 
-class DomainsSketchPlugin(interface.BaseSketchAnalyzer):
-    """Sketch analyzer for domains."""
+class PhishyDomainsSketchPlugin(interface.BaseSketchAnalyzer):
+    """Sketch analyzer for phishy domains."""
 
-    NAME = 'domains'
+    NAME = 'phishy_domains'
```
kiddinn: once #767  is in, please add the DEPENDENCIES to depend on the domain sketch analyzer

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -156,7 +158,8 @@ def _get_tld(self, domain):
         """
         return '.'.join(domain.split('.')[-2:])
 
-    def _strip_www(self, domain):
+    @staticmethod
```
kiddinn: again... same here

timesketch/lib/analyzers/phishy_domains.py
berggren: heh, as discussed offline. Making it static is best practise when you don't need a reference to the object it self. That will reduce the amount of objects created. but this is moot now, because these methods are now refactored to utils.py instead.

timesketch/lib/analyzers/phishy_domains.py
berggren: See previous comment

timesketch/lib/analyzers/phishy_domains.py
berggren: Done in the latest refactor

timesketch/lib/analyzers/phishy_domains.py
berggren: See previous reply

berggren:
@kiddinn PTAL

berggren:
screenshot 2019-01-04 at 13 23 30

timesketch/lib/analyzers/domain.py
berggren: good catch, fixed.

timesketch/lib/analyzers/phishy_domains.py
berggren: Done in the latest refactor

timesketch/lib/analyzers/utils_test.py
```diff
@@ -0,0 +1,56 @@
...
+
+from timesketch.lib.testlib import BaseTest
+from timesketch.lib.analyzers import utils
+
+
+class TestAnalyzerUtils(BaseTest):
+    """Tests the functionality of the utilities."""
+
+    def __init__(self, *args, **kwargs):
+        super(TestAnalyzerUtils, self).__init__(*args, **kwargs)
```
kiddinn: why even define __init__ when you are just calling super there....

timesketch/lib/emojis.py
```diff
@@ -33,6 +33,7 @@
     'SKULL_CROSSBONE': emoji('&#x2620', 'Suspicious entry'),
     'UNLOCK': emoji('&#x1F513', 'Logoff activity'),
     'WASTEBASKET': emoji('&#x1F5D1', 'Deletion activity'),
+    'FISHING_POLE': emoji('&#x1F3A3', 'Phishing')
```
kiddinn: maybe add this in an alpha order

timesketch/lib/analyzers/utils_test.py
berggren: Ah, this is the scaffolder that adds this. Also I want to run the init on the BaseTestClass here.

timesketch/lib/emojis.py
berggren: Done.

PR #770, opened 2019-01-07
New importer in tsctl

berggren:
This PR removes some old code from tsctl and add support for importing timelines using the Celery workers instead. This enables analyzers to run as expected.


tsctl
```diff
@@ -27,21 +29,18 @@ from flask_script import Server
 from flask_script import Option
 from flask_script import prompt_bool
 from flask_script import prompt_pass
-
 from sqlalchemy.exc import IntegrityError
+from werkzeug.exceptions import Forbidden
```
kiddinn: isn't this a new dependency?

tsctl
```diff
@@ -134,7 +126,7 @@ class GroupManager(Command):
         super(GroupManager, self).__init__()
 
     # pylint: disable=arguments-differ, method-hidden
-    def run(self, add, remove, group_name, user_name):
+    def run(self, remove, group_name, user_name):
         """Add the user to the group."""
```
kiddinn: maybe add a docstring here

tsctl
```diff
@@ -616,28 +309,136 @@ class SearchTemplateManager(Command):
...
+        Option('--timeline_name', '-n', dest='timeline_name',
+               required=False),
+    )
+
+    def __init__(self):
+        super(ImportTimeline, self).__init__()
+
+    # pylint: disable=arguments-differ, method-hidden
+    def run(self, file_path, sketch_id, username, timeline_name):
+        """This is the run method."""
```
kiddinn: maybe add a docstring here

tsctl
```diff
@@ -616,28 +309,136 @@ class SearchTemplateManager(Command):
...
+
+        file_path = os.path.realpath(file_path)
+        file_path_no_extension, extension = os.path.splitext(file_path)
+        extension = extension.lstrip('.')
+        filename = os.path.basename(file_path_no_extension)
+
+        if not os.path.isfile(file_path):
+            sys.exit('No such file: {0:s}'.format(file_path))
+
+        if extension not in ('plaso', 'csv', 'jsonl'):
```
kiddinn: what about .json instead of just .jsonl?

tsctl
```diff
@@ -616,28 +309,136 @@ class SearchTemplateManager(Command):
...
+            sys.exit('Unknown extension: {0:s}'.format(file_path))
```
kiddinn: can we put the supported extensions in a variable/list and then print them out here... extension FOO is not in the approved list of LIST, or something like that?

tsctl
```diff
@@ -616,28 +309,136 @@ class SearchTemplateManager(Command):
...
+        if sketch_id:
+            try:
+                sketch = Sketch.query.get_with_acl(sketch_id, user=user)
+            except Forbidden:
+                pass
+
+        if not timeline_name:
+            timeline_name = unicode(filename.replace('_', ' '))
+            # Remove sketch ID if present in the filename.
+            if timeline_name.split()[0].isdigit():
```
kiddinn: can you make this more readable by splitting it into two lines, not have the .split() and [] and isdigit all in one line

tsctl
berggren: No, this is a requirement from Flask, the web framework we are using. It has always been there.. Just a bit lower level :)

tsctl
berggren: Yeah, tsctl is a weird leftover from early days. These arguments are passed down to the Command class from the flask-script package. Flaks-script have actually been deprecated so we need to port it to flask.cli anyways. This will be done in a followup PR though as this is a bit involved.

tsctl
berggren: There used to be a json importer but this PR actually removes that. It is not used and is not very good at big data sets (read everything into memory). jsonl (json-line) is the actual json importer which solves this by having every json document on its own row. So, I rather keep it as jsonl to make it explicit what the format is.

tsctl
berggren: Done.

tsctl
berggren: Done.

berggren:
@kiddinn PTAL

PR #771, opened 2019-01-09
Adding the ability to request return fields in the API.

kiddinn:
This way in the API you can specify which fields you want to get back from the search query.


timesketch/api/v1/resources.py
```diff
@@ -596,6 +596,13 @@ def post(self, sketch_id):
         if form.validate_on_submit():
             query_dsl = form.dsl.data
             query_filter = form.filter.data
+            return_fields = form.fields.data
+
+            print 'FORM DATA IS: {}'.format(return_fields)
```
berggren: Remove this print

PR #773, opened 2019-01-09
Send email notification when import is done

berggren:
This PR will send an email whenever a timeline has been indexed and analyzed.
* New configuration added to timesketch.conf
* Only sends emails to whitelisted users.
* Will summarise analysis results.

### DIFF
--- a/timesketch/lib/utils.py
+++ b/timesketch/lib/utils.py
@@ -190,20 +190,39 @@ def get_validated_indices(indices, sketch_indices):
 
 
 def send_email(subject, body, to_username, html=False):
+    """Send email using configure SMTP server.
+
+    Args:
+        subject: Email subject string.
+        body: Email message body.
+        to_username: User to send email to.
+        html: Boolean indicating if the email body should be sent as html.
+
+    Raises:
+        RuntimeError if not properly configured or if the recipient user is no
+        in the whitelist.
+    """
     email_enabled = current_app.config.get('ENABLE_EMAIL_NOTIFICATIONS')
     email_domain = current_app.config.get('EMAIL_DOMAIN')
     email_smtp_server = current_app.config.get('EMAIL_SMTP_SERVER')
+    email_from_user = current_app.config.get('EMAIL_FROM_ADDRESS')
+    email_user_whitelist = current_app.config.get('EMAIL_USER_WHITELIST', [])
 
     if not email_enabled:
-        raise RuntimeError('EMAIL NOTIFICATIONS are not enabled, aborting.')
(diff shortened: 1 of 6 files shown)


timesketch/lib/utils.py
```diff
@@ -181,3 +187,55 @@ def get_validated_indices(indices, sketch_indices):
     if exclude:
         indices = [index for index in indices if index not in exclude]
     return indices
+
+
+def send_email(subject, body, to_username, html=False):
```
kiddinn: since the html is a bool, why not have it as use_html=False... makes it more "bool" like

timesketch/lib/utils.py
```diff
@@ -181,3 +187,55 @@ def get_validated_indices(indices, sketch_indices):
...
+
+    if not email_smtp_server:
+        raise RuntimeError('Email SMTP server is not configured, aborting.')
+
+    # Only send mail to whitelisted usernames.
+    if to_username not in email_user_whitelist:
+        raise RuntimeError('User {0:s} is not in the email whitelist, '
+                           'not sending email.'.format(to_username))
+
+    from_address = '{0:s}@{1:s}'.format(email_from_user, email_domain)
```
kiddinn: no checks for the email_from_user variable... should we have a default value here?

email_from_user  = curre.....get('EM....', 'timesketch')

so it becomes timesketch@<DOMAIN> as default... since you do check teh domain and make sure that is set

timesketch/lib/utils.py
```diff
@@ -181,3 +187,55 @@ def get_validated_indices(indices, sketch_indices):
...
+
+    if not email_domain:
+        raise RuntimeError('Email domain is not configured, aborting.')
+
+    if not email_smtp_server:
+        raise RuntimeError('Email SMTP server is not configured, aborting.')
+
+    # Only send mail to whitelisted usernames.
+    if to_username not in email_user_whitelist:
+        raise RuntimeError('User {0:s} is not in the email whitelist, '
```
kiddinn: why raise a RuntimeError if the user is not whitelisted?

Does this crash things on the other end, or?

timesketch/lib/utils.py
berggren: Good point, done.

timesketch/lib/utils.py
berggren: Good catch, fixed.

timesketch/lib/utils.py
berggren: I agree, removed the raising.

berggren:
@kiddinn PTAL

PR #781, opened 2019-01-11
Tag known CDN domains

berggren:
Check each domain against list of known CDNs and tag if match.

screenshot 2019-01-11 at 15 53 34


timesketch/lib/analyzers/domain.py
```diff
@@ -69,16 +70,28 @@ def run(self):
         satellite_emoji = emojis.get_emoji('SATELLITE')
         for domain, count in domain_counter.iteritems():
             emojis_to_add = [satellite_emoji]
+            tags_to_add = []
             text = '{0:s} seen {1:d} times'.format(domain, count)
 
+            is_known_cdn = utils.get_cdn_providers(domain)
```
kiddinn: is_known_cdn sounds like a boolean to me... but it seems to be a list

timesketch/lib/analyzers/domain.py
```diff
@@ -69,16 +70,28 @@ def run(self):
...
+                    cdn_counter[cdn] += 1
+
             for event in domains.get(domain, []):
+                event.add_tags(tags_to_add)
                 event.add_emojis(emojis_to_add)
                 event.add_human_readable(text, self.NAME, append=False)
-                event.add_attributes({'domain_count': count})
+                new_attributes = {'domain_count': count}
+                if is_known_cdn:
+                    new_attributes['cdn_provider'] = ' '.join(is_known_cdn)
```
kiddinn: why not keep it just as a list?

then you can do checks to see if CDN is in that list...

timesketch/lib/analyzers/utils.py
```diff
@@ -20,6 +20,143 @@
...
+    '.cdninstagram.com': 'Facebook',
+    '.fbcdn.net': 'Facebook',
+    '.fastly.net': 'Fastly',
+    '.fastlylb.net': 'Fastly',
+    '.nocookie.net': 'Fastly',
+    '.cdn.gocache.net': 'GoCache',
+    '.doubleclick.net': 'Google',
+    '.googleusercontent.com': 'Google',
+    '.gstatic.com': 'Google',
+    'googlehosted.com': 'Google',
```
kiddinn: there are few of these domains... that is that don't start with a dot, shouldn't they all start with a dot?

in your check function you would then match on 

evilgooglehosted.com

That would be flagged as a CDN network... whereas I think you are just looking for goodie.googlehosted.com

timesketch/lib/analyzers/domain.py
berggren: This is true. I have change this to return a string of the provider name instead. It should always be just one, but if for whatever reason there are multiple returned they will be joined to one string.

timesketch/lib/analyzers/domain.py
berggren: See above. Refactored this in the utility function instead.

timesketch/lib/analyzers/utils.py
berggren: Good catch. Yeah I cleaned up the data and added dots.

timesketch/lib/analyzers/utils.py
```diff
@@ -20,6 +20,143 @@
...
+    '.tl88.net': 'Akamai China CDN',
+    '.gslb.tbcache.com': 'Alimama',
+    '.cloudfront.net': 'Amazon CloudFront',
+    '.aads-cn.net': 'Aryaka',
+    '.aads-cng.net': 'Aryaka',
+    '.aads1.net': 'Aryaka',
+    '.azion.net': 'Azion',
+    '.azioncdn.com': 'Azion',
+    '.azioncdn.net': 'Azion',
+    'bo.lt': 'BO.LT',
```
kiddinn: this is still not with a dot

timesketch/lib/analyzers/utils.py
```diff
@@ -20,6 +20,143 @@
...
+    '.cloudflare.com': 'Cloudflare',
+    '.cotcdn.net': 'Cotendo CDN',
+    '.systemcdn.net': 'Edgecast',
+    '.transactcdn.net': 'Edgecast',
+    '.v1cdn.net': 'Edgecast',
+    '.v2cdn.net': 'Edgecast',
+    '.v3cdn.net': 'Edgecast',
+    '.v4cdn.net': 'Edgecast',
+    '.v5cdn.net': 'Edgecast',
+    'edgecastcdn.net': 'Edgecast',
```
kiddinn: missing a dot

PR #783, opened 2019-01-11
Adding a login analyzer.

kiddinn:
This is the first attempt/version of a login/logoff analyzer.

There are several things that need to be added:

- More Windows EVTX events
- Windows EVT events
- Linux events
- Mac OS X events

This is therefore just the first iteration of this. With more PRs to come to improve it shortly.

### DIFF
--- a/timesketch/lib/analyzers/login.py
+++ b/timesketch/lib/analyzers/login.py
@@ -34,6 +34,7 @@ def parse_evtx_logoff_event(event, string_list, emoji):
     Returns:
         int: 1 if the event got processed correctly, 0 otherwise.
     """
+    emojis_to_add = [emoji]
     if not len(string_list) == 5:
         return 0
 
@@ -46,9 +47,14 @@ def parse_evtx_logoff_event(event, string_list, emoji):
     attributes['logon_type'] = LOGON_TYPES.get(
         logon_type_code, LOGON_TYPES.get(u'0'))
 
+    # Want to add an emoji in case this is a screensaver unlock.
+    if logon_type_code == '7':
+        screen_emoji = emojis.get_emoji('screen')
+        emojis_to_add.append(screen_emoji)
+
     event.add_tags(['logoff-event'])
     event.add_attributes(attributes)
-    event.add_emojis([emoji])
+    event.add_emojis(emojis_to_add)
     return 1
 
 
@@ -65,6 +71,7 @@ def parse_evtx_logon_event(event, string_list, string_parsed, emoji):
         int: 1 if the event got processed correctly, 0 otherwise.
     """
     attributes = {}
+    emojis_to_add = [emoji]
 
     if not string_parsed:
         string_parsed = {}
@@ -78,6 +85,11 @@ def parse_evtx_logon_event(event, string_list, string_parsed, emoji):
     attributes['logon_type'] = LOGON_TYPES.get(
         logon_type_code, LOGON_TYPES.get(u'0'))
 
+    # Want to add an emoji in case this is a screensaver unlock.
+    if logon_type_code == '7':
+        screen_emoji = emojis.get_emoji('screen')
+        emojis_to_add.append(screen_emoji)
+
     attributes['username'] = '{0:s}/{1:s}'.format(
         string_parsed.get('target_user_id', 'N/A'),
         string_parsed.get('target_user_name', 'Unknown'))
@@ -86,7 +98,7 @@ def parse_evtx_logon_event(event, string_list, string_parsed, emoji):
 
     event.add_tags(['logon-event'])
     event.add_attributes(attributes)
-    event.add_emojis([emoji])
+    event.add_emojis(emojis_to_add)
     return 1
 
 
@@ -112,6 +124,7 @@ class LoginSketchPlugin(interface.BaseSketchAnalyzer):
         Returns:
             String with summary of the analyzer result
         """
+        screenshot_emoji = emojis.get_emoji('camera')
         login_emoji = emojis.get_emoji('lock')
         logoff_emoji = emojis.get_emoji('unlock')
         login_counter = 0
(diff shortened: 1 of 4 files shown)


timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+        self.index_name = index_name
+        super(LoginSketchPlugin, self).__init__(index_name, sketch_id)
+
+    def run(self):
+        """Entry point for the analyzer.
+
+        Returns:
+            String with summary of the analyzer result
+        """
+        screenshot_emoji = emojis.get_emoji('camera')
```
berggren: This one is not used anywhere?

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+        login_emoji = emojis.get_emoji('lock')
```
berggren: Shouldn't "login" be unlock and logout be "lock"?

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+        logoff_emoji = emojis.get_emoji('unlock')
+        login_counter = 0
+        logoff_counter = 0
+
+        # TODO: Add EVT lookups, ID 528 for logon and 538, 540 for logoff.
+        # TODO: Add RDP EVT lookups, ID 682 for logon and 683 for logoff.
+        query = (
```
berggren: We probably want to add 4647 here as well (User initiated logoff)

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+                except ValueError:
+                    logging.warning((
+                        'Unable to convert EVTX identifier to an integer, '
+                        'value is {0:s}').format(identifier))
+                    continue
+
+            if identifier == 4624:
+                login_counter += parse_evtx_logon_event(
+                    event, strings, strings_parsed, login_emoji)
+            elif identifier == 4634:
```
berggren: and 4647?

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+    attributes = {}
+    emojis_to_add = [emoji]
+
+    if not string_parsed:
+        string_parsed = {}
+        if not len(string_list) == 23:
+            return 0
+        string_parsed['target_user_id'] = string_list[4]
+        string_parsed['target_user_name'] = string_list[5]
+        string_parsed['hostname'] = string_list[11]
```
berggren: Do we have the src attributes here as well? There is "WorkstationName" and "IpAddress" in 4624 that holds the information where the login came from (localhost or blank for local logins).

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+    u'5': u'WindowsService',
+    u'7': u'UnlockScreen',
+    u'8': u'NetworkCleartext',
+    u'9': u'NewCredentials',
+    u'10': u'RemoteInteractive',
+    u'11': u'CachedInteractive'
+}
+
+
+def parse_evtx_logoff_event(event, string_list, emoji):
```
berggren: Can we refactor this to only do the parsing/extraction in this function and return the "attributes" dict instead of updating the event? Feels like we are doing a little too much here. Same for the parse_evtx_logon_event() method below.

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,178 @@
...
+        screen_emoji = emojis.get_emoji('screen')
+        emojis_to_add.append(screen_emoji)
+
+    event.add_tags(['logoff-event'])
+    event.add_attributes(attributes)
+    event.add_emojis(emojis_to_add)
+    return 1
+
+
+def parse_evtx_logon_event(event, string_list, string_parsed, emoji):
```
berggren: string_parsed is not in the docstring

timesketch/lib/analyzers/login.py
kiddinn: yes it should... good catch... this was an honest mistake on my behalf ;)

timesketch/lib/analyzers/login.py
kiddinn: done

timesketch/lib/analyzers/login.py
kiddinn: I noticed as well....

timesketch/lib/analyzers/login.py
kiddinn: yes... there are... I'll add them in there as well,... wasn't sure if we should be extracting those as well, but I agree, they can be useful

kiddinn:
did some updates, still running few additional tests, with newer EVTX files to make sure things are running correctly before asking for another round of reviews

timesketch/lib/analyzers/login.py
berggren: Cool, yeah I use them when I create the graph stuff.

kiddinn:
ok, ready for another round

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def parse_evtx_logoff_event(string_list):
+    """Parse logoff events and return a dict with attributes.
+
+    Args:
+        string_list: a list of strings extracted from the Event Log.
+
+    Returns:
+        Dict with attributes parsd out of the logoff events.
```
berggren: s/parsd/parsed

timesketch/lib/analyzers/login.py
```diff
@@ -0,0 +1,182 @@
...
+
+def parse_evtx_logon_event(string_list, string_parsed):
+    """Parse logon events and return a count of event processed.
+
+    Args:
+        string_list: a list of strings extracted from the Event Log.
+        string_parsed: a dict with strings extracted from the Event log.
+
+    Returns:
+        Dict with attributes parsd out of the logon events.
```
berggren: s/parsd/parsed

timesketch/lib/analyzers/login.py
berggren: I don't see the WorkstationName and/or IpAddress being extracted. Are they present in the 4624 strings?

PR #784, opened 2019-01-14
tests(travis): remove deprecated option

RDIL:
Removing the line has the same effect it had, and Travis is going to remove the `false` option soon so it is better to get rid of it.


RDIL:
I can sign the CLA when I get home.

RDIL:
recheck googlebot

Issue #736, opened 2018-12-15
Add the option of getting python pandas DataFrame returned back in analyzers/API client


kiddinn:
First step is to determine if getting back DataFrame objects is possible in the analyzers
  + Main problem here is to determine how to mark events/add attributes, etc, from the DataFrame

Adding the option to get back DataFrame in the API Client is at least something that should be an option.

kiddinn:
by adding the _id, _type and _index fields into the DataFrame columns there is an option to create a new Event object inside the analyzers to get back the ability to mark/comment/etc on events.

kiddinn:
this has been added in, with #787

PR #787, opened 2019-01-15
Added pandas support into the API and analyzers interface.

kiddinn:
This is part of #736 , both adding the ability to requests pandas instead of dict from search results in the API client as well as inside the analyzer interface for sketch analyzers.

The PR adds the _id, _type and _index fields to the pandas DataFrame in order to "get back" an event in the interface, with the option of 

```
event_dict = dict(_id=ID, _type=TYPE, _index=INDEX)
event = interface.Event(event_dict, self.datastore)
```

Where the ID, TYPE and INDEX are derived from the pandas DataFrame's columns _id, _type and _index.

### DIFF
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -22,8 +22,8 @@ import yaml
 import pandas
 
 from flask import current_app
-from timesketch.lib.datastores.elastic import ElasticsearchDataStore
 from timesketch.lib import definitions
+from timesketch.lib.datastores.elastic import ElasticsearchDataStore
 from timesketch.models import db_session
 from timesketch.models.sketch import Event as SQLEvent
 from timesketch.models.sketch import Sketch as SQLSketch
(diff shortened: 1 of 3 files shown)


timesketch/lib/analyzers/interface.py
```diff
@@ -432,6 +436,73 @@ def __init__(self, index_name, sketch_id):
         self.sketch = Sketch(sketch_id=sketch_id)
         super(BaseSketchAnalyzer, self).__init__(index_name)
 
+    def event_pandas(
```
berggren: This looks good, but the method looks very similar to event_stream(). Could we refactor this into a parameter on event_stream() instead, with a bool to return pandas or the generator?

something like:
```event_stream(...., return_pandas=False)```

WDYT

kiddinn: spoke offline, mainly due to the fact that event_stream is a generator but pandas is a single object that gets returned, and I don't think we should be mixing those (pylint will be very unhappy)... makes the API confusing

kiddinn: I Also thought about just calling the event_stream and create a list there, that is also an option, but it also has some overhead there, some things are done slightly differently in this function than in the event stream one.

berggren: Ack, I agree.

PR #803, opened 2019-01-23
Analyzer index performance fix

berggren:
With the current implementation the updated event will be re-indexed in ES on every action, e.g. add_tag, add emoji etc. This results in way too many ES index requests and put a lot of strain on the server.

This PR fixes this by only sending the re-index action once for every analyzer.

### DIFF
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -117,6 +117,7 @@ class Event(object):
 
         Args:
             event_dict: (optional) Dictionary with updated event attributes.
+            Defaults to self.updated_event.
         """
         if event_dict:
             event_to_commit = event_dict
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -94,6 +94,8 @@ class Event(object):
         self.datastore = datastore
         self.sketch = sketch
 
+        self.updated_event = {}
+
         try:
             self.event_id = event['_id']
             self.event_type = event['_type']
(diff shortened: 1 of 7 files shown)


timesketch/lib/analyzers/interface.py
```diff
@@ -103,14 +105,33 @@ def __init__(self, event, datastore, sketch=None):
...
+
+        Args:
+            event_dict: (optional) Dictionary with updated event attributes.
+
+        Returns:
+            None if the event dictionary is empty.
+        """
+        event_to_commit = self.updated_event
+        if event_dict:
+            event_to_commit = event_dict
```
kiddinn: well you set the event_to_commit to self.updated_Event and then you overwrite here... maybe better to have:

```
if event_dict:
    event_to_commit  = event_dict
else:
    event_to_commit = self.updated_event
```

timesketch/lib/analyzers/interface.py
```diff
@@ -103,14 +105,33 @@ def __init__(self, event, datastore, sketch=None):
...
+        self.updated_event.update(event)
+
+    def commit(self, event_dict=None):
+        """Commit an event to Elasticsearch.
+
+        Args:
+            event_dict: (optional) Dictionary with updated event attributes.
+
+        Returns:
+            None if the event dictionary is empty.
```
kiddinn: well... technically you never return anything....

timesketch/lib/analyzers/interface.py
```diff
@@ -103,14 +105,33 @@ def __init__(self, event, datastore, sketch=None):
...
         Args:
             event: Dictionary with new or updated values.
         """
+        self.updated_event.update(event)
+
+    def commit(self, event_dict=None):
+        """Commit an event to Elasticsearch.
+
+        Args:
+            event_dict: (optional) Dictionary with updated event attributes.
```
kiddinn: maybe add that if not supplied it will default to self.updated_event

timesketch/lib/analyzers/interface.py
```diff
@@ -136,14 +157,17 @@ def add_label(self, label, toggle=False):
         updated_event = self.datastore.set_label(
             self.index_name, self.event_id, self.event_type, self.sketch.id,
             user_id, label, toggle=toggle, single_update=False)
-        self._update(updated_event)
+        self.commit(updated_event)
```
kiddinn: why have the commit here?

what if the analyzer does add_attributes more than once? wouldn't it be better to just update the `self.updated_event` and commit in the end?

timesketch/lib/analyzers/interface.py
berggren: Done.

timesketch/lib/analyzers/interface.py
berggren: Done.

timesketch/lib/analyzers/interface.py
berggren: I can't update the event, because this is a special scripted update to set the label. It has to go in as it's own re-index event. I call self.commit(event) with the event and commit() will then just add it to it's queue and not touch self.updated_event.

kiddinn: ack

PR #807, opened 2019-01-25
Fixed problem that the celery does not start automatically.

piwikjapan:
Fixed problem that the celery does not start automatically.


vagrant/bootstrap.sh
```diff
@@ -116,6 +116,7 @@ mkdir -p /var/{lib,log,run}/celery
 chown $RUN_AS_USER /var/{lib,log,run}/celery
 cp "${VAGRANT_PATH}"/celery.service /etc/systemd/system/
 cp "${VAGRANT_PATH}"/celery.conf /etc/
+cp "${VAGRANT_PATH}"/tmpfiles.d/celery.conf /usr/lib/tmpfiles.d/
```
piwikjapan: / var/run/celery disappears when ubuntu starts up. And some functions of timesketch will not work. So, remake it.

vagrant/celery.service
```diff
@@ -1,13 +1,14 @@
...
-After=network.target
+After=network.target redis.service
 
 [Service]
 Type=forking
 User=vagrant
 Group=vagrant
 EnvironmentFile=-/etc/celery.conf
 WorkingDirectory=/var/lib/celery
+ExecStartPre=/bin/sleep 30
```
berggren: This seem to be a bit fragile. Is this to wait for Redis to start up?

vagrant/bootstrap.sh
berggren: ack, good catch.

berggren:
ping @piwikjapan can you take a look at the comment?

vagrant/celery.service
piwikjapan: Yes, I am looking for a nice way. It was the result of my Dell E4310.
Do you have any good idea?

piwikjapan:
ping  @berggren Yes, I commented on 4 Mar.

vagrant/celery.service
berggren: Sorry for the delay, but I think we can drop the sleep here seeing we adding the wait for redis.service. WDYT?

piwikjapan: I leave this to you.
There is no problem if you can successfully start the timesketch.

piwikjapan:
We were on vacation set by the Japanese government. Please see inline comment.

berggren:
Thanks for contributing!

PR #811, opened 2019-01-30
Added a browser timeframe analyzer plugin

kiddinn:
Added a timeframe analyzer plugin.

What this analyzer does is the following:

- Extract hour from each browser event
- Aggregate based on the hour, counting each occurrence
- Finding a threshold of what an active hour constitutes.
- Construct a range of hours where the most activity occurs 
- Tag the events that fall outside of that range.

### DIFF
--- a/timesketch/lib/analyzers/browser_timeframe.py
+++ b/timesketch/lib/analyzers/browser_timeframe.py
@@ -82,7 +82,7 @@ def fix_gap_in_list(hour_list):
     return sorted(hours)
 
 
-def get_hours(frame):
+def get_active_hours(frame):
     """Return a list of the hours with the most activity within a frame.
 
     Args:
@@ -158,7 +158,7 @@ class BrowserTimeframeSketchPlugin(interface.BaseSketchAnalyzer):
         data_frame['hour'] = pd.to_numeric(
             data_frame.datetime.dt.strftime('%H'))
 
-        activity_hours = get_hours(data_frame)
+        activity_hours = get_active_hours(data_frame)
         data_frame_outside = data_frame[~data_frame.hour.isin(activity_hours)]
 
         for event in utils.get_events_from_data_frame(
--- a/timesketch/lib/analyzers/browser_timeframe.py
+++ b/timesketch/lib/analyzers/browser_timeframe.py
@@ -1,15 +1,120 @@
 """Sketch analyzer plugin for browser timeframe."""
 from __future__ import unicode_literals
 
-from timesketch.lib import emojis
+import pandas as pd
+
 from timesketch.lib.analyzers import interface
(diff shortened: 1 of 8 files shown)


timesketch/lib/analyzers/browser_timeframe.py
```diff
@@ -0,0 +1,203 @@
...
+import pandas as pd
+
+from timesketch.lib import emojis
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+from timesketch.lib.analyzers import utils
+
+
+def get_runs(hour_list):
+    """Returns a list of runs from a list of numbers.
```
berggren: What is a run? :)

timesketch/lib/analyzers/browser_timeframe.py
```diff
@@ -0,0 +1,203 @@
...
+    def run(self):
+        """Entry point for the analyzer.
+
+        Returns:
+            String with summary of the analyzer result
+        """
+        # TODO: Once we can identify user generated events this should be
+        # updated to include all user generated events instead of focusing
+        # solely on browser events.
+        query = 'source_short:"WEBHIST"'
```
berggren: To cover plaso exported CSVs we should add source as well, e.g.:

```query = 'source_short:"WEBHIST" OR source:"WEBHIST"' ```

File: timesketch/lib/analyzers/browser_timeframe.py [hunk 3]
```diff
@@ -0,0 +1,203 @@
...
+
+        return_fields = ['timestamp', 'url', 'tag', '__ts_emojis']
```
berggren: You doesn't use the "__ts_emojis" anywhere so you can drop it from the returned fields.

timesketch/lib/analyzers/browser_timeframe.py
```diff
@@ -0,0 +1,203 @@
...
+            query_string=query, return_fields=return_fields)
+
+        if not data_frame.shape[0]:
+            return 'No browser events discovered.'
+
+        sleeping_emoji = emojis.get_emoji('SLEEPING_FACE')
+
+        # This query filters out all timestamps that have a zero timestamp as
+        # well as those that occure after 2035-01-01, this may need to be
+        # changed in the future.
```
berggren: can we make this 2038-01-01 to account for the Y2K38 bug ;)

timesketch/lib/analyzers/browser_timeframe.py
berggren: Aha! magic ;)

timesketch/lib/analyzers/browser_timeframe.py
berggren: Great!

PR #814, opened 2019-02-04
Trying to extract IP addresses and usernames from potential bruteforce attempts

kovakina:
Not sure if it makes sense limiting IP extraction to ssh logs only, or make it more universal


config/features.yaml
```diff
@@ -61,3 +61,19 @@ github_accounts:
         aggregate: True
         tags: ['Github Account']
         emojis: ['ID_BUTTON']
+
+ip_address:
+        query_string: 'source_short:"LOG"'
+        attribute: 'ipv4'
```
kiddinn: ok... so "source_short:"LOG"' could result in quite a lot of events selected.... but not nearly all of those have an ipv4 attribute... is it possible to change the query_string to something like:

"""
query_string: 'source_short:"LOG" AND _exists_:"ipv4"'
"""

That way you should only get back the events that at least have that attribute set.

config/features.yaml
```diff
@@ -61,3 +61,19 @@ github_accounts:
...
+        store_as: 'ip_address'
+        re: '\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}'
+        aggregate: True
+        tags: ['IP address']
+
+username:
+        query_string: 'source_short:"LOG" AND body:"Invalid user"'
+        attribute: 'username'
+        store_as: 'username'
```
kiddinn: Do we want to overwrite the username attribute here?

Or use the same as we have in other places of storing this as "extracted_username"

config/features.yaml
```diff
@@ -61,3 +61,19 @@ github_accounts:
...
+        re: 'Invalid user ([A-z-\d]{1,39}) from'
```
kiddinn: why the max length of 39? is it better to have

[^ ]+

or just [A-z-\d]+ ???

config/features.yaml
```diff
@@ -61,3 +61,19 @@ github_accounts:
         aggregate: True
         tags: ['Github Account']
         emojis: ['ID_BUTTON']
+
+ip_address:
+        query_string: 'source_short:"LOG"'
+        attribute: 'ipv4'
+        store_as: 'ip_address'
+        re: '\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}'
```
obsidianforensics: This will catch invalid IPs - 666.789.000.999 for example. Also, if there already is an ipv4 attribute, why do we need to extract it again?

kiddinn:
why did you close the PR?

kovakina:
After some consideration, I'm moving these features extraction back into bruteforce scorer itself, as i need them in specific situations and for a specific types of scenarios, that might not be applicable for other uses

PR #824, opened 2019-02-07
Remove unused dep (pycypher) and small bugfix for heatmap

berggren:


timesketch/lib/testlib.py
```diff
@@ -129,8 +129,10 @@ def search(self, *args, **kwargs):
         """Mock a search query.
 
         Returns:
-            A dictionary with search result.
+            A dictionary with search result or integer if count is requested.
         """
+        if kwargs.get('count'):
+            return 4711
```
kiddinn: what is this 4711???

can you at least put a comment here explaining that

kiddinn:
otherwise LGTM

timesketch/lib/testlib.py
berggren: Used sometimes instead of 17, on occasions when you want to denote a slightly larger n
number. Probably comes from the name of 'genuine' Eau-de-cologne, 'No. 4711 ".

Swe hacker jargon..

;)

Issue #823, opened 2019-02-07
Python3 release


berggren:
Python3 support is wip. This issue is to track this work

joachimmetz:
This looks completed.

kiddinn:
this is completed

PR #829, opened 2019-02-08
Fixes for python3 compatibility

kiddinn:
This round mainly focuses on use of decode. 

Part of #823


kiddinn:
Not requesting code review right now, still doing tests, also wanted to have travis run on the code first.

timesketch/lib/py2to3.py
```diff
@@ -0,0 +1,28 @@
+# -*- coding: utf-8 -*-
+"""The Python 2 and 3 compatible type definitions.
+
+File copied from the plaso project.
+"""
+import sys
```
Onager: Don't use this, use six instead, it has all the same functionality.

kiddinn: ack.... much better to just use six... done that

kiddinn:
removed py2to3 in favor of just using six... ready for a review

PR #831, opened 2019-02-09
Moved pylint to stand-alone CI test target

joachimmetz:


joachimmetz:
@berggren can you have a look, so we can move the GIFT PPA move forward

PR #833, opened 2019-02-09
Multiple changes for code flagged by pylint 2.1.1

joachimmetz:
Multiple changes for code flagged by pylint 2.1.1

Also added `score=no` to `.pylintrc` to provide compacter view of pylint output.


joachimmetz:
@berggren there are various places where there are trailing commas in the code. I assumed these should not be there and removed them.

PR #844, opened 2019-02-26
Allow comma separated list of usernames to share to

berggren:
Allow comma separated username list when sharing.


timesketch/views/sketch.py
```diff
@@ -110,12 +110,13 @@ def overview(sketch_id):
...
         if permission_form.username.data:
-            username = permission_form.username.data
-            base_username = username.split('@')[0]
-            user = User.query.filter_by(username=base_username).first()
-            if user:
-                sketch.grant_permission(permission='read', user=user)
-                sketch.grant_permission(permission='write', user=user)
+            usernames = permission_form.username.data
+            for username in usernames.split(','):
+                base_username = username.split('@')[0]
```
kiddinn: maybe add

```
if not base_username:
    continue
```

berggren: There will always be a base_username. The call to the DB will return None if user does not exist so we are good. But I realized that we need to strip() all blanks so I added that to the base_username.

berggren:
@kiddinn PTAL

PR #848, opened 2019-03-08
New frontend

berggren:
This is a rewrite of the frontend code from Angular to Vue. In order to simplify the build environment and to make it easier for non-frontend developers to add new feature we are moving to use the Vue JS framework. The learning curve is small and on-boarding new developers will hopefully be easier.

The old Angular (mixed 1 and 2) will still be the default frontend for the time being. To test the new frontend there is a magic setting in the main __init__.py file to control it. This frontend is not done yet, but it is going to be a WiP for the coming weeks.

**Major changes:**
* Angular -> Vue
* Typescript -> JS
* Bootstrap CSS -> Bulma CSS

Build with yarn+webpack.

**Early screenshot:**
screenshot 2019-03-08 at 15 27 54


timesketch/__init__.py
```diff
@@ -36,6 +36,10 @@
 from timesketch.views.home import home_views
 from timesketch.views.sketch import sketch_views
 from timesketch.views.auth import auth_views
+from timesketch.views.spa import spa_views
+
+# Set to true to use the new Vue.js based frontend.
+USE_NEW_FRONTEND = True
```
kiddinn: should this not be set to False by default? Or read from a config file...

timesketch/frontend/dist/index.html
```diff
@@ -0,0 +1 @@
+<!DOCTYPE html><html lang=en><head><meta name=csrf-token content="{{ csrf_token() }}"><meta charset=utf-8><meta http-equiv=X-UA-Compatible content="IE=edge"><meta name=viewport content="width=device-width,initial-scale=1"><link rel=icon href=/dist/favicon.ico><title>Timesketch</title><link href=/dist/css/app.2334801b.css rel=preload as=style><link href=/dist/js/app.870b2c78.js rel=preload as=script><link href=/dist/js/chunk-vendors.3bf7e5ac.js rel=preload as=script><link href=/dist/css/app.2334801b.css rel=stylesheet></head><body><div id=app></div><script src=/dist/js/chunk-vendors.3bf7e5ac.js></script><script src=/dist/js/app.870b2c78.js></script></body></html>
```
kiddinn: is this auto-generated?

if not, then having this "prettified" would make it more legible... if this is automatically generated, then please ignore my comment

timesketch/__init__.py
berggren: Good catch! Yes it should definitely be set to False by default :)
Fixing.

timesketch/frontend/dist/index.html
berggren: Correct, this is auto-generated and should be checked in.

timesketch/__init__.py
berggren: Fixed

Issue #855, opened 2019-03-24
Update pylintrc


joachimmetz:
Update .pylintrc to closer match that of the log2timeline projects

* [x] update .pylintrc file https://github.com/google/timesketch/pull/856
  * fix issue pylint python 3.5.2 versus pip python 3.5.6
* [ ] enable load-plugins=pylint.extensions.docparams
* [ ] do not ignore invalid-name
* [ ] do not ignore no-member

berggren:
Closing this out as the pylintrc has been updated.

PR #856, opened 2019-03-24
Updated .pylintrc and changes for linter #855

joachimmetz:
* removed code flagged as useless-super-delegation
* changed code flagged as broad-except
* additional overrides


joachimmetz:
@berggren please have a look, I've switched to use GIFT PPA packages

PR #857, opened 2019-03-28
Adding a term aggregator.

kiddinn:
Adding a new aggregator that allows for filtering the result set before aggregating... two ways of evoking:

1. Send in a query string that you wish to use to filter result set prior to aggregation
2. Send in a query DSL string you wish to use to filter result set prior to aggregation.

Then you need to denote what field or attribute you want to be aggregated.

This can be used for instance to query for all events with a certain tag, eg by doing:

```
<aggregator_object>.run(field='domain', query='tag: "phishy-domain"')
```

This would aggregate only on the events that pass the query (so only events with that specific tag) and then aggregate based on the field name `domain`.


berggren:
I like the added feature of the query to choose what documents to aggregate on. I would like to generalise this aggregator though. I think this should be a query_aggregator instead just for tags. If we break this down it is basically an aggregator that takes a query to get the set of documents to aggregate, and then it does a terms aggregation on a chosen field. 

Something like this:

```
aggregator.run(query, field)
...
```

Then, if you want to aggregate in tags you just:

```
aggregator.run(query='tag:foo-bar-baz', field='data_type')
```

And if you wanna do the same for any other field you just change your query.
WDYT?

berggren:
and if we find tags to be super common to aggregate on we can add a helper for that.
```query_aggregator.all_tags_for_sketch(sketch_id=1)```

or something.

berggren:
Oh, and we need to support passing in query_filter and query_dsl as well :)

berggren:
For number 1:
_Aggregate all tags, that is give an overview of what tags are in a timeline._

This is just the bucket aggregator
```bucket_aggregator.run(field='tag')```

So we can drop that from this IMO.

kiddinn:
so first of all... 

last comment about the tag.. yes that is completely unnecessary if we are to generalize this, this was just so that we have all "tag" aggregation in one place.. however since I agree with you that we should generalize this I've removed all mentions of the word tag.... removed that first condition since that is already solved with bucket aggregator and added support for queries and query_dsl...

kiddinn:
PTAL

PR #860, opened 2019-04-03
Added API calls for aggregations as well as enabling client access to aggregations (step 1)

kiddinn:
Adding to the API:
+ placeholder for listing all saved aggregations.
+ placeholder for fetching a saved aggregation
+ a method to run aggregation queries (similar to explore, except for aggregation)

Adding to the client:
+ Method for submitting aggregation queries.


api_client/python/timesketch_api_client/client.py
```diff
@@ -367,6 +367,20 @@ def _build_pandas_dataframe(self, search_response):
 
         return data_frame
 
+    def _get_aggregation_buckets(self, entry, name=''):
+        """Yields all buckets from a aggregation result object."""
```
berggren: Add args

timesketch/api/v1/resources.py
```diff
@@ -695,10 +700,140 @@ def post(self, sketch_id):
...
+
+        if aggregator_name and aggregator_parameters:
+            agg_class = aggregator_manager.AggregatorManager.get_aggregator(
+                aggregator_name)
+            if not agg_class:
+                return {}
+            if not aggregator_parameters:
+                aggregator_parameters = {}
+            aggregator = agg_class(sketch_id=sketch_id)
+            time_before = time.time()
```
berggren: You get this in the response back in the "took" value. Why calculating this yourself?

timesketch/api/v1/resources.py
```diff
@@ -695,10 +700,140 @@ def post(self, sketch_id):
...
+            result = {
+                'aggregation_result': {
+                    aggregator_name: buckets
+                }
+            }
+            meta = {
+                'method': 'aggregator_run',
+                'name': aggregator_name,
+                'es_time': time_after - time_before,
+                'timed_out': False
```
berggren: Can it never time out? :)

timesketch/api/v1/resources.py
```diff
@@ -695,10 +700,140 @@ def post(self, sketch_id):
...
+            aggregation_id: Integer primary key for an agregation database model
+        """
+        # TODO: Implement once we have an aggregation model in the datastore.
+        abort(HTTP_STATUS_CODE_FORBIDDEN)
+
+
+class AggregationExploreResource(ResourceMixin, Resource):
+    """Resource to send an aggregation request."""
+
+    RESULT_FIELDS = frozenset(['_shards', 'hits', 'timed_out', 'took'])
```
berggren: It looks like you are filtering these out? In that case, rename to somthing like:
```REMOVE_FIELDS```

File: timesketch/models/sketch.py [hunk 5]
```diff
@@ -63,6 +63,15 @@ def __init__(self, name, description, user):
         self.description = description
         self.user = user
 
+    @property
+    def get_named_aggregations(self):
+        """Get named aggregations.
+
+        Get named aggregations, i.e. only aggregations that have a name.
+        """
+        # Implement a storage for aggregations.
```
berggren: Add TODO here as well.

api_client/python/timesketch_api_client/client.py
kiddinn: done

timesketch/api/v1/resources.py
kiddinn: no, it's infallible ;)

I kept this here since it is provided by the other request.. I just removed this, since it provides no value ;)

timesketch/api/v1/resources.py
kiddinn: because this is running an aggregator class, and they return back an AggregationResult object, which does not have that stored... and aggregators may just run that aggregation logic in code and not rely on ES at all, so there is no guarantee that a counter is maintained there

timesketch/api/v1/resources.py
kiddinn: ack

timesketch/models/sketch.py
kiddinn: ack

kiddinn:
done with round

timesketch/api/v1/resources.py
berggren: ok, got it.

PR #866, opened 2019-04-10
Markdown stories

berggren:
Implements Markdown support for the story feature in the new frontend code. This replaces the old HTML based stories and add the benefit of being data driven. This will enable API support for creating stories.

* Markdown support with GitHub flavoured styling
* Inline saved view support

Screenshot 2019-04-12 at 11 07 06

Part of: #867


timesketch/views/spa.py
```diff
@@ -16,13 +16,28 @@
...
 # Register flask blueprint
 spa_views = Blueprint('spa_views', __name__)
 
 
+@spa_views.route(
+    '/sketch/<int:sketch_id>/explore/view/<int:view_id>/', methods=['GET'])
+@login_required
+# pylint: disable=unused-argument
+def redirect_view(sketch_id, view_id):
+    """Generates the template.
```
kiddinn: this feels a bit weird....

the function is called `redirect_view` but the docstring says `generate the template` and then all it does is return a `redirect` call.. so the function name and content seem to match but the docstring sounds a bit weird

PR #872, opened 2019-04-23
MUS2019-CTF Colab 

obsidianforensics:


notebooks/MUS2019_CTF.ipynb
```diff
@@ -0,0 +1,1120 @@
...
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/obsidianforensics/timesketch/blob/ctf-colab/notebooks/MUS2019_CTF.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
kiddinn: wondering whether this gets automatically changed once this gets merged... otherwise it is pointing to your branch

notebooks/MUS2019_CTF.ipynb
```diff
@@ -0,0 +1,1120 @@
...
+      },
+      "cell_type": "markdown",
+      "source": [
+        "I'm going to focus on how to find the answers to the CTF questions after all the processing has been done. I'll quickly summarize the processing steps I did to get to the state when I pick up my walkthrough.\n",
+        "\n",
+        "I started off by processing the provided E01 image with a basic log2timeline command; nothing special added:\n",
+        "\n",
+        "\n",
+        "```\n",
+        "ryan:~$ log2timeline.py MUS2019-CTF.plaso MUS-CTF-19-DESKTOP-001.E01\n",
```
kiddinn: I would remove this "ryan" and just have

$ log2....

notebooks/MUS2019_CTF.ipynb
```diff
@@ -0,0 +1,1120 @@
...
+    },
+    {
+      "metadata": {
+        "id": "hvCl6L7ZQrsZ",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "ctf = sketches[0]"
```
kiddinn: I'm wondering whether something like:

```
ctf = [x for x in sketches if x.name == 'MUSCTF 2019'][0]
```

is better, since you are picking the first sketch in the list... what if another sketch is added to the demo server?

notebooks/MUS2019_CTF.ipynb
```diff
@@ -0,0 +1,1120 @@
...
+    {
+      "metadata": {
+        "id": "5R57xnzQGzEB",
+        "colab_type": "code",
+        "colab": {}
+      },
+      "cell_type": "code",
+      "source": [
+        "# The ActiveTimeBias is in minutes, so divide by -60 (I don't know why it's stored negative): \n",
+        "420 / -60"
```
kiddinn: although this is a bit more complicated text... should we do this more "programmatically" ? food for thought... something like:

```
time_string = list(set(ts_results.message))[0]
_, _, data_string = time_string.partition(']')
data = {}
items = data_string.split(':')
key = items[0].strip()

for line in items[1:-1]:
  words = line.split()
  data[key] = ' '.join(words[:-1])
  key = words[-1]
data[key] = items[-1]

print('The UTC offset is: {0:d}'.format(int(data.get('ActiveTimeBias', '0')) / int(data.get('DaylightBias', '1'))))
```

notebooks/MUS2019_CTF.ipynb
obsidianforensics: Yeah I'm not sure - I can do another PR after the merge with the new one if it doesn't?

notebooks/MUS2019_CTF.ipynb
obsidianforensics: Style choice, but done.

notebooks/MUS2019_CTF.ipynb
obsidianforensics: I was trying to show how to pick a sketch from the list of sketches, that's why it is two steps. I've added some clarification to the text above, let me know what you think.

notebooks/MUS2019_CTF.ipynb
obsidianforensics: I'd rather keep it simpler for the target audience. The notebook is supposed to show a "live" investigation, and writing that much code for a small one-time operation seems like overkill to me.

I also think it'd be good to take another look at how plaso parses these out (separate entries per key, rather than all into one string).

notebooks/MUS2019_CTF.ipynb
obsidianforensics: I changed it to what I think the master path should be - I can fix afterwards if wrong.

obsidianforensics:
PTAL

notebooks/MUS2019_CTF.ipynb
kiddinn: ack

notebooks/MUS2019_CTF.ipynb
kiddinn: ok... agree... but maybe change the text a bit to show or talk about where those numbers come from...

PR #882, opened 2019-05-05
Added Debian packaging files

joachimmetz:


berggren:
@joachimmetz let me know when this one is ready for review.

joachimmetz:
@berggren the main parts are there, please have a look at the PR. There is likely more tweaking to be done to make sure everything is running out of the box.

config/dpkg/control
```diff
@@ -0,0 +1,45 @@
...
+Description: Python 3 module of Timesketch
+ Timesketch is a web based tool for collaborative forensic
+ timeline analysis. Using sketches you and your collaborators can easily
+ organize timelines and analyze them all at the same time.  Add meaning
+ to your raw data with rich annotations, comments, tags and stars.
+
+Package: timesketch-server
+Architecture: all
+Depends: apt-transport-https, elasticsearch, gunicorn3, openjdk-8-jre-headless, postgresql, python3-psycopg2, python3-timesketch (>= ${binary:Version}), ${python3:Depends}, ${misc:Depends}
```
berggren: I'm not sure we want to depend on ES and Java here. The ES servers can run independently from the TS servers. IMO this should be handle by the deployment process rather than installing it here.

joachimmetz:
@joachimmetz note to self make ES and java recommends

config/dpkg/control
berggren: moved to recommends

PR #890, opened 2019-05-15
Testing for None should be done by is

deralexxx:
instead of == according to @lgtm which I used to check the repo (might add some more issues)


berggren:
Yeh, this could have side effects in SQLalchemy. I rather keep as is. LEt me know if you wanna reopen.

Issue #896, opened 2019-05-28
Sketch object not committed before setting ACL


berggren:
In some combinations of SQLAlchemy versions we get an error when trying to set ACL on a non-committed sketch object.

berggren:
Fixed.

Issue #886, opened 2019-05-07
sqlalchemy.exc.StatementError when adding new sketch from ui


garanews:
I'm getting
`
sqlalchemy.exc.StatementError

sqlalchemy.exc.StatementError: (sqlalchemy.exc.InvalidRequestError) Can't resolve value for column sketch.id on object <Sketch at 0x7f675e9aeb38>; no value has been set for this column
[SQL: SELECT sketch_accesscontrolentry.id AS sketch_accesscontrolentry_id, sketch_accesscontrolentry.created_at AS sketch_accesscontrolentry_created_at, sketch_accesscontrolentry.updated_at AS sketch_accesscontrolentry_updated_at, sketch_accesscontrolentry.permission AS sketch_accesscontrolentry_permission, sketch_accesscontrolentry.parent_id AS sketch_accesscontrolentry_parent_id, sketch_accesscontrolentry.user_id AS sketch_accesscontrolentry_user_id, sketch_accesscontrolentry.group_id AS sketch_accesscontrolentry_group_id 
FROM sketch_accesscontrolentry 
WHERE %(param_1)s = sketch_accesscontrolentry.user_id AND sketch_accesscontrolentry.group_id IS NULL AND sketch_accesscontrolentry.permission = %(permission_1)s AND %(param_2)s = sketch_accesscontrolentry.parent_id]
[parameters:  [{}]]
`

when try to add new sketch.
I already dropped postgresql DB and recreated withtout success.
Everything was working until I deleted last sketch from ui.
I tried to install again from git (keeping current config file) with both python2 and pyton3 but same error.

Any help?

garanews:
I tried to modify /usr/local/lib/python3.6/dist-packages/timesketch-20190207-py3.6.egg/timesketch/models/acl.py in this way:

```# Check access for user.
        ace = self.AccessControlEntry.query.filter_by(
            user=user,
            group=None,
            permission=permission,
            #parent=self
            ).all()
```

Now I am able to create the sketch but only the first one created is shown in gui (the others are accessible  specifying the direct url in path, like ..../sketch/7/)
Creating a new user the behavior is the same, only the first sketch created is shown.
I am sure the parent=self is needed, but leaving it causes the error posted before.

Any suggestion?

lprat:
Hi,
Maybe relation with unicode problem (warning at first line):

[...]

PR #897, opened 2019-05-28
Commit Sketch object to DB before setting ACL

berggren:
Addresses #896 - Commit Sketch to DB before setting ACL.


berggren:
related #886

PR #900, opened 2019-05-29
Yeti indicator analyzer

tomchop:
This PR introduces a Yeti indicator analyzer.

The analyzer will retrieve all indicators of type `x-regex` from a remote Yeti instance and run them over a timeline. Matching events are tagged and views are created per-actor.

### DIFF
--- a/timesketch/lib/analyzers/yetiindicators.py
+++ b/timesketch/lib/analyzers/yetiindicators.py
@@ -1,0 +1,130 @@
+"""Index analyzer plugin for threatintel."""
+from __future__ import unicode_literals
+
+from flask import current_app
+import requests
+
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+from timesketch.lib import emojis
+
+
+def build_query_for_indicators(indicators):
+    """Builds an ElasticSearch query for Yeti indicator patterns.
+
+    Prepends and appends .* to the regex to be able to search within a field.
+
+    Returns:
+      The resulting ES query string.
+    """
+    query = []
+    for domain in indicators:
+        query.append('domain:/.*{0:s}.*/'.format(domain['pattern']))
+    return ' OR '.join(query)
+
+
+class YetiIndicators(interface.BaseSketchAnalyzer):
+    """Index analyzer for Yeti threat intel indicators."""
+
+    NAME = 'yetiindicators'
+    DEPENDENCIES = frozenset(['domain'])
+
+    def __init__(self, index_name, sketch_id):
+        """Initialize the Index Analyzer.
+
+        Args:
+            index_name: Elasticsearch index name
+        """
+        super(YetiIndicators, self).__init__(index_name, sketch_id)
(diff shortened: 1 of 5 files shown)


timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -0,0 +1,129 @@
...
+
+                events = self.event_stream(query_string=query,
+                                           return_fields=[])
+
+                name = intrusion_set['name']
+                for event in events:
+                    found = True
+                    event.add_emojis([emojis.get_emoji('SKULL')])
+                    event.add_tags([name])
+                    event.commit()
```
berggren: Move event.commit() after event.comment

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -0,0 +1,129 @@
...
+
+from flask import current_app
+import requests
+
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+from timesketch.lib import emojis
+
+
+def build_query_for_indicators(indicators):
```
berggren: This only get Yeti domain indicators. Do you plan on supporting more types? e.g. file hashes?
If that is the case we should probably have a mapping between Yeti indicator type and ES field to query. Then we can build this up dynamicaly in this function. WDYT?

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -0,0 +1,129 @@
...
+    query = []
+    for domain in indicators:
+        query.append('domain:/.*{0:s}.*/'.format(domain['pattern']))
+    return ' OR '.join(query)
+
+
+class YetiIndicators(interface.BaseSketchAnalyzer):
+    """Index analyzer for Yeti threat intel indicators."""
+
+    NAME = 'yetiindicators'
```
berggren: Do you plan on having more Yeti analyzers? If you only plan on updating this one I opt for renaming this (filename and class and NAME) to simply "yeti".

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -0,0 +1,129 @@
...
+                actors_found.append(name)
+                self.sketch.add_view(
+                    'Domain activity for actor {0:s}'.format(name),
+                    self.NAME,
+                    query_string=query)
+
+        if actors_found:
+            return '{0:d} actors were found! [{1:s}]'.format(
+                len(actors_found), ', '.join(actors_found))
+        return 'No indicators were found in the timeline.'
```
berggren: s/were/was

timesketch/lib/analyzers/yetiindicators.py
berggren: Huh, interesting. I'll have a look.

timesketch/lib/analyzers/yetiindicators.py
berggren: ack

PR #902, opened 2019-05-31
Testing for None should use the 'is' operator.

deralexxx:
Code health better readability: Testing for None should use the 'is' operator.
https://help.semmle.com/wiki/display/PYTHON/Testing+equality+to+None


berggren:
Have you checked that this doesn't have any side effects in the SQL queries? This is or SQLAlchemy and is following the best practices for that package.

I rather opt not to change this as it can introduce hard-to-debug errors

berggren:
This could have side effects in SQLalchemy. I rather keep as is. LEt me know if you wanna reopen.

PR #906, opened 2019-06-06
more emojis. more horsepower. more meatballs.

jeanbrand0n:


timesketch/lib/emojis.py
```diff
@@ -25,18 +25,46 @@
 
 
 EMOJI_MAP = {
+    '8BALL': emoji('&#1F3B1', 'N/A'),
```
kiddinn: why are we adding all these emojis in here if we don't have a purpose for them? that is they are all marked as N/A?

can we somehow find purpose for those?

berggren:
After some consideration, let's add emojis when they are needed instead, with purpose documented. I'll close this PR.

PR #909, opened 2019-06-12
Dynamic analyzers

berggren:


timesketch/lib/tasks.py
```diff
@@ -115,21 +119,20 @@ def _get_index_analyzers():
...
 
-    # Exit early if index analyzers are disabled.
-    if not current_app.config.get('ENABLE_INDEX_ANALYZERS'):
+    if not index_analyzers:
         return None
 
-    for analyzer_name, analyzer_cls in manager.AnalysisManager.get_analyzers():
+    for analyzer_name, analyzer_cls in manager.AnalysisManager.get_analyzers(
+            index_analyzers):
         kwarg_list = analyzer_cls.get_kwargs()
```
kiddinn: wasn't get_kwargs() removed from the interface?

timesketch/lib/tasks.py
```diff
@@ -181,37 +186,78 @@ def build_index_pipeline(file_path, timeline_name, index_name, file_extension,
...
-    # Exit early if sketch analyzers are disabled.
-    if not current_app.config.get('ENABLE_SKETCH_ANALYZERS', False):
+    # Exit early if no sketch analyzers are configured to run.
+    if not (analyzer_names or auto_analyzers):
         return None
 
-    for analyzer_name, analyzer_cls in manager.AnalysisManager.get_analyzers():
-        kwarg_list = analyzer_cls.get_kwargs()
+    if not analyzer_names:
+        analyzer_names = auto_analyzers
```
kiddinn: ok... so you are forgetting to actually use auto_analyzers except in the case where there are no analyzers defined...

should this not be

```
auto_analyzers = current....get('AUTO+SKETCH_ANALYZERS', [])
...
if analyzer_names:
  analyzer_names.extend(auto_analyzers)
else:
  analyzer_names = auto_analyzers
```

That is to extend the analyzer names with the auto ones? or is this meant to be called twice, once in the beginning where we have just the auto analyzers and then a potentially second/third/etc times when it is manually run? 
if that is the case I would argue that this should be

```
if not analyzer_names:
  analyzer_names = current_app.config.get('AUTO_SKETCH_ANALYZERS', []

if not analyzer_names:
  return
```

and just not define the `auto_analyzers` attribute at all

timesketch/lib/tasks.py
```diff
@@ -181,37 +186,78 @@ def build_index_pipeline(file_path, timeline_name, index_name, file_extension,
...
         Celery group with analysis tasks or None if no analyzers are enabled.
     """
     tasks = []
+    auto_analyzers = current_app.config.get('AUTO_SKETCH_ANALYZERS', None)
 
-    # Exit early if sketch analyzers are disabled.
-    if not current_app.config.get('ENABLE_SKETCH_ANALYZERS', False):
+    # Exit early if no sketch analyzers are configured to run.
+    if not (analyzer_names or auto_analyzers):
         return None
```
kiddinn: no need to define None....

`return` does return None

timesketch/lib/tasks.py
```diff
@@ -181,37 +186,78 @@ def build_index_pipeline(file_path, timeline_name, index_name, file_extension,
...
+    analyzers = manager.AnalysisManager.get_analyzers(analyzer_names)
+    for analyzer_name, analyzer_cls in analyzers:
         if not analyzer_cls.IS_SKETCH_ANALYZER:
             continue
 
-        if kwarg_list:
-            for kwargs in kwarg_list:
-                tasks.append(run_sketch_analyzer.s(
-                    sketch_id, analyzer_name, **kwargs))
+        if analyzer_kwargs:
```
kiddinn: what are you doing when it comes to auto analyzers? some of them might require some kwargs? how are you passing them along?

docker/Dockerfile
```diff
@@ -35,10 +35,10 @@ RUN sed -i -e '/pyyaml/d' /tmp/timesketch/requirements.txt
 RUN pip3 install /tmp/timesketch/
 
 # Copy the Timesketch configuration file into /etc
-RUN cp /tmp/timesketch/timesketch.conf /etc
+RUN cp /tmp/timesketch/data/timesketch.conf /etc
 # Copy Timesketch config files into /etc/timesketch
 RUN mkdir /etc/timesketch
-RUN cp -r /tmp/timesketch/config/* /etc/timesketch/
+RUN cp -r /tmp/timesketch/data/features.yaml /etc/timesketch/
```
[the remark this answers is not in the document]
berggren: Good catch, this was a leftover. Fixed.

timesketch/lib/tasks.py
berggren: I know, linter complains though because there are other returns in this function. I rather be explicit here than ignoring linter.

timesketch/lib/tasks.py
berggren: Good catch, it is the latter. Changed per your suggestion which is clearer.

timesketch/lib/tasks.py
berggren: I was planning on adding TODOs for this, but missed adding that. Good catch! I decided to fix this now instead as this is a small fix. You will be able to pass in kwargs to auto analyzers via config, and for any other manually run analyzer there will be a config option to specify default kwargs per analyzer as well. Explicitly provided kwargs will always take precedence though.
```
AUTO_SKETCH_ANALYZERS_KWARGS = {'domain': {'foo': 'bar'}} 
ANALYZERS_DEFAULT_KWARGS = {'domain': {'foo': 'bar'}} 
```

timesketch/lib/tasks.py
berggren: Fixed.

kiddinn:
There is still the case of the test that fails...

```
Checking: timesketch/lib/tasks.py
************* Module timesketch.lib.tasks
timesketch/lib/tasks.py:127:23: W0612: Unused variable 'analyzer_cls' (unused-variable)
The command "./config/travis/run_with_timeout.sh 30 ./config/travis/runtests.sh" exited with 4.
```

berggren:
@kiddin can you flip the LGTM bit? Tests pass now

PR #918, opened 2019-06-26
GCP servicekey analyzer for stackdriver logs

pstirparo:
Gcp servicekey analyzer for stackdriver logs, as produced by dftimewolf collector (https://github.com/log2timeline/dftimewolf/blob/master/dftimewolf/lib/collectors/stackdriver.py)


timesketch/lib/analyzers/gcp_servicekey.py
```diff
@@ -0,0 +1,71 @@
...
+        self.index_name = index_name
+        super(GcpServiceKeySketchPlugin, self).__init__(index_name, sketch_id)
+
+    def run(self):
+        """Entry point for the analyzer.
+
+        Returns:
+            String with summary of the analyzer result
+        """
+        query = ('principalEmail:*gserviceaccount*')
```
berggren: Hm, can we make this query a bit cheaper? using "*" is a bit expensive. Dowe have a data_type attribute maybe?

timesketch/lib/analyzers/gcp_servicekey.py
```diff
@@ -0,0 +1,71 @@
...
+            query_string=query, return_fields=return_fields)
+
+        gcp_servicekey_counter = 0
+
+        for event in events:
+            # Fields to analyze.
+            methodName = event.source.get('methodName')
+
+            if 'CreateServiceAccount' in methodName:
+                event.add_star()
```
berggren: Do we want to star by default? Maybe just label? Also, use a Tag instead of a label. The label is sketch specific and will not follow if the timeline is added to any future sketch.

pstirparo: done:
- removed all stars by default
- changed all labels into tags

timesketch/lib/analyzers/gcp_servicekey.py
pstirparo: Unfortunately this version of parsed logs from dftimewolf has less fields to use compared to the previous one. I can remove one * from the query and make it "principalEmail:*gserviceaccount.com"

berggren: Ok, no worries. lets keep as it is.

timesketch/lib/analyzers/gcp_servicekey.py
berggren: The stars and labels are still there. Did you push? :)

PR #925, opened 2019-07-07
New UI: Time filters

berggren:
Add time range filter

Screenshot 2019-07-07 at 17 45 54


berggren:
@kiddinn tests fixed.

Issue #927, opened 2019-07-09
Changes to /data/timesketch.conf cause error


mrezqi:
This error occurred when trying to build timesketch using docker-compose. The configuration file changed in the latest version is causing errors. It is causing indentation errors. I believe it is originated from this block:
```
#-------------------------------------------------------------------------------
# Analyzers.

# Which analyzers to run automatically.
AUTO_INDEX_ANALYZERS = []
AUTO_SKETCH_ANALYZERS = []

# Optional specify any default arguments to pass to analyzers.
# The format is:
# {'analyzer1_name': {
      'param1': 'value'
    },
    {'analyzer2_name': {
        'param1': 'value'
      }
    }
  }
AUTO_SKETCH_ANALYZERS_KWARGS = {}
ANALYZERS_DEFAULT_KWARGS = {}
```

kiddinn:
this should be fixed by now

PR #928, opened 2019-07-11
Fix for issue #927

jadams:
only the first line was commented out of a block which throws errors


jadams:
I signed it!

jadams:
I signed it! again

joachimmetz:
@jadams thx for the fix. @berggren and @kiddinn have low availability at this point in time. I've assinged @berggren to have a look when he's available.

PR #941, opened 2019-08-07
Expert sessionizer

katemacleod:
Implemented web activity expert sessionizer.

- Added BaseExpertSessionizerPlugin as a base class for the expert sessionizers.
- Added WebActivitySessionizerSketchPlugin
- Changed sessionizer.py so that the attributes for non-expert sessioned events follows the same format as expert sessioned ones.
- Updated sessionizer_tests.py to reflect this.

### DIFF
--- a/data/timesketch.conf
+++ b/data/timesketch.conf
@@ -138,7 +138,7 @@ NEO4J_PASSWORD = '<NEO4J_PASSWORD>'
 
 # Which analyzers to run automatically.
 AUTO_INDEX_ANALYZERS = []
-AUTO_SKETCH_ANALYZERS = []
+AUTO_SKETCH_ANALYZERS = ['sessionizer']
 
 # Optional specify any default arguments to pass to analyzers.
 # The format is:
--- a/data/timesketch.conf
+++ b/data/timesketch.conf
@@ -152,7 +152,10 @@ AUTO_SKETCH_ANALYZERS = []
 # }
 AUTO_SKETCH_ANALYZERS_KWARGS = {}
 ANALYZERS_DEFAULT_KWARGS = {}
+<<<<<<< Updated upstream
 
+=======
+>>>>>>> Stashed changes
 
 # Add all domains that are relevant to your enterprise here.
 # All domains in this list are added to the list of watched
--- a/data/timesketch.conf
+++ b/data/timesketch.conf
@@ -152,10 +152,6 @@ AUTO_SKETCH_ANALYZERS = []
 # }
 AUTO_SKETCH_ANALYZERS_KWARGS = {}
 ANALYZERS_DEFAULT_KWARGS = {}
-<<<<<<< Updated upstream
-
-=======
->>>>>>> Stashed changes
 
 # Add all domains that are relevant to your enterprise here.
 # All domains in this list are added to the list of watched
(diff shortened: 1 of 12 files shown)


data/timesketch.conf
```diff
@@ -138,7 +138,7 @@ NEO4J_PASSWORD = '<NEO4J_PASSWORD>'
 
 # Which analyzers to run automatically.
 AUTO_INDEX_ANALYZERS = []
-AUTO_SKETCH_ANALYZERS = []
+AUTO_SKETCH_ANALYZERS = ['web_activity_sessionizer']
```
tomchop: Is this a remnant of your dev environment setup?

timesketch/lib/analyzers/expert_sessionizers.py
```diff
@@ -0,0 +1,21 @@
...
+
+from __future__ import unicode_literals
+
+from timesketch.lib.analyzers import manager
+from timesketch.lib.analyzers.sessionizer import SessionizerSketchPlugin
+
+class BaseExpertSessionizerSketchPlugin(SessionizerSketchPlugin):
+    """Base class for the expert sessionizers."""
+    def annotateEvent(self, event, session_num):
+        event.add_attributes({'session_id': {self.session_type:session_num}})
```
tomchop:
```suggestion
        event.add_attributes({'session_id': {self.session_type: session_num}})
```

timesketch/lib/analyzers/sessionizer.py
```diff
@@ -15,56 +15,46 @@ class SessionizerSketchPlugin(interface.BaseSketchAnalyzer):
...
-                                 'sessionizer', query_string=query)
+                                 self.NAME, query_string=self.query)
         except StopIteration:
             pass
 
         return ('Sessionizing completed, number of session created:'
                 ' {0:d}'.format(session_num))
 
+    def annotateEvent(self, event, session_num):
+        event.add_attributes({'session_id': {'all_events':session_num}})
```
tomchop: Since you're extracting things like `query` and `max_time_diff_micros`, wouldn't it make sense to also extract a `session_type` attribute and use it here instead of hardcoding `'all_events'`? That way you don't even have to override the function in the expert sessionizers

timesketch/lib/analyzers/base_sessionizer_test.py
```diff
@@ -0,0 +1,214 @@
...
+    """
+    Returns an instance of Event, based on the MockDataStore event_dict
+    example.
+
+    Args:
+        event_id: Desired ID for the Event.
+        quantity: The number of Events to be generated.
+        time_diffs: A list of time differences between the generated
+        Events.
+
```
tomchop: Pleas add the docstring for `source_attrs` here

timesketch/lib/analyzers/expert_sessionizers_test.py
```diff
@@ -2,102 +2,47 @@
...
                 message,
                 'Sessionizing completed, number of session created: 1')
 
             ds = MockDataStore('test', 0)
             event1 = (ds.get_event('test_index', '0', stored_events=True))
-            self.assertEqual(event1['_source']['session_id'],
-                             {'web_activity': 1})
-            # checking event with id '101' as 100 events have been inserted
-            # as 'padding' (see _create_mock_event())
+            self.assertEqual(event1['_source']['source_short'], 'WEBHIST')
```
tomchop: I think it's worth keeping the check that the session_id also matches what you expect *in addition* to checking for the source. Sorry if I wasn't clear :)

Issue #956, opened 2019-08-12
Sequence session analyzer


tomchop:
Sequence session analyzer is a base class for analyzers that are specialized in detecting a sequence of events occuring close in time.

PR #960, opened 2019-08-14
Add sequence sessionizer.

nnyx7:
The provided sequence sessionizer works like an interface for specific
sequence sessionizers.
It resolves issue #956.


timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
+"""Sequenced activity sessionizing sketch analyzer plugin."""
+
+from __future__ import unicode_literals
+
+#from timesketch.lib.analyzers import manager
```
tomchop: Remove commented line

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+class SequenceSessionizerSketchPlugin(
+        sessionizer.SessionizerSketchPlugin):
+    """Base class for specialized sequenced activity sessionizings
+    sketch plugins.
+
+    Attributes:
+        event_seq_name: The name of the event sequence.
+        event_seq: List of dictionaries of attributes describing the events
```
tomchop:
```suggestion
        event_seq: List of dictionaries of attributes describing the events.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+        event_storage: List of events.
+        num_event_to_find: Number that shows which event from the sequence
+        should be matched.
```
tomchop: Please indent with 4 spaces when doing a newline in the `Args` section
```suggestion
            should be matched.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+        recording: Shows if currently are finded and stored events that match
+        the specified sequence of events.
```
tomchop:
```suggestion
            the specified sequence of events.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+        return_fields: List of name of event attributes, should be specified
+        in the inheriting sessionizers. It's mandatory to contains 'timestamp'.
```
tomchop:
```suggestion
            in the inheriting sessionizers. It must contain 'timestamp'.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+    event_seq_name = None
+    event_seq = None
+    event_storage = []
+    num_event_to_find = 0
+    recording = False
+    return_fields = ['timestamp']
+    session_num = 0
+
+    def run(self):
+        """Entry point for the analyzer. Allocates each event between the
```
tomchop: You can insert a newline before `Allocates` to have a neater docstring:

```
"""Entry point for the analyzer

Allocates each event...

Returns:
    String containing...
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+        Returns:
+            String containing the name of the event sequence and the
+            number of sessions created.
+        """
+        if self.event_seq_name is None:
+            raise RuntimeError('No event_seq_name provided.')
+        if self.event_seq is None or self.event_seq == []:
+            raise RuntimeError('No event_seq provided.')
+        if 'timestamp' not in self.return_fields:
+            raise RuntimeError('No \'timestamp\' in return_fields.')
```
tomchop: Alternatively, you can use double quotes around `"timestamp"` to avoid escaping single quotes

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+                self.process_event(event)
+                last_timestamp = curr_timestamp
+        except StopIteration:
+            pass
+
+        self.query = self.get_query_string()
+
+        self.sketch.add_view('Session view', self.NAME, query_string=self.query)
+
+        return ('Sessionizing completed, number of {0} session created:'
```
tomchop:
```suggestion
        return ('Sessionizing completed, number of {0:s} session created:'
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Args:
+            event: Event to annotate.
+            session_num: Session number for the event.
+        """
+        event.add_attributes({self.event_seq_name: session_num})
+        event.commit()
+
+    def process_event(self, event):
+        """Process event depending on if the event is significant for the
```
tomchop: Same here, try to have a docstring with a brief one-line description, and insert details in a paragraph underneath.

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+
+            if self.num_event_to_find == len(self.event_seq):
+                self.flush_events()
+            return
+
+        if self.recording:
+            self.event_storage.append(event)
+
+    def flush_events(self, drop=False):
+        """Flush the events in the event_storage.
```
tomchop: Maybe write something more descriptive here? "Annotates or clears events in event_storage and resets session recording state"

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,152 @@
...
+        self.num_event_to_find = 0
+        self.event_storage = []
+
+    def match_event(self, event):
+        """Compare event with the event to search for in the event sequence.
+
+        Args:
+            event: Event to compare with the one in the event sequence.
+
+        Returns: If event matches the event to search for in the event
```
tomchop: Returns should be on a newline:

```
Returns:
    If event matches...
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,391 @@
...
+                                   2,
+                                   one_session_args,
+                                   time_diffs=[max_allowed])):
+            index = 'test_index'
+            sketch_id = 1
+            sessionizer = ManyEventsSequenceSessionizer(index, sketch_id)
+            message = sessionizer.run()
+            self.assertEqual(
+                message,
+                'Sessionizing completed, number of {0} session created: 1'.
```
tomchop:
```suggestion
                'Sessionizing completed, number of {0:s} session created: 1'.
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,391 @@
...
+
+            for i in range(1, 100):
+                event = (ds.get_event('test_index', str(i),
+                                      stored_events=True))
+                self.assertNotIn(sessionizer.event_seq_name, event)
+
+
+def _create_mock_event(event_id, quantity, attributes, time_diffs=None):
+    """
+    Returns an instance of Event, based on the MockDataStore event_dict
```
tomchop: Does this really return an instance of Event? Or rather generates a list of Event objects?

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,391 @@
...
+    example and the given attributes and the time differences.
+
+    Args:
+        event_id: Desired ID for the Event.
+        quantity: The number of Events to be generated
```
tomchop:
```suggestion
        quantity: The number of Events to be generated.
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,391 @@
...
+        attributes: A list of dictionaries with attributes for the generated
+        events.
```
tomchop:
```suggestion
            events.
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,391 @@
...
+            sketch_id = 1
+            sessionizer = ManyEventsSequenceSessionizer(index, sketch_id)
+            message = sessionizer.run()
+            self.assertEqual(
+                message,
+                'Sessionizing completed, number of {0} session created: 1'.
+                format(sessionizer.event_seq_name))
+
+            ds = MockDataStore('test', 0)
+            event1 = (ds.get_event('test_index', '0', stored_events=True))
```
tomchop: Why the extra parentheses around the `ds.get_event()` call? I see this in many places in the code

nnyx7:
I changed the function create_eventObj to be more flexible.
There was a bug. The event_template object was not deep copied and changed every time which was not problem before because of the way the function create_eventObj was written, but I had to fix it for the currect version.

nnyx7:
The SequenceSessionizerSketchPlugin is dependent from the added additional changes in commit 5233e50f3a0aea0488554aadb2b7dc56e7876a43. Those changes are the same like the ones in the Expert sessionizer pull request.

nnyx7:
A session_type is new attribute with default value that SequenceSessionizerSketchPlugin inherits from SessionizerSketchPlugin. It behaved the same like the attribute event_seq_name so I replace event_seq_name with session_type.

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+        Returns:
+            String containing the name of the event sequence and the
+            number of sessions created.
+        """
+        if self.session_type is None:
+            raise RuntimeError('No session_type provided.')
+        if self.event_seq is None or self.event_seq == []:
+            raise RuntimeError('No event_seq provided.')
+        # If return_fields in none, then all attributes are provided.
+        if self.return_fields is not None:
```
tomchop: Didn't we discuss building a minimal list of `return_fields` depending on the dicts in `event_seq`? It's OK to leave it for a future PR but in this case please add a `# TODO` comment.

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+            if 'timestamp' not in self.return_fields:
+                raise RuntimeError('No "timestamp" in return_fields.')
+            if len(self.return_fields) <= 1:
+                raise RuntimeError('No return_fields provided.')
```
tomchop: Either change the condition to `<= 0` or change the error message to `"Not enough return_fields provided"`

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+                                   return_fields=self.return_fields)
+
+        last_timestamp = 0
+
+        try:
+            first_event = next(events)
+            last_timestamp = first_event.source.get('timestamp')
+            self.process_event(first_event)
+
+            for event in events:
```
tomchop: You can probably change this loop to

```python
            last_timestamp = None
            for event in events:
                curr_timestamp = event.source.get('timestamp')
                if last_timestamp and (curr_timestamp - last_timestamp > self.max_time_diff_micros):
                    self.flush_events(drop=True)
                self.process_event(event)
                last_timestamp = curr_timestamp
```

and skip the bloc where you call `next(events)`. Or am I missing something?

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+        except StopIteration:
+            pass
+
+        self.query = self.get_query_string()
+
+        self.sketch.add_view('Session view',
+                             self.NAME,
+                             query_string=self.query)
+
+        return ('Sessionizing completed, number of {0:s} session created:'
```
tomchop:
```suggestion
        return ('Sessionizing completed, number of {0:s} sessions created:'
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+            session_num: Session number for the event.
+        """
+        event.add_attributes({self.session_type: session_num})
+        event.commit()
+
+    def process_event(self, event):
+        """Process event depending on if the event is significant for the
+        searched event sequence.
+
+        Event is significant if it matches the current
```
tomchop: Please try to reach the max number of words in a line before breaking.

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if self.recording:
+            self.event_storage.append(event)
+
+    def flush_events(self, drop=False):
+        """Annotates or clears events in event_storage according to the flag
+        drop and resets session recording state.
+
+        Args:
+            drop: Indicator if the stored events should not be commited.
```
tomchop: This sentence is a bit unclear.

```suggestion
            drop: If True, the stored events will not be committed to the database.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def match_event(self, event):
+        """Compare event with the event to search for in the event sequence.
+
+        Args:
+            event: Event to compare with the one in the event sequence.
+
+        Returns:
+            If event matches the event to search for in the event
+        sequence.
```
tomchop:
```suggestion
              sequence.
```

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+        """
+        event_to_match = self.event_seq[self.num_event_to_find]
+
+        for key, value in event_to_match.items():
+            if value != event.source.get(key):
+                return False
+        return True
+
+    def get_query_string(self):
```
tomchop: If this function is meant to be overriden by other classes, please specify it in the docstring.

timesketch/lib/analyzers/sequence_sessionizer.py
```diff
@@ -0,0 +1,158 @@
...
+        """Generate query string for all events allocated with session_type
+        attribute.
+
+        Returns:
+            Query string for Elasticsearch.
+        """
+        query_string = self.session_type + ':*'
```
tomchop:
```suggestion
        return self.session_type + ':*'
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,430 @@
...
+
+class TestSequenceSessionizerPlugin(BaseTest):
+    """Tests the functionality of the sequence sessionizing sketch analyzer."""
+    @mock.patch('timesketch.lib.analyzers.interface.ElasticsearchDataStore',
+                MockDataStore)
+    def test_session_type_none(self):
+        """Test session_type is not None."""
+        index = 'test_index'
+        sketch_id = 1
+        self.assertRaises(RuntimeError, sessionizer=\
```
tomchop: It's better to write it this way:

```python
self.assertRaises(
    RuntimeError, 
    sessionizer=NoneSessionTypeSequenceSessionizer(index, sketch_id))
```

Same for the other assertions in the tests.

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,430 @@
...
+                               'event_stream',
+                               return_value=_create_mock_event(
+                                   0, 2, one_session_args, time_diffs=[1])):
+            index = 'test_index'
+            sketch_id = 1
+            sessionizer = ManyEventsSequenceSessionizer(index, sketch_id)
+            message = sessionizer.run()
+            self.assertEqual(
+                message,
+                'Sessionizing completed, number of {0:s} session created: 1'.
```
tomchop: Quick note: if you fix the string in the previous comments, you'll have to adjust this string too.

```suggestion
                'Sessionizing completed, number of {0:s} sessions created: 1'.
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,430 @@
...
+            # significant events from the event sequence considered as a session
+            # are part of the significant events' session.
+            for i in range(0, 101):
+                event = ds.get_event('test_index', str(i), stored_events=True)
+                self.assertEqual(event['_source'][sessionizer.session_type], 1)
+
+    @mock.patch('timesketch.lib.analyzers.interface.ElasticsearchDataStore',
+                MockDataStore)
+    def test_multiple_sessions(self):
+        """Test multiple sessions are finded and allocated correctly."""
```
tomchop:
```suggestion
        """Test multiple sessions are found and allocated correctly."""
```

timesketch/lib/analyzers/sequence_sessionizer_test.py
```diff
@@ -0,0 +1,430 @@
...
+            message = sessionizer.run()
+            self.assertEqual(
+                message,
+                'Sessionizing completed, number of {0:s} session created: 2'.
+                format(sessionizer.session_type))
+
+            ds = MockDataStore('test', 0)
+            for i in range(0, 100):
+                event = ds.get_event('test_index', str(i), stored_events=True)
+                self.assertEqual(event['_source'][sessionizer.session_type], 1)
```
tomchop: In tests, I would rather be explicit here and use the actual hardcoded string instead of the variable `sessionizer.session_type` when possible.

timesketch/lib/analyzers/sequence_sessionizer.py
nnyx7: My logic there is that if someone chosed to have all of the fields, he probably needs all of them.
If someone didn't specified the return_fields in his sessionizer, then return_fields remains ['timestamp'] and I am going to fill the rest of the fields that are needed. And a field is needed if event from event_seq has it.
If someone specified return_fields but not with the right attributes for some reason, I am not removing the not needed ones, I am just adding the attributes that are needed but missing. So in that case return_fields can end up with not used attributes.
But the another side of situation is if the person needs to access other attributes too, then to removing them is unexpected behaviour.
In the first situation I am adding extra complexity but making sure that the code will work as expected in the sequence_sessionizer. In the second situation, I can break someone's logic.

nnyx7: What do you think it's better to be done?

PR #971, opened 2019-09-05
Session resource

katemacleod:
This is for access from the frontend.

- Session class added in Sketch.py
- add_session method added in interface.py
- sessionizer.py updated to use this method


timesketch/frontend/dist/index.html
```diff
@@ -1 +1 @@
...
\ No newline at end of file
+<!DOCTYPE html><html lang=en><head><meta name=csrf-token content="{{ csrf_token() }}"><meta charset=utf-8><meta http-equiv=X-UA-Compatible content="IE=edge"><meta name=viewport content="width=device-width,initial-scale=1"><link rel=icon href=/dist/favicon.ico><title>Timesketch</title><link href=/dist/css/chunk-common.045a93d9.css rel=preload as=style><link href=/dist/js/chunk-common.276d8a50.js rel=preload as=script><link href=/dist/js/chunk-vendors.1f9d0459.js rel=preload as=script><link href=/dist/js/index.0c45ccdf.js rel=preload as=script><link href=/dist/css/chunk-common.045a93d9.css rel=stylesheet></head><body><div id=app></div><script src=/dist/js/chunk-vendors.1f9d0459.js></script><script src=/dist/js/chunk-common.276d8a50.js></script><script src=/dist/js/index.0c45ccdf.js></script></body></html>
```
katemacleod: This should not be here but I can't find a way to sensibly remove it - permission is denied whenever I try to change it.

berggren:
Why do we need to add the sessions to the SQL database? I try to avoid adding stuff to it due to the complexity with schema changes. How do you envision accessing this in the frontend? Let's try to do this without the SQL database, and just with ES and JS.

Is it because you want to track start & end times?

katemacleod:
> Why do we need to add the sessions to the SQL database? I try to avoid adding stuff to it due to the complexity with schema changes. How do you envision accessing this in the frontend? Let's try to do this without the SQL database, and just with ES and JS.
> 
> Is it because you want to track start & end times?

I thought querying elasticsearch and processing the results would take too long (due to having to find the start and end times for each session without knowing how many sessions / types of sessions there will be). I will try it using elasticsearch and let you know how I get on!

berggren:
Sounds good! I would like to know a little bit more on the use-case here. Let's have a quick chat on Monday.

katemacleod:
> Sounds good! I would like to know a little bit more on the use-case here. Let's have a quick chat on Monday.

Hi, I've managed to do this by querying elasticsearch, fixing the bug I had yesterday. The only issue is that the names for the types of sessions are hard coded, meaning that if someone writes another analyzer they would have to add it to the list for it to appear in the frontend. I can't see another way of doing this using elasticsearch - what are your thoughts?

berggren:
Hard coding the names is fine, I can come up with a more dynamic way at a later time.

timesketch/api/v1/resources.py
```diff
@@ -1979,3 +1980,105 @@ def get(self, searchindex_id):
...
+                                'field': 'timestamp'
+                            }
+                        }
+                    }
+                }
+            }
+        }
+
+        for session_type in session_types:
+            id_agg_spec['aggregations']['term_count']['terms']['field'] =\
```
berggren: Avoid "\" to break the line. Maybe make ```id_agg_spec['aggregations']['term_count']['terms']['field']``` a variable and then break the line?

timesketch/api/v1/resources.py
```diff
@@ -1979,3 +1980,105 @@ def get(self, searchindex_id):
...
+                'session_id.%s.keyword' % session_type
```
berggren: use:
```
 'session_id.{}.keyword'.format(session_type)
```

timesketch/api/v1/resources.py
```diff
@@ -1979,3 +1980,105 @@ def get(self, searchindex_id):
...
+            session_count = 0
+
+            for bucket in buckets:
+                if session_count == MAX_SESSIONS:
+                    isTruncated = True
+                    break
+                session_count += 1
+
+                session_id = bucket['key']
+                timestamp_agg_spec['aggregations']['timestamp_range'] \
```
berggren: Same here, avoid using "\" to break.

timesketch/frontend/dist/index.html
berggren: It's fine, I'll rebuild the UI later.

PR #972, opened 2019-09-05
SSH bruteforce session analyzer

katemacleod:
Solution for issue #953


katemacleod:
The majority of the changes in base_sessionizer_test.py, and all of the changes to sessionizer_test.py are the same as #977, they are just included here  so the tests pass.

PR #973, opened 2019-09-11
SSH sessionizing sketch analyser

nnyx7:
The SSH sessionizer is labeling all events from a SSH connection with
the client's IP address and port.

### DIFF
--- a/timesketch/lib/analyzers/ssh_sessionizer_test.py
+++ b/timesketch/lib/analyzers/ssh_sessionizer_test.py
@@ -7,6 +7,8 @@ import mock
 from timesketch.lib.analyzers.ssh_sessionizer import SSHSessionizerSketchPlugin
 from timesketch.lib.testlib import BaseTest
 from timesketch.lib.testlib import MockDataStore
+# TODO _create_mock_event will be renamed in another pull request. It's name
+# should be also changed here.
 from timesketch.lib.analyzers.sequence_sessionizer_test \
     import _create_mock_event
 
@@ -65,15 +67,13 @@ class TestSSHSessionizerPlugin(BaseTest):
         message = sessionizer.run()
         self.assertEqual(
             message,
-            'Sessionizing completed, number of {0:s} sessions created: 1'.
-            format(sessionizer.session_type))
-
+            'Sessionizing completed, number of ssh_session sessions created: 1'
+        )
         session_id = '1.1.1.1_1'
         #pylint: disable=unexpected-keyword-arg
         event = datastore.get_event('test_index', '0', stored_events=True)
-        self.assertEqual(
-            event['_source']['session_id'][sessionizer.session_type],
-            session_id)
+        self.assertEqual(event['_source']['session_id']['ssh_session'],
+                         session_id)
 
     @mock.patch('timesketch.lib.analyzers.interface.ElasticsearchDataStore',
                 MockDataStore)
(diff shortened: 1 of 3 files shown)


timesketch/lib/analyzers/ssh_sessionizer_test.py
```diff
@@ -0,0 +1,175 @@
...
+
+from __future__ import unicode_literals
+
+import mock
+
+from timesketch.lib.analyzers.ssh_sessionizer import SSHSessionizerSketchPlugin
+from timesketch.lib.testlib import BaseTest
+from timesketch.lib.testlib import MockDataStore
+from timesketch.lib.analyzers.sequence_sessionizer_test \
+    import _create_mock_event
```
tomchop: functions with a leading `_` are supposed to be private. Please change the name of the function if you want to import it here.

timesketch/lib/analyzers/ssh_sessionizer_test.py
```diff
@@ -0,0 +1,175 @@
...
+        sessionizer = SSHSessionizerSketchPlugin(index, sketch_id)
+        sessionizer.datastore.client = mock.Mock()
+        datastore = sessionizer.datastore
+
+        _create_mock_event(datastore, 0, 1, one_ssh_session_args)
+
+        message = sessionizer.run()
+        self.assertEqual(
+            message,
+            'Sessionizing completed, number of {0:s} sessions created: 1'.
```
tomchop: You can use the hardcoded type here (and in other tests), since this test will only be executed with `SSHSessionizerSketchPlugin`

timesketch/lib/analyzers/ssh_sessionizer_test.py
```diff
@@ -0,0 +1,175 @@
...
+                MockDataStore)
+    def test_session_starts_with_connection_event(self):
+        """Test a session is created if it starts with SSH connection event."""
+        index = 'test_index'
+        sketch_id = 1
+        sessionizer = SSHSessionizerSketchPlugin(index, sketch_id)
+        sessionizer.datastore.client = mock.Mock()
+        datastore = sessionizer.datastore
+
+        _create_mock_event(datastore, 0, 1, one_ssh_session_args)
```
tomchop: How are you calling `_create_mock_event` here? it seems like the arguments are not the same. From what I have:

```python
def _create_mock_event(datastore,
                       event_id,
                       quantity,
                       time_diffs=None,
                       source_attrs=None):
    """
    Loads in the datastore mock events that based on the given arguments.

    Args:
        datastore: An instance of MockDataStore.
        event_id: Desired ID for the Event.
        quantity: The number of Events to be generated.
        time_diffs: A list of time differences between the generated
            Events.
        source_attrs: Dictionary of attributes to add to the source of the
            generated events.
    """
```

timesketch/lib/analyzers/ssh_sessionizer_test.py
```diff
@@ -0,0 +1,175 @@
...
+        message = sessionizer.run()
+        self.assertEqual(
+            message,
+            'Sessionizing completed, number of {0:s} sessions created: 1'.
+            format(sessionizer.session_type))
+
+        session_id = '1.1.1.1_1'
+        #pylint: disable=unexpected-keyword-arg
+        event = datastore.get_event('test_index', '0', stored_events=True)
+        self.assertEqual(
```
tomchop: Also please use the hardcoded value here

timesketch/lib/analyzers/ssh_sessionizer_test.py
nnyx7: I'm calling _create_mock_event from timesketch/lib/analyzers/sequence_sessionizer_test.

PR #964, opened 2019-08-20
Logon sessions sketch analyzer

katemacleod:
Solution to issue #955

Issue #952, opened 2019-08-12
Screensaver lock / unlock analyzer


tomchop:

jaegeral:
@tomchop / @katemacleod is still an open thing?

berggren:
Closing this out for now seeing there hasn't been any update in a while.

PR #976, opened 2019-09-13
Evtx sessionizer

katemacleod:
Replaces previous pull request #964 (changes to tests needed to be made for compatibility with other changes. These tests would be overwritten by this pull request anyway). Solution for issues #955 and #952


tomchop:
Manually setting CLA to yes because we rebased from the GitHub UI

PR #977, opened 2019-09-16
Remove multiple inheritance in test classes

nnyx7:

PR #981, opened 2019-09-19
Psexec sessionizer

nnyx7:
This pull request depends on #977.


tomchop:
Adding the `cla: yes` label as the signers got messed up by github

PR #982, opened 2019-09-25
Fixing aggregations in the API client

kiddinn:
Removing the need to depend on the aggregation manager, and thus removing dependencies on timesketch library in the API client.

Also adding support for storing and listing stored aggregations.

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -25,7 +25,6 @@ import requests
 from requests.exceptions import ConnectionError
 import altair
 import pandas
-import numpy
 from .definitions import HTTP_STATUS_CODE_20X
 
 
@@ -728,7 +727,7 @@ class Sketch(BaseResource):
             self.api.api_root, self.id)
 
         form_data = {
-            'name': aggregator_name,
+            'name': name,
             'description': description,
             'agg_type': aggregator_name,
             'chart_type': chart_type,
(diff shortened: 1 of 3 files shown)


api_client/python/timesketch_api_client/client.py
```diff
@@ -688,6 +709,43 @@ def run_aggregator(
...
+        }
+
+        response = self.api.session.post(resource_url, json=form_data)
+        response_dict = response.json()
+
+        objects = response_dict.get('objects', [])
+        if not objects:
+          return
+
+        _ = self.lazyload_data(refresh_cache=True)
```
berggren: no need to declare a variable here

api_client/python/timesketch_api_client/client.py
```diff
@@ -874,32 +932,23 @@ def table(self):
...
+            self.name, self.parameters, as_pandas=False)
+
+        meta = data.get('meta', {})
+        vega_spec = meta.get('vega_spec')
+
+        if not vega_spec:
+            return altair.Chart(pandas.DataFrame()).mark_point()
 
-        return chart_obj.generate()
+        vega_string = json.dumps(vega_spec)
```
berggren: s/vega_string/vega_spec_string/

api_client/python/timesketch_api_client/client.py
kiddinn: ignored after offline chat

api_client/python/timesketch_api_client/client.py
kiddinn: ack

PR #990, opened 2019-10-04
Update docker-compose to correct elasticsearch volume 

exFill:
**Problem:** Elasticsearch data was not persistent across down/up commands using docker-compose yaml file provided.

**Solution:** 

1. Research found [this](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docker.html) documentation which revealed  the docker volume for elasticsearch was referencing the incorrect source directory.

1. After fixing the volume reference, testing found a directory permissions problem with /var/lib/elasticsearch.  So added documentation to pre-create the directory and set appropriate permissions.

1. Additional documentation was added to highlight a needed check for max_map_count -- see [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docker.html#docker-cli-run-prod-mode).

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -19,13 +19,6 @@ cd timesketch
 ### Verify the Kernel Settings
 Follow the official instructions [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docker.html#docker-cli-run-prod-mode)
 
-### Create elasticsearch directory
-
-```
-mkdir /var/lib/elasticsearch
-chown 1000:1000 /var/lib/elasticsearch
-```
-
 ### Build and Start Containers
 
 ```shell
--- a/docker/README.md
+++ b/docker/README.md
@@ -16,6 +16,16 @@ Follow the official instructions [here](https://docs.docker.com/compose/install/
 git clone https://github.com/google/timesketch.git
 cd timesketch
 ```
+### Verify the Kernel Settings
+Follow the official instructions [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docker.html#docker-cli-run-prod-mode)
+
+### Create elasticsearch directory
+
+```
+mkdir /var/lib/elasticsearch
+chown 1000:1000 /var/lib/elasticsearch
+```
+
 ### Build and Start Containers
 
 ```shell
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -39,7 +39,7 @@ services:
       - "9300:9300"
     restart: always
     volumes:
-      - /var/lib/elasticsearch:/usr/share/elasticsearch/data/elasticsearch
+      - /var/lib/elasticsearch:/usr/share/elasticsearch/data
 
   postgres:
     image: postgres
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -28,10 +28,11 @@ services:
       - ../:/usr/local/src/timesketch/
 
   elasticsearch:
+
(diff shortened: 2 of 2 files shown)


docker/README.md
```diff
@@ -16,6 +16,16 @@ Follow the official instructions [here](https://docs.docker.com/compose/install/
...
 cd timesketch
 ```
+### Verify the Kernel Settings
+Follow the official instructions [here](https://www.elastic.co/guide/en/elasticsearch/reference/6.4/docker.html#docker-cli-run-prod-mode)
+
+### Create elasticsearch directory
+
+```
+mkdir /var/lib/elasticsearch
+chown 1000:1000 /var/lib/elasticsearch
```
berggren: 1000? I guess this needs to be the docker user?

exFill: uid,gid=1000 is the elasticsearch user inside the elasticsearch docker.  I wasn't sure how to cleanly do this without using the raw uid/gid.  I haven't tried, but I guess you could also run something like this after the docker-compose up command:

`sudo docker exec -ti docker_elasticsearch_1 chown elasticsearch:elasticsearch /var/share/elasticsearch/data`

This is the effect that is needed, but I need to see if this works before changing up the readme to this command.

exFill: So because the docker-compose yaml file has a "restart: always" and the elasticsearch docker keeps failing because it can't create a node directory in the volume directory /var/lib/elasticsearch, there doesn't seem to be time enough to execute the command cleanly within docker before it restarts:

```
$ sudo docker exec -it docker_elasticsearch_1 chown elasticsearch:elasticsearch /usr/share/elasticsearch
Error response from daemon: Container 5efcf17ad8c6b8e78c105eb73d6f01668cbbfe8bcbf1d8d779812ef39cbe2ab0 is restarting, wait until the container is running
```

Precreating the /var/lib/elasticsearch directory and setting the owner to uid,gid=1000 seems to be the only option I can come up with.  Does someone else have a cleaner way of doing this in docker?

```
$ sudo docker exec -it docker_elasticsearch_1 ls -al /usr/share/elasticsearch/data 
total 12
drwxr-xr-x  3 elasticsearch elasticsearch 4096 Oct  7 22:34 .
drwxrwxr-x 11 elasticsearch root          4096 Oct  7 22:34 ..
drwxrwxr-x  3 elasticsearch root          4096 Oct  7 22:34 nodes

$ ls -al /var/lib/elasticsearch 
total 12
drwxr-xr-x  3 1000  1000  4096 Oct  8 09:34 .
drwxr-xr-x 83 root  root  4096 Oct  8 09:27 ..
drwxrwxr-x  3 1000  root  4096 Oct  8 09:34 nodes
```

docker/docker-compose.yml
```diff
@@ -28,10 +28,11 @@ services:
       - ../:/usr/local/src/timesketch/
 
   elasticsearch:
+
+    environment:
+      - TAKE_FILE_OWNERSHIP=1
```
exFill: After reviewing the repo for elasticsearch Docker I found the [entrypoint](https://github.com/elastic/elasticsearch/blob/master/distribution/docker/src/docker/bin/docker-entrypoint.sh) which referenced an environment variable at line 95 used to chmod the data directory to the correct uid,gid values.  

Adding that enviroment variable did the trick and now /var/lib/elasticsearch does not need to be precreated and chown 1000:1000 does not need to be run on it.

berggren: Sounds good!

Issue #997, opened 2019-10-18
Run analyzer on already imported timelines


taishi8117:
We should have a feature in the UI to support running analyzers on imported timelines.

kiddinn:
This depends on #998, that is having an API to manually run them.

Pushing that in, and then we can add the UI element.

kiddinn:
assigning the issue to @berggren due to the UI element.

The backend is ready, API is in and client supports this right now.

berggren:
UI is ready and in review: #1019

PR #998, opened 2019-10-18
Adding the ability to manually run an analyzer in the API

kiddinn:
Adding two things:

+ API to manually run an analyzer on a timeline
+ Updates to the API client to call this new API

This is part of #997


api_client/python/timesketch_api_client/client.py
```diff
@@ -637,6 +637,47 @@ def explore(self,
...
+                further parameters.
+            timeline_id: the data store index for the timeline. This is optional
+                and only required if timeline_name is not set.
+            timeline_name: the name of the timeline in the timesketch UI. This
+                is optional and only required if timeline_id is not set.
+
+        Returns:
+            A string with the results of the API call to run the analyzer.
+        """
+        if not timeline_id and not timeline_name:
```
berggren: This is risky, there can be multiple timelines with the same name in a sketch. We don't enforce unique names today. I say that you need to specify the ID here to be on the safe side.

timesketch/api/v1/resources.py
```diff
@@ -1658,6 +1660,85 @@ def post(self):
...
+    """Resource to get all timelines for sketch."""
+
+    @login_required
+    def get(self, sketch_id):
+        """Handles GET request to the resource.
+
+        Returns:
+            A list of all available analyzer names.
+        """
+        _ = sketch_id
```
berggren: You want to do ```sketch = Sketch.query.get_with_acl(sketch_id)``` here as well to ensure that the caller has the correct permissions

timesketch/api/v1/resources.py
```diff
@@ -1658,6 +1660,85 @@ def post(self):
...
+            return abort(
+                HTTP_STATUS_CODE_FORBIDDEN,
+                'User does not have write permission on the sketch.')
+
+        timeline_name = form.timeline_name.data
+        timeline_id = form.timeline_id.data
+
+        search_index = None
+        if timeline_name:
+            for timeline in sketch.timelines:
```
berggren: Hmm, i see what you are doing here. You will analyze all timelines with the name. Is this what the caller expects?

timesketch/api/v1/resources.py
kiddinn: ack

api_client/python/timesketch_api_client/client.py
kiddinn: changed this a bit... removed the entire notion of timeline_name from the actual API and here I'm going to check if you use timeline name and there is only one result, then I'll proceed.

timesketch/api/v1/resources.py
kiddinn: changed

kiddinn:
ok, ready for another round

PR #999, opened 2019-10-18
Fixed the TypeError: delimiter must be string, not unicode error.

wajihyassine:
Fixed the TypeError: delimiter must be string, not unicode error importing data into timesketch by adding a check for the delimiter type and decoding it if it isn't a string as csv.DictReader does not accept any other type but string for the delimiter argument.


kiddinn:
can you fix the linter error?

PR #1004, opened 2019-10-23
Added more upload options to API client

kiddinn:
Adding:
 - Ability to upload pandas data frames.
 - Changed the API to allow appending data to an index
 - Changed the AIP to allow for only indexing newly added data, and no analyzer runs
 - Added an update streamer to allow for streaming results from parsing to TS.

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -15,6 +15,7 @@
 from __future__ import unicode_literals
 
 import json
+import tempfile
 import uuid
 
 # pylint: disable=wrong-import-order
@@ -26,6 +27,11 @@ from requests.exceptions import ConnectionError
 import altair
 import pandas
 from .definitions import HTTP_STATUS_CODE_20X
+
+
+def format_data_frame_row(row, format_message_string):
+    """Return a formatted data frame using a format string."""
+    return format_message_string.format(**row)
 
 
 class TimesketchApi(object):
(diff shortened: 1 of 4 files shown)


kiddinn:
This is still WIP, just uploading to demonstrate a bit where I was going with this.

Still need to test it (haven't tested it once), which will most likely results in some unit tests being added as well as major or minor changes to the actual code.

api_client/python/timesketch_api_client/client.py
```diff
@@ -503,6 +504,67 @@ def upload(self, timeline_name, file_path):
...
+            string_items = []
+            for column in data_frame.columns:
+                if 'time' in column:
+                    continue
+                elif 'timestamp_desc' in column:
+                    continue
+                string_items.append('{0:s} = {{0!s}}'.format(column))
+            format_message_string = ' '.join(string_items)
+
+        response = None
```
berggren: maybe rename to streamer_response ?

api_client/python/timesketch_api_client/client.py
```diff
@@ -503,6 +504,67 @@ def upload(self, timeline_name, file_path):
...
+            streamer.set_message_format_string(format_message_string)
+
+            streamer.add_data_frame(data_frame)
+            response = streamer.response
+
+        if not response:
+            return 'No return value.'
+
+        return_lines = []
+        for timesketch_object in response.get('objects', []):
```
berggren: sketch_object?

timesketch/api/v1/resources.py
```diff
@@ -1368,6 +1369,50 @@ def post(self, sketch_id):
                 annotations, status_code=HTTP_STATUS_CODE_CREATED)
 
 
+class AnalyzerPipelineResource(ResourceMixin, Resource):
```
berggren: Maybe we want to extend this to also accept an analyzer name? So we can run a specific analyzer instead of all.

timesketch/api/v1/resources.py
```diff
@@ -1407,45 +1452,58 @@ def post(self):
...
+                    description=searchindex.description,
+                    sketch=sketch,
+                    user=current_user,
+                    searchindex=searchindex)
+                timeline.set_status('processing')
+                sketch.timelines.append(timeline)
+                db_session.add(timeline)
+                db_session.commit()
+
+        stream = form.enable_stream.data
```
berggren: s/stream/enable_stream/

timesketch/api/v1/resources.py
```diff
@@ -1407,45 +1452,58 @@ def post(self):
...
         if not isinstance(index_name, six.text_type):
             index_name = codecs.decode(index_name, 'utf-8')
 
         file_path = os.path.join(upload_folder, filename)
         file_storage.save(file_path)
 
-        # Create the search index in the Timesketch database
-        searchindex = SearchIndex.get_or_create(
+        # Check if search index already exists.
+        searchindex = SearchIndex.query.filter_by(
```
berggren: hm, why not use get_or_create() here?

timesketch/api/v1/routes.py
```diff
@@ -53,6 +54,7 @@
     (SketchListResource, '/sketches/'),
     (SketchResource, '/sketches/<int:sketch_id>/'),
     (AnalyzerRunResource, '/sketches/<int:sketch_id>/analyzer/'),
+    (AnalyzerPipelineResource, '/sketches/<int:sketch_id>/analyzer/auto_run/'),
```
berggren: auto_run? I would call it /run/ and accept list of analyzer names as well. If none is provided, run all analyzers, WDYT?

docs/UploadDataViaAPI.md
```diff
@@ -0,0 +1,201 @@
+# Create Timeline From Other Sources
```
berggren: Kudos for great documentation :)

api_client/python/timesketch_api_client/client.py
```diff
@@ -503,6 +504,67 @@ def upload(self, timeline_name, file_path):
             searchindex=timeline['searchindex']['index_name'])
         return timeline_obj
 
+    def upload_data_frame(
+            self, data_frame, timeline_name, format_message_string=''):
```
berggren: Can we have format strings loaded from a YAML file? Maybe key off of the type (name) of the data?
Fine to put in a TODO and follow up in a later PR.

timesketch/api/v1/resources.py
kiddinn: I don't want to create one here if it doesn't exist, only querying whether it exists or not.

timesketch/api/v1/resources.py
kiddinn: done

timesketch/api/v1/routes.py
kiddinn: this is removed, expanded the use case of the analyzer run resource instead.

kiddinn:
@berggren PTAL

timesketch/api/v1/resources.py
berggren: cool, SG

timesketch/api/v1/resources.py
berggren: ack

timesketch/api/v1/resources.py
```diff
@@ -1795,15 +1816,30 @@ def post(self, sketch_id):
                     'No timeline was found, make sure you\'ve got the correct '
                     'timeline ID or timeline name.'))
 
-        analyzer_name = form.analyzer_name.data
-        analyzer_kwargs = form.analyzer_kwargs.data
+        analyzer_name = form.get('analyzer_name')
```
berggren: This is a list, so let's call it "analyzer_names"

kiddinn: (offline discussions)

PR #1007, opened 2019-10-24
New UI: Comment support

berggren:
Add and display comments.

Screenshot 2019-10-24 at 14 35 05


timesketch/frontend/dist/login.html
```diff
@@ -15,4 +15,8 @@
...
\ No newline at end of file
+<<<<<<< HEAD
```
kiddinn: I'm assuming the <<< JHEAD is still an artifact from merging...

berggren: fixed

PR #1015, opened 2019-10-28
New UI: Sharing controls

berggren:
This PR adds sharing controls to the new UI.

Screenshot 2019-10-28 at 22 19 03


kiddinn:
There are some merge conflicts that need to be resolved

timesketch/api/v1/resources.py
```diff
@@ -2148,6 +2158,84 @@ def get(self, searchindex_id):
...
+                HTTP_STATUS_CODE_FORBIDDEN,
+                'The user does not have write permission on the sketch.')
+
+        for username in form.get('users', []):
+            base_username = username.split('@')[0]
+            base_username = base_username.strip()
+            user = User.query.filter_by(username=base_username).first()
+            if user:
+                sketch.grant_permission(permission='read', user=user)
+                sketch.grant_permission(permission='write', user=user)
```
kiddinn: do we not want to have this more granular? That is have a config of what you want to grant here? 

want to grant just read, or both? I see some instance where I would prefer to just grant a read permission to a sketch...

timesketch/api/v1/resources.py
```diff
@@ -2148,6 +2158,84 @@ def get(self, searchindex_id):
...
+
+    @login_required
+    def post(self, sketch_id):
+        """Handles POST request to the resource.
+
+        Args:
+            sketch_id: Integer primary key for a sketch database model
+        """
+        sketch = Sketch.query.get_with_acl(sketch_id)
+        form = request.json
```
kiddinn: why not use the forms and have the request.json call directly?

timesketch/api/v1/resources.py
```diff
@@ -2148,6 +2158,84 @@ def get(self, searchindex_id):
...
+            if user:
+                sketch.grant_permission(permission='read', user=user)
+                sketch.grant_permission(permission='write', user=user)
+
+        for group in form.get('groups', []):
+            group = Group.query.filter_by(name=group).first()
+            # Only add groups publicly visible or owned by the current user
+            if not group.user or group.user == current_user:
+                sketch.grant_permission(permission='read', group=group)
+                sketch.grant_permission(permission='write', group=group)
```
kiddinn: same here... with the more granular controls

timesketch/api/v1/resources.py
```diff
@@ -2148,6 +2158,84 @@ def get(self, searchindex_id):
...
+
+        for username in form.get('remove_users', []):
+            user = User.query.filter_by(username=username).first()
+            sketch.revoke_permission(permission='read', user=user)
```
kiddinn: if more granular, you would have to then check permissions first and remove those that exist

timesketch/api/v1/resources.py
berggren: Yes, very much. But this will be done at a later stage. I'll put a todo in the code.

timesketch/api/v1/resources.py
berggren: As discussed offline - Trying to move away from the old forms library for validation. We pass json nowadays, and forms don't support list of strings for example.

timesketch/api/v1/resources.py
berggren: Ack.

timesketch/api/v1/resources.py
berggren: yes. Opened up an issue for it: https://github.com/google/timesketch/issues/1016

berggren:
Done, @kiddinn PTAL

Issue #1017, opened 2019-10-31
Bug in the importer (py3 vs py2)


kiddinn:
When importing a file, came across:

```
  File "...ib/python2.7/site-packages/timesketch/lib/tasks.py", line 467, in run_csv_jsonl
    for event in read_and_validate(source_file_path):
  File ".../lib/python2.7/site-packages/timesketch/lib/utils.py", line 66, in read_and_validate_csv
    with open(path, 'r', encoding='utf-8') as fh:
TypeError: 'encoding' is an invalid keyword argument for this function
```

This is an issue when run on py2 (open there doesn't support encoding option)

Solution to use io.open instead of open

Another issue that has come up is:

```
Traceback (most recent call last):
  File "...lib/python2.7/site-packages/timesketch/lib/tasks.py", line 467, in run_csv_jsonl
    for event in read_and_validate(source_file_path):
  File ".../lib/python2.7/site-packages/timesketch/lib/utils.py", line 81, in read_and_validate_csv
    for row in reader:
  File "/usr/lib/python2.7/csv.py", line 108, in next
    row = self.reader.next()
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2019' in position 92: ordinal not in range(128)
```

This is another py2 issue

kiddinn:
fixed

PR #1018, opened 2019-10-31
Fixing #1017 and one other minor issue.

kiddinn:
Adding a fix to #1017, if you attempt to upload a file and the TS backend is still running py2 then you'll have an exception raised. So adding backward compatibility there as well as fixing a minor issue in the importer.


kiddinn:
Both issues are now fixed and have been tested.

PR #1028, opened 2019-11-13
Initial Sigma support

Onager:
Initial work for Sigma integration.

Adding more mappings would be ongoing work, to make analyzer compatible with different rulesets. I can add more tests in a follow up PR, once some of the analyzer testing code is complete/merged.


timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -0,0 +1,96 @@
...
+from sigma.backends import elasticsearch as sigma_elasticsearch
+import sigma.configuration as sigma_configuration
+from sigma.parser import collection as sigma_collection
+
+
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+
+
+class SigmaPlugin(interface.BaseIndexAnalyzer):
```
berggren: Note: I think we want to move this to a SketchAnalyzer later, so it can be run from a sketch as well as creating Views etc. This can be changed later though seeing the only action in this version is tagging.

Onager: Changed this to a sketch analyzer, thought I think we'll want to think more about this in the future. If we use rules that contain aggregations for example, we don't necessarily want to aggregate over the whole sketch if contains data from multiple machines.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -0,0 +1,96 @@
...
+        for rule_filename in os.listdir(rules_path):
+            tag_name, _ = rule_filename.rsplit('.')
+            full_path = os.path.join(rules_path, rule_filename)
+            with open(full_path, 'r') as rule_file_content:
+                query = rule_file_content.read()
+                parser = sigma_collection.SigmaCollectionParser(
+                    query, self.sigma_config, None)
+                results = parser.generate(sigma_backend)
+                for result in results:
+                    print(result)
```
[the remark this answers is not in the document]
Onager: Refactored this.

timesketch/lib/analyzers/sigma_tagger.py
berggren: Ah, yeah. So you can choose what timelines to operate on, so it's totally fine to just run on the current timeline (this is the default).

berggren:
@Onager Seeing that sigmatools is py3 only, it fails (naturally) when run on py2. Until we remove py2 support we need to check for this in the tests, and also not register the analyzer if run on py2. 

When this is fixed this one is ready to merge.

berggren:
This is a good oppurtunity to sunset py2 support. I'll do that in another PR before merging this one.

PR #1038, opened 2019-11-19
Adding aggregation information to the API and API client.

kiddinn:
Adding the ability to get more information about aggregation objects into the API and API client.

aggregator


api_client/python/timesketch_api_client/client.py
```diff
@@ -322,6 +322,51 @@ def get_sketch(self, sketch_id):
         """
         return Sketch(sketch_id, api=self)
 
+    def get_aggregator_information(self, name='', as_pandas=False):
```
berggren: can we name this get_aggregator_info instead to be a little more sparse?

PR #1039, opened 2019-11-19
Cleaning up analyzer tags

berggren:
we want them all to adhear to foo-bar format.


timesketch/lib/analyzers/domain.py
```diff
@@ -110,10 +110,10 @@ def run(self):
                 cdn_counter[cdn_provider] += 1
 
             if domain in common_domains:
-                tags_to_add.append('common_domain')
+                tags_to_add.append('common-domain')
```
kiddinn: can we just skip this tag?

I'm wondering if this is providing value or if it is just noise? we can add an attribute maybe, to make it easier to filter things out instead, WDYT?

timesketch/lib/analyzers/ntfs_timestomp.py
```diff
@@ -121,16 +121,14 @@ def run(self):
 
         timestomps = 0
         for file_info in file_infos.values():
-            if self.handle_timestomp(file_info):
+            if self.is_suspicious(file_info):
```
kiddinn: I know this is not the correct place for this, but since you are making changes, can you change the docstring for the `is_suspicious` function, since it is longer than a single line, and it shouldn't be.

timesketch/lib/analyzers/domain.py
berggren: I would love to get less tags :) so SGTM. I'll move this to an attribute instead.

timesketch/lib/analyzers/ntfs_timestomp.py
berggren: Done.

berggren:
Done, @kiddinn PTAL

Issue #1052, opened 2019-12-11
Sorting order dropdown shows duplicates


vikahl:
The sorting order dropdown shows both the selected order as well asc/desc resulting in duplicates.

<img width="338" alt="Screenshot 2019-12-11 at 15 28 03" src="https://user-images.githubusercontent.com/3051877/70629956-0ae25d00-1c2b-11ea-857e-96ec4ad4ba16.png">

PR #1054, opened 2019-12-12
Fix duplicate sorting order in sketch explore sort order dropdown

vikahl:
Fixes: #1052

Vue will handle the selected value by itself, it does not need to be explicitly specified. See https://vuejs.org/v2/guide/forms.html#Select


vikahl:
@googlebot I signed it!

PR #1055, opened 2019-12-12
Added dynamic form definition for the feature extraction analyzer.

kiddinn:
This is the first step to add the ability to run analyzer with custom configs in the UI, in this case to extract a specific feature that is not yet defined in the config file.


kiddinn:
need to make few minor adjustments to the form

kiddinn:
> need to make few minor adjustments to the form

and ready now for a review

Issue #1060, opened 2019-12-17
Uploading 1GB plus plaso file to timesketch fails


shannaniggans:
I am running timesketch in docker with docker-compose. When i log onto the UI and upload plaso files, files over 1GB failed with the following messages. The standard psort messages, but the last 2 lines are only present on files that are larger than 1GB and the file upload times out. 

Ubuntu server 18.04, latest version of docker, ran update/upgrade on everything before running up docker and docker-compose. Have 8GB assigned to the VM, updated vm.max_map_count=262144 else it doesn't run anyway.

`timesketch_1     | [2019-12-17 05:22:54,924: INFO/MainProcess] Received task: timesketch.lib.tasks.run_plaso[1fde0509-4b67-4c16-b490-7d7f3ee09b6a]
timesketch_1     | [2019-12-17 05:22:54,929: INFO/ForkPoolWorker-1] Index timeline [PCIC_191127_PSG_004] to index [66e678b734cc4e3d8f9cf50ad48c72be] (source: plaso)
timesketch_1     | [2019-12-17 05:23:00,096: INFO/ForkPoolWorker-1] Task timesketch.lib.tasks.run_plaso[1fde0509-4b67-4c16-b490-7d7f3ee09b6a] succeeded in 5.169042253000043s: '2019-12-17 05:22:59,007 [INFO] (MainProcess) PID:49 <data_location> Determined data location: /usr/share/plaso
timesketch_1     | 2019-12-17 05:22:59,127 [INFO] (MainProcess) PID:49 <timesketch_out> Timeline name: <redacted>
timesketch_1     | 2019-12-17 05:22:59,128 [INFO] (MainProcess) PID:49 <timesketch_out> Owner of the timeline: None
timesketch_1     | Traceback (most recent call last):
timesketch_1     |   File "/usr/bin/psort.py", line 85, in <module>
timesketch_1     |     if not Main():
timesketch_1     |   File "/usr/bin/psort.py", line 67, in Main
timesketch_1     |     tool.ProcessStorage()
timesketch_1     |   File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 501, in ProcessStorage
timesketch_1     |     self._CheckStorageFile(self._storage_file_path)
timesketch_1     |   File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 115, in _CheckStorageFile
timesketch_1     |     logger.warning('Appending to an already existing storage file.')
[...]

Issue #870, opened 2019-04-20
Unable to import plaso data into Timesketch


CdtDelta:
This might be related to #836 

I'm running plaso from the docker hub image (plaso - log2timeline version 20190331), and timesketch from the docker instance in the repo (I'm pulling from source). I'm able to create an image with plaso, but not able to upload them into timesketch.

I'm running log2timeline with the following command:
docker run -ti --name=plaso --mount type=bind,source="$(pwd)"/<Case Folder>,target=/cases log2timeline/plaso log2timeline.py --status_view window --logfile /cases/Admin/Timeline/<computer-name>/<computer-name>.log /cases/Admin/Timeline/<computer-name>/<computer-name>.plaso /cases/Evidence/<computer-name>

The source data is triage files NOT a hard drive image. Plaso completes successfully but with some errors parsing data.

I'm running timesketch with the following command:
sudo docker-compose up
from the timesketch/docker directory

When I try to upload the plaso file to timesketch I get the following error:

 2019-04-20 15:21:18,123 [INFO] (MainProcess) PID:98 <data_location> Determined data location: /usr/share/plaso
2019-04-20 15:21:18,167 [INFO] (MainProcess) PID:98 <timesketch_out> Timeline name: <computer>
2019-04-20 15:21:18,167 [INFO] (MainProcess) PID:98 <timesketch_out> Owner of the timeline: None
Traceback (most recent call last):
  File "/usr/bin/psort.py", line 68, in <module>
    if not Main():
  File "/usr/bin/psort.py", line 54, in Main
    tool.ProcessStorage()
  File "/usr/lib/python2.7/dist-packages/plaso/cli/psort_tool.py", line 492, in ProcessStorage
    self._CheckStorageFile(self._storage_file_path)
  File "/usr/lib/python2.7/dist-packages/plaso/cli/psort_tool.py", line 116, in _CheckStorageFile
    logger.warning('Appending to an already existing storage file.')
  File "/usr/lib/python2.7/logging/__init__.py", line 1186, in warning
    self._log(WARNING, msg, args, **kwargs)
  File "/usr/lib/python2.7/logging/__init__.py", line 1293, in _log
    self.handle(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 1303, in handle
    self.callHandlers(record)
  File "/usr/lib/python2.7/logging/__init__.py", line 1343, in callHandlers
    hdlr.handle(record)
[...]

PR #1067, opened 2019-12-20
Docker install: run from /tmp directory

gaelmuller:
In the Docker installation, make sure the celery worker is running from the `/tmp` directory. This makes sure that `psort` will be able to create its log files.

Fixes https://github.com/google/timesketch/issues/1060 & https://github.com/google/timesketch/issues/870


jorlamd:
Thanks for pointing me in this direction, you are correct. What do you think about putting a `WORKDIR /tmp ` before the ENTRYPOINT in the dockerfile instead? This could make sense, since the entrypoint.sh code just deals with editing the main config file and has no directory dependent tasks.

Issue #841, opened 2019-02-24
Error with local mounts (Docker for Mac)


tlianza:
This seems similar to #591 , but I'm not using Windows I'm using native Docker for Mac. 

```
ERROR: for docker_elasticsearch_1  Cannot start service elasticsearch: b'Mounts denied: \r\nThe path /var/lib/elasticsearch\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'
Starting docker_redis_1         ... error
ERROR: for docker_postgres_1  Cannot start service postgres: b'Mounts denied: \r\nThe path /var/lib/postgresql/data\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'

Starting docker_neo4j_1         ... error
/docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'

ERROR: for docker_neo4j_1  Cannot start service neo4j: b'Mounts denied: \r\nThe path /var/lib/neo4j/data\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'

ERROR: for elasticsearch  Cannot start service elasticsearch: b'Mounts denied: \r\nThe path /var/lib/elasticsearch\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'

ERROR: for postgres  Cannot start service postgres: b'Mounts denied: \r\nThe path /var/lib/postgresql/data\r\nis not shared from OS X and is not known to Docker.\r\nYou can configure shared paths from Docker -> Preferences... -> File Sharing.\r\nSee https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.\r\n.'

[...]

PR #1068, opened 2019-12-27
Provide example commands docker mac shares

deralexxx:
The description might be okay for advanced users, but it might be handy if people could c&p stuff. The commands are from https://github.com/google/timesketch/issues/841 - so it seems at least one person had this issue before.

Let me know if that is something not wanted to be covered in the readme and / or more OS should be explained.


berggren:
Hm, the docker-compose will set the shares up, no? This documentation is in great need of some more attention for sure :)

I'm working on replacing the docker stuff here with something more robust in the near future.

deralexxx:
Nope it did not set them up (at least for me)

PR #1075, opened 2020-01-07
Cleanup dependencies and tests

berggren:
Clean up how we manage dependencies and how Travis runs the tests.

* Move old UI build config from  root -> timesketch/ui/
* Use PyPi for Travis (we are not 100& ready to switch to the new dpkg version just yet)
* Clean up requirements.txt and remove the old PipUtils version
* Small bugfix in one of the analyzers sneaked in

We will switch to DPKG and dependency management in the future.


api_client/python/timesketch_api_client/sketch_test.py
```diff
@@ -48,9 +49,3 @@ def test_get_timelines(self):
         self.assertIsInstance(timelines, list)
         self.assertEqual(len(timelines), 2)
         self.assertIsInstance(timelines[0], timeline_lib.Timeline)
-
-    def test_explore(self):
```
berggren: Removing this as it lacks mocked results, and hangs when run on travis.

PR #1086, opened 2020-01-28
First pass at fixing docker build

Onager:


berggren:
This has been fixed in #1090 - the old UI that was used in the Docker build has been deprecated and replaced with the implementation, so the fixes in this PR is not needed anymore.

Thanks!

PR #1099, opened 2020-02-10
Changed the upload API to support chunks and direct entries, added functionality into the importer

kiddinn:
This PR does few things:

1. Adds support in the upload API for file chunks, so that a large file can be split up and imported as a multiple web requests
2. Changed the importer when it imports plaso files to use the new chunk behavior and split the file up.
3. Add support in the upload API to upload events without needing a file upload
4. Changed the default behavior of the importer to upload using this new mechanism for all streaming and file parsing (besides plaso uploads)
5. Increment the API client version

There is also #1105, that adds a front-end to the importer library, a simple CLI to upload files as well as splitting up JSON files.

Future PRs would be to change the web uploader to use chunking, as well as clean up the code (once the two PRs are merged) and support re-uploads of failed chunks.


kiddinn:
will follow up with another PR that add re-uploads of failed chunks and better error handling

api_client/python/timesketch_api_client/importer.py
```diff
@@ -51,7 +63,54 @@ def __init__(self, entry_threshold=None):
...
+                else:
+                    try:
+                        date = datetime.datetime.utcfromtimestamp(value / 1e6)
+                        my_dict['datetime'] = date.isoformat()
+                        break
+                    except ValueError:
+                        continue
+
+        # We don't want to include any columns that start with an underscore.
+        underscore_columns = [x for x in my_dict if x.startswith('_')]
```
berggren: I guess this is due to the "special" TS fields, like __ts_label ? 

I wonder if we will miss anything else that might have underscore in the name?

api_client/python/timesketch_api_client/importer.py
```diff
@@ -108,37 +167,136 @@ def _reset(self):
...
             'enable_stream': not end_stream,
             'index_name': self._index,
+            'events': data_frame.to_json(orient='records', lines=True),
         }
 
-        response = self._sketch.api.session.post(
-            self._resource_url, files=files, data=data)
+        response = self._sketch.api.session.post(self._resource_url, data=data)
+        self._chunk += 1
+        # TODO: Add in the ability to re-upload failed file.
```
berggren: +1 also make sure we set a limit on how many retries we do etc.

api_client/python/timesketch_api_client/importer.py
```diff
@@ -210,10 +362,11 @@ def add_dict(self, entry):
         if not isinstance(entry, dict):
             raise TypeError('Entry object needs to be a dict.')
 
-        if self._count >= self._threshold:
+        if self._count >= self._threshold_entry:
             self.flush(end_stream=False)
             self._reset()
 
+        self._fix_dict(entry)
```
berggren: There is a fair amount of fixing in the method call here - can you add a comment outlining what is happening, and refer to the method for more details?

timesketch/api/v1/resources.py
```diff
@@ -1522,49 +1521,29 @@ def post(self, sketch_id):
...
+            enable_stream, file_path='', events='', meta=None):
+        """Creates a full pipeline for an uploaded file and returns the results.
+
+        Args:
+            file_extension: the extension of the uploaded file.
+            timeline_name: name the timeline will be stored under in the
+                           datastore.
+            index_name: the Elastic index name for the timeline.
+            sketch: Instance of timesketch.models.sketch.Sketch
+            sketch_id: integer with the sketch ID.
```
berggren: Both sketch and sketch_id?

timesketch/api/v1/resources.py
```diff
@@ -1603,24 +1582,188 @@ def post(self):
...
+            form: a dict with the configuration for the upload.
+            sketch: Instance of timesketch.models.sketch.Sketch
+            index_name: the Elastic index name for the timeline.
+
+        Returns:
+            A timeline if created otherwise a search index in JSON (instance
+            of flask.wrappers.Response)
+        """
+        timeline_name = form.get('name', 'unknown_events')
+        file_extension = 'jsonl'
```
berggren: this only applies to .jsonl files?

api_client/python/timesketch_api_client/importer.py
kiddinn: this is due to those kind of field names. Do you have suggestions? Have a disclaimer?

api_client/python/timesketch_api_client/importer.py
kiddinn: yes, we will absolutely do that

api_client/python/timesketch_api_client/importer.py
kiddinn: done

api_client/python/timesketch_api_client/importer.py
kiddinn: done

timesketch/api/v1/resources.py
kiddinn: should not be needed... agree, removed

timesketch/api/v1/resources.py
kiddinn: no... you are adding events, and events need to be JSON. This is a way of telling the uploader that the events in question are json line formatted, not csv formatted. since that' show events are uploaded.

PR #1104, opened 2020-02-13
Updated l2tdevtools configuration and dpkg files

joachimmetz:


joachimmetz:
@berggren note that timesketch.ini is used by l2tdevtools to generate the dpkg files

PR #1111, opened 2020-02-21
Sketch Analyzer for Windows artefacts of application crashes

binglot:
The analyzer identifies Windows artefacts of application crashes and also checks if the crash reporting has been explicitly disabled. 

This is my first commit to someone else's repo so please be gentle on me ;) I haven't had a chance to create any tests yet, I can work on it next week.


kiddinn:
ok, first of all can you create the tests?

```
======================================================================
FAIL: Test analyzer.
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/mock/mock.py", line 1330, in patched
    return func(*args, **keywargs)
  File "/home/travis/build/google/timesketch/timesketch/lib/analyzers/crash_test.py", line 24, in test_analyzer
    self.assertEqual(True, False)
AssertionError: True != False
```

I'll review the PR as well. but you'll need to add tests to it

timesketch/lib/analyzers/__init__.py
```diff
@@ -18,6 +18,7 @@
 from timesketch.lib.analyzers import browser_search
 from timesketch.lib.analyzers import browser_timeframe
 from timesketch.lib.analyzers import chain
+from timesketch.lib.analyzers import crash
```
kiddinn: can you name this something else, since this is only for windows crash files, wondering whether better to have it `win_crash` so it is more explicit.

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
+"""Sketch analyzer plugin for crashes."""
```
kiddinn: ...plugin for Windows crash artifacts.

or something like that

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+
+from timesketch.lib import emojis
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+
+
+class CrashesSketchPlugin(interface.BaseSketchAnalyzer):
+    """Sketch analyzer for Windows Application Crashes."""
+
+    NAME = 'crashes'
```
kiddinn: and the name then `win_crashes`

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+            'File - WER Report' : (
+                'data_type:"fs:stat"',
+                'filename:"/Microsoft/Windows/WER/"',
+                'filename:/((Non)?Critical|AppCrash)_.*/',
+                'file_entry_type:("directory" or "2")',
+            ),
+            'Registry - Crash Reporting' : (
+                'data_type:"windows:registry:key_value"',
+                r'key_path:r"\\Control\\CrashControl"',
+                'values:("LogEvent: REG_DWORD_LE 0" OR "SendAlert: REG_DWORD_LE 0" OR "CrashDumpEnabled: REG_DWORD_LE 0")',
```
kiddinn: line too long,

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+            ),
+            'File - WER Report' : (
+                'data_type:"fs:stat"',
+                'filename:"/Microsoft/Windows/WER/"',
+                'filename:/((Non)?Critical|AppCrash)_.*/',
+                'file_entry_type:("directory" or "2")',
+            ),
+            'Registry - Crash Reporting' : (
+                'data_type:"windows:registry:key_value"',
+                r'key_path:r"\\Control\\CrashControl"',
```
kiddinn: why is the key_path a raw string? and the same with the control\crashontrol?

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+                'values:("LogEvent: REG_DWORD_LE 0" OR "SendAlert: REG_DWORD_LE 0" OR "CrashDumpEnabled: REG_DWORD_LE 0")',
+            ),
+            'Registry - Error Reporting' : (
+                'data_type:"windows:registry:key_value"',
+                r'key_path:"\\Software\\Microsoft\\PCHealth\\ErrorReporting"',
```
kiddinn: same here, why is key_path a raw string?

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+                'values:("DoReport: REG_DWORD_LE 0" OR "ShowUI: REG_DWORD_LE 0")',
+            ),
+        }
+
+        # Creator of nested Elastic Search queries.
```
kiddinn: why define all of these functions inside the run function, and not just as a separate function?

also this comment is not needed, and all functions lack docstrings

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Generator of events based on your query.
+        events = self.event_stream(
+            query_string=query, return_fields=return_fields)
+
+        # Get the executable filenames.
+        # Examples: 
+        #   "[...]\WER\ReportQueue\AppCrash_notepad.exe_8d163e29d3960561ca2e9723640cd8fff5c2ad5_cab_0a5810ac"
+        #   "[...]/WER/ReportQueue/NonCritical_x64_d473a376adfb18a7b165c5e3c26de43cd8bccb_cab_07fd2620"
+        #   "Strings: ['iexplore.exe', '8.0.7601.17514' [...]"
```
kiddinn: these comments are not providing value, remove them

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+        self.index_name = index_name
+        super(CrashesSketchPlugin, self).__init__(index_name, sketch_id)
+
+    def run(self):
+        """Entry point for the analyzer.
+
+        Returns:
+            String with summary of the analyzer result
+        """
+        query_elements = {
```
kiddinn: I'm wondering whether it makes more sense, or makes it a bit easier to change this file in the future to have this as a class constant

QUERY_ELEMENTS = ....

and thus you don't have to edit anything int eh run function, that can just parse this structure

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+        #   "[...]\WER\ReportQueue\AppCrash_notepad.exe_8d163e29d3960561ca2e9723640cd8fff5c2ad5_cab_0a5810ac"
+        #   "[...]/WER/ReportQueue/NonCritical_x64_d473a376adfb18a7b165c5e3c26de43cd8bccb_cab_07fd2620"
+        #   "Strings: ['iexplore.exe', '8.0.7601.17514' [...]"
+        re_filename = re.compile(r'(?:\\|\/)(?:AppCrash|Critical|NonCritical)_(.+?\.exe)_[a-f0-9]{16,}|\'([^\']+\.exe)\'', re.I)
+
+        # Container for filenames of crashed applications.
+        filenames = dict()
+
+        for event in events:
+
```
kiddinn: unneeded extra line

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+            data_type = event.source.get('data_type')
+
+            # Search event log entries for filenames of crashed applications.
+            if data_type == 'windows:evtx:record':
+                message = event.source.get('message')
+
+                fn = extract_filename(message, re_filename)
```
kiddinn: s/fn/filename/g

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+                    # The regex can match on full file paths and filenames,
+                    # so only return the filename.
+                    return min([m for m in match.groups() if m])
+            return None
+
+        # For adding information to identified entries.
+        def mark_as_crash(event, filename):
+            event.add_star()
+            event.add_attributes({'crash_app': filename})
+            event.add_label('crash')
```
kiddinn: why add a label here? why not just add a tag instead?

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+                'It could be indicative of attacker activity. '
+                'For details refer to page 16 from '
+                'https://assets.documentcloud.org/documents/3461560/Google-Aquarium-Clean.pdf.')
+
+            # Commit the event to the datastore.
+            event.commit()
+
+        # Create a saved view with our query.
+        if filenames:
+            self.sketch.add_view('Crash activity', 'app_crashes', query_string=query)
```
kiddinn: so if you have he query string `query` then you'll get a whole bunch of events from this view that are not really crashes, since you have checks here, and only if the checks suceced then you add something to the event, for instance of you had added a tag instead of a label you could do  

```
query_string='tag:"crash"`
```

and then you just get the correct events

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+
+        # Get the executable filenames.
+        # Examples: 
+        #   "[...]\WER\ReportQueue\AppCrash_notepad.exe_8d163e29d3960561ca2e9723640cd8fff5c2ad5_cab_0a5810ac"
+        #   "[...]/WER/ReportQueue/NonCritical_x64_d473a376adfb18a7b165c5e3c26de43cd8bccb_cab_07fd2620"
+        #   "Strings: ['iexplore.exe', '8.0.7601.17514' [...]"
+        re_filename = re.compile(r'(?:\\|\/)(?:AppCrash|Critical|NonCritical)_(.+?\.exe)_[a-f0-9]{16,}|\'([^\']+\.exe)\'', re.I)
+
+        # Container for filenames of crashed applications.
+        filenames = dict()
```
kiddinn: why is this a dict? you don't seem to do anything with the actual values, this is all just

filenames[filename] = ''

so why not have this as

filenames = set()

or something like that?

timesketch/lib/analyzers/crash.py
```diff
@@ -0,0 +1,178 @@
...
+
+                fn = extract_filename(filename, re_filename)
+                if not fn:
+                    continue
+                filenames[fn] = ''
+                mark_as_crash(event, fn)
+
+            # Check if the crash reporting has been disabled in the registry.
+            elif data_type == 'windows:registry:key_value':
+                event.add_star()
```
kiddinn: are you sure we want to star any event that has a crash element to it?

kiddinn:
remember to do `git pull` before working on the code again, since I just merged the PR with the chan ges that have happened to the codebase since last review

kiddinn:
and also remember, once you are done,  please request a new review from me, by clicking the "circle" icon next to my name in the "Reviewers" section, otherwise I may not see that you are ready for another round

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+        query = self.formulate_query(self.QUERY_ELEMENTS)
+
+        return_fields = ['data_type', 'message', 'filename']
+
+        # Generator of events based on your query.
+        events = self.event_stream(
+            query_string=query, return_fields=return_fields)
+
+        # Container for filenames of crashed applications.
+        filenames = list()
```
kiddinn: why have this as a list only to do `set` operation on it at the end? why ont have this as a set to begin with

```
...
filenames = set()
...
filenames.add(filename)
```

or have this as a `collections.Counter` if you want to have both a unique set of filenames and a count of them (but only if you want to use the count\)

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+
+            # Tag entries that show the crash reporting has been disabled.
+            elif data_type == 'windows:registry:key_value':
+                event.add_comment(
+                    'WARNING: The crash reporting was disabled. '
+                    'It could be indicative of attacker activity. '
+                    'For details refer to page 16 from '
+                    'https://assets.documentcloud.org/documents/3461560/'
+                    'Google-Aquarium-Clean.pdf.')
+                event.commit()
```
kiddinn: in this case there is no need to continue evaluatingn the rest... I woudl rather opt for starting here

```
if data_type == 'windows:regist....`:
  event.add_comment(....
  event.commit()
  continue

if data_type == 'windows:evtx:record':
...

```
etc, etc, since now you are adding a comment, then doing a commit and then checkingn filename, which you know will not work, since you are just calling a function on an empty strign (or a none as the code stands now), so you might as well just do a `continue` so that you can just move ahead and evaluate the next record

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+                filenames.append(filename)
+                event.commit()
+
+        # Dedup the filenames of crashed apps
+        filenames = set(filenames)
+
+        # Create a saved view with our query.
+        if filenames:
+            self.sketch.add_view(
+                'Windows Crash activity', 'win_crash', query_string=query)
```
kiddinn: why have the entire query here that will produce events that do not end up getting tagged? why not just search for the events with the crash tag?

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+
+        return 'Windows Crash analyzer completed, ' + \
+            '{0:d} crashed application{1:s} identified: {2:s}'.format(
+                len(filenames),
+                's' if len(filenames) > 1 else '',
+                ', '.join(filenames))
```
kiddinn: do we want all the filenames on this return string? what if there are 200 filenames returned, this string would make the logs completely unreadable, and it does not provide any value really, remember this is displayed in the ui andn is in the logs on the server, no need to contain specifics, more just the number of applications, etc

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+        """Generates the Elasticsearch query.
+
+        Args:
+            elements: Dictionary with a list of conditions
+
+        Returns:
+            The Elasticsearch query
+        """
+        conditions = list()
+        for e in elements.values():
```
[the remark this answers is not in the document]
binglot: Sure, done

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,195 @@
...
+        return ' OR '.join(conditions)
+
+    def extract_filename(self, text):
+        """Finds filenames of crashed applications using a regular expression.
+
+        Args:
+            text: String that might contain the filename
+
+        Returns:
+            The string with filename if found
```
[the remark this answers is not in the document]
binglot: Great suggestion!

timesketch/lib/analyzers/win_crash.py
binglot: Good point, I replaced list() with set().

timesketch/lib/analyzers/win_crash.py
binglot: Changed the code as per your suggestion. I tried to avoid "continue" as it makes the code less readable for me.

timesketch/lib/analyzers/win_crash.py
binglot: Oh my, you did suggest this before! My bad, it escaped me. It should be correct now.

timesketch/lib/analyzers/win_crash.py
binglot: I see your point. From testing a few Plaso files, I had a very small number of application names that were returned. Once the analyzer finished running, it was actually quite helpful for me to see the filenames that crashed in the web UI (iirc, it was called "history") before diving into the analysis. If you don't mind, I'd leave it as it is.

kiddinn: ok, but we'll remove this if it becomes an issue

timesketch/lib/analyzers/win_crash.py
```diff
@@ -0,0 +1,194 @@
+"""Sketch analyzer plugin for Windows crash artefacts."""
+from __future__ import unicode_literals
+
+import re
+
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+
```
kiddinn: should be two empty lines before the class declaration

(update: I just made the edits myself)

PR #1113, opened 2020-02-25
Fix an issue with the importer

kiddinn:
This PR does few things:

+ Increment the version of the API client
+ Reduce the entry threshold from 100k to 50k
+ Add a sleep command to fix issues with a race condition with post requests
+ Fix an issue with threshold setting
+ Add the option to define threshold in the importer tool
+ Adding a try/except around bulk inserts into ES, handling timeout issues.


api_client/python/timesketch_api_client/importer.py
```diff
@@ -199,6 +199,9 @@ def _upload_data_buffer(self, end_stream):
             'prepare.'.format(time.time() - start_time))
 
         response = self._sketch.api.session.post(self._resource_url, data=data)
+        # To prevent unexpected errors with connection refusal adding a quick
+        # sleep.
```
berggren: Can you add a TODO here to investigate the underlying issue and get rid of this sleep when ready?

kiddinn: done

PR #1118, opened 2020-02-27
Fix start of Elasticsearch

wincentbalin:
Run Elasticsearch as a single node. Documented at https://hub.docker.com/_/elasticsearch .

Without this fix, Elasticsearch does not find other nodes and restarts itself forever.


wincentbalin:
> @googlebot I signed it!

PR #1127, opened 2020-03-10
Replaces PASSWORD with PASSWORD_FILE (fixes #1126)

### LINKED ISSUE #1126: Passwords for Docker Container Configuration Should Be Read From File
Passing password by `ENV` is bad practice and should be replaced by using `docker secrets`. To accomplish this, the current environment variables could be extended by `POSTGRES_PASSWORD_FILE` and `TIMESKETCH_PASSWORD_FILE` which read the file from the appropriate path (usually `/run/secrets/secret-name`).

einfallstoll:
This should replace the PASSWORD with PASSWORD_FILE environment variables, while keeping backward compatibility, improve existing security and add a note to upgrade.

### DIFF
--- a/docker/docker-entrypoint.sh
+++ b/docker/docker-entrypoint.sh
@@ -10,8 +10,12 @@ if [ "$1" = 'timesketch' ]; then
   fi
 
   # Set up the Postgres connection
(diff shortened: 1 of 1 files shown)


einfallstoll:
@googlebot I signed it!

PR #1129, opened 2020-03-10
Stats on indices

berggren:
This PR adds count stats per index when searching and it shows counters and size on disk in the timeline list.

Screenshot 2020-03-10 at 10 21 17

Screenshot 2020-03-10 at 10 13 21

Screenshot 2020-03-10 at 10 07 57


timesketch/api/v1/resources.py
```diff
@@ -410,6 +410,16 @@ def get(self, sketch_id):
             for t in sketch.active_timelines
         ]
 
+        # Get event count and size on disk for each index in the sketch.
+        stats_per_index = {}
+        es_stats = self.datastore.client.indices.stats(
+            index=sketch_indices, metric='docs, store')
+        for index_name, stats in es_stats['indices'].items():
```
kiddinn: can we do here

for .... in es_stats.get('indices', {}).items()

just to be more error tolerant.

timesketch/api/v1/resources.py
```diff
@@ -410,6 +410,16 @@ def get(self, sketch_id):
...
+            stats_per_index[index_name] = {
+                'count': stats['total']['docs']['count'],
```
kiddinn: also here... do we know that these will always exist?

```
stats.get('total', {}).get('docs', {}).get('count', 0)
```
or something like that

timesketch/api/v1/resources.py
```diff
@@ -826,10 +846,18 @@ def post(self, sketch_id):
...
-                aggregations=None,
+                aggregations=index_stats_agg,
                 return_fields=return_fields,
                 enable_scroll=enable_scroll)
 
+        # Get number of matching documents per index.
+        count_per_index = {}
+        try:
+            for bucket in result['aggregations']['indices']['buckets']:
+                count_per_index[bucket['key']] = bucket['doc_count']
```
kiddinn: ok, so the way this works now is that if a single key doesn't exist a KeyError will be raised, can happen in several different places here, and then nothing more will be stored.... what if you actually get back the `result['aggregations']['indices']['buckets']`, so you'll get some `bucket` attributes... and then a single bucket will not have a `key` value (may not happen), but the subsequent ones would... that would mean you don't get all the values you could have gotten... would it be better to make the assignment more error prone?

```
key = bucket.get('key')
if key:
    count_per_index[key] = bucket.get('doc_count')
```

or something like that
count_per_index[

timesketch/api/v1/resources.py
berggren: Done.

timesketch/api/v1/resources.py
berggren: Done

timesketch/api/v1/resources.py
berggren: Good catch, fixed now.

berggren:
@kiddinn PTAL

PR #1132, opened 2020-03-11
Create stories from analyzers

berggren:
This PR adds the ability to create and author a Story from analyzers.

```
        view = self.sketch.add_view(
            view_name='Story test', analyzer_name=self.NAME,
            query_string='foobar')
        story = self.sketch.add_story(title='Story from analyzer')
        story.add_text('## This is a markdown title')
        story.add_view(view)
        story.add_text('This is another paragraph')
```


timesketch/lib/analyzers/interface.py
```diff
@@ -351,6 +353,26 @@ def add_view(self, view_name, analyzer_name, query_string=None,
...
+        Args:
+            title: The name of the view.
+
+        Raises:
+            ValueError: If both query_string an query_dsl are missing.
+
+        Returns:
+            An instance of a Story object.
+        """
+
```
kiddinn: no need for this extra space/line

timesketch/lib/analyzers/interface.py
```diff
@@ -351,6 +353,26 @@ def add_view(self, view_name, analyzer_name, query_string=None,
...
+        story = SQLStory.get_or_create(
+            title=title, content=json.dumps([]), sketch=self.sql_sketch,
```
kiddinn: I don't see a reason to call `json.dumps([])` here, that is a constant... calling `json.dumps([])` will produce the string: `'[]'`, and therefore just have `content='[]', ` should suffice

timesketch/lib/analyzers/interface.py
berggren: done

timesketch/lib/analyzers/interface.py
berggren: good catch, fixed.

berggren:
PTAL @kiddinn

PR #1134, opened 2020-03-11
Adding support for working with stories from the API client, as well as adding aggregation support in the backend

kiddinn:
This PR does quite a few things:

1. Increments client API version number
2. Adds a Story class that can add views and text to a story
3. Adds the ability to delete a story from a sketch in the API
4. Adds some logging into the API client for failed resource fetches
5. Adds the ability to create and list stories in the sketch object for the API client
6. Adds ability to manipulate and work with stories and individual blocks in the API client
7. Adds the ability for analyzers to add aggregation to stories.
8. Fixes an issue where API client tests weren't run and fixes broken tests

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -166,7 +166,7 @@ class Sketch(resource.BaseResource):
         story_dict = response_json.get('objects', [{}])[0]
         return story.Story(
             story_id=story_dict.get('id', -1),
-            sketch_id=self.id,
+            sketch=self,
             api=self.api)
 
     def list_aggregations(self):
@@ -250,7 +250,7 @@ class Sketch(resource.BaseResource):
         for story_dict in stories:
             story_list.append(story.Story(
                 story_id=story_dict.get('id', -1),
-                sketch_id=self.id,
+                sketch=self,
                 api=self.api))
         return story_list
 
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -211,14 +211,38 @@ class Sketch(resource.BaseResource):
                 return aggregation_obj
         return None
 
+    def get_story(self, story_id=None, story_title=None):
+        """Returns a story object that is stored in the sketch.
+
+        Args:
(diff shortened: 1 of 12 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -143,6 +144,31 @@ def _build_pandas_dataframe(self, search_response, return_fields=None):
...
+        data = {
+            'title': title,
+            'content': ''
+        }
+
+        response = self.api.session.post(resource_url, json=data)
+        response_json = response.json()
+        story_dict = response_json.get('objects', [{}])[0]
+        return story.Story(
+            story_id=story_dict.get('id', -1),
```
berggren: -1 ?

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -185,14 +211,38 @@ def get_aggregation(self, aggregation_id):
                 return aggregation_obj
         return None
 
+    def get_story(self, story_id=None, story_title=None):
+        """Returns a story object that is stored in the sketch.
+
+        Args:
+            story_id: an integer indicating the ID of the story to
+                be fetched. Defaults to None.
```
berggren: Isn't this argument required?

api_client/python/timesketch_api_client/sketch.py
kiddinn: don't remember why, cjhanged to zero

api_client/python/timesketch_api_client/sketch.py
kiddinn: no, either one... if need to provide either one, if you provide story_title the first instance of a story with that title will be returned... made that a bit more clear in the documentation.

kiddinn:
responded to comments

api_client/python/timesketch_api_client/sketch.py
berggren: hmm, is this really what we want? Oh, a Story for that sketch right? that would make sense

berggren: Ok, this is for the sketch. All good

PR #1141, opened 2020-03-18
New analyzer that adds tag/emojis to events based on config.

berggren:
This PR adds a simple analyzer to tag/emoji events based on a config file. It can also create saved views.

In addition, an updated SQL schema for View and SearchTemplate has been done, adding a description field.


timesketch/lib/analyzers/tagger.py
```diff
@@ -0,0 +1,83 @@
...
+            config: A dict that contains the configuration See data/tags.yaml
+                for fields and documentation of what needs to be defined.
+
+        Returns:
+            String with summary of the analyzer result.
+        """
+        query = config.get('query_string')
+        query_dsl = config.get('query_dsl')
+        create_view = config.get('create_view', False)
+        view_name = config.get('view_name')
```
kiddinn: I would prefer to have this as

```
view_name = config.get('view_name', name)
```

instead of later doing

```
if not view_name:
  view_name = name
```

then you can skip that check

data/tags.yaml
```diff
@@ -0,0 +1,21 @@
...
+#
+# The tagger works in the way that the query is run and the emojis and
+# tags that are defined are applied to the matching events.
+# A view is created if the tag is added (only if there are results).
+#
+# Example:
+#
+# test_tagger:
+#        query_string: 'test'
+#        attribute: 'domain'
```
kiddinn: OK, so this is nowhere used, and I'm not sure how you would even use it in this case....

(the only place this is used is in the tests.... and to check it is empty... so please remove it

timesketch/lib/analyzers/tagger.py
berggren: Good idea, done.

data/tags.yaml
berggren: oops, removed this elsewhere but missed this one. Good catch :) Fixed now.

berggren:
@kiddinn PTAL

timesketch/lib/analyzers/tagger_test.py
```diff
@@ -0,0 +1,51 @@
...
+class TestTaggerPlugin(BaseTest):
+    """Tests the functionality of the analyzer."""
+
+    def _config_validation(self, config):
+        """Validate that all items of a config are valid."""
+        query = config.get('query_string', config.get('query_dsl'))
+        self.assertIsNotNone(query)
+        self.assertIsInstance(query, str)
+
+        attribute = config.get('attribute')
```
kiddinn: you can remove the attribute from the test as well, since there is no need to test something that isn't there

timesketch/lib/analyzers/tagger_test.py
```diff
@@ -0,0 +1,51 @@
...
+
+    # TODO: Add tests for the tagger.
+    def test_config(self):
+        """Tests that the config is valid."""
+        test_config = yaml.safe_load("""
+        place_holder:
+          query_string: '*'
+          tags: ['place-holder']
+          emojis: ['ID_BUTTON']
+          attribute: 'message'
```
kiddinn: also here

timesketch/lib/analyzers/tagger_test.py
berggren: fixed

timesketch/lib/analyzers/tagger_test.py
berggren: fixed

PR #1144, opened 2020-03-19
Fix a bug if there are no stories in a sketch and add properties to View

kiddinn:
Fixes a bug when trying to call `list_stories` on a sketch with no stories in it.

Also adding few more properties of a view object.

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -344,13 +344,13 @@ class Sketch(resource.BaseResource):
             self.api.api_root, self.id)
         response = self.api.session.get(resource_url)
         response_json = response.json()
-        objects = response_json.get('objects')
-        if not objects:
+        story_objects = response_json.get('objects')
+        if not story_objects:
             return story_list
 
-        if not len(objects) == 1:
+        if not len(story_objects) == 1:
             return story_list
-        stories = objects[0]
+        stories = story_objects[0]
         for story_dict in stories:
             story_list.append(story.Story(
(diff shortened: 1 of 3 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -344,7 +344,13 @@ def list_stories(self):
             self.api.api_root, self.id)
         response = self.api.session.get(resource_url)
         response_json = response.json()
-        stories = response_json.get('objects', [[]])[0]
+        objects = response_json.get('objects')
```
berggren: maybe story_objects ?

api_client/python/timesketch_api_client/view.py
```diff
@@ -39,15 +39,58 @@ def __init__(self, view_id, view_name, sketch_id, api):
...
+            name: String with the attribute name.
+            default_value: The default value if the attribute does not exit,
+                defaults to None.
+            refresh: If set to True then the data will be refreshed.
+
+        Returns:
+            The dict value of the key "name".
+        """
+        view = self.lazyload_data(refresh_cache=refresh)
+        objects = view.get('objects')
```
berggren: view_objects?

api_client/python/timesketch_api_client/sketch.py
kiddinn: done

api_client/python/timesketch_api_client/view.py
kiddinn: done

PR #1160, opened 2020-04-01
More Chart support - added linechart and href

kiddinn:
This PR adds the following features into TS:

1. Line chart is added
2. Charts now define a URL href and have a calculate transformation, allowing them to be interactive, and clicking a bar on a chart will take you to explore and view those events.


kiddinn:
Removed group support from this PR and thus it is ready for review

PR #1166, opened 2020-04-07
Adding aggregation groups

kiddinn:
This PR adds aggregation groups to TS, which can produce a chart that consists of several other charts combined.

The PR adds the concept of an aggregation group, which requires schema changes to the DB. Aggregation groups is simply a placeholder for a group of aggregations and definitions on how they should be combined and some specific parameters to control them.

There is also some other features added, such as the ability to define color to the chart, adding support into the API client and into stories.


kiddinn:
This is ready for a review.

Not sure what to do with the. CLA, since I'm pretty sure I've signed it ;)

api_client/python/timesketch_api_client/aggregation.py
```diff
@@ -272,3 +276,172 @@ def generate_chart(self):
...
+
+    @property
+    def chart(self):
+        """Property that returns an altair Vega-lite chart."""
+        if not self._aggregations:
+            return altair.Chart()
+        return self.generate_chart()
+
+    @property
+    def description(self):
```
berggren, line 315: These just returns the properties. Can we make them public and remove these?

E.g. _description --> description

api_client/python/timesketch_api_client/aggregation.py
```diff
@@ -272,3 +276,172 @@ def generate_chart(self):
...
+    def get_charts(self):
+        """Returns a list of altair Chart objects from each aggregation."""
+        return [x.chart for x in self._aggregations]
+
+    def get_tables(self):
+        """Returns a list of pandas DataFrame from each aggregation."""
+        return [x.table for x in self._aggregations]
+
+    def to_pandas(self):
+        """Returns a pandas DataFrame."""
```
berggren, line 439: Can you add some more words around what the DataFrame containes?

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -208,6 +208,26 @@ def add_to_acl(self, user_list=None, group_list=None, make_public=False):
...
+        for group_dict in data.get('objects', []):
+            if not group_dict.get('id'):
+                continue
+            group = aggregation.AggregationGroup(
+                sketch=self, api=self.api)
+            group.from_dict(group_dict)
+            groups.append(group)
+        return groups
+
     def list_aggregations(self):
```
berggren, line 231: Add returns

timesketch/api/v1/resources.py
```diff
@@ -1064,6 +1080,148 @@ def post(self):
...
+                'group sketch ID ({1:d})'.format(sketch.id, group.sketch_id))
+            abort(HTTP_STATUS_CODE_FORBIDDEN, msg)
+
+        if not sketch.has_permission(user=current_user, permission='read'):
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN,
+                'The user does not have read permission on the sketch.')
+
+        result_chart = None
+        how = group.how
```
berggren, line 1118: how? how what? :)

berggren: Is this only for chart orientation? If yes, can we call this chart_orientation instead?

timesketch/api/v1/resources.py
```diff
@@ -1236,6 +1395,98 @@ def post(self, sketch_id):
...
+                'No JSON data, unable to process request to create '
+                'a new aggregation group.')
+
+        aggregation_string = form.get('aggregations', '')
+        if not aggregation_string:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Unable to create an empty group.')
+
+        agg_list = [int(x) for x in aggregation_string.split(',')]
```
berggren, line 1467: s/x/agg maybe?

timesketch/lib/analyzers/interface.py
```diff
@@ -324,6 +325,34 @@ def add_aggregation(
         db_session.commit()
         return aggregation
 
+    def add_aggregation_group(self, name, description='', view_id=None):
+        """Add aggregation Group to the sketch.
+
+        Args:
+            name: the name of the aggregation run.
+            description: description of the aggregation, visible in the UI,
```
berggren, line 333: nit: Optional description of the aggregation (visible in the UI)

timesketch/lib/analyzers/interface.py
```diff
@@ -401,6 +430,102 @@ def get_all_indices(self):
...
+        """Commit changes to DB."""
+        self.group.how = self._how
+        self.group.parameters = self._parameters
+        db_session.add(self.group)
+        db_session.commit()
+
+    def set_how(self, how='layer'):
+        """Sets how charts should be joined.
+
+        Args:
```
berggren, line 481: aha, layered as well..

I still think "how" is a little difficult to understand. Maybe "combine_as" or something similar?

berggren: or, "layered_as" ?

timesketch/migrations/versions/b11dea30835c_.py
```diff
@@ -0,0 +1,46 @@
...
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        'aggregation',
+        sa.Column('aggregationgroup_id', sa.Integer(), nullable=True))
+    op.create_foreign_key(
+        None, 'aggregation', 'aggregationgroup',
+        ['aggregationgroup_id'], ['id'])
+    op.drop_column('aggregationgroup', 'agg_type')
```
berggren, line 25: Hmm, I wonder if these are needed, or if it is because of previous interation of your DB migration?

berggren: or wait.. did we have the model before?

timesketch/models/sketch.py
```diff
@@ -431,19 +436,62 @@ def __init__(self, name, description, agg_type, parameters, chart_type,
...
 
+class AggregationGroup(
+        AccessControlMixin, LabelMixin, StatusMixin, CommentMixin, BaseModel):
+    """Implements the Aggregation Group model."""
+    name = Column(Unicode(255))
+    description = Column(UnicodeText())
+    aggregations = relationship(
+        'Aggregation', backref='aggregationgroup', lazy='select')
+    parameters = Column(UnicodeText())
+    how = Column(Unicode(15))
```
berggren, line 463: You can add some more chars here. Not much impact on the DB, but will future proof it a bit if we need to add more "hows" later.

api_client/python/timesketch_api_client/aggregation.py
kiddinn: so I prefer not to do that, since I will be doing a follow up CL where I add a setter to this function, which will trigger an update of the database as soon as you change the parameter... for now, this makes this a read-only parameter... if I expose it with `.description` then it becomes a public parameter, meaning read/write and nothing changes if I modify it.

api_client/python/timesketch_api_client/aggregation.py
kiddinn: ack...

api_client/python/timesketch_api_client/sketch.py
kiddinn: there is a return section there....

timesketch/api/v1/resources.py
kiddinn: how is `how should the charts be combined`... and yes, we can call it that instead... but `chart_orientation` is a pretty long parameter... how about just `orientation` instead of how?

timesketch/api/v1/resources.py
kiddinn: changed this a bit

timesketch/lib/analyzers/interface.py
kiddinn: ack

timesketch/lib/analyzers/interface.py
kiddinn: I changed how into orientation, hopefully that has "saved some troubles"

timesketch/migrations/versions/b11dea30835c_.py
kiddinn: no, this is needed, since I'm adding aggregationgroup into the database for other modules...

is this drop column needed? I don't know, this is automatically generated code, I just assume this is done "just in case"

timesketch/models/sketch.py
kiddinn: ah... well..I can add a bit more, (i did change this to orientation as well).... if altair/vega decides to add a new orientation that is longer

kiddinn:
responded to comments, the only outstanding one is the auto generated migratio nscript, but that is being resolved offline

timesketch/migrations/versions/b11dea30835c_.py
kiddinn: file dopped... new one replaces it

kiddinn:
ok, replaced the migration file, it's your turn now

berggren:
CLA is broken - kiddi is a googler. All good for the record.

PR #1168, opened 2020-04-15
Fixing typo Installation re. libffi

dov-csis:
Not sure, but I think `libffi-dev` should have been left intact maybe? `libffi3-dev` doesn't seem to exist on my Ubuntu.


dov-csis:
@googlebot I signed it!

dov-csis:
@googlebot I signed it!

PR #1171, opened 2020-04-20
Safe Browsing API url analyzer

dov-csis:
An analyzer for **url** fields against the [Google Safe Browsing API](https://developers.google.com/safe-browsing/v4/lookup-api).

- The API key should be set in `timesketch.conf` as `SAFEBROWSING_API_KEY`
- It tries to parse a valid URL from the content of the **url** field - we've seen some logs where there was a short log message prior to `http://...`
- It is possible to whitelist ~glob-like~ URL masks specified in a file called `safebrowsing_whitelist.yaml`
- Without a whitelist file, by default it uses the [domain analyzer's whitelist](https://github.com/google/timesketch/blob/master/data/timesketch.conf#L180)
----


berggren:
Thanks for contributing! I'll review this asap.

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+import fnmatch
+import logging
+import re
+
+import pkg_resources
+import requests
+from flask import current_app
+
+from timesketch.lib.analyzers import interface, manager
```
kiddinn, line 12: let's split this into two lines

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class SafeBrowsingSketchPlugin(interface.BaseSketchAnalyzer):
+    """Sketch analyzer for Safe Browsing."""
+
+    NAME = 'safebrowsing'
+
+    _SAFE_BROWSING_BULK_LIMIT = 500
+    _URL_WHITELIST_CONFIG = 'safebrowsing_whitelist.yaml'
+    _SAFEBROWSING_ENTRY_KEEP = frozenset(
```
kiddinn, line 22: for formatting can you rather do:

```
_SAFEBROWSING_ENTRY_KEEP = frozenset([
    'platformType',
    'threadType',
])

```

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+        Returns:
+            String with the URL
+        """
+        m = self._URL_BEGINNING_RE.search(url_entry)
+
+        if m:
+            return m.group(1)
+
+        return None
```
kiddinn, line 137: this is not a string... please be consistent with the return... return ''

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+        return results
+
+    def _sanitize_url(self, url_entry):
+        """Finds http[s]:// in 'url_entry' and returns its content from there.
+
+        Args:
+            url_entry: a URL, with some other characters before and after
+
+        Returns:
+            String with the URL
```
kiddinn, line 130: ... or an empty string if not found.

something like that

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+                continue
+
+            urls.setdefault(url, []).append(event)
+
+        # Exit early if there are no URLs in the data set to analyze.
+        if not urls:
+            return 'No URLs to analyze.'
+
+        if not self._safebrowsing_api_key:
+            return 'Safe Browsing API requires an API key!'
```
kiddinn, line 171: I would move this up much higher... no need to do even query TS or check anything first... this should be the first check

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+        if not url_whitelist:
+            domain_analyzer_whitelisted = current_app.config.get(
+                'DOMAIN_ANALYZER_WHITELISTED_DOMAINS',
+                [],
+            )
+            for domain in domain_analyzer_whitelisted:
+                url_whitelist.add('*.%s/*' % domain)
+
+        logging.info(
+            '{:d} entries on the whitelist.'.format(len(url_whitelist)),
```
kiddinn, line 190: use numbers here (consistentn with the rest of the codebase)

```
'{0:d} entries....
```

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+            lookup_urls.append(url)
+
+        try:
+            safebrowsing_results = self._do_safebrowsing_lookup(
+                lookup_urls,
+                safebrowsing_platforms,
+                safebrowsing_types,
+            )
+        except requests.HTTPError:
+            return "Couldn't reach the Safe Browsing API."
```
kiddinn, line 219: Be consistent in using quotes, use a single one instead of a double

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+        for url, events in urls.items():
```
kiddinn, line 221: shouldn't this be `for url.... in lookup_urls` since those are the only ones you actually did a lookup on.. no need to check the whitelisted ones here, is there?

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+            if url in safebrowsing_results:
```
kiddinn, line 222: save indentation...

```
if url not in safebrowsing_results:
    continue
```

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+                    if threat_type:
+                        tags.append(
+                            'google-safebrowsing-%s' % threat_type.lower(),
+                        )
+
+                    event.add_tags(tags)
+                    event.add_attributes(
+                        {
+                            'google-safebrowsing-threat': safebrowsing_result,
```
kiddinn, line 237: you add the dict directly as an attribute... would it be better to "expand" it or make it "prettier""...

convert it to a string or something... or have it as a json dump... or separate attributes.... or a list "key:value, key:value,..."

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+                        },
+                    )
+                    event.commit()
+
+        return '{:d} Safe Browsing result(s) on {:d} URL(s), ' \
```
kiddinn, line 242: same here as before' {0:d} Safe Bro.... {1:d}....`

kiddinn: and no need to use the backslash...

```
return (
    '{0:d} Safe Browsing result(s) on {1:d} URL(s), ' 
    '(2:d} whitelisted').format(.....)
```

kiddinn:
But otherwise great analyzer, thank you so much for this PR.

I also wanted to check with you whether you didn' twant to add some more code in there, to generate a `view` or `aggregation`, or maybe both?

So before your analyzer is completed, you can add a view by doing:
```
self.sketch.add_view(
    view_name='Safebrowsing Entries', analyzer_name=self.NAME,
    query_string:'tag:"google-safebrowsing-url"',
    additional_fields=['domain', 'url', 'google-safebrowsing-threat'])
```

Then you can also add an aggregation... something like

```
params = {
    'query_string': 'tag:"google-safebrowsing-url"',
    'field': 'google-safebrowsing-threat',
}
self.sketch.add_aggregation(
    name='Top 20 Safebrowsing Threats', agg_name='query_bucket',
    agg_params=params, view_id=view.id, chart_type='hbarchart',
    description='Created by the SafeBrowsing analyzer, label='interesting')
```

Something like that... 

(FYI the "`label=...` will not work until #1176 is pushed, and you re-synced your client)

kiddinn:
I'm also missing the changes to the actual config file... can you add the lines there as well, with some comments for documentation?

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+            'Timesketch',
+        )
+
+        self._google_client_version = current_app.config.get(
+            'SAFEBROWSING_CLIENT_VERSION',
+            pkg_resources.get_distribution('timesketch').version,
+        )
+
+    def _is_url_whitelisted(self, url, whitelist):
+        """Does a glob-match against the whitelist.
```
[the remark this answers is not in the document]
dov-csis: Indeed, that comment is wrong. Otherwise, we think `fnmatch` is good enough for URL wildcards here. Do you think we should keep the functionality as it is now or should we also replace `fnmatch` with something else?

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+                },
+            }
+
+            r = requests.post(
+                endpoint,
+                params={'key': self._safebrowsing_api_key},
+                json=body,
+            )
+
+            r.raise_for_status()
```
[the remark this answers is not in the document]
dov-csis: What do you think, which approach would be the most fitting to handle a `HTTPError` coming from here?

timesketch/lib/analyzers/safebrowsing.py
dov-csis: Which one format of these would you find the prettiest for the contents of this `dict`?

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,250 @@
...
+
+            result = r.json()
+            if result and 'matches' in result:
+                for match in result['matches']:
+                    threat_result = match.copy()
```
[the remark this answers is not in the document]
dov-csis: To avoid changing the dictionary when also iterating it.

timesketch/lib/analyzers/safebrowsing.py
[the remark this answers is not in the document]
kiddinn: I think fnmatch should be fine here, just to match the docstring to what is actually being done

requirements.txt
```diff
@@ -15,6 +15,7 @@ flask_wtf==0.14.2
 google-auth==1.7.0
 google_auth_oauthlib==0.4.1
 gunicorn==19.10.0  # Note: Last version to support py2
+httmock==1.3.0
```
kiddinn, line 18: since you are adding a requirement here, can you also add it into the config/dpkg into the debian packages so that all end to end tests will survive

timesketch/lib/analyzers/safebrowsing.py
kiddinn: a gentle ping about this one

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,277 @@
...
+
+                event.add_tags(tags)
+                event.add_attributes(
+                    {
+                        'google-safebrowsing-threat': safebrowsing_result,
+                    },
+                )
+                event.commit()
+
+        return (
```
kiddinn, line 268: I also had some other comments , but I'll put them here.

No need to create any aggregations? Or views? Or an automatic story?

kiddinn:
any updates?

timesketch/lib/analyzers/safebrowsing.py
dov-csis: No, both are correct. Changing it anyway!

timesketch/lib/analyzers/safebrowsing.py
dov-csis: Thanks, we might add something like these in another PR.

dov-csis:
@kiddinn All issues are addressed!

timesketch/lib/analyzers/safebrowsing.py
```diff
@@ -0,0 +1,288 @@
...
+            'SAFEBROWSING_API_KEY')
+
+        self._google_client_id = current_app.config.get(
+            'SAFEBROWSING_CLIENT_ID',
+            'Timesketch',
+        )
+
+        self._google_client_version = current_app.config.get(
+            'SAFEBROWSING_CLIENT_VERSION',
+            pkg_resources.get_distribution('timesketch').version,
```
kiddinn, line 62: you can now also just do

```
from timesketch import version

...

version.get_version()
```

to the get the TS version information, since it's moved outside of the package resource

dov-csis: Wow, this is useful, so we won't need to import `pkg_resources` in this file. We'll follow up!

PR #1176, opened 2020-04-24
Adding labels to aggregations, adding ability to filter based on labels.

kiddinn:
Adding labels to aggregations. This adds the ability to label aggregations with arbitrary labels. 

Also adding the ability to filter aggregations based on labels in the API client.

### DIFF
--- a/timesketch/models/annotations.py
+++ b/timesketch/models/annotations.py
@@ -135,6 +135,45 @@ class LabelMixin(object):
                               parent=relationship(self)))
         return relationship(self.Label)
 
+    def add_label(self, label, user=None):
+        """Add a label to an object.
+
+        Each entry can have multible labels.
+
+        Args:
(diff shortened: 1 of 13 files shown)


timesketch/models/annotations.py
```diff
@@ -135,6 +135,57 @@ def labels(self):
...
+        Returns:
+            A list of strings with all the applied labels.
+        """
+        if not self.labels:
+            return []
+
+        return [x.label for x in self.labels]
+
+    @property
+    def label_string(self):
```
berggren, line 179: Can you get rid of this one, and add a `as_string' argument to get_labels() instead?

kiddinn: explained offline, that is the need for a property for the marshalling to work (`to_json` call on the aggregation), since the aggregation fields require a property to work.

PR #1177, opened 2020-04-27
Change the behavior of the Yeti analyzer

tomchop:


timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -39,48 +27,58 @@ def __init__(self, index_name, sketch_id):
...
+            item['compiled_regexp'] = re.compile(item['pattern'])
+            self.intel[item['id']] = item
+
+    def mark_event(self, indicator, event, neighbors):
+        """Anotate an event with data from indicators and neighbors.
+
+        Tags with skull emoji, adds a comment to the event.
+        """
+        event.add_emojis([emojis.get_emoji('SKULL')])
+        event.add_tags([n['name'] for n in neighbors])
```
berggren, line 73: Out of curiosity, what is the name of a neighbour? Just wanna make sure we don't end up with too long tag names which will clutter the UI.

tomchop: It's the human-readable name of whatever neighbor it is. As an example, the names of TTPs imported through MITRE ATT&CK are the names of the actual TTPs: https://attack.mitre.org/

So yeah, tag names could end up being long? More importantly (and I just though about this) they could also contain spaces, which might make searching for them a bit tricky. WDYT? We could slugify the names

berggren: Yeah, sluggify would be great!
I rather do it now, before any saved views etc wil depend on the tag name

berggren:
@tomchop ping

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -91,41 +94,38 @@ def run(self):
...
-
-        if actors_found:
-            return '{0:d} actors were found! [{1:s}]'.format(
-                len(actors_found), ', '.join(actors_found))
-        return 'No indicators were found in the timeline.'
+        self.get_indicators('x-regex')
+
+        entities_found = set()
+
+        events = self.event_stream(query_string='*',
```
berggren, line 101: All events? Not even some sort of filter, on data_type maybe?

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -91,41 +94,38 @@ def run(self):
...
+                                   return_fields=['message'])
+        total_matches = 0
+        matching_indicators = set()
+        for event in events:
+            for _id, indicator in self.intel.items():
+                regexp = indicator['compiled_regexp']
+                if regexp.search(event.source['message']):
```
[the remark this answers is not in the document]
tomchop: This makes sense. This is only for 'regex' type of indicators which are super generic and can be applied to almost anything. I picked the message source, because it's what humans read and therefore probably what people would base themselves on to write the regexes.

timesketch/lib/analyzers/yetiindicators.py
tomchop: I am happy to keep all events here if we're gonna use the generic regex approach. Since (for now) we don't really know "where" to look for a specific regex, it's probably better to pass it on all events. 

This would change if the Yeti object had specified a location where the regex is relevant via a tag or an extended field. But this is ontology work I'm not ready to dive in on this PR :)

berggren: Ack, SG.

timesketch/lib/analyzers/yetiindicators.py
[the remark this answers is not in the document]
berggren: SG

PR #1181, opened 2020-05-05
Archive sketch functionality

kiddinn:
Adding the ability to export sketches and archive/unarchive sketches.

This PR adds the ability to:

 + Archive a sketch
 + Export data from a sketch (first pass)
 + Unarchive a sketch
 + API client improvements

Future PRs will include updates mainly to the story export functionality. That is to improve the export functionality (there are few TODOs in the code for those placeholders)

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -1122,9 +1122,9 @@ class Sketch(resource.BaseResource):
             'action': 'archive'
         }
         response = self.api.session.post(resource_url, json=data)
-        answer = response.status_code in definitions.HTTP_STATUS_CODE_20X
-        self._archived = answer
-        return answer
+        return_status = response.status_code in definitions.HTTP_STATUS_CODE_20X
+        self._archived = return_status
+        return return_status
 
     def unarchive(self):
         """Unarchives a sketch and return boolean whether it was succesful."""
@@ -1138,6 +1138,6 @@ class Sketch(resource.BaseResource):
             'action': 'unarchive'
         }
         response = self.api.session.post(resource_url, json=data)
-        answer = response.status_code in definitions.HTTP_STATUS_CODE_20X
-        self._archived = answer
-        return answer
+        return_status = response.status_code in definitions.HTTP_STATUS_CODE_20X
+        self._archived = return_status
+        return return_status
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -57,6 +57,7 @@ class Sketch(resource.BaseResource):
         """
         self.id = sketch_id
         self.api = api
+        self._archived = None
         self._sketch_name = sketch_name
         self._resource_uri = 'sketches/{0:d}'.format(self.id)
         super(Sketch, self).__init__(api=api, resource_uri=self._resource_uri)
@@ -636,10 +637,14 @@ class Sketch(resource.BaseResource):
 
         Raises:
             ValueError: if unable to query for the results.
-            RuntimeError: if the query is missing needed values.
+            RuntimeError: if the query is missing needed values, or if the
+                sketch is archived.
         """
         if not (query_string or query_filter or query_dsl or view):
             raise RuntimeError('You need to supply a query or view')
+
+        if self.is_archived():
+            raise RuntimeError('Unable to query an archived sketch.')
 
         if not query_filter:
             query_filter = {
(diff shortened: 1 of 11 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -57,6 +57,7 @@ def __init__(self, sketch_id, api, sketch_name=None):
         """
         self.id = sketch_id
         self.api = api
+        self._archived = None
```
berggren, line 60: s/None/False

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1085,3 +1097,47 @@ def add_event(self, message, timestamp, timestamp_desc):
             self.api.api_root, self.id)
         response = self.api.session.post(resource_url, json=form_data)
         return response.json()
+
+    def is_archived(self):
+        """Return a boolean indicating whether the sketch has been archived."""
+        if self._archived is not None:
```
berggren, line 1103: Why not simply:
```
if not self._archived:
  return
```

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1085,3 +1097,47 @@ def add_event(self, message, timestamp, timestamp_desc):
...
+            logger.error('Sketch already archived.')
+            return False
+
+        resource_url = '{0:s}/sketches/{1:d}/archive/'.format(
+            self.api.api_root, self.id)
+        data = {
+            'action': 'archive'
+        }
+        response = self.api.session.post(resource_url, json=data)
+        answer = response.status_code in definitions.HTTP_STATUS_CODE_20X
```
berggren, line 1125: s/answer/return_status

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1085,3 +1097,47 @@ def add_event(self, message, timestamp, timestamp_desc):
...
+            logger.error('Sketch wasn\'t archived.')
+            return False
+
+        resource_url = '{0:s}/sketches/{1:d}/archive/'.format(
+            self.api.api_root, self.id)
+        data = {
+            'action': 'unarchive'
+        }
+        response = self.api.session.post(resource_url, json=data)
+        answer = response.status_code in definitions.HTTP_STATUS_CODE_20X
```
berggren, line 1141: same here

data/timesketch.conf
```diff
@@ -29,6 +29,12 @@ SQLALCHEMY_DATABASE_URI = 'postgresql://<USERNAME>:<PASSWORD>@localhost/timesket
 ELASTIC_HOST = '127.0.0.1'
 ELASTIC_PORT = 9200
 
+# Define what labels should be defined that make it so that a sketch and
+# timelines will not be deleted. This can be used to add a list of different
+# labels that ensure that a sketch and it's associated timelines cannot be
+# deleted.
+LABELS_TO_PREVENT_DELETION = ['lithold']
```
berggren, line 36: maybe call the default example label something else? 'preserve', or 'protected'
WDYT?

importer_client/python/timesketch_import_client/helper_test.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2019 Google Inc. All rights reserved.
```
berggren, line 1:
```suggestion
# Copyright 2020 Google Inc. All rights reserved.
```

timesketch/api/v1/resources.py
```diff
@@ -408,6 +467,291 @@ def post(self):
...
+                }
+                zip_file.writestr(
+                    'views/{0:s}.meta'.format(name), data=json.dumps(meta))
+
+            # TODO (kiddi): Add in aggregation group support.
+            # TODO (kiddi): Add in support for comments/stars/labels.
+            # TODO (kiddi): Add in support for all tagged events (includes
+            # a metadata file with a list of all tags and their count).
+
+        if is_archived:
```
berggren, line 679: should be `if not`, right?

timesketch/api/v1/resources.py
```diff
@@ -408,6 +467,291 @@ def post(self):
...
+        sketch.remove_label(ARCHIVE_LABEL)
+
+        indexes_to_open = []
+        for timeline in sketch.timelines:
+            timeline.remove_label(ARCHIVE_LABEL)
+            search_index = timeline.searchindex
+            search_index.remove_label(ARCHIVE_LABEL)
+            indexes_to_open.append(search_index.index_name)
+
+        self.datastore.client.indices.open(','.join(indexes_to_open))
```
berggren, line 708: Let's add a TODO here about moving this to elastic.py datastore abstraction.
```suggestion
        # TODO: Move to lib/datastores/elastic.py
        self.datastore.client.indices.open(','.join(indexes_to_open))
```

api_client/python/timesketch_api_client/sketch.py
kiddinn: no, since I want the value to be filled with a "not filled in" criteria to begin with. since i specifically only do the "GET" request if the value is None.. if this would be set to False, then I would just always send "False" as the responsne to `is_archived`... difficult to distinguish between an actual False value and a False initial value

api_client/python/timesketch_api_client/sketch.py
kiddinn: because I want to specifically check whether the value is None... otherwise I'm checking a boolean.

if this is None,; then I know I haven't actually checked, and then it is time for me to check using the API, if it is True/False, then I know I've checked before and I can just return that value (that is this is just checking the initial value)

api_client/python/timesketch_api_client/sketch.py
kiddinn: done

api_client/python/timesketch_api_client/sketch.py
kiddinn: done

data/timesketch.conf
kiddinn: sure... we can do that.... maybe both?

importer_client/python/timesketch_import_client/helper_test.py
kiddinn: done

timesketch/api/v1/resources.py
kiddinn: no.. in the beginning there is a check to see if the `archive` label is applied. 

If it is, then this variable `is_archived` is set... this means that at the beginning of this phase `is_archived` is set to True... that means in order to do the export I needed to `unarchive` and therefore I need to `archive` again now....

whether this variable should be named something more "clear" is another question.... (renamed, is this more clear?)

timesketch/api/v1/resources.py
kiddinn: we can do that... there are quite a few other `client.indices.X` calls in the resources file, not sure if you want to move all of them.

api_client/python/timesketch_api_client/sketch.py
berggren: ack

api_client/python/timesketch_api_client/sketch.py
berggren: ok

data/timesketch.conf
berggren: Yeah, lets do both

timesketch/api/v1/resources.py
berggren: ah, missed that one. Cool, all good.

timesketch/api/v1/resources.py
berggren: Yeah, let me take care of that later :)

data/timesketch.conf
kiddinn: that was already done

PR #1196, opened 2020-05-16
Remove $ from the README.md

jaegeral:
Remove $ from the README.md to be consistent. In other readme files we also do not write $, thus removing it here, makes it also easier to copy & paste the line


jaegeral:
@googlebot I signed it!

PR #1198, opened 2020-05-16
Fix count in linux sigma rule

jaegeral:
The current count does not work, it raises a SigmaParserError.

That has two reasons as far as I understood, the condition is in the wrong place, and also the way count was proposed in the yaml file is not covered in Sigma, while testing the original one, I got the following exception:
````
  File "/usr/local/lib/python3.6/dist-packages/sigma/parser/rule.py", line 41, in parse_sigma
    conditions = self.parsedyaml["detection"]["condition"]
KeyError: 'condition'

During handling of the above exception, another exception occurred:

  File "/usr/local/src/timesketch/timesketch/lib/analyzers/interface.py", line 803, in run_wrapper
    result = self.run()
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/sigma_tagger.py", line 84, in run
    rule_file_content, self.sigma_config, None)
  File "/usr/local/lib/python3.6/dist-packages/sigma/parser/collection.py", line 61, in __init__
    self.parsers.append(SigmaParser(yamldoc, config))
  File "/usr/local/lib/python3.6/dist-packages/sigma/parser/rule.py", line 29, in __init__
    self.parse_sigma()
  File "/usr/local/lib/python3.6/dist-packages/sigma/parser/rule.py", line 49, in parse_sigma
    raise SigmaParseError("No condition found")
sigma.parser.exceptions.SigmaParseError: No condition found
````

My fix should solve that problem.

### DIFF
--- a/data/linux/recon_commands.yaml
+++ b/data/linux/recon_commands.yaml
@@ -20,5 +20,5 @@ detection:
     - 'hostname -f'
     - 'netstat -nltupw'
     - 'cat /proc/net/*'
+  condition: keywords | count() > 3
 timeframe: 30m
-conditions: count > 3

### CODE REVIEW

**Comment by jaegeral** [comment]
@Onager can you check if the count that you implemented was working for you? Or has the Sigma specification changed since?

(for reference, I am testing it on a fresh created timesketch dev)

**Comment by jaegeral** [comment]
@googlebot I signed it!

File: data/linux/recon_commands.yaml [hunk 1]
```diff
@@ -20,5 +20,5 @@ detection:
     - 'hostname -f'
     - 'netstat -nltupw'
     - 'cat /proc/net/*'
+  condition: keywords | count() > 3
 timeframe: 30m
```
Onager, line 24: It looks like 'timeframe' should also be part of the detection collection

jaegeral:
> Not sure what happened here, though I suspect a change to sigma parsing. This rule worked as far as I can remember.
> 
> It looks the tests need to be extended here as well

Do you want me to look into that (So far I have not touched the tests portion of your sigma code)

berggren:
> > Not sure what happened here, though I suspect a change to sigma parsing. This rule worked as far as I can remember.
> > It looks the tests need to be extended here as well
> 
> Do you want me to look into that (So far I have not touched the tests portion of your sigma code)

@jaegeral if you have the cycles, yea please update the tests :)

data/linux/recon_commands.yaml
Onager: Ping @jaegeral - https://github.com/Neo23x0/sigma/wiki/Specification

jaegeral:
so actually because we are gonna remove the rules directly stored in TImesketch , this PR should be obsolete and once https://github.com/google/timesketch/pull/1231 is merged, I will create another PR to remove the current sigma rules from the repo.

berggren:
Ack, closing this PR

PR #1215, opened 2020-05-25
Adding an analyzer result object to API client.

kiddinn:
This PR adds a simple analyzer result object to the API client. When running an analyzer instead of returning back a JSON blob with the session ID an object is returned, which has functions to fetch the status of the analyzer being run, as well as getting the results back.


api_client/python/timesketch_api_client/analyzer.py
```diff
@@ -0,0 +1,111 @@
...
+            return {}
+
+        data = response.json()
+
+        objects = data.get('objects')
+        if not objects:
+            return {}
+
+        for result in objects[0]:
+            result_id = result.get('analysissession_id', -1)
```
berggren, line 50: -1 ?

berggren: Why not None?

api_client/python/timesketch_api_client/error.py
```diff
@@ -27,3 +27,11 @@ def error_message(response, message=None, error=RuntimeError):
         text = soup.p.string
     raise error('{0:s}, with error [{1:d}] {2!s} {3:s}'.format(
         message, response.status_code, response.reason, text))
+
+
+class Error(Exception):
```
berggren, line 32: Pretty generic name. Maybe ApiClientError?

api_client/python/timesketch_api_client/analyzer.py
kiddinn: No particular reason, other than to get an integer, might as well just have it default to none

api_client/python/timesketch_api_client/error.py
kiddinn: this is not an error to use... that iis this is following "proper way of doing things", see: https://docs.python.org/3/tutorial/errors.html#user-defined-exceptions

```
When creating a module that can raise several distinct errors, a common practice is to create a base class for exceptions defined by that module, and subclass that to create specific exception classes for different error conditions:
```

This is how we have done things in other projects, eg. plaso, etc

berggren: okm

PR #1227, opened 2020-05-26
Disable mans import task

berggren:
The recent addition of a .mans file importer raises a couple of concerns.

1. The library is depending on ciso8601 which is a C library with a SWIG wrapper. This breaks our Docker builds and adds unnecessary dependencies.
2. It does raw ES inserts, which means that we put a lot of trust into how the library deals with ES. It's a bit fragile, and I would like to be able to get a stream of events (either dict or Dataframes) and then use TS internal ES interface.

This PR disables the feature for now until we have sorted these things out with @garanews . I will raise a couple of issues in the mans_to_es project.


berggren:
Hey @garanews head-up. I need to temporary disable this feature because it makes building our docker containers a bit too complex. I filed https://github.com/LDO-CERT/mans_to_es/issues/10 - let me know what you think and I hope we can enable this importer soon again.

berggren:
@garanews I also added this FR for mans_to_es
https://github.com/LDO-CERT/mans_to_es/issues/11

kiddinn:
can you fix/resolve the conflicts and update the branch?

PR #1229, opened 2020-05-27
Added an import helper for reading configs.

kiddinn:
Breaking out all parsing of configs and the use of them away from the importer and into a separate import helper.

Also changing a bit the behavior in the API client for fetching analyzer results.


importer_client/python/timesketch_import_client/helper.py
```diff
@@ -0,0 +1,155 @@
+# Copyright 2019 Google Inc. All rights reserved.
```
berggren, line 1:
```suggestion
# Copyright 2020 Google Inc. All rights reserved.
```

PR #1231, opened 2020-05-27
Add support for Sigma rules

jaegeral:
This is a clean version of the sigma commit, without polluting rules and such.


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "data/sigma"]
```
berggren, line 1: Don't do a sub module here. This will just make packaging difficult and risk of getting old versions of the rules.

I opt for skipping this alltogether, and let the server admin take care of syncing whatever rules they want. We can (in a later PR) add a helper script that syncs various sigma repositories that can be run on the server, either manually or via cron.

WDYT?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -24,6 +26,8 @@ class SigmaPlugin(interface.BaseSketchAnalyzer):
     # this file.
     _RULES_PATH = ''
 
+    # A list of fields to include in the view output.
+    _FIELDS_TO_INCLUDE = ['message']
```
berggren, line 30: You don't need this. Message is the default field if nothing is provided.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -55,7 +59,7 @@ def run_sigma_rule(self, query, tag_name):
         events = self.event_stream(
             query_string=query, return_fields=return_fields)
         for event in events:
-            event.add_tags([tag_name])
+            event.add_tags(["sigma_"+tag_name])
```
berggren, line 62:
```suggestion
            event.add_tags(['sigma_{0:s}'.format(tag_name)])
```

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
-                    number_of_tagged_events = self.run_sigma_rule(
-                        result, tag_name)
-                    tags_applied[tag_name] += number_of_tagged_events
+
+        # The extension to search for
+        exten = '.yml'
+
+        for dirpath, dirnames, files in os.walk(rules_path):
+            for rule_filename in files:
+                if rule_filename.lower().endswith(exten):
```
berggren, line 86:
```suggestion
                if rule_filename.lower().endswith('yml'):
```

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
-
-                for result in results:
-                    logging.info(
-                        '[sigma] Generated query {0:s}'.format(result))
-                    number_of_tagged_events = self.run_sigma_rule(
-                        result, tag_name)
-                    tags_applied[tag_name] += number_of_tagged_events
+
+        # The extension to search for
+        exten = '.yml'
```
berggren, line 82:
```suggestion

```

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
+
+        for dirpath, dirnames, files in os.walk(rules_path):
+            for rule_filename in files:
+                if rule_filename.lower().endswith(exten):
+                    print(os.path.join(dirpath, rule_filename))
```
berggren, line 87: Remove?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
+                    if os.path.isdir(os.path.join(rules_path, rule_filename)): # if a sub dir is found, append it to be scanned for rules
```
berggren, line 88: Place comment above instead

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
+            }
+            agg_obj = self.sketch.add_aggregation(
+                name='Top 20 Sigma tags', agg_name='field_bucket',
+                agg_params=agg_params, view_id=view.id, chart_type='hbarchart',
+                description='Created by the Sigma analyzer')
+
+
+            story = self.sketch.add_story("Sigma Rule hits")
+            story.add_text(
+                utils.BROWSER_STORY_HEADER, skip_if_exists=True)
```
berggren, line 154: BROWSER_STORY_HEADER?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
+                    tag_name, _ = rule_filename.rsplit('.')
+                    tags_applied[tag_name] = 0
+                    rule_file_path = os.path.join(dirpath, rule_filename)
+                    rule_file_path = os.path.abspath(rule_file_path)
+                    logging.info('[sigma] Reading rules from {0!s}'.format(
+                        rule_file_path))
+                    with open(rule_file_path, 'r') as rule_file:
+                        try:
+                            rule_file_content = rule_file.read()
+                        except UnicodeDecodeError as exception:
```
berggren, line 103: Why do we get a UnicodeDecodeError here? Shouldn't we support unicode files?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,123 @@ def run(self):
...
+                        parser = sigma_collection.SigmaCollectionParser(
+                            rule_file_content, self.sigma_config, None)
+                        try:
+                            results = parser.generate(sigma_backend)
+                        except NotImplementedError as exception:
+                            logging.error(
+                                'Error generating rule in file {0:s}: {1!s}'.format(
+                                    rule_file_path, exception))
+                            continue
+                        except Exception as exception:
```
berggren, line 117: This is very broad. Can we narrow it down? I guess this is for when the rule cannot be read?

jaegeral: True, but while testing I had several exception when running all rules from Sigma, from values being null etc. So there is basically no way to tell which Exceptions we want to catch, as those exceptions are raised by the sigma project.

.gitmodules
jaegeral: sure, is it okay to then add the data/sigma folder into .gitignore?
For the time being I will then add sections in one or more docs to explain how to pull the repo and or place the own rules.

jaegeral: ok adressed in https://github.com/jaegeral/timesketch/commit/c778e1de1ed53e1127518b296768246603ca8db2

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: ack, done

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: ack done

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: true, removing it

jaegeral:
So besides the one open conversation of the broad exception handling (that I would like to keep, while testing it I at least had the following exceptions:
- Aggregations not implemented for this backend
- 'ascii' codec can't decode byte 0xe2 in position 532: ordinal not in range(128)
- List values must be strings or numbers

This should be fine for another round of reviews.
ping @berggren

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -3,10 +3,15 @@
 
 import logging
 import os
+import time
+import elasticsearch
```
jaegeral, line 7: @berggren heads up, not sure if that is okay to use / import here, but below I tried to use the most accurate exception type for es timeout, that is the reason behind importing it here

berggren: That should be fine

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -70,47 +74,128 @@ def run(self):
...
+                                'Error generating rule in file {0:s}: {1!s}'.format(
+                                    rule_file_path, exception))
+                            continue
+
+                        for sigma_rule in parsed_sigma_rules:
+                            try:
+                                simple_counter += 1
+                                # TODO fix that in Sigma hack to get rid of nested stuff
+                                # https://github.com/google/timesketch/issues/1199#issuecomment-639475885
+                                sigma_rule = sigma_rule.replace(".keyword:", ":")
```
berggren, line 125: Why not query .keyword here? .keyword is the raw string in ES, and I guess that the sigma rule depends on that?

jaegeral: Ah that is a weird one, because A sigma rule might have nested values, so if you are searching for username.keyword, the field in timesketch would actually be only username. So searching for username.keyword does not match.

The Sigma project also not sure how to solve that: https://github.com/Neo23x0/sigma/issues/705

And there is no way to make nested mappings either in the yaml config file to say:
username.keyword from Sigma matches username in timesketch.

That is why I introduced the (not so nice and elegant) hack here.

jaegeral: (https://github.com/Neo23x0/sigma/tree/master/tools#elasticsearch-or-elk)

PR #1233, opened 2020-05-27
Update Dockerfile-prod to install Timesketch from GIFT PPA

joachimmetz:


docker/build/Dockerfile-prod
```diff
@@ -1,25 +1,36 @@
...
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get -y update
+RUN apt-get -y install apt-transport-https apt-utils
+RUN apt-get -y install libterm-readline-gnu-perl software-properties-common
+RUN add-apt-repository -y ppa:gift/$PPA_TRACK
+
+RUN apt-get -y update
+RUN apt-get -y upgrade
```
berggren, line 18:
```suggestion

```

Doing an upgrade is considered a bit dangerous, because some packages cannot upgrade inside an unprivileged container. We trust that the Ubuntu image is being kept up2date.

docker/build/Dockerfile-prod
```diff
@@ -1,25 +1,36 @@
...
 
 # Install Plaso
-RUN add-apt-repository ppa:gift/stable
-RUN apt-get update && apt-get -y install \
-    plaso-tools \
- && rm -rf /var/lib/apt/lists/*
+RUN apt-get -y install plaso-tools
+
+# Install Timesketch
+RUN apt-get -y install python3-timesketch
```
berggren, line 27:
```suggestion
RUN apt-get -y install timesketch-server
```

Should we install timesketch-server here, as discussed offline?

docker/build/Dockerfile-prod
joachimmetz: We've been using this with Plaso for a while now, and has not caused issues so far. The Dockerfile CI test should catch breakage early.

joachimmetz: For context: https://github.com/log2timeline/plaso/blob/master/config/docker/Dockerfile#L22

berggren: Ack - lets keep it here then, and remove if we get problems later down the road.

PR #1242, opened 2020-05-29
Adding the ability to add tags to events.

kiddinn:
Adding the ability in the API to add tags to a list of events (similar to comments/labels). This is added into the annotation API call.

Also added this ability into the API client.

### DIFF
--- a/timesketch/api/v1/resources.py
+++ b/timesketch/api/v1/resources.py
@@ -1854,6 +1854,100 @@ class EventResource(ResourceMixin, Resource):
         }
         return jsonify(schema)
 
+class EventTaggingResource(ResourceMixin, Resource):
+    """Resource to fetch and set tags to an event."""
+
+    @login_required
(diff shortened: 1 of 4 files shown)


timesketch/api/v1/resources.py
```diff
@@ -1850,6 +1854,100 @@ def get(self, sketch_id):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST, 'Tags need to be a list')
+
+        if not all([isinstance(x, str) for x in tags_to_add]):
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Tags need to be a list of strings')
+
+        events = form.get('events', [])
+        for _event in events:
```
berggren, line 1907: We should move this to the elastic datastore instead, seeing we are using this from more than one place. Let's do that later though.

PR #1245, opened 2020-06-02
Replace term whitelist with allowlist

jaegeral:
This is mostly a refactoring thing, I searched for whitelist and replaced it with allowlist.


jaegeral:
Context for that: https://developers.google.com/style/inclusive-documentation#features-and-users

berggren:
Yes, good point. Let's get rid of the term whitelist. Allow list is not necessarily the right term for all of these though. It's context sensitive. I'll add in comments in the code where I think we should name it something else.

We also need to take care of not breaking folks who have a config file already. Please add a warning to logs saying that the config parameter "whitelist" has been deprecated and urge folks to update their config file.

config/dpkg/timesketch-server.timesketch.default
```diff
@@ -110,7 +110,7 @@ GOOGLE_OIDC_HOSTED_DOMAIN = None
 # If populated only these users (email addresses) will be able to login to
 # this server. This can be used when access should be limited to a specific
 # set of users.
-GOOGLE_OIDC_USER_WHITELIST = []
+GOOGLE_OIDC_USER_ALLOWLIST = []
```
berggren, line 113: GOOGLE_OIDC_ALLOWED_USERS

config/dpkg/timesketch-server.timesketch.default
```diff
@@ -174,7 +174,7 @@ DOMAIN_ANALYZER_WATCHED_DOMAINS_SCORE_THRESHOLD = 0.75
 
 # A list of domains that are frequent source of false positives
 # in the "phishy" domain comparison, mostly CDNs and similar.
-DOMAIN_ANALYZER_WHITELISTED_DOMAINS = ['ytimg.com', 'gstatic.com', 'yimg.com', 'akamaized.net', 'akamaihd.net', 's-microsoft.com', 'images-amazon.com', 'ssl-images-amazon.com', 'wikimedia.org', 'redditmedia.com', 'googleusercontent.com', 'googleapis.com', 'wikipedia.org', 'github.io', 'github.com']
+DOMAIN_ANALYZER_ALLOWLISTED_DOMAINS = ['ytimg.com', 'gstatic.com', 'yimg.com', 'akamaized.net', 'akamaihd.net', 's-microsoft.com', 'images-amazon.com', 'ssl-images-amazon.com', 'wikimedia.org', 'redditmedia.com', 'googleusercontent.com', 'googleapis.com', 'wikipedia.org', 'github.io', 'github.com']
```
berggren, line 177: DOMAIN_ANALYZER_EXCLUDE_DOMAINS

config/dpkg/timesketch-server.timesketch.default
```diff
@@ -201,7 +201,7 @@ EMAIL_FROM_USER = 'nobody'
 EMAIL_SMTP_SERVER = 'localhost'
 
 # Only send emails to these users.
-EMAIL_USER_WHITELIST = []
+EMAIL_USER_ALLOWLIST = []
```
berggren, line 204: EMAIL_RECIPIENTS

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -41,8 +41,8 @@ def __init__(self, index_name, sketch_id):
 
         self.domain_scoring_threshold = current_app.config.get(
             'DOMAIN_ANALYZER_WATCHED_DOMAINS_SCORE_THRESHOLD', 0.75)
-        self.domain_scoring_whitelist = current_app.config.get(
-            'DOMAIN_ANALYZER_WHITELISTED_DOMAINS', [])
+        self.domain_scoring_allowlist = current_app.config.get(
```
berggren, line 44: Can't come up with anything rightnow, but this one need to reflect what it is. Some other name.. I'll get back on that

timesketch/views/auth.py
```diff
@@ -264,22 +264,22 @@ def validate_api_token():
     validated_email = token_json.get('email')
 
     # Check if the authenticating user is part of the allowed domains.
-    domain_whitelist = current_app.config.get('GOOGLE_OIDC_HOSTED_DOMAIN')
-    if domain_whitelist:
+    domain_allowlist = current_app.config.get('GOOGLE_OIDC_HOSTED_DOMAIN')
```
berggren, line 267: hosted_domains

timesketch/lib/analyzers/phishy_domains.py
jaegeral: DOMAIN_ANALYZER_DONOTCHECK_DOMAINS would be what this means the most maybe?

jaegeral: > Yes, good point. Let's get rid of the term whitelist. Allow list is not necessarily the right term for all of these though. It's context sensitive. I'll add in comments in the code where I think we should name it something else.
> 
> We also need to take care of not breaking folks who have a config file already. Please add a warning to logs saying that the config parameter "whitelist" has been deprecated and urge folks to update their config file.

Just to verify I understand correctly, do you suggest a warning for people who update should get a warning, or when someone tries to get a "new" value but has an "old" config.

The first I would do in the setup.py. The later would be an if clause at every call that I changed. Or do we have a better way to "age" configs? If there is a PR that did something similar, happy to copy that.

berggren: So, we rarely change old things in the config (maybe never?), just add new things.
If we rename this config parameter, it will break installations that have an old version of the config. 

We need to warn on this in the __init__.py file, where all the Flask init stuff happens.

berggren: > DOMAIN_ANALYZER_NOCHECK_DOMAINS would be what this means the most maybe?

DOMAIN_ANALYZER_EXCLUDE_DOMAINS
and
self.domain_scoring_exclude_domains

WDYT?

jaegeral: ACK; I will give it a try and come up with a warning etc.

Would it be an idea to introduce a version number in the config file that we can use looking forward to check such things?

berggren: Yeah, it's complicated :) The config file is meant to be copied by the server admin, and it could in theory contain additions etc. Also, when new stuff is introduced to the standard config most folks will just copy it over to their old config file.

So I think that a version on the config isn't worth it.

berggren: So, you need to do a check in all places where the config variable is used, and warn there, telling the user that it has been deprecated. You need to support both names for a while.

timesketch/__init__.py
```diff
@@ -70,6 +70,10 @@ def create_app(config=None):
         os.environ['TIMESKETCH_SETTINGS'] = config
         try:
             app.config.from_envvar('TIMESKETCH_SETTINGS')
+            if app.config.get('EMAIL_RECIPIENTS') is None:
```
berggren, line 73: None is default if not found. 

```suggestion
            if not app.config.get('EMAIL_RECIPIENTS'):
```

timesketch/__init__.py
```diff
@@ -70,6 +70,10 @@ def create_app(config=None):
...
+                sys.stderr.write(
+                    "Warning, please update your timesketch.conf "
+                    "format. https://github.com/google/timesketch/pull/1245")
```
berggren, line 76:
```suggestion
                    "Warning, EMAIL_USER_WHITELIST has been deprecated. "
                    "Please update timesketch.conf.")
```

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -41,8 +41,17 @@ def __init__(self, index_name, sketch_id):
...
         self.domain_scoring_threshold = current_app.config.get(
             'DOMAIN_ANALYZER_WATCHED_DOMAINS_SCORE_THRESHOLD', 0.75)
-        self.domain_scoring_whitelist = current_app.config.get(
-            'DOMAIN_ANALYZER_WHITELISTED_DOMAINS', [])
+        self.domain_scoring_exclude_domains = current_app.config.get(
+            'DOMAIN_ANALYZER_EXCLUDE_DOMAINS', [])
+
+        """ TODO remove that after a 6 months, this following check is to ensure
+        backportability of config file
+        """
```
berggren, line 49:
```suggestion
        # TODO: remove that after a 6 months, this following check is to ensure
        # compatibility of the config file.
```

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -41,8 +41,17 @@ def __init__(self, index_name, sketch_id):
...
+        if len(self.domain_scoring_exclude_domains) == 0:
+            print("Warning, please update your timesketch.conf format. https://github.com/google/timesketch/pull/1245")
```
berggren, line 51:
```suggestion
            print("Warning, DOMAIN_ANALYZER_WHITELISTED_DOMAINS has been deprecated. "
                    "Please update timesketch.conf.")
)
```

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -244,9 +253,9 @@ def run(self):
                 text = 'Domain {0:s} is similar to {1:s}'.format(
                     domain, ', '.join(similar_text_list))
                 if any(domain.endswith(
-                        x) for x in self.domain_scoring_whitelist):
-                    tags_to_add.append('whitelisted-domain')
-                    whitelist_encountered = True
+                        x) for x in self.domain_scoring_exclude_domains):
+                    tags_to_add.append('allowlisted-domain')
```
berggren, line 257: s/allowlisted-domain/known-domain

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -262,12 +271,12 @@ def run(self):
                 view_name='Phishy Domains', analyzer_name=self.NAME,
                 query_string='tag:"phishy-domain"')
 
-            if whitelist_encountered:
+            if allowlist_encountered:
                 self.sketch.add_view(
-                    view_name='Phishy Domains, excl. whitelist',
+                    view_name='Phishy Domains, excl. allowlist',
```
berggren, line 276:
```suggestion
                    view_name='Phishy Domains, excl. known domains',
```

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -262,12 +271,12 @@ def run(self):
...
                     analyzer_name=self.NAME,
                     query_string=(
-                        'tag:"phishy-domain" AND NOT tag:"whitelisted-domain"'))
+                        'tag:"phishy-domain" AND NOT tag:"allowlisted-domain"'))
```
berggren, line 279:
```suggestion
                        'tag:"phishy-domain" AND NOT tag:"known-domain"'))
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
             return abort(
                 HTTP_STATUS_CODE_UNAUTHORIZED,
                 'Domain {0:s} is not allowed to authenticate against this '
                 'instance.'.format(domain))
 
-    user_whitelist = current_app.config.get('GOOGLE_OIDC_USER_WHITELIST')
-    # Check if the authenticating user is on the whitelist.
-    if user_whitelist:
-        if validated_email not in user_whitelist:
+    user_allowlist = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
```
berggren, line 276:
```suggestion
    allowed_users = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
+    """ TODO remove that after a 6 months, this following check is to ensure
+           backportability of config file
+           """
```
berggren, line 279:
```suggestion
    # TODO: Remove that after a 6 months, this following check is to ensure
    # compatibility of config file
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
+    if len(user_allowlist) == 0:
+        print("Warning, please update your timesketch.conf format. https://github.com/google/timesketch/pull/1245")
```
berggren, line 281:
```suggestion
        print("Warning, GOOGLE_OIDC_USER_WHITELIST has been deprecated. "
                  "Please update timesketch.conf.")
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
+        user_allowlist = current_app.config.get(
+            'GOOGLE_OIDC_USER_WHITELIST', [])
+
+    # Check if the authenticating user is on the allowlist.
+    if user_allowlist:
```
berggren, line 286:
```suggestion
    if allowed_users:
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
+        if validated_email not in user_allowlist:
```
berggren, line 287:
```suggestion
        if validated_email not in allowed_users:
```

timesketch/views/auth.py
```diff
@@ -264,22 +264,30 @@ def validate_api_token():
...
             return abort(
                 HTTP_STATUS_CODE_UNAUTHORIZED,
-                'Unauthorized request, user not in whitelist')
+                'Unauthorized request, user not in allowlist')
```
berggren, line 290:
```suggestion
                'Unauthorized request, user not allowed')
```

timesketch/views/auth.py
```diff
@@ -358,14 +366,14 @@ def google_openid_connect():
             'Unable to validate request, with error: {0!s}'.format(e))
 
     validated_email = decoded_jwt.get('email')
-    user_whitelist = current_app.config.get('GOOGLE_OIDC_USER_WHITELIST')
+    user_allowlist = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
```
berggren, line 369:
```suggestion
    allowed_users = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
```

timesketch/views/auth.py
```diff
@@ -358,14 +366,14 @@ def google_openid_connect():
...
 
-    # Check if the authenticating user is on the whitelist.
-    if user_whitelist:
-        if validated_email not in user_whitelist:
+    # Check if the authenticating user is on the allowlist.
+    if user_allowlist:
```
berggren, line 372:
```suggestion
    if allowed_users:
```

timesketch/views/auth.py
```diff
@@ -358,14 +366,14 @@ def google_openid_connect():
             'Unable to validate request, with error: {0!s}'.format(e))
 
     validated_email = decoded_jwt.get('email')
-    user_whitelist = current_app.config.get('GOOGLE_OIDC_USER_WHITELIST')
+    user_allowlist = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
 
-    # Check if the authenticating user is on the whitelist.
-    if user_whitelist:
-        if validated_email not in user_whitelist:
+    # Check if the authenticating user is on the allowlist.
```
berggren, line 371:
```suggestion
    # Check if the authenticating user is allowed.
```

timesketch/views/auth.py
```diff
@@ -358,14 +366,14 @@ def google_openid_connect():
...
+    if user_allowlist:
+        if validated_email not in user_allowlist:
```
berggren, line 373:
```suggestion
        if validated_email not in allowed_users:
```

timesketch/views/auth.py
```diff
@@ -358,14 +366,14 @@ def google_openid_connect():
...
             return abort(
                 HTTP_STATUS_CODE_UNAUTHORIZED,
-                'Unauthorized request, user not in whitelist')
+                'Unauthorized request, user not in allowlist')
```
berggren, line 376:
```suggestion
                'Unauthorized request, user not allowed')
```

timesketch/lib/analyzers/phishy_domains.py
jaegeral: LGTM, done in https://github.com/google/timesketch/pull/1245/commits/1989e31b5ea2efadfe816fa0ed739270c1d07483

kiddinn:
there are some merge conflicts that you need to resolve

kiddinn:
you may need to change the `app.py` file now, since I removed the code from the old `__init__` file. please re-sync and check the `app.py`

jaegeral:
Checks are moved to the app.py file-

timesketch/__init__.py
```diff
@@ -24,7 +24,4 @@ def create_app(config=None):
 
     Returns:
         Application object (instance of flask.Flask).
-    """
-    # pylint: disable=import-outside-toplevel
-    from timesketch import app
-    return app.create_app(config)
+    """
```
kiddinn, line 27: why did you delete these lines?

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -41,8 +41,17 @@ def __init__(self, index_name, sketch_id):
...
             'DOMAIN_ANALYZER_WATCHED_DOMAINS_SCORE_THRESHOLD', 0.75)
-        self.domain_scoring_whitelist = current_app.config.get(
-            'DOMAIN_ANALYZER_WHITELISTED_DOMAINS', [])
+        self.domain_scoring_exclude_domains = current_app.config.get(
+            'DOMAIN_ANALYZER_EXCLUDE_DOMAINS', [])
+
+        # TODO: remove that after a 6 months, this following check is to ensure
+        # compatibility of the config file.
+        if len(self.domain_scoring_exclude_domains) == 0:
+            print("Warning, DOMAIN_ANALYZER_WHITELISTED_DOMAINS has been deprecated. "
```
kiddinn, line 50: please use single quotes, keep the consistency (' vs ")

timesketch/views/auth.py
```diff
@@ -265,22 +265,30 @@ def validate_api_token():
...
-    user_whitelist = current_app.config.get('GOOGLE_OIDC_USER_WHITELIST')
-    # Check if the authenticating user is on the whitelist.
-    if user_whitelist:
-        if validated_email not in user_whitelist:
+    allowed_users = current_app.config.get('GOOGLE_OIDC_ALLOWED_USERS')
+    # TODO: Remove that after a 6 months, this following check is to ensure
+    # compatibility of config file
+    if len(user_allowlist) == 0:
+        print("Warning, GOOGLE_OIDC_USER_WHITELIST has been deprecated. "
+                  "Please update timesketch.conf.")
```
kiddinn, line 282: single vs double quotes

kiddinn: also probably better to use the logger instead of print

timesketch/__init__.py
jaegeral: hm guess they have been deleted by the refactoring, I dit not intentionally mess this this part of the init file at all.

timesketch/views/auth.py
jaegeral: done, used logging, not sure if correct usage in auth.py, and needed to import logging in the analyzer, but that seems reasonable.

timesketch/lib/analyzers/phishy_domains.py
jaegeral: done

timesketch/__init__.py
kiddinn: but you are deleting this with this PR, so I would like to ask you to revert that, otherwise TS will stop working for few tools

PR #1246, opened 2020-06-02
Adding the ability to export sketches in the API client

kiddinn:
Adding the ability to export sketches in the API client, as well as a minor bug fix

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -1158,7 +1158,8 @@ class Sketch(resource.BaseResource):
                 '{0:s} first'.format(directory))
 
         if not file_path.lower().endswith('.zip'):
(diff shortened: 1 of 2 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1141,3 +1141,39 @@ def unarchive(self):
...
+        Raises:
+            RuntimeError: if sketch cannot be exported.
+        """
+        directory = os.path.dirname(file_path)
+        if not os.path.isdir(directory):
+            raise RuntimeError(
+                'The directory needs to exist, please create: '
+                '{0:s} first'.format(directory))
+
+        if not file_path.lower().endswith('.zip'):
```
berggren, line 1160: Can we just append .zip in this case?

kiddinn: sure, we can do that....

PR #1251, opened 2020-06-04
Refactor archive API as well as expanding export support

kiddinn:
Expanding upon the export support in the archive API call.

This PR brings about:

- Refactor of the API
- Exports of starred events
- Exports of all events with comments (and their comments)
...


timesketch/api/v1/archive.py
```diff
@@ -114,6 +270,126 @@ def post(self, sketch_id):
             HTTP_STATUS_CODE_NOT_FOUND,
             'The action: [{0:s}] is not supported.'.format(action))
 
+    def _get_all_events_with_a_label(self, label, sketch):
+        """Returns a DataFrame with events in a sketch with a certain label.
+
+        Args:
+            label (string): the label string to search for.
+            sketch...
```
berggren, line 278:
```suggestion
            sketch (timesketch.models.sketch.Sketch): a sketch object.
```

Issue #1254, opened 2020-06-05
Psort ingestion error 'no attribute create_app'


cshanahan:
I understand Timesketch installation is a moving target right now, and fixes are being developed.

I installed Timesketch using pip from Git clone today. Timesketch seems to be installed correctly. I can run the server (tsctl runserver) and log into the app. However, I cannot use Plaso psort to ingest a Plaso DB file into Timesketch.

I am using Plaso 20200430, elasticsearch 7.5.1 and python3-elasticsearch 7.6.0 (I can't find a package 7.5.x). I am running Plaso and Timesketch on the same system, Ubuntu 20.04.

Is there a way to test whether all required packages are installed and set to the correct version, aside from what the current requirements file lists? Similar to "log2timeline.py --troubles"

My command
```
$ psort.py -z "UTC" -o timesketch --name SOMENAME --index SOMEINDEX --username USERNAME MY_PLASO_FILE.plaso "date > '2020-04-01 00:00:00' AND date < '2020-05-31 23:59:59'"
```

Python debug output
```
2020-06-04 20:32:22,238 [INFO] (MainProcess) PID:7741 <data_location> Determined data location: /usr/share/plaso
Traceback (most recent call last):
  File "/usr/bin/psort.py", line 95, in <module>
    if not Main():
  File "/usr/bin/psort.py", line 30, in Main
    if not tool.ParseArguments(sys.argv[1:]):
  File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 405, in ParseArguments
    self.ParseOptions(options)
  File "/usr/lib/python3/dist-packages/plaso/cli/psort_tool.py", line 494, in ParseOptions
    self._output_module = self._CreateOutputModule(options)
  File "/usr/lib/python3/dist-packages/plaso/cli/tool_options.py", line 178, in _CreateOutputModule
    output_module = output_manager.OutputManager.NewOutputModule(
  File "/usr/lib/python3/dist-packages/plaso/output/manager.py", line 139, in NewOutputModule
    return output_class(output_mediator)
  File "/usr/lib/python3/dist-packages/plaso/output/timesketch_out.py", line 40, in __init__
    self._timesketch = timesketch.create_app()
AttributeError: module 'timesketch' has no attribute 'create_app'
```

kiddinn:
This is due to #1248 where we moved functions, including `create_app` away from `timesketch` to `timesketch.app`.

This is a bug that needs to be fixed on plaso side, I will take care of that .

[...]

PR #1255, opened 2020-06-05
Hot patching the timesketch __init__ file until a new plaso release comes out

kiddinn:
This is a small fix to make sure #1254 is not affecting current Timesketch installations.


kiddinn:
This is a temporary change that will be removed with the next release of plaso

Issue #1016, opened 2019-10-29
Granular ACL controls


berggren:
Default sharing permissions are "read". There should be an option for other permissions as well.

jaegeral:
Closing due to lack of activity

PR #1262, opened 2020-06-08
Adding more granular ACL controls

kiddinn:
This PR allows for more granular controls of granting ACLs to a sketch.

ATM it is only possible to add users/groups to a sketch, with the permissions granted being only default values. See #1016

What this PR does is:

1. Moving collaborator API call out of resources files.
2. Adding the options in the API to define what permissions an user or a group is granted
3. Adding the ability into the API client to define the set of permissions to grant


kiddinn:
Closed this since there is another PR in the making to split the resources file into smaller pieces. Will wait until that is done and then pick up

PR #1264, opened 2020-06-08
Refactor API

berggren:
The resources.py file started to get too big. This PR split up the API resources in their own files.


timesketch/api/v1/resources/__init__.py
```diff
@@ -0,0 +1,241 @@
+# Copyright 2015 Google Inc. All rights reserved.
```
kiddinn, line 1: first of all, let's update the date here...

but the secnod question... why is this in `__init__` why not in a separate `resources` file or `common.py` or something like that?

berggren: Done.

for the __init__, I'll keep it like this for now to not break all the imports etc. We can change later.

@kiddinn PTAL

kiddinn: __init__ vs common vs something else was discussed offline... also this is inside resources/__init__ not just __init__ so this is fine

PR #1267, opened 2020-06-11
Changing the behavior of the API calls for event tagging

kiddinn:
This PR changes the approach to event tagging to doing some bulk lookups to better test against large tagging requests.

There was an issue with the previous tagging approach which resulted in both too many requests to the backend, since there was a single ES search per event that needed to be tagged. Now searches are done in bulk, just as the actual changes to the document is. 

Also made some changes in increasing the size of the bulk requests (increased the threshold value)

### DIFF
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -343,6 +343,22 @@ class EventTaggingResource(resources.ResourceMixin, Resource):
 
         events = form.get('events', [])
         event_df = pd.DataFrame(events)
+
+        for field in ['_id', '_type', '_index']:
+            if field not in event_df:
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    'Events need to have a [{0:s}] field associated '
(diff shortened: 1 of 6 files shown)


kiddinn:
Marking as a draft, since I saw one thing I need to fix before I'm ready.

timesketch/api/v1/resources/event.py
```diff
@@ -291,43 +345,140 @@ def post(self, sketch_id):
...
-                    'Unable to tag event: {0:s}, couldn\'t find the '
-                    'event.'.format(_event['_id']))
-                continue
 
-            source = ds_events[0].get('_source', {})
-            existing_tags = source.get('tag', [])
-            new_tags = list(set().union(existing_tags, tags_to_add))
-
-            if set(existing_tags) == set(new_tags):
+                size = len(should_list) + 100
```
berggren, line 408: why +100 here?

kiddinn: added a comment

Issue #1270, opened 2020-06-12
`add_index` option no longer functions


cshanahan:
**TS Version**
Using Timesketch from Git master; installed on 2020-06-10
Using Plaso 20200430 from GIFT PPA

**STEPS TO REPRODUCE**
- Use psort from Plaso to ingest a plaso DB file into Elasticsearch
- Run `tsctl` with `add_index` option to associate an existing index with Timesketch
    - `tsctl add_index --index <my elastic index name> --name <name in timesketch> --user <my timesketch username>`
- `tsctl add_index` command completes without error
- Run `tsctl run_server` log in to the app and create a new Sketch
- From the Sketch UI the new index is not available as an option to add to the Sketch (it doesn't appear in the UI)

The index in Elasticsearch is fine. I can explore the index using other apps, including Kibana.

If it helps, Elasticsearch is pinned to v7.5.1

kiddinn:
in this case, have you tried to use the timesketch importer? that can be used from every machine that can reach teh web interface of TS.

cshanahan:
> in this case, have you tried to use the timesketch importer? that can be used from every machine that can reach teh web interface of TS.

I have never had much success with the import function. I almost always work with Plaso DB files, not JSONL or CSV.

Can you provide an example command using _import_ ?

For example, I just performed a test using the same Plaso DB file. I copied the Plaso DB file to the Timesketch server and ran the following (sanitized) command:

- tsctl import -h
usage: tsctl import [-?] --file FILE_PATH [--sketch_id SKETCH_ID] [--username USERNAME] [--timeline_name TIMELINE_NAME]

- tsctl import --file MY_TIMELINE.plaso --sketch_id importtest --username MY_USERNAME --timeline_name MY_TIMELINE                

And this was the traceback:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/sqlalchemy/engine/base.py", line 1245, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.8/dist-packages/sqlalchemy/engine/default.py", line 581, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.InvalidTextRepresentation: invalid input syntax for type integer: "importtest"
LINE 3: WHERE sketch.id = 'importtest'
                          ^
The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/tsctl", line 8, in <module>
    sys.exit(main())
[...]

Issue #1269, opened 2020-06-12
Delete Sketch from UI doesn't remove Sketch from DB


cshanahan:
**TS Version**
Using Timesketch from Git master; installed on 2020-06-10

**STEPS TO REPRODUCE**
Run `tsctl runserver` and log into the app. Select an existing Sketch or create a new (test) Sketch. Select Delete from the UI drop-down option. Sketch appears to be deleted correctly and no errors are displayed. The Sketch is no longer available from the UI.
From CLI, run `tsctl list_sketches' and the Sketch that I deleted is still listed in the output.

kiddinn:
The sketch does not get deleted from the DB, and should not, it gets set into a `delete` state, and that should preclude it from being listed in `list_sketches` function. I'll fix that.

cshanahan:
> The sketch does not get deleted from the DB, and should not, it gets set into a `delete` state, and that should preclude it from being listed in `list_sketches` function.

That makes sense, thank you.

Issue #1268, opened 2020-06-12
[FR] Add version flag to tsctl


cshanahan:
It would be helpful if there was a version option available in tsctl, for example:
`tsctl version`

Perhaps this would also make it better when filing bug reports and feature requests. We could be more specific about the version of TS in use.

PR #1272, opened 2020-06-14
Various changes to tsctl and the API

kiddinn:
This changes few things:

1. Adds version information to tsctl, #1268
2. Adds version information to the API and adds a query to the API client to look up the backend version.
3. Adding status check to `tsctl list_sketches`, #1269
4. Adding a `description` property to the index API client object.
5. Adding some error handling into the API for fetching sketch to deal with potentially missing searchindexes in timelines
6. Adding some error handling into the `add_index` option in tsctl, see #1270
7. Changes to how searchindexes are deleted in the API 
8. Some other error handling or checks added to the API.

### DIFF
--- a/timesketch/api/v1/resources/searchindex.py
+++ b/timesketch/api/v1/resources/searchindex.py
@@ -47,7 +47,8 @@ class SearchIndexListResource(resources.ResourceMixin, Resource):
             List of search indices in JSON (instance of flask.wrappers.Response)
         """
         indices = SearchIndex.all_with_acl(current_user).all()
-        return self.to_json([i for i in indices if i.get_status.status != 'deleted'])
+        return self.to_json(
+            [i for i in indices if i.get_status.status != 'deleted'])
 
     @login_required
     def post(self):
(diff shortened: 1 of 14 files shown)


timesketch/api/v1/resources/searchindex.py
```diff
@@ -126,6 +126,7 @@ def delete(self, searchindex_id):
                 HTTP_STATUS_CODE_FORBIDDEN,
                 '\n'.join(error_strings))
 
+        searchindex.set_status(status='deleted')
         db_session.delete(searchindex)
```
berggren, line 130: Why setting the status when you delete the object next?

berggren: Do we really want to delete the object from the datastore? Or just mark it as deleted? This is what happens when you delete a sketch. It is not fully deleted from SQL, just marked as such.

timesketch/api/v1/routes.py
```diff
@@ -33,6 +33,7 @@
 from .resources.sketch import SketchResource
 from .resources.sketch import SketchListResource
 from .resources.archive import SketchArchiveResource
+from .resources.information import TSVersionResource
```
berggren, line 36: IMO, rename to VersionResource

timesketch/api/v1/resources/searchindex.py
kiddinn: this is the behavior as it has been so far... but that is a good question, I think we should just rather close the index and mark it as deleted.

timesketch/api/v1/routes.py
kiddinn: agree

Issue #1199, opened 2020-05-18
Improve Sigma support


jaegeral:
To track it here, there is a basic support for sigma in TS, but the vast majority of rules from the Sigma project would not run out of the box.

Reasons for that:
- missing field mapping (esp. if sketch was created based on plaso data)

Some todos on that side:
- [x] Add documentation how to use it / how to create rules
- [x] create basic field mappings to cover plaso parsed evtx files
- [x] Add a story / aggregation after it was ran, to be able to get an idea of what is found

jaegeral:
- [ ] Provide links to sample data

jaegeral:
-[ ] have a folder only for internal rules that people could deploy

jaegeral:
- [ ] Ignore deprecated folders in sigma rule folder, otherwise it would include https://github.com/Neo23x0/sigma/tree/master/rules/windows/deprecated which isncludes a rule that causing to many tags

jaegeral:
There is an issue that IN some rules, it is checking for 
```
Message.content = ""
```

When setting 
```
Message : xml_string
```
 in the mapping file, it will end in a rule that searches for:
```
xml_string.content
```
Will of course fail, this will be adressed with a small hack, that is not really nice.

jaegeral:
- [ ] Set a wait timer in case to many Timesketch queries are caused for ES.

jaegeral:
- [ ] Fix issue with rule: 2b30fa36-3a18-402f-a22d-bf4ce2189f35 to generate to many hits.

PR #1286, opened 2020-06-26
Sigma cleanup

jaegeral:
Friday cleanup in Sigma code.

The current linux rules in the repo where basically broken, and given the discussions with @berggren we decided to not ship any sigma rules (yet) but I decided to ship one rule with a sample event file (which is part of this PR) to enable folks to test the functionality overall.

* introduce one sigma rule to show how it works
* modify the sigma mapping

This PR also will change the expected way people should store their rules (for now)

* refactored the analyzer names to only have three types:
** sigma_sample (with the one that is shipped with timesketch)
** sigma_test (for all files that people deveop on their instance)
** sigma_all (for all other sigma rules they have under data/sigma/

That should make it much cleaner and less broken and easier to test by the maintainers as the literally only have to add the shipped jsonl file and hit the run sigma-sample analyzer to run.


kiddinn:
you'll need to update the branch, first of all.. and then secondly to fix the tests... unit tests are failing:
```
ERROR: Test analyzer.
  File "/usr/lib/python3/dist-packages/mock/mock.py", line 1369, in patched
    return func(*newargs, **newkeywargs)
  File "/timesketch/timesketch/lib/analyzers/sigma_tagger_test.py", line 27, in test_analyzer
    _ = sigma_tagger.LinuxRulesSigmaPlugin(
AttributeError: module 'timesketch.lib.analyzers.sigma_tagger' has no attribute 'LinuxRulesSigmaPlugin'
Ran 132 tests in 27.629s
```

jaegeral:
To pass all the tests, I adjusted some lines and refactored minor things to increase readability of the code.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -175,19 +174,19 @@ def run(self):
         return output_string
 
 
-class LinuxRulesSigmaPlugin(SigmaPlugin):
+class SampleRulesSigmaPlugin(SigmaPlugin):
```
berggren, line 177: Do we need this? Is it for testing? Sample of what? :)
It clutters the UI with this in the available analyzer list.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -175,19 +174,19 @@ def run(self):
...
     """Sigma plugin to run Linux rules."""
 
-    _RULES_PATH = '../../../data/sigma/rules/linux'
+    _RULES_PATH = '../../../data/sigma_sample'
 
-    NAME = 'sigma_linux'
+    NAME = 'sigma_sample'
 
-class WindowsRulesSigmaPlugin(SigmaPlugin):
+class AllRulesSigmaPlugin(SigmaPlugin):
```
berggren, line 184: Do we intend to have all Sigma rules under this analyzer? If that is the case, can we call it simply `SigmaPlugin`?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -175,19 +174,19 @@ def run(self):
...
     """Sigma plugin to run Windows rules."""
 
-    _RULES_PATH = '../../../data/sigma/rules/windows'
+    _RULES_PATH = '../../../data/sigma/rules/'
 
-    NAME = 'sigma_windows'
+    NAME = 'sigma_all'
 
 class TestRulesSigmaPlugin(SigmaPlugin):
```
berggren, line 191: Same here - do we want this to be visible in the UI?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -104,41 +104,40 @@ def run(self):
...
-                                    rule_file_path, exception))
+                                'Error generating rule in file {0:s}: {1!s}'
+                                .format(rule_file_path, exception))
                             continue
 
                         for sigma_rule in parsed_sigma_rules:
                             try:
                                 simple_counter += 1
-                                # TODO fix that in Sigma hack to get rid of nested stuff
+                                # TODO fix that in Sigma to remove nested stuff
```
kiddinn, line 118: what does this mean? `to remove nested stuff` what stuff? what nested

timesketch/lib/analyzers/sigma_tagger.py
kiddinn: agree

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -196,6 +195,6 @@ class TestRulesSigmaPlugin(SigmaPlugin):
 
     NAME = 'sigma_test'
 
-manager.AnalysisManager.register_analyzer(LinuxRulesSigmaPlugin)
-manager.AnalysisManager.register_analyzer(WindowsRulesSigmaPlugin)
+manager.AnalysisManager.register_analyzer(SampleRulesSigmaPlugin)
+manager.AnalysisManager.register_analyzer(AllRulesSigmaPlugin)
 manager.AnalysisManager.register_analyzer(TestRulesSigmaPlugin)
```
kiddinn, line 200: why do we need three here?

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: Yeah so the rationale behind is that we have no specific folder for OS, so I assumed it is better to have one for developing rules that would not be pushed to github.

However I see the issue here.
Would it make more sense if we say we have one Analyzer in the UI and it grabs all rules in the folder?

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: this nested stuff: https://github.com/google/timesketch/issues/1199#issuecomment-639475885

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: resolved by removing

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: (the adjustment of the comment was only to address too long lines)

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: fixed

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: resolved by removing

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -107,46 +107,37 @@ def run(self):
...
+                                'Error generating rule in file {0:s}: {1!s}'
+                                .format(rule_file_path, exception))
                             continue
 
                         for sigma_rule in parsed_sigma_rules:
                             try:
                                 simple_counter += 1
-                                # TODO fix that in Sigma hack to get rid of
-                                # nested stuff. Also see:
+                                # TODO fix that in Sigma to remove nested stuff
```
berggren, line 118:
```suggestion
                                # TODO Investigate how to handle .keyword fields in Sigma.
```

timesketch/lib/analyzers/sigma_tagger.py
berggren: Adjusted the comment.

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: good suggestion, added

PR #1298, opened 2020-07-09
Adding the `admin` function to the User model

kiddinn:
What this PR does is to add an `admin` bool to the User model, and then change the API to allow limited viewing of data if you have the admin bit set.

Having admin rights gives you the access to:

1. See all sketches in the store (name/description/sketch ID)
2. Archive all sketches in the store

For now, that's it.

Additionally there is some more error handling added to the API client to deal with sketches that don't have full information associated to them (due to looking at them as an admin but not having access to the underlying data).

Finally a functionality is added to `tsctl` to set the admin bit for a user (and remove it)


timesketch/api/v1/resources/sketch.py
```diff
@@ -64,9 +64,12 @@ def get(self):
         Returns:
             List of sketches (instance of flask.wrappers.Response)
         """
-        filtered_sketches = Sketch.all_with_acl().filter(
-            not_(Sketch.Status.status == 'deleted'),
-            Sketch.Status.parent).order_by(Sketch.updated_at.desc()).all()
+        if current_user.admin:
```
berggren, line 67: Do we really want to list all sketches here? Wont this mess up the UI for all admin users?

timesketch/api/v1/resources/sketch.py
```diff
@@ -110,14 +113,49 @@ def post(self):
 class SketchResource(resources.ResourceMixin, Resource):
     """Resource to get a sketch."""
 
+    def _get_sketch_for_admin(self, sketch):
+        """Missing a docstring."""
```
berggren, line 117: It is indeed missing a docstring :)

kiddinn: it is really missing it... that is so true

kiddinn: done

timesketch/api/v1/resources/sketch.py
kiddinn: see screenshots sent to you offline.

API client has been changed to accommodate this, but I assume we need to make some UI changes, at least change the messaging, or have it in a similar way as the `archived` mode.. we can change the status to "admin_unknown" or something like that, and in the UI check if that's the status and then display a similar windows as you see when you visit an archived sketch... that is you can see the name/description of the sketch and then have the option of archiving it, that's it.. since ethat's all you're supposed to see anyway.

WDYT?

Can you make the UI change afterwards.. I can change the status or whatever needs to be done to make it easy to identify in the UI.

kiddinn:
answered, comments, PTAL @berggren

kiddinn:
I did change the status name, to make it easier to do the UI change, let me know WDYT @berggren

timesketch/api/v1/resources/sketch.py
berggren: ack, as discussed offline. sounds good.

PR #1302, opened 2020-07-10
Updated dpkg files

joachimmetz:


dependencies.ini
```diff
@@ -162,7 +162,7 @@ rpm_name: python3-google-auth-oauthlib
 
 [gunicorn]
 dpkg_name: python3-gunicorn
-minimum_version: 19.10.0
```
berggren, line 165: As discussed offline - 19.7.1 is the one shipped with Ubuntu 18.04. hence the downgrade here.

1.19.x should be fine.

PR #1304, opened 2020-07-14
Minor changes to test tools

kiddinn:
Minor updates to reflect changes to interface.


test_tools/timesketch/lib/analyzers/interface.py
```diff
@@ -62,7 +63,13 @@ def get_config_path(file_name):
...
         return path
 
-    return None
+    path = os.path.join(
+        os.path.dirname(__file__), '..', '..', '..', '..', 'data', file_name)
+    path = os.path.abspath(path)
+    if os.path.isfile(path):
+        return path
+
+    return ''
```
jaegeral, line 72: What is the reason you return '' instead of None before?

kiddinn: since you do expect a path to come back.... in the analyzer code you often use something like `open(file_path)` and then you fail if you dont get a string (well you will fail as well since fthe path does not exist, but the error code might make more sense)

jaegeral: ack sounds reasonable.

PR #1316, opened 2020-08-04
Added scrolling support into export API call

kiddinn:
If the results of a query exceed the size limit of said query the current API does not contain those results. For instance if a view defines a limit of 40 or 200 records in a query, then the resulting exported file does not include more entries.

This PR fixes that issue and includes scrolling support as well as to overwrite the size limit to reduce the amount of search queries needed to get all the data.


timesketch/api/v1/resources/archive.py
```diff
@@ -470,21 +470,48 @@ def _export_view(self, view, sketch, zip_file):
...
+            enable_scroll=True,
             indices=indices)
 
-        fh = _query_results_to_filehandle(result, sketch)
+        scroll_id = result.get('_scroll_id', '')
+        if scroll_id:
+            data_frame = _query_results_to_dataframe(result, sketch)
+
+            total_count = result.get(
+                'hits', {}).get('total', {}).get('value', '')
```
tomchop, line 497: What if this fails? (and `total_count` is `''`)? The comparison in line 500 will raise an exception

kiddinn: yes, changed a bit

kiddinn:
@tomchop  PTAL

Issue #1106, opened 2020-02-14
Export events


berggren:
Need to be able to:
* export selected events
* export all events with a specific label, e.g. Starred events

kiddinn:
do we want this to be exported just as a CSV, or is the idea to support other mechanism?

kiddinn:
that is, do we want this to be modular, eg. story export style, or do we just want to stick to one format, either CSV or JSONL?

jaegeral:
To add to this thing, having export for CSV and JSONL would be even better.

kiddinn:
we already have the option in the UI to do export, which will export `all` the things...

we also have the option to use the API client to do the search, which comes back as a DataFrame, which can be easily exported into whatever format.

What is missing is really having in the UI an option after doing a search to export the results from that search.

This could be achieved using the explore API call, to add an option of `export` or `action = export` or something like that. If that option is set then the results are not sent back but instead exported (can use the same code as the other export functionality)

Maybe best to split up the export functionality into a common file, that both the export API call and the Explore one can use, that would be an easy way to support this. Then just to add some ui dust/magic to make it visible to the user.

Passimist:
I struggle to find the option in the UI to export something for hours now. @kiddinn could you please tell me if this has to be configurated somehow? I installed the latest timesketch version following the guide. I've seen the export button in screenshots but for me its nowhere to be found.

rocket-ops:
@Passimist, I believe currently the UI export is only available in the Dev version.

adulau:
MISP export of

- selected events
- all events with a defined label (starred)

There is already a MISP object template for time-sketch (which can be adjusted if needed):

https://www.misp-project.org/objects.html#_timesketch_timeline

The idea would be able to share a timeline via the MISP and benefit from the timeline view in MISP. This can be done easily if a first_seen/last_seen is set for each the timesketch_timeline object for each of the exported event. As a complementary step, a relationship can be added (https://www.misp-project.org/objects.html#_relationships) such as "followed-by" and "preceding-by" which are already used by the mactime import.

kiddinn:
> I struggle to find the option in the UI to export something for hours now. @kiddinn could you please tell me if this has to be configurated somehow? I installed the latest timesketch version following the guide. I've seen the export button in screenshots but for me its nowhere to be found.

@Passimist  Sorry for the late answer, but yes you'll need the current dev version of TS, since the release does not have that feature as of yet, we'll create a new release shortly that contains this export feature.

[...]

PR #1317, opened 2020-08-05
Added the ability to export events (not yet in the UI)

kiddinn:
This implements the ask in #1106, that is in the backend.

What is missing is to add the UI controls to make it visible.

A new parameter to the `sketch.explore` is also added to make it possible to export searches via the API client to files...

```
sketch.explore(..... , file_name='/tmp/foo.zip')
```

Takes in the regular parameters for an explore function, just adds the ability to indicate a filename.


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -797,6 +805,12 @@ def explore(self,
                 response, message='Unable to query results',
                 error=ValueError)
 
+        if file_name:
+            # TODO: Add in scrolling support (requires changes in API as well).
```
berggren, line 809: I thought scrolling was fixed?

kiddinn: this was a copy/paste

PR #1318, opened 2020-08-05
Initial stab at API client documentation

kiddinn:
This is the first version/first attempt at creating documentation for the API Client


docs/APIClient.md
```diff
@@ -0,0 +1,208 @@
+# API Client
+
+The API client is a set of Python libraries that can be used to interact with the REST API of Timesketch from notebooks or scripts. It takes
+care of setting up authentication, sending the API calls to the server, error handling, and presentation.
+
+This is not meant to be a full fledged documentation of the API client, but rather a starting point for people on how to use the API client in
```
jaegeral, line 6: Maybe something like:

docs/APIClient.md
```diff
@@ -0,0 +1,208 @@
...
+## Connecting to a Sketch
+
+There are two ways of getting a sketch object, either by listing all available sketches or by fetching a specific one.
+To fetch a specific one use the `get_sketch` function:
+
+```
+sketch = ts_client.get_sketch(SKETCH_ID)
+```
+
+All you need is the sketch ID, eg:
```
jaegeral, line 65: There is inconsistent use of "you" vs "the user"

docs/APIClient.md
```diff
@@ -0,0 +1,208 @@
...
+
+Each of these properties can be accessed using `sketch.<PROPERTY>`, eg. `sketch.name`, or `sketch.labels`
+
+## Explore Data
+
+The sketch object has several ways to explore data, via aggregations or searches.
+
+### Views
+
+The first is simply to list up all views... the function `list_views` returns back View object for all views. An example overview would be:
```
jaegeral, line 89: "To list all the views use the `list_views` function. This functions returns a View object for all views....."

docs/APIClient.md
kiddinn: maybe something like?????

jaegeral: Hm interesting that the proposal is gone

"This documentation will give an overview for the most common use cases. Some methods implemented in the API client might not be covered here".

Idea of that change is to start the sentence positive.

kiddinn: SGTM

PR #1321, opened 2020-08-06
Added an EVTX Gap analyzer and fixed logging (filter out ES logs)

kiddinn:
Added an analyzer that inspects EVTX files in attempt to find gaps in the Event logs, potentially discovering someone deleting/removing EVTX records.

Other changes:

+ Added `timeline_name` to the analyzer interface.
+ Changes to the logger configuration, to filter out elasticsearch logging records that were filling the logs
+ Changed all instances where the root logger was being used in favor of a named logger.
+ Fixed missing title and other information from exported stories.


timesketch/lib/analyzers/win_evtxgap.py
```diff
@@ -0,0 +1,296 @@
...
+        current_days = [int(day) for day in event_count.day.values]
+        missing_days = list(set(all_days).difference(set(current_days)))
+
+        missing_ranges = list(get_range(missing_days, all_days))
+
+        if not (missing_ranges and record_gaps):
+            return (
+                'No gaps were identified in the EVTX logs, that is not to say '
+                'there aren\'t any gaps, just that these tests weren\'t able '
+                'to find ones.')
```
berggren, line 153: s/ones/any

Issue #1314, opened 2020-08-02
Review from UserGuide


jaegeral:
We need a complete review of the User-Guide.

E.g.: 
- [ ] mark add manual events as currently not available.
- [x] mark Heatmaps as currently not available

jaegeral:
Done

PR #1329, opened 2020-08-12
remove heatmap and manual events from userdoc

jaegeral:
Addressing https://github.com/google/timesketch/issues/1314 and updating the table of contents.


jaegeral:
I kind of don't want to remove it completely as it might confuse people, that is why I chose to strike it. Lmk if it should go completely.

PR #1330, opened 2020-08-12
Add tabbed view of sketch list

berggren:
This PR splits up the sketch list view in different tabs to make the UI a bit cleaner.

Screenshot 2020-08-12 at 14 56 20


timesketch/frontend/src/views/Home.vue
```diff
@@ -62,15 +62,37 @@ limitations under the License.
...
+                  <ts-sketch-list :sketches="myArchivedSketches"></ts-sketch-list>
+                </div>
+              </div>
+            </b-tab-item>
+
+        </b-tabs>
       </div>
     </div>
 
+    <!--
```
jaegeral, line 95: Why do you want to keep that as a comment?

berggren: good catch, removed now.

PR #1331, opened 2020-08-13
UI: Display Data Sources per Timeline & Make them Clickable

### LINKED ISSUE #1214: UI: Display Data Sources per Timeline & Make them Clickable
Details
-
The request consists of 3 parts:

1) For each of the timelines that part of a sketch, it would be great to have a list of all data sources that it includes and for each of them the number of entries. Ideally, those data sources should be converted from technical (e.g. "syslog:cron:task_run") into more human readable format (e.g. "Cron Jobs").
2) Each of these data sources to be clickable. Upon click, it would open the Search with pre-populated filter for that data type. Ideally, I'd like the view to be different and instead of showing default columns, show all of the columns for that data type.
3) Have a tick-box next to each of the data sources so that multiple of them can be selected. The ideal behaviour would be to be able to select multiple using the check-boxes, or click on any particular one. The difference could be that selecting multiple data sources at once would display them in a "summary" view, while clicking on each separately would show a detailed view (i.e. a view with all the columns for that data type).

Mock-up Solution
-
TS Data Sources (1 of 2)
-
TS Data Sources (2 of 2)

Reasoning
-
**Helps:**
- Prioritise
- Break down analysis
- Determine completeness (analysis & collection)

**To consider:**
- Focused collection VS Comprehensive collection

binglot:
Fixes #1214 :
- for each of the timelines it lists all data sources that it includes
- each of these data sources is clickable
- have a tick-box next to each of the data sources so that multiple of them can be selected


binglot:
Here's a demo:

TS-1

TS-2

TS-3

binglot:
# FYI @berggren - I've added expand/collapse to the Data Source list.

TS-expand_collapse

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
```diff
@@ -135,6 +135,32 @@ limitations under the License.
       Added {{ timeline.updated_at | moment("YYYY-MM-DD HH:mm") }}
       <span class="tag is-small" :title="meta.stats[timeline.searchindex.index_name]['count'] + ' events in index'">{{ meta.stats[timeline.searchindex.index_name]['count'] | compactNumber }}</span>
     </span>
+
+    <div v-if="timelineStatus === 'ready'" class="is-size-6 small-top-margin">
+      <a @click="toggleExpansion(timeline)">
```
berggren, line 140: You can simplify this a bit by using buefy. We are supporting Buefy components that implements some useful logic and UI interactions. See: https://buefy.org/documentation/collapse

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
```diff
@@ -135,6 +135,32 @@ limitations under the License.
...
+        <span class="icon">
+            <i :class="[isExpanded(timeline) ? 'fas fa-minus-circle' : 'fas fa-plus-circle']"></i>
```
berggren, line 142: What do you think of putting the expand/collapse control on the same line as `Added 2020-06-05 08:50 ...`? That way we don't get a new line after each timeline in the list.

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
```diff
@@ -168,6 +194,7 @@ import ApiClient from '../../utils/RestApiClient'
 import TsAnalyzerListDropdown from './AnalyzerListDropdown'
 import TsAnalyzerSessionDetail from './AnalyzerSessionDetail'
 import TsAnalyzerHistory from './AnalyzerHistory'
+import router from '../../router'
```
berggren, line 197: You don't have to import the router. It is already present in the Vue instance. See my comment below for an example.

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
```diff
@@ -237,6 +266,26 @@ export default {
...
+    openFilteredTimeline: function (index, dataTypes) {
+      let searchQuery = ''
+      for (let i = 0; i < dataTypes.length; i++) {
+        const dt = dataTypes[i];
+        if (i != 0) {
+          searchQuery += ' OR '
+        }
+        searchQuery += 'data_type:"' + dt + '"'
+      }
+      router.push({name: 'SketchExplore', query: { index: index, q: searchQuery }})
```
berggren, line 279: Use the router already present in the Vue instance:
`this.$router.push({name: 'SketchExplore', query: { index: index, q: searchQuery }})`

binglot: Done

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
binglot: @berggren - how about this?
TS-removed-whitespace

binglot: Thanks Johan! As discussed on the call, I've implemented it as a button on the right side instead of a separate text below.

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
binglot: I'll leave it as a todo for another merge request :)

berggren: 👍

Issue #1332, opened 2020-08-13
Pagination is broken if events/page changes


berggren:
There is a bug when the frontend code doesn't reset the scroll/pagination when you change the number of events to show per page.

Repro: paginate, then change the result/page to be bigger than the result set. This will break the pagination and it get stuck. Also reports the wrong numbers, e.g. 1001-480 / 480

berggren:
Fixed.

PR #1340, opened 2020-08-18
Pagination bugfix

### LINKED ISSUE #1165: Make sort direction a toggle
When viewing a timeline, you have the option in the top right to sort by asc or desc. I don't see there ever being more than those two options, so the dropdown selector seems overkill. It would save a click to make clicking on the "asc" or "desc" text toggle it to the other sorting direction.

berggren:
This PR fix a bug in pagination where the page location wasn't updated when page size was changed. The initial fix is to reset the pagination and search to display the first page as soon as page size changes.

To keep the location for the user we need to investigate another pagination component as it is not trivial with the current solution.

As a small bonus that sneaked in: Rename Fields --> Customize columns in the UI for a more intuitive wording.

Fixes #1165 #1332


berggren:
Travis error is a timeout, nothing is actually error out. Merging.

PR #1343, opened 2020-08-18
UI: Add export button for search results

berggren:
This PR adds a button to download/export search result as a CSV.

Screenshot 2020-08-18 at 12 16 23


berggren:
@kiddinn I added a loading spinner. PTAL

PR #1347, opened 2020-08-25
Display datetime based on timestamp

berggren:
In the UI we are currently rendering the datetime field based on the datetime string in the event. This is fragile if the datetime string doesn't contain a timezone indication. Then the render will be local time.

This PR fixes this by always using the timestamp (millis since unix epoc) instead when parsing/render.

It also contains a couple of UI polish:
* Show second resolution by default (with a control to show micros)
* Cut off long timeline names to reduce clutter. They will be available in full if hovered upon.


kiddinn (approved):
LGTM

PR #1349, opened 2020-08-26
Optimize UI for quicker archive

berggren:
The UI is a bit sluggish when archiving. This PR moves some of the delay to the background and makes the UI respond faster.
Also adds a loading indicator when loading the home page.


kiddinn (approved):
LGTM

PR #1350, opened 2020-08-26
Filter for events that have comments

berggren:
This PR adds the ability to filter for events that have comments on them. This is similar to the current Starred filter.

Screenshot 2020-08-26 at 09 31 40

Screenshot 2020-08-26 at 11 04 48


kiddinn (approved):
LGTM

PR #1354, opened 2020-08-28
DRAFT: Changing chain behavior

kiddinn:
*[this PR is very much a WIP ATM, it will not work, it will not do what it claims it does... it needs quite a bit of change, but the idea is described in the text]*

This PR will change the chain behavior to make it so that each chain has a defined TYPE.

Each event will not be able to have more than one chain PER type (unless it is session, an event can be part of more than a single session).

This means that if a plugin chains two events together, and then another plugin ties two other events together, but they share the same type, then these two chains will be merged into one.

A single event can then be part of more than one session, but only one `event` chain for instance. That means that the data structure for the `chains` attribute needs to change. It will still be a list of dicts, but the dict content will change a bit, it will have a list of plugins that generated the chain, and a combined count of the leafs.

If an event is considered to be a base event in a chain but gets merged into another chain, it will be turned into a leaf (there maybe some exceptions, this is still a WIP)


kiddinn:
One thing to keep in mind, is in a merger of chains, what dictates when a base is turned into a leaf? Is it just an order of business, first one gets to decide? or is there a temporal element to this? or some other conditions that control it?

PR #1355, opened 2020-09-01
Analyzer run description

jaegeral:
Refactoring variable names and make sure it is better explained that analyzer_run does not actually run search queries.


timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -125,9 +125,9 @@ def run(self):
                                 logger.info(
                                     '[sigma] Generated query {0:s}'
                                     .format(sigma_rule))
-                                sum_of_tagged_events = self.run_sigma_rule(
+                                tagged_events_counter = self.run_sigma_rule(
                                     sigma_rule, tag_name)
-                                tags_applied[tag_name] += sum_of_tagged_events
+                                tags_applied[tag_name] += tagged_events_counter
                             except elasticsearch.TransportError \
```
berggren, line 131: Can you take the opportunity to refactor so we don't need the "\" . Just to stay consistent with the rest of the codebase.

test_tools/analyzer_run.py
```diff
@@ -22,6 +22,9 @@
   $ python analyzer_run.py --test_file test_file.txt \
       ../timesketch/lib/analyzers/my_analyzer.py MyAnalyzerSketchPlugin
 
+Remark: The tool ignores Elastic Search queries. It should be fed with
```
berggren, line 25:
```suggestion
Remark: The tool ignores Elasticsearch queries. It should be fed with
```

berggren (commented):
Just some small nits, otherwise LGTM

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: Not sure if I understand which " you mean? You mean: ```query_string='tag:"sigma*"')``` ?

berggren: damn it, GH removed my \

berggren: I meant, remove the / :)

berggren (changes requested):
PTAL

kiddinn (approved):
LGTM

PR #1357, opened 2020-09-03
Small update to the 'Show data types' button

binglot:
1) Fixed the fact that the disabled button 'Show data types' was still clickable. 
2) Changed the collapse implementation to use Buefy.


berggren (approved):
LGTM

PR #1358, opened 2020-09-03
Minor changes to the way date is parsed when a custom event is added

kiddinn:
Adding an error handling to the custom `add_event`, to catch badly formed datestrings.

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -1268,7 +1268,8 @@ class Sketch(resource.BaseResource):
 
         Args:
             message: A string that will be used as the message string.
-            date: A string with the timestamp of the message.
+            date: A string with the timestamp of the message. This should be
+                in a human readable format, eg: "2020-09-03T22:52:21".
             timestamp_desc : Description of the timestamp.
             attributes: A dict of extra attributes to add to the event.
             tags: A list of strings to include as tags.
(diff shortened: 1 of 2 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1268,7 +1268,8 @@ def add_event(
 
         Args:
             message: A string that will be used as the message string.
-            date: A string with the timestamp of the message.
+            date: A string with the timestamp of the message. This should be
```
joachimmetz, line 1271: Why not ISO 8601 format ?

kiddinn: well, because we use `dateutil.parser` to parse the string, which supports much more than just ISO 8601, you can still use something like "Mon Sep 6 14:32:12' and get a timestamp out of it... which is not ISO 8601.

Is that preferable to use ISO 8601, yes of course. Might we want to move to dfdatetime and maybe enforce ISO 8601, maybe, but for now you can basically input whatever string the dateutil parser supports.

joachimmetz: "Mon Sep 6 14:32:12" that looks more like RFC 1123

I would standardize here on ISO 8601 to prevent unintended conversions issues https://github.com/dateutil/dateutil/issues 

I see no issues with the changes in this PR, but I opt we have a broader conversation about this.

berggren: Agree to have a broader conversation about this. Deffering that outside of this PR. :)

berggren (approved):
Small nit, otherwise LGTM

PR #1363, opened 2020-09-09
Adding the ability to define return fields in the search_by_label function.

kiddinn:


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1222,11 +1222,20 @@ def tag_events(self, events, tags, verbose=False):
...
+            self, label_name, return_fields=None, max_entries=None,
+            as_pandas=False):
         """Searches for all events containing a given label.
 
         Args:
             label_name: A string representing the label to search for.
+            return_fields (str): A comma separated string with a list of fields
+                that should be included in the response. Optional and defaults
+                to None.
+            max_entries (int): Optional integer denoting a best effort to limit
```
berggren, line 1235: Why doesn't this honor the 'size' attribute in the query_filter?

berggren (changes requested):

api_client/python/timesketch_api_client/sketch.py
kiddinn: this is different... this is the max integer... since we have a ES limit of 10k, this is to be able to provide a larger number yet not "complete the entire scrolling window" if you don't want it.

So if I have 200k results, and I get only 40k back, I can increase the max entries window to let's say 100k, which would not give me all the records, but at least first 100k of it.

This is irrelevant of the size in the query filter, that does not allow you to get more than ES is capable of providing. This has been part of the explore function for quite some time now, and I'm just adding it to the search by label.

berggren: got it. SGTM

berggren (approved):

PR #1365, opened 2020-09-10
UI: Updating time filters

### LINKED ISSUE #1338: Editable timefilters
Adding one or multiple time filters is easy, but they cannot be edited. This fact makes it difficult to quickly change the time range while exploring.

This feature request is to make the time filter chip clickable and editable.

binglot:
Fix #1338. Allow the users to update the time filters that they've created by clicking on them. Also added some UX improvements, such as closing the dropdown when the filter is added/updated, or auto-populating the end time input based on the start time.


timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -116,6 +117,9 @@ export default {
       this.startDateTime = startDateTimeMoment.format(dateTimeTemplate)
       this.endDateTime = endDateTimeMoment.format(dateTimeTemplate)
 
+      // Change focus to the 'endDateTime' input
+      event.target.parentNode.parentNode.parentNode.getElementsByClassName('input')[1].focus()
```
berggren, line 121: This is a little difficult to follow. If you need to get the DOM element you can use the Ref. See this for an example:
https://vuejs.org/v2/guide/components-edge-cases.html#Accessing-Child-Component-Instances-amp-Child-Elements

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -130,6 +134,19 @@ export default {
       this.$emit('addChip', this.chip)
       this.startDateTime = ''
       this.endDateTime = ''
+
+      // Close the menu
+      this.$parent.$parent.isActive = false;
```
berggren, line 139: This is difficult to reason about. What prop is being mutated here? In what component. Can you emit an event and catch that in the calling component instead?

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -130,6 +134,19 @@ export default {
...
+    },
+    update: function() {
+      if (!(this.startDateTime && this.endDateTime)) {
+        return
+      }
+      this.chip['value'] = this.startDateTime + ',' + this.endDateTime;
+      this.$emit('updateChip', this.chip)
+
+      // Close the menu
+      this.$parent.$parent.isActive = false;
```
berggren, line 149: Same here

berggren (changes requested):

berggren:
First round done, @binglot PTAL

binglot:
@berggren - ready for the 2nd round of review :)

berggren (approved):
LGTM

PR #1366, opened 2020-09-10
Several fixes: workflow changed, size in query_filter is honored in API client, special types are stripped in uploads and bulk insert errors are surfaced

kiddinn:
This PR makes several changes:

+ GH Action Workflow is changed to work on PR instead of just a push to the repo
+ API client now honors the `size` and `terminate_after` in query filters.
+ Special columns are stripped from uploads (_id, _doc,...)
+ Errors in bulk uploads are now surfaced to the end user 
+ Rewrite the datetime field in the importer to make sure dates are properly formed before uploading.
+ Changed the UI to surface the error in importing.

One of the issues with the API Client was that if you supplied a `query_filter` which had a `size` or `terminate_after` field it did not get honored. What happened is that ES returned back only records with the size of `size/terminate_after` but since scrolling was enabled, searches continued until all search results were returned back (instead of one query of size 10k until results are done we had one query of size `size` until all results were done). This meant that search queries using these size restrictions took much longer to complete, since they entailed lot more queries before results were returned as well as returning unexpected amount of events back.

Another issue is bulk upload of events. TS ignored the feedback from ES and therefore lost all tracking of whether ES actually indexed the event or not. Now these are surfaced both in the errors logs as well as to the end user via the description field.

There was also an issue with importing data, where the datetime field was set in the dataset but it wasn't formed properly. In that case the importer didn't do anything, since it was already present and uploaded it as is, with Elastic then later throwing it away since it didn't conform to proper date parsing.


importer_client/python/timesketch_import_client/importer.py
```diff
@@ -174,6 +177,21 @@ def _fix_data_frame(self, data_frame):
             if 'timestamp' in data_frame:
                 data_frame['datetime'] = data_frame['timestamp'].dt.strftime(
```
berggren, line 178: You wanna make sure that timestamp is an int here. I encountered this when hacking on the CLI client and all of a sudden the timestamp was a str and this raised an error.

`data_frame['timestamp'].astype(int).dt.strftime ... `

timesketch/lib/utils.py
```diff
@@ -73,6 +73,18 @@ def _parse_tag_field(row):
     return [row]
 
 
+def _scrub_special_tags(dict_obj):
+    """Remove Elastic specific fields from a dict."""
+    if '_id' in dict_obj:
```
berggren, line 78: Maybe something like:

```
fields_to_remove = ['_id', '_type', '_index', '_source']
for field in fields_to_remove:
    if field in dict_obj:
        _ = dict_obj.pop(field)
```

berggren (changes requested):

importer_client/python/timesketch_import_client/importer.py
kiddinn: done

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
```diff
@@ -34,6 +34,7 @@ limitations under the License.
                 <li>Original name: {{ timeline.searchindex.name }}</li>
                 <li>Added by: {{ timeline.searchindex.user.username }}</li>
                 <li>Added: {{ timeline.searchindex.created_at | moment("YYYY-MM-DD HH:mm") }}</li>
+		<li v-if="timelineStatus === 'ready'">Import Error Message: <b>{{ timeline.searchindex.description }}</b></li>
```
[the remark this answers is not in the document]
kiddinn: well... these are error messages for a timeline that is `ready`, but with errors... the fact is that you can have imported 90% of the events, with few errors, so it's ready but it does contain some issues....

And the text is either goign to be empty (or the name of the timeline if this was uploaded prior to this PR) or. it is goign to contain an error message, .... but your call

timesketch/lib/utils.py
kiddinn: sure, can do that (this was copy / paste) yours is a bit cleaner ;)

kiddinn:
PTAL

timesketch/frontend/src/components/Sketch/TimelineListItem.vue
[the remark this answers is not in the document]
kiddinn: changed it a bit

berggren (approved):

PR #1389, opened 2020-10-02
Adding the ability to check your own permission to a sketch.

kiddinn:
Adding the ability to check your own ACL on a sketch (instead of relying on the `global` ACL and then attempting to do group expansions)


berggren (commented):
lgtm

PR #1396, opened 2020-10-06
Adding in a manual vega spec aggregator

kiddinn:
This PR adds in a manual aggregator that accepts Vega-lite specs to generate charts.

Can be called with:

```
chart = alt.Chart(data)......

...
parameters = {
    'data': chart.to_dict(),
    'title': 'some title',
    'supported_charts': 'foobar',
    'chart_title': 'my title'
}
agg = sketch.run_aggregator(
    aggregator_name='manual_vega',
    aggregator_parameters=parameters)
```

This would generate an aggregation object that can the viewed by using:
```
agg.chart
```

Or saved using:
```
agg.title = 'This little aggregation of mine'
agg.name =. 'This is the name of the aggregation'
agg.save()
```

And that way the manual altair chart can now be included in a story.

### DIFF
--- a/timesketch/lib/aggregators/__init__.py
+++ b/timesketch/lib/aggregators/__init__.py
@@ -17,3 +17,4 @@
 from timesketch.lib.aggregators import bucket
 from timesketch.lib.aggregators import feed
 from timesketch.lib.aggregators import term
+from timesketch.lib.aggregators import vega
--- a/timesketch/lib/aggregators/vega.py
+++ b/timesketch/lib/aggregators/vega.py
@@ -1,0 +1,172 @@
+# Copyright 2020 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Vega Spec aggregation.
+
+This file provides a manual vega spec aggregator. This is useful for
+analyzers that generate specific vega specs (or altair charts) that
+are not yet supported by other aggregators and/or charts. This can
+also be used by those manually generating charts in notebooks.
+
+This analyzer is not useful for UI and therefore is "hidden" from
+views.
+"""
+from __future__ import unicode_literals
+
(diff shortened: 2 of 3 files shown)


timesketch/lib/aggregators/vega.py
```diff
@@ -0,0 +1,172 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Vega Spec aggregation.
+
+This file provides a manual vega spec aggregator. This is useful for
+analyzers that generate specific vega specs (or altair charts) that
+are not yet supported by other aggregators and/or charts. This can
+also be used by those manually generating charts in notebooks.
+
+This analyzer is not useful for UI and therefore is "hidden" from
```
berggren, line 21:
```suggestion
This aggregator is not useful for UI and therefore is "hidden" from
```

timesketch/lib/aggregators/vega.py
```diff
@@ -0,0 +1,172 @@
...
+        Returns:
+            Dict with aggregation result.
+        """
+        datasets = self._spec.get('datasets', {})
+        values = []
+
+        for dataset in datasets.values():
+            values.extend(list(dataset))
+
+        return {
```
berggren, line 66: maybe use dict() here

timesketch/lib/aggregators/__init__.py
```diff
@@ -17,3 +17,4 @@
 from timesketch.lib.aggregators import bucket
 from timesketch.lib.aggregators import feed
 from timesketch.lib.aggregators import term
+from timesketch.lib.aggregators import vega
```
berggren, line 20: Is this an aggregation or chart? i.e. should it be in the charts/ dir instead?

berggren (changes requested):

timesketch/lib/aggregators/__init__.py
kiddinn: no, this is an aggregator, since you need to be able to add it into stories ,etc... and you want it to be displayed like other aggregators, if you would have this as a chart, then it would need to be added as part of another aggregator, and that would not work.

timesketch/lib/aggregators/vega.py
kiddinn: so you prefer this to be 

```
return dict(values=values)
```

rather than

```
return {
    'values': values,
}
```
???

kiddinn:
back to you @berggren

timesketch/lib/aggregators/vega.py
berggren: Kind of, but no strong opinion :) Feel free to keep.

timesketch/lib/aggregators/__init__.py
berggren: ack

PR #1398, opened 2020-10-07
Added support for optional keyword "rdomain" in SSH log line

itsmvd:
Included rdomain in the regex. Previously all log lines that had rdomain at the end of the line would be ignored.


itsmvd:
Will resubmit, needs some testing

PR #1403, opened 2020-10-09
Added support for optional keyword rdomain

itsmvd:
rdomain is sometimes logged in SSH logs. Previously any log line that contained rdomain would be skipped.


kiddinn (approved):
LGTM

PR #1404, opened 2020-10-09
Manually add/filter labels

berggren:
This PR adds the ability to manual add labels to events. Labels differs from tags in the way that they are scoped to a specific sketch. With this change it is now also possible to filter on labels.

Apart from this new feature there are also a few smaller bug fixes with the filter system.

Screenshot 2020-10-10 at 00 42 45

Screenshot 2020-10-10 at 00 43 10

Screenshot 2020-10-10 at 00 44 14


timesketch/api/v1/resources/sketch.py
```diff
@@ -341,7 +341,9 @@ def get(self, sketch_id):
                 x for x, y in analyzer_manager.AnalysisManager.get_analyzers()
             ],
             mappings=list(mappings),
-            stats=stats_per_index
+            stats=stats_per_index,
+            labels=self.datastore.get_sketch_labels(sketch.id, sketch_indices),
+            sketch_labels=[label.label for label in sketch.labels]
```
kiddinn, line 346: `sketch.labels` is just the labels themselves, we already have the `label_string` with the same information, is this needed?

timesketch/lib/datastores/elastic.py
```diff
@@ -415,6 +412,65 @@ def search_stream(self, sketch_id=None, query_string=None,
             for event in result['hits']['hits']:
                 yield event
 
+    # pylint: disable=line-too-long
```
kiddinn, line 415: why this pylint disable?

timesketch/lib/datastores/elastic.py
```diff
@@ -415,6 +412,65 @@ def search_stream(self, sketch_id=None, query_string=None,
...
+        Returns:
+            List with label names.
+        """
+        # This is a workaround to return all labels by setting the max buckets
+        # to something big. If a sketch has more than this amount of labels
+        # the list will be incomplete but it should be uncommon to have >10k
+        # labels in a sketch.
+        max_labels = 10000
+        label_aggregation = {
+            "aggs": {
```
kiddinn, line 432: can you use single quotes, so that we are consistent in the codebase?

kiddinn (commented):
few comments

timesketch/api/v1/resources/sketch.py
berggren: label_string() gives me a json encoded string of the list of labels. I opted to do this instead of deserialize. I'm jsonifying this so I need a dict.

berggren: I did however rename it to filter_labels to make it clearer what it is.

timesketch/lib/datastores/elastic.py
berggren: Some line overflow in the filter dict, but to remain readably and consistent I opted to ignore this lint error. I moved it closer to where it ignores it though.

timesketch/lib/datastores/elastic.py
berggren: Good catch, fixed.

kiddinn (changes requested):
can you fix the tests?

```
  File "/timesketch/timesketch/api/v1/resources/sketch.py", line 345, in get
    filter_labels=self.datastore.get_filter_labels(
AttributeError: 'MockDataStore' object has no attribute 'get_filter_labels'
```

kiddinn (approved):
LGTM

Issue #1423, opened 2020-10-21
Add API for managing labels on timelines


rgayon:
we have remove_sketch_label and add_sketch_label, would like to have remove_timeline_label & add_timeline_label

Issue #1422, opened 2020-10-21
Adding large sets of data via importer can cause multiple timelines to be created with same data


jaegeral:
Uploading large (several 100 k events) sets using the API importer can end in "multiple" timelines.

That created timelines will all have the same name and all the same data in it, so I expect it is not really duplicated in ES but only multiple entries in the TS specific daatabase.

Also if you click one of the Timelines to activate or filter out, all the others are activated / Deactivated.

````
[2020-10-13 13:13:20,033] timesketch.tasks/INFO Index timeline [all_packets.csv] to index [151af5f32cfb447285c4c447403b0c53] (source: jsonl) 
[2020-10-13 13:13:24,278] timesketch.tasks/INFO Index timeline [all_packets.csv] to index [151af5f32cfb447285c4c447403b0c53] (source: jsonl) 
[2020-10-13 13:13:25,892] timesketch.elasticsearch/WARNING Attempting to create an index that already exists (151af5f32cfb447285c4c447403b0c53 - True) 
[2020-10-13 13:13:32,891] celery.worker.strategy/INFO Received task: timesketch.lib.tasks.run_csv_jsonl[bd7d9535-4678-4ecb-9549-caa171e32309]
[2020-10-13 13:14:02,435] timesketch.tasks/INFO Index timeline [all_packets.csv] to index [151af5f32cfb447285c4c447403b0c53] (source: jsonl) 
[2020-10-13 13:14:05,871] timesketch.tasks/INFO Index timeline [all_packets.csv] to index [151af5f32cfb447285c4c447403b0c53]
````

So it seems the script correctly detects that the index already exists.

![gh_import_issue](https://user-images.githubusercontent.com/741037/96725801-954e2e00-13b1-11eb-89c0-7729e556353c.png)

Note: have not verified if it happens also with web uploader.

**Comment by kiddinn** [comment]
Looking at an example of this issue: 

One timeline:
```
Detailed information for network_pcap_with_scapy

Elasticsearch index: 3fec3837e59f42bfa9dae06bb3fdd6f1
Number of events: 86K (86000)
Size on disk: 361.72MB (361716008)
Original name: network_pcap_with_scapy
Added by: timavordur
Added: 2020-10-21 19:27
```

Another one:
```
Detailed information for network_pcap_with_scapy

Elasticsearch index: 3fec3837e59f42bfa9dae06bb3fdd6f1
Number of events: 86K (86000)
Size on disk: 361.72MB (361716008)
[...]

PR #1426, opened 2020-10-21
Fixes to the API and API client

kiddinn:
This PR includes few fixes to the API and API client:

+ Fix a bug in uploads where multiple timelines could be generated, #1422
+ Adding an API to add/remove labels from timelines, #1423
+ Making it a requirement to have write access to a sketch to be able to run an analyzer (since it modifies the data)

### DIFF
--- a/timesketch/api/v1/resources/timeline.py
+++ b/timesketch/api/v1/resources/timeline.py
@@ -242,12 +242,12 @@ class TimelineResource(resources.ResourceMixin, Resource):
                 abort(
                     HTTP_STATUS_CODE_BAD_REQUEST, (
                         'Label needs to be a JSON string that '
-                        'converts a list of strings [{}] {}'.format(type(labels), labels)))
+                        'converts to a list of strings.'))
             if not all([isinstance(x, str) for x in labels]):
                 abort(
                     HTTP_STATUS_CODE_BAD_REQUEST, (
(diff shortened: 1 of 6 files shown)


timesketch/api/v1/resources/timeline.py
```diff
@@ -198,10 +230,61 @@ def post(self, sketch_id, timeline_id):
...
+                        'Label needs to be a JSON string that '
+                        'converts to a list of strings.'))
+            if not all([isinstance(x, str) for x in labels]):
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST, (
+                        'Label needs to be a JSON string that '
+                        'converts to a list of strings (not all strings)'))
+
+            label_action = form.label_action.data
+            if label_action not in ('add', 'remove'):
```
berggren, line 253: Why don't we use proper HTTP verbs here? e.g.

http.delet()
http.post()

I know this is how we have designed the backend API, but I'm thinking if we would like to clean that up later. Not in this PR obviously :)

berggren (approved):
Just one question, but otherwise LGTM

timesketch/api/v1/resources/timeline.py
kiddinn: ack

PR #1431, opened 2020-10-27
Fix a bug in the API client

kiddinn:
Fixing a bug in the API Client.

If `verify=True` is set in the code (which is turned on by default) this is called:
```
session.verify = True
```

However that is not how the API works, SSL certificate verification is turned on by default, and according to the [tools documentation](https://requests.readthedocs.io/en/master/user/advanced/#ssl-cert-verification) it should be:

```
s.verify = '/path/to/certfile'
```

For a specific Certificate file verification. Since this is set to `True` in the code the following errors occurs when using `timesketch` as the authentication mechanism:

```
OSError: Could not find a suitable TLS CA certificate bundle, invalid path: True
```

Also changing how attributes are returned, to make it easier to search for them.

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -297,7 +297,9 @@ class TimesketchApi(object):
         if auth_mode == 'http-basic':
             session.auth = (username, password)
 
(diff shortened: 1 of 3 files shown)


berggren (approved):
LGTM

PR #1433, opened 2020-10-29
Stripping pagination from saved views.

kiddinn:
If you create a view, or a saved search from the UI while not viewing the first page of a result the pagination information is not removed.

This causes the views to fail in the API client (`from` is not known in the query_filter) as well as in stories.

This PR strips the pagination from the `query_filter` before a view is updated or created.

There is also a small fix for updating the index in the story.blocks for the API client.


berggren (approved):
LGTM

PR #1444, opened 2020-11-02
Single event notebook

jaegeral:


notebooks/add_a_single_event_to_a_sketch.ipynb
```diff
@@ -0,0 +1,187 @@
...
+  },
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "metadata": {
+        "id": "view-in-github",
+        "colab_type": "text"
+      },
+      "source": [
+        "<a href=\"https://colab.research.google.com/github/google/timesketch/blob/master/notebooks/Timesketch_add_a_single_event_to_a_sketch.ipynb\" target=\"_parent\"><img src=\"https://colab.research.google.com/assets/colab-badge.svg\" alt=\"Open In Colab\"/></a>"
```
kiddinn, line 23: this link is incorrect, since the notebook is not Timesketch_add_a_single_event_to_a_sketch.ipynb

so 
s/Timesketch_add_a_single_event_to_a_sketch.ipynb/add_a_single_event_to_a_sketch.ipynb

notebooks/add_a_single_event_to_a_sketch.ipynb
```diff
@@ -0,0 +1,187 @@
...
+      "cell_type": "code",
+      "metadata": {
+        "id": "ZoKUE_mbwOyy"
+      },
+      "source": [
+        "# @title Install dependencies\n",
+        "# @markdown Only execute if not already installed and running a cloud runtime\n",
+        "!pip install -q timesketch_api_client\n",
+        "!pip install -q vt-py\n",
+        "!pip install -q nest_asyncio"
```
kiddinn, line 46: why are these required? That is the vt-py and nest_asyncio???

also can you hide teh code in the forms?

notebooks/add_a_single_event_to_a_sketch.ipynb
```diff
@@ -0,0 +1,187 @@
...
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "_En91RhFQ5l8"
+      },
+      "source": [
+        "# @title Sketch Configuration\n",
+        "# @markdown Set the sketch ID for the rest of the notebook\n",
+        "\n",
+        "Sketch_id = 7193 #@param {type:\"integer\"}"
```
kiddinn, line 78: can yo make this default parameter 1, something that is more likely to exist?

notebooks/add_a_single_event_to_a_sketch.ipynb
```diff
@@ -0,0 +1,187 @@
...
+      "execution_count": null,
+      "outputs": []
+    },
+    {
+      "cell_type": "code",
+      "metadata": {
+        "id": "uO7KPHOrNh4_"
+      },
+      "source": [
+        "sketch = ts.get_sketch(Sketch_id)"
```
kiddinn, line 101: why not combine these two in a single cell, that is get_client and get_sketch

kiddinn: I'm also missing all the instructions in `get_client`, if people are running this for the first time, or use the cloud colab kernel they will not have a config already set, so they 'll need to enter some data to get information, probably better to include some instructions

kiddinn (changes requested):
few comments

notebooks/add_a_single_event_to_a_sketch.ipynb
jaegeral: done

notebooks/add_a_single_event_to_a_sketch.ipynb
jaegeral: combined it into one cell. 

For the "more description" I am not 100 % sure I would agree here, so if we think the data requested when prompted is not enough, I opt we provide more info when get prompted instead of adding more explanation in every call we make to that function.

What do you think?

kiddinn: well, pros and cons... ATM people are confused what to answer when prompted with `auth_mode`... 

I agree that there should be more explanations here, might need to add that, at least for these most common options.
but in this case here, since this is not using the demo server I'm fine with this as is for now...

kiddinn (approved):
LGTM

PR #1445, opened 2020-11-02
APIClient doc link to Colab notebooks

jaegeral:
There was a remaining TODO in the APIclient doc to link to a Notebook (that imho existed) so I fixed that, as well as link to the general notebook folder in the TImesketch repo.


kiddinn (approved):
LGTM

PR #1447, opened 2020-11-03
Few updates to the MUS2019 CTF Colab notebook

kiddinn:
The colab does not fully work, but at least I fixed few issues/bugs with it, that needed to be updated after new plaso versions and new TS API versions.


obsidianforensics (approved):

obsidianforensics (approved):
LGTM

PR #1452, opened 2020-11-04
Analyzer feature extract

lprat:
Hi,

I added option for improve extract.
Exemple, if you want extract multi ip from field (field contains 3, 4 IP address), you should keep all result in list format or text...

4 news options:
  - overwrite_store_as (boolean default: True)
    - Can be used if you don't want overwrite field if exist (set to False) 
  - overwrite_and_merge_store_as (boolean default: False)
    - Can be used if you want overwrite field and merge the existing value (set to True) 
  - store_type_list (boolean default: False)
    - Can be used if you want to store result in List type
  - keep_multimatch (boolean default: False)
    - Can be used if you want to keep multi result extracted from regex (ex: multi IP extracted)

I will create new PR for IP extract (in features.yaml).

Thank


kiddinn:
Hi, thank you for the PR. Could you start by fixingn the linter issues?

```
importer_client/python/timesketch_import_client timesketch
************* Module timesketch.lib.analyzers.feature_extraction
timesketch/lib/analyzers/feature_extraction.py:108:0: C0301: Line too long (97/80) (line-too-long)
timesketch/lib/analyzers/feature_extraction.py:124:0: C0301: Line too long (85/80) (line-too-long)
timesketch/lib/analyzers/feature_extraction.py:199:0: C0301: Line too long (88/80) (line-too-long)
timesketch/lib/analyzers/feature_extraction.py:242:34: C0326: Exactly one space required after comma
        return_fields = [attribute,store_as]
                                  ^ (bad-whitespace)
timesketch/lib/analyzers/feature_extraction.py:285:0: C0301: Line too long (95/80) (line-too-long)
timesketch/lib/analyzers/feature_extraction.py:292:34: C0326: Exactly one space required around assignment
                        resultList=[]
                                  ^ (bad-whitespace)
timesketch/lib/analyzers/feature_extraction.py:298:0: C0301: Line too long (93/80) (line-too-long)
timesketch/lib/analyzers/feature_extraction.py:302:0: C0301: Line too long (95/80) (line-too-long)
ERROR: InvocationError for command /timesketch/.tox/pylint/bin/pylint --rcfile=.pylintrc api_client/python/timesketch_api_client importer_client/python/timesketch_import_client timesketch (exited with code 16)
___________________________________ summary ____________________________________
```

data/features.yaml
```diff
@@ -14,6 +14,10 @@
 #       tags: []
 #       create_view: False
 #       aggregate: False
+#       overwrite_store_as: True
+#       overwrite_and_merge_store_as: False
+#       store_type_list: False
+#       keep_multimatch: False
```
kiddinn, line 20: can you explain a bit what these mean in the comment section (see other fields and their explanations below in the file)

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +265,43 @@ def extract_feature(self, name, config):
                 continue
 
             event_counter += 1
-            event.add_attributes({store_as: result[0]})
+            #verify if store_as exist
```
kiddinn, line 268: this comment serves no purpose, let's get rid of it

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +265,43 @@ def extract_feature(self, name, config):
...
+            if store_type and not overwrite_store_as:
+                continue
+            if isinstance(store_type, six.text_type):
+                store_type_list = False
+            elif isinstance(store_type, (list, tuple)):
+                store_type_list = True
+            if store_type_list:
+                if keep_multimatch:
+                    if store_type and overwrite_and_merge_store_as:
+                        #don't add duplicate value
```
kiddinn, line 279: same with this comment

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +265,43 @@ def extract_feature(self, name, config):
...
+                        resultList = list(set(store_type) | set(result))
```
kiddinn, line 280: we use snake case for variables, so this should be `result_list`

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +265,43 @@ def extract_feature(self, name, config):
...
+                        event.add_attributes({store_as: resultList})
+                    else:
+                        event.add_attributes({store_as: result})
+                else:
+                    if overwrite_and_merge_store_as and store_type and result[0] in store_type:
+                        continue
+                    event.add_attributes({store_as: [result[0]]})
+            else:
+                if keep_multimatch:
+                    #TODO: add option to choice join char
```
kiddinn, line 290: keep a space between # and the TODO

also can you further elaborate on what the todo is about?

kiddinn (changes requested):
brief overview, I'll do a deeper review once linter issues and these have been fixed

timesketch/lib/analyzers/feature_extraction.py
lprat: I removed, it was for change join char by option:
```
event.add_attributes({store_as: ','.join(result)})
```

timesketch/lib/analyzers/feature_extraction.py
lprat: ok

kiddinn:
(for future references, when you are ready for another round, please click the circle next to my name in the review pane, that will let me know that the PR is again ready for a review round)

Screen Shot 2020-11-05 at 15 13 04

but I'll go ahead and review the change now

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -161,6 +193,11 @@ def extract_feature(self, name, config):
         query = config.get('query_string')
         query_dsl = config.get('query_dsl')
         attribute = config.get('attribute')
+        store_type_list = config.get('store_type_list', False)
+        keep_multimatch = config.get('keep_multimatch', False)
+        overwrite_store_as = config.get('overwrite_store_as', True)
+        overwrite_and_merge_store_as = \
```
kiddinn, line 199: don't use \

rather do

```
overwrite_and_merge_store_as = config.get(
    'overwrite_and_merge_store_as', False)
```

This is more consistent with the rest of the codebase

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +266,46 @@ def extract_feature(self, name, config):
                 continue
 
             event_counter += 1
-            event.add_attributes({store_as: result[0]})
+            store_type = event.source.get(store_as)
```
kiddinn, line 269: I would say that this is not a `store_type` but rather the original value... 

so rather say

`old_value = eve....store_as)` or `current_value = ....` or something more descriptive here.

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +266,46 @@ def extract_feature(self, name, config):
...
+            if store_type and not overwrite_store_as:
+                continue
+            if isinstance(store_type, six.text_type):
+                store_type_list = False
+            elif isinstance(store_type, (list, tuple)):
+                store_type_list = True
+            if store_type_list:
```
kiddinn, line 276: here you are starting to wade into the land of too much indentation and I would opt for you to create a new function

```
def _get_attribute_value(current_value, new_value, .....): 
....
```
(better to have this function at the module level, not at the class level, or have it at the class level and decorate it as a static method, since you don't need `self` and this is going to be potentially called quite a few times, so better to attempt to speed things up at the same time as you are making it more readable)

and then call it like this:

```
if store_type_list:
  attribute_value = self._get_attribute_value(current_value, result, .....
else:
  attribute_value = result[0]

event.add_attributes({store_as: attribute_value})
```

That way you are not mixing together the changes to the event and figuring out what the attribute it (separation of concerns) and you are not using up too much indentation that makes the code hard to read.

kiddinn: you can also make it much easier in many ways by returning early if certain conditions are met.

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -229,7 +266,46 @@ def extract_feature(self, name, config):
...
+                if keep_multimatch:
+                    if store_type and overwrite_and_merge_store_as:
+                        result_list = list(set(store_type) | set(result))
+                        event.add_attributes({store_as: result_list})
+                    else:
+                        event.add_attributes({store_as: result})
+                else:
+                    if overwrite_and_merge_store_as and \
```
kiddinn, line 284: again, don't use \ to divide between lines

here if you use the function you can a) save a lot of indentation and b) return early to avoid all this

kiddinn (changes requested):
few comments here, one fairly large that will make you change some of structure of teh change here. but I think it will make the code easier to read.

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
         return ', '.join(return_strings)
 
+    @staticmethod
+    def _get_attribute_value(
+            current_val,
+            extracted_value,
+            keep_multi,
+            merge_values,
+            type_list):
+        """Return value to store.
```
kiddinn, line 188:
```
Returns the attribute value as it should be stored.
```

and then a <nit>one empty line after the doc string before Args section</nit>

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+        Args:
+            current_val: current value of store_as.
+            extracted_value: values matched from regexp (type list).
+            keep_multi: choise if you keep all match from regex (type boolean).
+            merge_values: choise if you merge value from extracted
+                 and current (type boolean).
+            type_list: choise if you store values in list type(type boolean).
+        Returns:
```
kiddinn, line 196: <nit>
one empty line before Returns section
</nit>

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+                if keep_multi:
+                    new_value = list(set(current_val) | set(extracted_value))
+                elif extracted_value[0] not in current_val:
+                    new_value = current_val + [extracted_value[0]]
+            else:
+                if keep_multi:
+                    new_value = extracted_value
+                else:
+                    new_value = [extracted_value[0]]
+        else:
```
kiddinn, line 215: this else is not needed... you can do
```
if type_list:
  # do things here
  return new_value

# then continue here and thus skip the else statement
```

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+        Returns:
+            Value to store
+        """
+        new_value = None
+        if not current_val:
+            merge_values = False
+        if len(extracted_value) == 1:
+            keep_multi = False
+        if type_list:
+            if merge_values:
```
kiddinn, line 205: I would opt here

```
if merge_values and keep_multi:
    return list(set(current_val) | ....

if merge_values and extracted_value[0] not in current_val:
  current_val.append(extracted_value[0])
  return current_val

```

kiddinn: that is, this code has a lot of if ... if else... if ... if elif ... and deep in.. .where you can save a lot of that by returning early and then skipping the else


also can you add unit tests for this function? Test it with few scenarios to make sure it is working as expected.

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+                    new_value = extracted_value
+                else:
+                    new_value = [extracted_value[0]]
+        else:
+            if merge_values:
+                if keep_multi:
+                    add_list = []
+                    for elem_match in extracted_value:
+                        if elem_match not in current_val:
+                            add_list.append(elem_match)
```
kiddinn, line 221: ok, so two things... if current_val is a comma separated string you can do:

```
current_list = [x.strip() fox x in current_val.split(',')]
new_set = set(current_list.extend(extracted_value))
return ','.join(list(new_set))
```

but if you just want to add this after the fact like you are doing here you can also do

```
extended_list = [x for x in extracted_value if x not in current_val]
extended_string = ','.join(extended_list)
return f'{current_val},{extended_string}'
```

since you are just extending teh list with the values that are not there

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+                    new_value = current_val + ',' + ','.join(add_list)
+                elif extracted_value[0] not in current_val:
+                    new_value = current_val + ',' + extracted_value[0]
```
kiddinn, line 224: agin here you can do

```
if extracted_value[0] in current_val:
  return current_val

return f'{current_val},{extracted_value[0]}'
```

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,57 @@ def run(self):
...
+            else:
+                if keep_multi:
+                    new_value = ','.join(extracted_value)
+                else:
+                    new_value = extracted_value[0]
```
kiddinn, line 229: again here...

```
if keep_multi:
  return ','.join(extracted_value)

return extracted_value[0]
```

etc,

kiddinn (changes requested):
ok, much better, now just to fix the new function and we should be good to go, and add some unit test to it, to make sure it is working as intended.

timesketch/lib/analyzers/feature_extraction.py
```diff
@@ -146,6 +178,54 @@ def run(self):
...
+            if merge_values:
+                if extracted_value[0] not in current_val:
+                    current_val.append(extracted_value[0])
+                return sorted(current_val)
+            if keep_multi:
+                return sorted(extracted_value)
+            return [extracted_value[0]]
+        if merge_values and keep_multi:
+            extended_list = [x for x in extracted_value if not re.match(
+                f"(^|[^0-9a-zA-Z]){x}([^0-9a-zA-Z]|$)", current_val)
```
kiddinn, line 217: I would opt to compile this RE

```
self._re_something_appropriate = re.compile(r'(......
```
and then here iuse

```
.... value if not self._re_whatever_name.match(current_val)
```

timesketch/lib/analyzers/feature_extraction_test.py
```diff
@@ -65,3 +69,125 @@ def test_config(self):
...
+
+    # Mock the Elasticsearch datastore.
+    @mock.patch(
+        'timesketch.lib.analyzers.interface.ElasticsearchDataStore',
+        MockDataStore)
+    def test_get_attribute_value(self):
+        """Test function _get_attribute_value()."""
+        analyzer = feature_extraction.FeatureExtractionSketchPlugin(
+            'test_index', 1)
+        #list, merge, multi
```
kiddinn, line 81: comment adds no value

timesketch/lib/analyzers/feature_extraction_test.py
```diff
@@ -65,3 +69,125 @@ def test_config(self):
...
+        current_val = ['hello']
+        extracted_value = ['hello']
+        # pylint: disable=protected-access
+        new_val = analyzer._get_attribute_value(
+            current_val,
+            extracted_value,
+            True,
```
kiddinn, line 88: to be more explicit and make it easier to read teh test can you rather do

```
new_val = analyzer._get_....(
    current_val=current_val,
    extracted_value=extracted_value,
    keep_multi=True,
    merge_values=True,
    type_list=True)

```

that way it will be easier to read the test

kiddinn (commented):
only minor comments, otherwise looks good.

timesketch/lib/analyzers/feature_extraction.py
lprat: Ok use split ',' instead of regexp

kiddinn (approved):
LGTM

PR #1453, opened 2020-11-05
Anaylzer IOC extract

lprat:
Hi,

It's analyzer for IOC check in events.
First, you create ioc DB by type: ipv4, hash, domain, hostname, filename, ... from MISP or/and your favorite Threat Intel source.
Secondly, check IOC on specified field name in all events.
If found ioc, push value in store_as and tags event.

Lionel


data/ioc/domain.json
```diff
@@ -0,0 +1 @@
+["evil.com"]
```
kiddinn, line 1: ok, so the definition is a yaml file, but now we are using a json for this part, is that confusing? or should this be:

```
- evil.com
- foo.com
- bar.com
...
```

to create a list?

data/ioc_extract.yaml
```diff
@@ -0,0 +1,66 @@
...
+# IOC TYPE:
+domain:
+        path_file_ioc: '/etc/timesketch/ioc/domain.json'
+        attributes: ['domain', 'tld']
+        attributes_contains: ['domain']
+        match_re: '([^a-zA-Z0-9]|^)$value$([^a-zA-Z0-9]|$)'
+        store_as: 'ioc'
+        tags: ['ioc_domain']
+        
+# hostname:
```
kiddinn, line 60: is this a TODO? maybe mark it as such then

data/ioc_extract.yaml
```diff
@@ -0,0 +1,66 @@
...
+#       attributes: []
+#       store_as:
+#       # Optional fields.
+#       attributes_contains: []
+#       match_re:
+#       re_flags: []
+#       tags: []
+#
+# Each definition needs to give IOC file (json format). Json sample:
+# ['ioc1', 'ioc2', 'iocx']
```
kiddinn, line 16: for json you need to use the double quote

data/ioc_extract.yaml
```diff
@@ -0,0 +1,66 @@
...
+# IOC DB can be make with script to collect IOC from MISP 
+# or other Threat Intel
+#
+# re_match is a regex to avoid false positive. By default is:
+# '([^a-zA-Z0-9]|^)$value$([^a-zA-Z0-9]|$)'
+# ex without re_match: '127.0.0.1' match '127.0.0.12'
+# ex with re_match by default: '127.0.0.1' not match '127.0.0.12'
```
kiddinn, line 23: ok, so the re_match is a check to eliminate?

Just from reading the config file I'm not quite sure of how it works, so you provide a JSON file with exact matches, but then you've got this regular expression match, that does what exactly? For the IOC where I've got al ist of 5k domains, how does the RE affect that?

also you can just define the elastic search to be more precise here.

data/ioc_extract.yaml
```diff
@@ -0,0 +1,66 @@
...
+# If there are tags defined they are also applied to that event.
+# ------------------------------------------------------------------------
+
+# IOC TYPE:
+domain:
+        path_file_ioc: '/etc/timesketch/ioc/domain.json'
+        attributes: ['domain', 'tld']
+        attributes_contains: ['domain']
+        match_re: '([^a-zA-Z0-9]|^)$value$([^a-zA-Z0-9]|$)'
+        store_as: 'ioc'
```
kiddinn, line 57: so here you can define a `store_as` where you pick the attribute to store to... so what's the difference between this and a feature extraction???

why not just have the ioc attribute as a fixed attribute, this is an IOC extractor, so I would expect IOC to be a data structure that is added to the attribute... (dict maybe?)

docs/Users-Guide.md
```diff
@@ -202,6 +203,12 @@ It will also add information about:
 
 - Known CDN providers (based on Timesketch config)
 
+#### IOC Extraction Analyzer
+
+The IOC extraction Analyzer check IOC from database (json) in all fieldnames content specified. Config file is "ioc_extract.yaml".
+
+You can use before "Feature Extraction Analyzer" to extract pattern (ip, url, hash, ...) in specified fields.
```
kiddinn, line 210: what does `you can use before "Feat..." to extract pattern?` mean? I'm not quite sure, can you elaborate?

Also, as this is written it is only domain names ATM.

data/ioc_extract.yaml
```diff
@@ -0,0 +1,66 @@
...
+        tags: ['ioc_domain']
+        
+# hostname:
+# ipv4:
+# email_address:
+# user_agent:
+# url:
+# hash_md5:
+# filename:
```
kiddinn, line 66: also, is it really necessary to have this YAML file? Might it not just be enough to have an analyzer that reads in the IOCs from the JSON files?

you create a JSON (or jsut a text file) file for each IOC, "domain.json", "ip.json"

data/ioc/domain.json
kiddinn: or in this case, since we are not really introducing any code or any complexities, isn't it enough to have this as a text file??? 

domain.txt with one line per domain

```
evil.com
evilfoo.com
evilbar.com
...
```

instead of building a json list, which will be long BTW... so a 5k file would be 
```
["evil.com", "evil1.com", "evil2.com", .... (5k entries)]
```

that is a single line with potentially 5k or even more domains.... depending on the amount of IOCs you are comparing against.

timesketch/lib/analyzers/interface.py
```diff
@@ -750,6 +750,28 @@ def __init__(self, index_name):
         if not hasattr(self, 'sketch'):
             self.sketch = None
 
+    def get_fields_list(
+            self, indices=None):
```
kiddinn, line 754: why is this on two lines?

```
def get_fields_list(self, indices=None)
```

timesketch/lib/analyzers/interface.py
```diff
@@ -750,6 +750,28 @@ def __init__(self, index_name):
...
+        """Indices get_mapping ElasticSearch.
```
kiddinn, line 755: this docstring is just confusing.. .what does `indices get_mapping elastic search` mean?

```
Returns a list of field names in an index.
```

Seems like a more accurate description.

timesketch/lib/analyzers/interface.py
```diff
@@ -750,6 +750,28 @@ def __init__(self, index_name):
...
+            List of fields name in index.
+        """
+        if not indices:
+            indices = [self.index_name]
+
+        # Refresh the index to make sure it is searchable.
+        for index in indices:
+            self.datastore.client.indices.refresh(index=index)
+
+        list_fieldnames = self.datastore.fields_list(
```
kiddinn, line 770: no need to store this as an attribute... just

```
return self.datastore.....
```

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            String with summary of the analyzer result.
+        """
+        config = self._config or interface.get_yaml_config(self.CONFIG_FILE)
+        if not config:
+            return 'Unable to parse the config file.'
+        query, ret_fields, ioc_db = self.parse_config(config)
+        if not query or not ret_fields:
+            return 'Error to make query'
+        if not ioc_db:
+            return 'Error to parse config'
```
kiddinn, line 120: I would opt for these error messages to be better explained...

```
return 'Error: unable to locate the IOC database'
```

Or something like that.

Also you want to add `logging.error('....` here to make sure that logs are also appearing in the backend.

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+        logger.info((
+            'Return field '
+            'error: {0!s}').format(ret_fields))
```
kiddinn, line 123: why are you doing this? this is done for every run, this does sound like a debugging session...

also it does not sound like an error and is logged as an `info`, not `error` nor `debug`.

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+        return self.extract_ioc(query, ret_fields, ioc_db)
+
+    def parse_config(self, ioc_config):
+        """Create IOC DB and make request
```
kiddinn, line 127: this does not match with the function name.. the name implies that you're just parsing a configuration... the doc string implies you are creating a database and making some sort of requests.... which requests, I'm, not sure, but the docstring seems to suggest that.

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+
+        Args:
+            config: A dict that contains the configuration for the ioc
+                extraction. See data/ioc.yaml for fields and further
+                documentation of what needs to be defined.
+
+        Returns:
```
kiddinn, line 134: so you can be more clear here... it returns a tuple with three entries, and then you go ahead and contine with the description you have here

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            query: A string that contains elastic query to return
+                events check ioc against.
+            return_fields: A list that contains names of field to return
+                from elastic request to check ioc against.
+            ioc_db: A dict contains extracted config and IOC files
```
kiddinn, line 139: a Dict `that` contains extrac....

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+        """
+        ioc_db = []
+        query = ""
```
kiddinn, line 142: to be consistent with the rest of the codebase, use single quotes

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+        return_fields = []
+        list_fieldnames = self.get_fields_list()
+        for name, config in iter(ioc_config.items()):
```
kiddinn, line 145: no need to use `iter`, we aer no longer providing py2 compatibility, this is no longer needed (will be removed from the codebase shortly, but should not be added with new code)

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            elm = {'name': name, 'found': 0}
```
kiddinn, line 146: what is `elm`?

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            path_file_ioc = config.get('path_file_ioc')
+            if not path_file_ioc:
+                logger.warning('No path for file IOC defined.')
```
kiddinn, line 149: what IOC file? you should include the name here...

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                continue
+            if not os.path.isfile(path_file_ioc):
+                logger.warning('file IOC defined not exists.')
```
kiddinn, line 152: Start with a capital letter (File...) and also include the name here (and for good measure also include teh file name to indicate what file specifically does not exist, will help sysadmins to know which file they forgot to include)

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                return ''
+            elm['ioc'] = None
+            with open(path_file_ioc) as json_file:
+                try:
+                    elm['ioc'] = json.load(json_file)
```
kiddinn, line 157: again I think a JSON is too heavy for this and not really needed... not if the intent is just to have a simple list here.

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                    if not isinstance(elm['ioc'], (list, tuple)):
+                        continue
+                    if not elm['ioc']:
+                        continue
+                except json.JSONDecodeError as exception:
+                    logger.warning((
+                        'File IOC error to parse: '
```
kiddinn, line 164: here you can first of all include what IOC the error applies to (helps with debugging) also you can include `exc_info=True` so that the stacktrace is included

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                        '{0!s}').format(exception))
+                    continue
+            elm['store_as'] = config.get('store_as')
+            if not elm['store_as']:
+                logger.warning('No attribute defined to store results in.')
```
kiddinn, line 169: again, you may want to provide additional information to maek this easier to debug

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                continue
+            elm['attributes'] = config.get('attributes')
+            attri_contains = config.get('attributes_contains')
```
kiddinn, line 172: attri_contains? can you make this variable name more clear

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            if attri_contains:
+                for fn in list_fieldnames:
```
kiddinn, line 174: ok, here you can just use a list comprehension...

this looks like to me that you have a list of field names... and you say

for each of the fields in a list of all the field names, go and check each of the attribute in the attribute container and check if it exists in the field... so if you've got:

```
list_fieldnames = ['apple', 'banana', 'grape']
attribute_container = ['applepie', 'ap', 'grape']
```

Then you do

```
fn = 'apple'
element['attributes'] = [
  'applepie' in 'apple' -> no
 'ap' in 'apple' -> YES
 'grape' in 'apple' -> No]

after first round
element['attributes'] = ['ap']

fn = 'banana'
element['attributes'] = [
  'applepie' in 'banana' -> no
 'ap' in 'banana' -> no
 'grape' in 'banana' -> No]

after second round
element['attributes'] = ['ap']

fn = 'grape'
element['attributes'] = [
  'applepie' in 'grape' -> no
 'ap' in 'grape' -> no
 'grape' in 'grape' -> yes]

final results

elment['attributes'] = ['ap', 'grape']
```

Is this the intention? Also seems very slow to go through these iterations, what about if list field names returns 60 fields (could very easily do so)??

or is the intention just to see if there is an exact match?

```
element['attributes'] = [x for x in attribute_container if x in list_fieldnames]
```

and then you don't have to have the `+=` which BTW is a list concatenation, which means for each time you call that loop you create a new list, leaving the old list behind in memory... so you 

list_a = []
list_a_ = []
create a new list with ['a']
overwrite list_a with the new list

...

instead of using 'append' or 'extend' on the list.

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                logger.warning('No attributes defined.')
+                continue
+            elm['tags'] = config.get('tags', [])
+            elm['regexp'] = config.get(
+                'match_re', '([^a-zA-Z0-9]|^)$value$([^a-zA-Z0-9]|$)')
+            regexp_flags = config.get('re_flags')
+            elm['re_flag'] = 0
+            if regexp_flags:
+                flags = set()
+                for flag in regexp_flags:
```
kiddinn, line 187: why not just do 

```
flags = [RE_FLAGS.get(x) for x in regexp_flags if x in RE_FLAGS]
```

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                    try:
+                        flags.add(getattr(re, flag))
+                    except AttributeError:
+                        logger.warning(
+                            'Unknown regular expression flag defined.')
+                        continue
+                elm['re_flag'] = sum(flags)
+            return_fields += elm['attributes'] + [elm['store_as']]
```
kiddinn, line 195: again, please don't use concatentation here... if thi sis a string use

return_fields = f'{return_fields}{elme....

if this a list, please use

```
return_fields.extend(element['attributes'])
return_fields.extend...
```

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+            ioc_db.append(elm)
+            for attribute in elm['attributes']:
+                if query:
+                    query += ' AND '
+                query += '_exists_:'+attribute
```
kiddinn, line 200: same here with queries

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+        return query, return_fields, ioc_db
+
+    def extract_ioc(self, query, ret_fields, ioc_db):
+        """Extract IOC from events.
```
kiddinn, line 204: this seems to be based on the "Returns string" the actual analyzer, or the run analyzer function... 

and shouldn't this be at least iocs instead of a single ioc?

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                elif isinstance(attribute_field, (list, tuple)):
+                    attribute_value = ','.join(attribute_field)
+                elif isinstance(attribute_field, (int, float)):
+                    attribute_value = attribute_field
+                else:
+                    attribute_value = None
+                if not attribute_value:
+                    continue
+                attributes_fmt[attribute] = attribute_value
+            for elm in ioc_db:
```
kiddinn, line 237: for element?

kiddinn: or what is `elm`?

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                ioc_field = event.source.get(elm['store_as'])
+                if ioc_field and not isinstance(ioc_field, (list, tuple)):
+                    logger.warning('IOC field exist but not list type')
```
kiddinn, line 240: might want to be more verbose, what IOC are we talking about?

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                    continue
+                if ioc_field:
+                    if elm['store_as'] in value_store:
+                        value_store[elm['store_as']] += ioc_field
```
kiddinn, line 244: why are we `+=` the store as field here?

timesketch/lib/analyzers/ioc_extraction.py
```diff
@@ -0,0 +1,280 @@
...
+                elif not elm['store_as'] in value_store:
+                    value_store[elm['store_as']] = []
+                for ioc in elm['ioc']:
+                    try:
+                        regexp_tmp = elm['regexp'].replace(
+                            '$value$', ioc)
+                        regexp = re.compile(regexp_tmp, flags=elm['re_flag'])
+                    except re.error as exception:
+                        # pylint: disable=logging-format-interpolation
+                        logger.warning((
```
kiddinn, line 256: that sounds more like an error instead of warning?

kiddinn (changes requested):
Starting you off with few comments, didn't get all the way to the end, but I think you've got enough to work with right now.

Also not quite sure this is the right approach, that is to define the YAML file, where maybe the other IOC config files could suffice, not sure what the YAML file is really providing with in this case.

lprat:
Replaced by https://github.com/google/timesketch/pull/1486

PR #1454, opened 2020-11-06
Improve the UX of Time Filters

binglot:
# The new feature:

Screenshot from 2020-11-06 08-12-25

---

# All the changes:
- Create a radio with 2 ways of adding time filters: 1) the date range, 2) a date and +/- min/hour/day
- Make the "edit" button look more similar to the "delete" button
- Fix the bug: the End Date input sometimes goes out of the "Add time range" box
- Fix the bug: if the first date contains the 'T' (e.g. "2020-01-02T00:00:00") and the second is before and doesn't have the 'T' (e.g. "2020-01-01 00:00:00") then the filter won't work.


stuhli:
Hi!
I was about to create an issue when I saw this PR. So maybe my issue fits better in here as obviously you are working on this at the moment?

Using the timefilter / time range in the newest Timesketch release (20201015) I observed the following behavior: when setting an invalid date the UI tells me "Invalid date" and discards all my input so I have to retype everything. In previous versions the UI simple told me that the date was invalid but let me fix my errors, which was much faster and more convenient. Would it be very cumbersome to restore the old behavior?

Also I miss the sometimes very useful feature to click on the datetime field of an timeline entry which pre-filled the start date of the time range filter? Is this intended or an error?

BTW: I love the new TS!!! I'll get in touch if I find more things.
Thanks for the effort!

binglot:
Hey @stuhli , thanks for your comments. I'll look into it later today if time allows and will try to address your comments.

@berggren - I just noticed that while doing a git merge I missed some of the code (for what the 'edit' buttons look like). I'll update the PR and let you know when it's done.

berggren:
@binglot : Got it. I'm doing the first round of review now. Let me know when your changes are in.

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -16,113 +16,219 @@ limitations under the License.
...
-  props: ['start', 'end', 'selectedChip'],
+  props: ['selectedChip'],
   data () {
     return {
-      startDateTime: this.start,
-      endDateTime: this.end,
-      chip: this.selectedChip
+      startDateTime: '',
+      endDateTime: '',
+      chip: this.selectedChip,
```
berggren, line 99: Seeing the passed chip (selectedChip) is an object this can lead to some unpredictable behavious because any update to the object in this component will propegate up to the parent component. I noticed that you do an explicit $emit later on in the code, so I assume that you want a local copy of the prop here instead?

I suggest you use spread to copy the prop into a local one:
```
chip: {...this.selectedChip}
```

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -16,113 +16,219 @@ limitations under the License.
...
+        {'text': 'Second', 'value' : 's'},
+        {'text': 'Minute', 'value' : 'm'},
+        {'text': 'Hour', 'value': 'h'},
+        {'text': 'Day', 'value': 'd'},
+      ],
+      selectedInterval: 'm',
+    }
+  },
+  created: function () {
+    if (!this.chip) {
```
berggren, line 114: You can also mark the prop (selectedChip) as required.

Something like:
```
 props: {
   selectedChip: {
     type: Object,
     required: true
   }
```

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -16,113 +16,219 @@ limitations under the License.
...
+      }
+      else if (this.isSelected('range')) {
+        if (this.startDateTime && this.endDateTime){
+          return true
+        }
+      }
+      return false
+    },
+    getPlaceholder: function (radioName) {
+      return (this.radio == radioName) ? '2019-07-07T10:00:01' : ''
```
berggren, line 150: Update the suggested placeholder to todays date instead? :)

timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -131,35 +237,51 @@ export default {
...
-      }
-
-      this.chip['value'] = this.startDateTime + ',' + this.endDateTime;
-      this.$emit('updateChip', this.chip)
-
       // Close the menu
       this.$emit('hideDropdown')
+    },
+    resetInterface: function() {
+      Object.assign(this.$data, this.$options.data()) // Credits to https://stackoverflow.com/a/38174780
```
berggren, line 277: Do you need the context here as well, seeing you are using `this`?
```
Object.assign(this.$data, this.$options.data.apply(this))
```

berggren (changes requested):
First round of comments

berggren:
@binglot First round done - PTAL

(note: do a git pull as I have merged updates from the main branch)

binglot:
I've updated the code for what the Edit button looks like.

image

timesketch/frontend/src/components/Sketch/TimeFilter.vue
binglot: What a great suggestion, thanks! There was indeed some code that was modifying the object that was passed in via the prop.

timesketch/frontend/src/components/Sketch/TimeFilter.vue
binglot: I don't think that's necessary because we treat it as optional and only use "selectedChip" to be able to update existing time filters. That's our way to distinguish if we're dealing with a new chip or if we're updating the existing one.

timesketch/frontend/src/components/Sketch/TimeFilter.vue
binglot: Why not! 👍

timesketch/frontend/src/components/Sketch/TimeFilter.vue
binglot: I'm not sure how `this.$options.data.apply(this)` is different to `this.$options.data()`? The idea is only to reset the object's state to that when it was first initiated. Since the way you suggested works to reset the component, I've updated the code to what you suggested.

binglot (commented):
Ready for the next round of review :)

binglot:
@stuhli - I've addressed your comments about the "Invalid date" behaviour. The code has a bunch of validation checks to prevent the timestamps from getting overwritten if the entered value is incorrect.

As for your 2nd point: "click on the datetime field of an timeline entry which pre-filled the start date of the time range filter", I'll check with @berggren to better understand what was the old behaviour and if we can restore it :)

Thanks again for your feedback and please keep it coming!

stuhli:
Thank you very much for your effort! Yeah, I think we'll stay in touch as we use TS very intensely. 😉

berggren (approved):
LGTM

PR #1455, opened 2020-11-06
Updating the notebook for Szechuan sauce to use picatrix magics (and a minor bug fix in the API client)

kiddinn:


berggren (approved):
LGTM

PR #1457, opened 2020-11-07
Minor updates to the szechuan sauce analysis notebook

kiddinn:
After some changes in the picatrix library, a bit of updates to the notebook to reflect that and to simplify.


berggren (approved):
LGTM

PR #1460, opened 2020-11-09
Update deploy_timesketch.sh

lxndrblz:
Updates URL to master branch.

I previously had errors generating the `config.env` using `deploy_timesketch.sh` script. It would just write a `404: Not Found` in the config file. Once I've updated the URL I could use the script for deploying timesketch using Docker again.


lxndrblz:
@googlebot I signed it!

berggren (approved):
LGTM, thanks for fixing!

PR #1462, opened 2020-11-10
Graph fixes

berggren:
Some fixes:
* Set indices to sketch active timelines by default
* Better UI with loading indicator when generating graph
* Indication when no graph could be generated.


kiddinn (approved):
Lgtm

PR #1463, opened 2020-11-12
API client changes: click removed from cli_input and adding aggregation into stories fixed.

kiddinn:
This PR makes few changes to the API client.

1. `click` is no longer used in the cli_input. This caused some issues when running on docker in jupyter notebooks.
2. Added typing into cli_input, config and story.
3. The `get_client` option in config now supports changing already existing config options.
4. Added hints for configuration parameters.
5. Fixed a bug when adding aggregations to stories.

### DIFF
--- a/api_client/python/timesketch_api_client/aggregation.py
+++ b/api_client/python/timesketch_api_client/aggregation.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Timesketch API client library."""
-from __future__ import unicode_literals
-
+import getpass
 import json
 import logging
 
@@ -44,6 +43,7 @@ class Aggregation(resource.BaseResource):
         self._sketch = sketch
         self._aggregator_data = {}
         self._labels = []
+        self._username = ''
         self._parameters = {}
         self.aggregator_name = ''
         self.chart_color = ''
@@ -52,7 +52,7 @@ class Aggregation(resource.BaseResource):
         self.view = None
         self.type = None
         resource_uri = 'sketches/{0:d}/aggregation/explore/'.format(sketch.id)
-        super(Aggregation, self).__init__(api, resource_uri)
+        super().__init__(api, resource_uri)
 
     def _get_aggregation_buckets(self, entry, name=''):
         """Yields all buckets from an aggregation result object.
@@ -107,6 +107,7 @@ class Aggregation(resource.BaseResource):
 
         self.aggregator_name = aggregator_name
         self.chart_color = parameters.get('chart_color', '')
+        self._parameters = parameters
 
         form_data = {
             'aggregator_name': aggregator_name,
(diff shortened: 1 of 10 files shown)


api_client/python/timesketch_api_client/aggregation.py
```diff
@@ -214,6 +219,18 @@ def lazyload_data(self, refresh_cache=False):
         # TODO: Implement a method to refresh cache.
         return self.resource_data
 
+    @property
+    def parameters(self):
```
berggren, line 223: Why not just make the attribute public? You are just returning it here, so no need for the property

api_client/python/timesketch_api_client/config.py
```diff
@@ -152,21 +57,41 @@ class ConfigAssistant:
...
+        'client_secret': 5,
+        'password': 6,
+    }
+
+    CONFIG_HINTS = {
+        'host_uri': 'URL of the Timesketch server',
+        'username': 'The username of the Timesketch user',
+        'password': 'Password of the chosen user.',
+        'auth_mode': (
+            'Authentication mode, valid choices are: "timesketch" '
```
berggren, line 74: I would opt to call it 'userpass' instead of 'timesketch' as it is a bit more clear on what it is.

api_client/python/timesketch_api_client/config.py
```diff
@@ -362,11 +290,164 @@ def save_config(self, file_path=''):
...
+
+    config_assistant.save_config()
+    credential_storage = crypto.CredentialStorage()
+    credentials = credential_storage.load_credentials(
+        config_assistant=config_assistant, password=token_password)
+
+    # Check if we are using username/password and we don't have credentials
+    # saved.
+    auth_mode = config_assistant.get_config('auth_mode')
+    if auth_mode != 'timesketch':
```
berggren, line 419:
```suggestion
    if auth_mode != 'userpass':
```

berggren (changes requested):

api_client/python/timesketch_api_client/aggregation.py
kiddinn: The main reason is that if it's public you can change it easily... but the change won't be reflected in the saved object.... by having it as a property it is now a read-only attribute, by then implementing a setter later I can then make sure that if you modify the parameters they will be saved in the actual aggregation object, that is the object will be updated on the server side.

api_client/python/timesketch_api_client/config.py
kiddinn: ok, agree.. changed it everywhere, since it needs to be changed in several places

api_client/python/timesketch_api_client/config.py
kiddinn: done

kiddinn (commented):
ok, done with round

api_client/python/timesketch_api_client/aggregation.py
berggren: ack, SG

berggren (approved):
LGTM

Issue #1465, opened 2020-11-12
.replace(".keyword:", ":") not working as expected


jaegeral:
Hi,

I tried to change uuid, and same result:
```
-------------------lnx3_susp_zenmap.yml-----------------
title: Suspicious Installation of Zenmap
id: 5266a592-b793-11ea-b3de-0242ac130005
description: Detects suspicious installation of Zenmap
references:
    - https://rmusser.net/docs/ATT&CK-Stuff/ATT&CK/Discovery.html
author: Alexander Jaeger
date: 2020/06/26
modified: 2020/06/26
logsource:
    product: linux
    service: shell
detection:
    keywords:
        # Generic suspicious commands
        - message: "apt-get install zmap"
    condition: keywords
falsepositives:
    - Unknown
level: high
-------------------------------------------------------------------
-------------------lnx_susp_zenmap.yml-----------------
title: Suspicious Installation of Zenmap
id: 5266a592-b793-11ea-b3de-0242ac130004
description: Detects suspicious installation of Zenmap
references:
    - https://rmusser.net/docs/ATT&CK-Stuff/ATT&CK/Discovery.html
author: Alexander Jaeger
date: 2020/06/26
modified: 2020/06/26
logsource:
    service: shell
detection:
    keywords:
        # Generic suspicious commands
        - '*apt-get install zmap*'
    condition: keywords
falsepositives:
    - Unknown
level: high
-------------------------------------------------------------------
------------------LOG CELERY------------
[2020-11-04 08:37:39,629] celery.pool/DEBUG TaskPool: Apply <function _fast_trace_task at 0x7f2dc8972f28> (args:('timesketch.lib.tasks.run_sketch_analyzer', '52b73a3a-158f-4cc9-8ec4-ee880593cb2d', {'lang': 'py', 
'task': 'timesketch.lib.tasks.run_sketch_analyzer', 'id': '52b73a3a-158f-4cc9-8ec4-ee880593cb2d', 'shadow': None, 'eta': None, 'expires': None, 'group': None, 'retries': 0, 'timelimit': [None, None], 'root_id': '
6009bdbb-f7f6-49d4-b1b0-64ec736903ce', 'parent_id': '6009bdbb-f7f6-49d4-b1b0-64ec736903ce', 'argsrepr': "('dd627756168a47869e7d5022216b1906', 1, 19, 'sigma')", 'kwargsrepr': '{}', 'origin': 'gen155@26fa720cf417',
[...]

PR #1468, opened 2020-11-13
Fix issue 1465 by removing hacky string replace  .keyword

jaegeral:
https://github.com/google/timesketch/issues/1465

(see comment in the issue discussion)


kiddinn (approved):
LGTM

PR #1471, opened 2020-11-16
Save and cache graphs

berggren:
This PR adds the ability to save subgraphs (selection of elements) as well as caching graph results.
It also adds two new DB models:
* Graph - saved graph based on a selected subgraph (similar to a saved view)
* GraphCache - cache per graph plugin and sketch. The cache can be manually refreshed, but will also refresh automatically if any timelines is added/removed.

Screenshot 2020-11-16 at 17 21 38


timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+            abort(HTTP_STATUS_CODE_NOT_FOUND, 'No sketch found with this ID.')
+
+        graphs = sketch.graphs
+        return self.to_json(graphs)
+
+    def post(self, sketch_id):
+        """Handles POST request to the resource."""
+        sketch = Sketch.query.get_with_acl(sketch_id)
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, 'No sketch found with this ID.')
```
kiddinn, line 59: OK, I would like to see a check here to see if you have write access to the sketch, since we are writing a graph to the database... please make sure you are not allowing everyone with read access to add this here

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+    """Resource to get all graphs."""
+
+    @login_required
+    def get(self, sketch_id, graph_id):
+        """Handles GET request to the resource.
+
+        Returns:
+            List of graphs in JSON (instance of flask.wrappers.Response)
+        """
+        sketch = Sketch.query.get_with_acl(sketch_id)
```
kiddinn, line 84: missing a check for sketch here... I would move the `if not sketch` check immediately after this, bail out early if there is no sketch

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+        graph = Graph.query.get(graph_id)
```
kiddinn, line 85: add a `if not graph` check as well

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, 'No sketch found with this ID.')
+
+        if not sketch.id == graph.sketch.id:
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND,
```
kiddinn, line 92: I would probably not use not found here but rather bad request

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+            user=current_user, sketch=sketch, name=str(name),
+            graph_elements=json.dumps(elements))
+        db_session.add(graph)
+        db_session.commit()
+
+        return self.to_json(graph)
 
 
 class GraphResource(resources.ResourceMixin, Resource):
+    """Resource to get all graphs."""
```
kiddinn, line 75: is it a resource to get "all" graphs or a specific one?

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+                'Graph does not belong to this sketch.')
+
+        response = self.to_json(graph).json
+        response['objects'][0]['elements'] = graph.graph_elements
+
+        return jsonify(response)
+
+
+class GraphCacheListResource(resources.ResourceMixin, Resource):
+    """Resource to get all graphs."""
```
kiddinn, line 102: all graphs or cache?

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+    def get(self):
+        """Handles GET request to the resource.
+
+        Returns:
+            List of graphs in JSON (instance of flask.wrappers.Response)
+        """
+        graph_plugins = manager.GraphManager.get_graphs()
+        graphs = []
+        for name, graph_class in graph_plugins:
+            graph_instance = graph_class()
```
kiddinn, line 114: why do you need to instantiate the class to get access to these constants? You should be able to use the class instance here

timesketch/api/v1/resources/graph.py
```diff
@@ -34,19 +39,90 @@ class GraphListResource(resources.ResourceMixin, Resource):
...
+                'display_name': graph_instance.DISPLAY_NAME,
+                'description': graph_instance.DESCRIPTION
+            }
+            graphs.append(graph_plugin)
+
+        return jsonify(graphs)
+
+
+class GraphCacheResource(resources.ResourceMixin, Resource):
     """Resource to get a graph."""
```
kiddinn, line 126: get a graph or get a cached graph?

timesketch/api/v1/resources/graph.py
```diff
@@ -61,8 +137,38 @@ def post(self, sketch_id):
...
+        plugin_name = form.get('plugin')
+        graph_config = form.get('config')
+        refresh = form.get('refresh')
+
+        sketch_indices = [
+            timeline.searchindex.index_name
+            for timeline in sketch.active_timelines
+        ]
+
+        cache = GraphCache.get_or_create(
```
kiddinn, line 149: ok, so do we want people with read access to the sketch to be able to do this? or should we add a check to see if you've go write access before proceeding?

timesketch/models/sketch.py
```diff
@@ -605,3 +607,73 @@ def __init__(
...
+            sketch (Sketch): The sketch that the graph is bound to.
+            name (str): Name of the graph.
+            description (str): Description of the graph.
+            graph_config (dict): Config used when generating the graph.
+            graph_elements (str): Graph in json string format.
+            graph_thumbnail (str): Image of graph in Base64 format.
+            num_nodes (int): Number of nodes in the graph.
+            num_edges (int): Number of edges in the graph.
+        """
+        super(Graph, self).__init__()
```
kiddinn, line 640: we should use the py3 way of using super.... 

super().__init__()

that is empty super calls (need to do that everywhere i the file)

kiddinn (changes requested):
few comments

timesketch/api/v1/resources/graph.py
berggren: Done

timesketch/api/v1/resources/graph.py
berggren: changed the wording. All saved graphs for a sketch.,

timesketch/api/v1/resources/graph.py
berggren: fixed.

timesketch/api/v1/resources/graph.py
berggren: done

timesketch/api/v1/resources/graph.py
berggren: good point, done.

timesketch/api/v1/resources/graph.py
berggren: Fixed, reworded.

berggren: Actually, I refactored this to be named GraphPluginListResource instead, which reflects what it actually is.

timesketch/api/v1/resources/graph.py
berggren: good point, fixed.

timesketch/api/v1/resources/graph.py
berggren: fixed

timesketch/api/v1/resources/graph.py
berggren: Yeah, for the cache it is ok for read only users to generate if it doesn't exist.

timesketch/models/sketch.py
berggren: Noted, will refactor in a later PR :)

berggren (commented):
PTAL

kiddinn (approved):
LGTM

PR #1480, opened 2020-11-18
Graph: Use ID from attributes

berggren:
This PR make it possible for Graph developers to set an ID string to use. This is useful if you don't want to rely on the label string to calculate the ID.


kiddinn (approved):
LGTM

PR #1486, opened 2020-11-19
Feature extract ioc

lprat:
Add option IOC check on Analyzer Feature extract.
This option can allow to check IOC during data extraction
IOC in file should be line by line, and you can to add comment context with symbol '|' after IOC.

replace PR: https://github.com/google/timesketch/pull/1453
Thank

### DIFF
--- a/data/features.yaml
+++ b/data/features.yaml
@@ -18,6 +18,8 @@
 #       overwrite_and_merge_store_as: False
 #       store_type_list: False
 #       keep_multimatch: False
+#       ioc_tag: []
+#       ioc_path:
 #
 # Each definition needs to define either a query_string or a query_dsl.
 #
@@ -54,6 +56,13 @@
 #
 # The keep_multimatch is an optional boolean that determines if we want to
 # store all matching results (default store first result).
+#
+# The ioc_path is an optional file path that contains ioc list to check 
+# on extracted data. The file content should be ioc line by line, and you
+# can to use optionnal info context if you add comment after symbol '|'.
+# E.G. my_ioc.com|source:https://threat_intel_website
+#
+# The ioc_tags is optional, you can add tags if ioc matched.
 #
 # The feature extraction works in the way that the query is run, and
 # the regular expression is run against the attribute to extract a value.
--- a/timesketch/lib/analyzers/feature_extraction.py
+++ b/timesketch/lib/analyzers/feature_extraction.py
@@ -4,6 +4,8 @@ from __future__ import unicode_literals
 import logging
 import re
 
+import os
+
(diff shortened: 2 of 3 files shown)


kiddinn (changes requested):
OK, after thinking about this some more and talking with @berggren about it we came to the conclusion that a better approach would be (and a more generic and not really IOC specific)

1. We don't extend or change the feature extraction analyzer. It will be used to extract features, that is if you don't have an IP address, but want to extract it from somewhere you can add configs into the feature extraction aS IS, no need to change that.

2. We extend the tagger analyzer. And let me explain the changes to the tagger.

We introduce a new config there called `compare_file` which would point to a file. The search query would be simple, as in something like `_exists_:"domain"` and that would produce all events that have the domain field set.

If (compare_file is optional) compare_file is not provided, all entries are tagged, as they are today. But if the `compare_file` option is there. A CSV file is parsed (using pandas.read_csv). The CSV file will have a mandatory field called `compare` or `name` (you can pick) and then all other fields are optional.

If the `compare_file` option is used, then that is an additional comparison, then the entry will need to both match the query_filter AND be inside the file. This can be a regular expression or a string. You can use built-in pandas operations to make the matches here.

If an entry matches both, that is both it comes from the search query filter and it matches an entry from the `compare_file` option ALL other fields in the CSV will be stored as fields in the events.

So let's take an example: 

```
ioc_domain_tagger:
  query_string: '_exists_:"domain"'
  tags: ['ioc']
  emojis: ['SPARKLES']
  create_view: true
  view_name: 'IOC matches'
  compare_file: 'domains.csv'
```

And the content of the CSV file would be:
```
name,context,notes
evil.com,"This comes from an infamous report","I'm not convinced of this one."
foobar.com,"This friend of mine told me about this one",
```

Then we would query for all domains, then compare teh domains against the list of domains in the compare file, so `evil.com` and `foobar.com` in this example. If there is a match there, then we add a tag to these events "ioc", add an emoji and we add two fields, `context` and `notes`. And the values of that would be what you see there.

I would suggest to have the values a list, instead of a string. And then you would append to it if the context/note hasn't been put there before.

The question is also whether we want to tie the context back to the root, if so you would just do

```
field: context
value: ["evil.com|This comes....", ]
```

Does this make sense?

That way we don't have just an IOC feature here, but we have something more flexible that can cover more scenarios/use-cases.

berggren:
Hi @lprat and @kiddinn What is the status of this PR?

lprat:
Sorry, the idea was abandoned.
The Best way is to use 'tag'  analyzer.
I close.

PR #1490, opened 2020-11-24
Added API client search object

kiddinn:
This PR implements few things:

1. Changes `super` calls to use py3 syntax in the API client
2. Changes class definitions to use py3 syntax in the API client
3. Introduces the `search.Search` object, which replaces all search functionality (the bread and butter of this PR)
4. Changes all occurrences of the view object to a search object.
5. Adds in some deprecation warnings to the view object.

The main change here is nr 3, this introduces a search.Search object that is responsible for all searches. This replaces really the `sketch.explore` call and the view object. This object can take in searches, defines chips, which can be edited, and tuned. And in turns provides an object that can be changed to a dataframe, dict or save the content to a file. It also provides the option to save the content of the search to the database (the old functionality of the view object, which was very limited and didn't provide all the options a view object can entail).

Chips are now classes on their own, defining functions and properties that are designed for each use case.

What needs to be done after this is to write up some documentation on how to conduct search using the API client, `docs/APISearch` or something like that.


api_client/python/timesketch_api_client/search.py
```diff
@@ -0,0 +1,840 @@
...
+        Raises:
+            ValueError: if unable to query for the results.
+            RuntimeError: if the query is missing needed values, or if the
+                sketch is archived.
+        """
+        super().from_explore(**kwargs)
+        if not (query_string or query_filter or query_dsl):
+            raise RuntimeError('You need to supply a query')
+
+        self._username = getpass.getuser()
```
berggren, line 540: Don't you wanna get the username from the .rc file?

api_client/python/timesketch_api_client/search.py
```diff
@@ -0,0 +1,840 @@
...
+        """Property that returns back the description of the saved search."""
+        return self._description
+
+    @description.setter
+    def description(self, description):
+        """Make changes to the saved search description field."""
+        self._description = description
+        self.commit()
+
+    def from_explore(  # pylint: disable=arguments-differ
```
berggren, line 506: from_explore? What does that mean? shouldn't this just be `search()` ?

api_client/python/timesketch_api_client/search.py
```diff
@@ -0,0 +1,840 @@
...
+        #self._searchtemplate = data.get('searchtemplate', 0)
+        #self._aggregations = data.get('aggregation', 0)
+
+        self._created_at = datetime.datetime.now(
+            datetime.timezone.utc).isoformat()
+        self._updated_at = self._created_at
+
+        self.resource_data = {}
+
+    def from_store(self, search_id):  # pylint: disable=arguments-differ
```
berggren, line 561: I would call this `from_saved_search`

api_client/python/timesketch_api_client/search_test.py
```diff
@@ -0,0 +1,135 @@
+# Copyright 2017 Google Inc. All rights reserved.
```
berggren, line 1:
```suggestion
# Copyright 2020 Google Inc. All rights reserved.
```

docs/APIClient.md
```diff
@@ -83,83 +83,231 @@ Each of these properties can be accessed using `sketch.<PROPERTY>`, eg. `sketch.
...
+
+```
+search_obj.*?
+```
+
+To find a list of all the possible configuration options.
+
+#### Chips
+
+Chips are the the "+ Time filter" or "+ Add label filter", etc elements you can
```
berggren, line 206:
```suggestion
Chips are the "+ Time filter" or "+ Add label filter", etc elements you can
```

berggren (changes requested):

api_client/python/timesketch_api_client/search.py
kiddinn: well, this is coming from the base object as well... to be consistent... we can name it something different, but we've already got the aggregation object defining that function name... 

we can just drop this also from the base class and rename this one

api_client/python/timesketch_api_client/search.py
kiddinn: hmmm.... so that would probably be better... get the actual logged in user, instead of the system user like we are doing right now.... let me change that.

api_client/python/timesketch_api_client/search.py
kiddinn: so this is to confirm with the base class... and to have it the same across... so in aggregations we've got `agg.from_store`, etc, etc,

kiddinn (commented):
done with this round

api_client/python/timesketch_api_client/search.py
berggren: changed to from_manual() as discussed offline

api_client/python/timesketch_api_client/search.py
berggren: changed to from_saved() as discussed offline

berggren (approved):
LGTM

PR #1493, opened 2020-11-25
Fix saved views in stories

berggren:
Event list for saved views are not rendered. Use query string instead of DSL to fix.


kiddinn (approved):
LGTM

PR #1497, opened 2020-12-01
Adding the ability to define multiple TS instances in the RC file.

kiddinn:
This is adding the ability to define different sections in the RC file to store more than one TS server connection information.

This provides the ability to connect to different TS instances.

Example code:

```
from timesketch_api_client import config
c1 = config.get_client()
```

this would get the default client configuration (section named `timesketch`) in the RC file and then you can do something like:

```
c2  = config.get_client(config_section='timesketch-demo')
```

To get a client that connects to another TS instance.

The RC file would look something like this:

```
[timesketch]
host_uri = '....
...

[timesketch-demo]
host_uri = '....
....
token_file_path = 'path_to_the_other_token_file'
...
```


berggren (approved):

berggren (changes requested):
Please fix linter, otherwise looks good

berggren (approved):

PR #1498, opened 2020-12-02
Minor changes to the search API object.

kiddinn:
This PR adds few things in the API client:

1. Fixes the `from_explore` function that should have been renamed last time
2. Chips weren't preserved when reading from saved searches, nor where they when the query filter was manually set. This is now fixed.
3. It is now possible to change the name and description of a timeline object in the API client.


berggren (approved):
LGTM

PR #1502, opened 2020-12-03
Sigma Rule maintain improvement

jaegeral:
This PR does two things:
- introduce an option to move bad sigma files to a dedicated folder
- refactors the verify_sigm.py tool to give the option to run it outside of Timesketch with the rule folder to check which ones are bad. The refactoring makes use of the new sigma_util class

Would like to get feedback if something like the "move sigma rules if they fail" is something we want to do.

Another option would be to move the "move rules away" function in the sigma_verify tool so that it can be done before deployment and is not causing any things during runtime.


test_tools/sigma_verify_rules.py
```diff
@@ -21,80 +21,20 @@
...
 
+
 import logging
 import os
 import argparse
 import sys
-import codecs
-import sigma.parser.exceptions
-import sigma.configuration as sigma_configuration
+# TODO: not sure if there is a better way
```
jaegeral, line 29: This obv causes a linter issue, but I am not sure how to fix that...

test_tools/sigma_verify_rules.py
```diff
@@ -21,80 +21,20 @@
...
+sys.path.append('../timesketch/lib/')
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
+import sigma_util as ts_sigma_lib
```
jaegeral, line 32: This causes a linter issue as it is not possible to import, I do not know how to fix it in a different way.

berggren: why not just `from timesketch.lib import sigma_util` ?

berggren (changes requested):

kiddinn:
you also ahve:

```
test_tools/sigma_verify_rules.py:20:73: C0303: Trailing whitespace (trailing-whitespace)
```

in the linter

test_tools/sigma_verify_rules.py
jaegeral: because it is in test_tools. So timesketch.lib would look into ```test_tools/timesketch/lib```or am I wrong?

kiddinn: the file is stored in test_tools/timesketch/lib and the way you run the tool is to run it within the test_Tool folder where you run it using `PYTHONPATH=. python ....` meaning that that is the only way to run this, otherwise this will fail.

and yes, it is stored there and thereore should be imported from there

kiddinn (commented):
.

jaegeral:
Should be good to be reviewed again @kiddinn

test_tools/sigma_verify_rules.py
```diff
@@ -17,84 +17,21 @@
 It also does not require you to have a full blown Timesketch instance.
 Default this tool will show only the rules that cause problems.
 Example way of running the tool:
-  $ sigma_verify_rules.py --config_file ../data/sigma_config.yaml ../data/linux/
+$ PYTHONPATH=. python3 test_tools/sigma_verify_rules.py --config_file
```
kiddinn, line 20: I would say that this should be 

```shell
cd test_tools
PYTHONPATH=. python3 test_tools/sigma_verify_rules.py --config......
...
```

that is you run it inside the `test_tools` folder, otherwise you are using the pythonpath as the root directory, which means you use the `actual` timesketch files, instead of the ones in the test folder.

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
 
-            rule_file_path = os.path.join(dirpath, rule_filename)
-            rule_file_path = os.path.abspath(rule_file_path)
 
-            if verify_rules_file(rule_file_path, sigma_config, backend):
-                return_verified_rules.append(rule_file_path)
-            else:
-                logger.info('File did not work{0:s}'.format(rule_file_path))
-                return_rules_with_problems.append(rule_file_path)
+def move_problematic_rule(filepath, move_to_path, reason=""):
```
kiddinn, line 86: single vs double quotes

`reason=''`

kiddinn: and do we want the reason to be optional?

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
+
+    Args:
+        filepath: path to the sigma rule that caused problems
+        move_to_path: path to move the problematic rules to
+        reason (optional): reason why file is moved
+
+    Returns:
+        Nothing
+    """
+
```
kiddinn, line 97: not an empty line here

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
+def move_problematic_rule(filepath, move_to_path, reason=""):
+    """ Moves a problematic rule to a subfolder so it is not used again
+
+    Args:
+        filepath: path to the sigma rule that caused problems
+        move_to_path: path to move the problematic rules to
+        reason (optional): reason why file is moved
+
+    Returns:
+        Nothing
```
kiddinn, line 95: skip the returns section if there is no return value

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
+    """
+
+    logging.info('Moving the rule: {0:s} to {1:s}'.format(
+        filepath, move_to_path))
+
+    os.makedirs(move_to_path, exist_ok=True)
```
kiddinn, line 101: want to have a try/except here in case something fails, for instance if you attempt to create a directory in a path that you don't have write permissions to?

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
+    file_object = open('{0:s}debug.log'.format(move_to_path), 'a')
```
kiddinn, line 102: I would rather use `os.path.join` here ...

```
debug_path = os.path.join(move_to_path, 'debug.log')
```

and also instead of using open/close, can you use `with`

```
with open(debug_path, 'a') as file_objec:
  file_object.write...

```

since there are guarantees that the file handle is properly closed when you use the `with` syntax, which are not provided by doing it this way

test_tools/sigma_verify_rules.py
```diff
@@ -119,51 +56,61 @@ def run_verifier(rules_path, config_file_path):
...
+    file_object.write('{0:s}\n{1:s}\n\n'.format(filepath, reason))
+    file_object.close()
+    os.rename(filepath, '{0:s}{1:s}'.format(
```
kiddinn, line 105: also enclose with try/except here

test_tools/sigma_verify_rules.py
```diff
@@ -214,6 +166,10 @@ def run_verifier(rules_path, config_file_path):
         print('### You should NOT import the following rules ###')
         print('### To get the reason per rule, re-run with --info###')
         for badrule in sigma_rules_with_problems:
+            if options.move_to_path:
+                move_problematic_rule(
+                    badrule, options.move_to_path,
+                    "sigma_verify_rules.py found an issue")
```
kiddinn, line 172: single vs double quotes

timesketch/lib/sigma_util.py
```diff
@@ -39,24 +39,25 @@ def get_sigma_config_file():
...
-    config_file = current_app.config.get('SIGMA_CONFIG')
+    if config_file is None:
+        config_file_path = current_app.config.get('SIGMA_CONFIG')
 
     if not config_file:
         raise ValueError(
             'SIGMA_CONFIG not found in config file')
 
-    if not os.path.isfile(config_file):
+    if not os.path.isfile(config_file_path):
```
kiddinn, line 49: config_file_path is not set if `config_file` is set...

so ou'll need to pick one

```
if config_file:
  config_file_path = config_file
else:
  config_file_path = current_app.conf....
```

something like that

timesketch/lib/sigma_util.py
```diff
@@ -72,7 +73,11 @@ def get_sigma_rules_path():
         ValueError: If SIGMA_RULES_FOLDERS is not found in the config file.
             or the folders are not readabale.
     """
-    rules_path = current_app.config.get('SIGMA_RULES_FOLDERS', [])
+    try:
+        rules_path = current_app.config.get('SIGMA_RULES_FOLDERS', [])
+    except RuntimeError:
+        # TODO: make that more flexible I have no idea what to best return here
+        return ['/']
```
kiddinn, line 80: can you please fix this before we submit this PR

timesketch/lib/sigma_util.py
```diff
@@ -92,11 +97,12 @@ def get_sigma_rules_path():
     return rules_path
 
 
-def get_sigma_rules(rule_folder):
+def get_sigma_rules(rule_folder, sigma_config=None):
     """Returns the Sigma rules for a folder including subfolders.
 
     Args:
         rule_folder: folder to be checked for rules
+        sigma_config: optional: pass a Sigma config object
```
kiddinn, line 105: fix the docstring, not having `:` twice... 

```
sigma_config: optional dict with Sigma config.
```

or something like that... what is `sigma config`? Is that a special object, is it a dict, is it? unclear from the docstring

timesketch/lib/sigma_util.py
```diff
@@ -145,17 +151,19 @@ def get_all_sigma_rules():
     return sigma_rules
 
 
-def get_sigma_rule(filepath):
+def get_sigma_rule(filepath, sigma_config=None):
     """ Returns a JSON represenation for a rule
 
     Args:
         filepath: path to the sigma rule to be parsed
+        sigma_config: optional: pass a Sigma config object
```
kiddinn, line 159: see earlier docstring comment

timesketch/lib/sigma_util.py
```diff
@@ -29,7 +29,7 @@
 logger = logging.getLogger('timesketch.lib.sigma')
 
 
-def get_sigma_config_file():
+def get_sigma_config_file(config_file=None):
```
kiddinn, line 32: are these changes reflected in the `test_tools/timesketch/lib/sigma_util.py` file as well?

kiddinn (changes requested):
few more comments

timesketch/lib/sigma_util.py
jaegeral: There is no sigma_util.py in the test_tools, my idea was to use the same, that is why this https://github.com/google/timesketch/pull/1502/files#r536000064 is from the root timesketch folder instead of the ```test_tools``` folder.

timesketch/lib/sigma_util.py
jaegeral: indeed, also mentioned the optional arg in the docstring, forgot that.

timesketch/lib/sigma_util.py
jaegeral: yes, will raise a ```ValueError```instead. This is mostly if used in unittests, the config file can not be read.

timesketch/lib/sigma_util.py
jaegeral: fixed

test_tools/sigma_verify_rules.py
jaegeral: so the idea here is to actually use the real timesketch files (the util file) otherwise I would need to copy the file over and always keep it sync.

test_tools/sigma_verify_rules.py
jaegeral: yes it shold be optional, made it that.

test_tools/sigma_verify_rules.py
jaegeral: good point, I will make it a broad OSError except and an log entry that no files have been moved, that should be good enough WDYT?

test_tools/sigma_verify_rules.py
jaegeral: as always, clever, yes!

kiddinn (approved):

PR #1503, opened 2020-12-03
Modification to enable connection to production Elastic instance

joshfrogers:
In order to connect to an SSL secured elastic instance, we need to provide authentication details as part of the connection.

This pull request adds the option to include authentication details in the config, failing back to unauthenticated/HTTP connection if SSL is not supplied.


timesketch/api/v1/resources/__init__.py
```diff
@@ -232,7 +232,12 @@ def datastore(self):
         """
         return ElasticsearchDataStore(
             host=current_app.config['ELASTIC_HOST'],
-            port=current_app.config['ELASTIC_PORT'])
+            port=current_app.config['ELASTIC_PORT'],
```
kiddinn, line 235: these are a lot of options that you need to read in, and configure and in a lot of places.

I think it would be much better to do this in a shared library, maybe just in the `timesketch.lib.datastores.elastic` or in some other file, maybe a new one, that can the import both this library as well as the flask configs, that would set it up, ... since you both have a lot of options you are reading and having conditions for SSL oro fallback to HTTP

timesketch/lib/aggregators_old_test.py
```diff
@@ -25,6 +25,6 @@ class TestAggregators(BaseTest):
 
     def test_heatmap(self):
         """Test to get heatmap data."""
-        es_client = MockDataStore('127.0.0,1', 4711)
+        es_client = MockDataStore('127.0.0,1', 4711, user='user', password='password', ssl=False, verify=True)
```
kiddinn: line too long

kiddinn (changes requested):
fwe comments

timesketch/lib/aggregators/interface.py
```diff
@@ -154,9 +154,21 @@ def __init__(self, sketch_id=None, index=None):
         if not sketch_id and not index:
             raise RuntimeError('Need at least sketch_id or index')
 
-        self.elastic = Elasticsearch(
-            host=current_app.config['ELASTIC_HOST'],
-            port=current_app.config['ELASTIC_PORT'])
+        host = current_app.config['ELASTIC_HOST']
+        port = current_app.config['ELASTIC_PORT']
+        user = current_app.config.get('ELASTIC_USER', None)
+        password = current_app.config.get('ELASTIC_PASSWORD', None)
```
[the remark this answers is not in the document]
joshfrogers: Updated and resolved in current build

timesketch/lib/aggregators_old_test.py
joshfrogers: Updated and resolved in current

timesketch/api/v1/resources/__init__.py
joshfrogers: Pulled the SSL config into timesketch/lib/datastores/elastic.py.

Looking at timesketch/lib/aggregators/interface.py we skip the datastore and use the Elasticsearch library directly.

Is there a reason behind this, or would we be okay to implement the methods used under ElasticsearchDataStore?

timesketch/lib/datastores/elastic.py
```diff
@@ -870,6 +884,29 @@ def flush_queued_events(self):
...
+        """Get Elasticsearch Field Mapping
+
+        Returns:
+            ElasticSearch Field Mapping as json object.
+        """
+        return self.client.indices.get_field_mapping(index=index,
+                                                     fields=fields)
+
+    def aggregation_shim(self, index, aggregation_spec):
+        """Shim ahead of moving the Helper method 'elastic_aggregation'
```
kiddinn, line 897: ok, so two things

a) can you explain what you mean by this being a `shim ahead of moving the helper...`

b) can you change the doc string here... if you look at the style guide (https://google.github.io/styleguide/pyguide.html#381-docstrings) a function docstring should be:

```
A docstring should give enough information to write a call to the function without reading the function’s code. The docstring should be descriptive-style ("""Fetches rows from a Bigtable.""") rather than imperative-style ("""Fetch rows from a Bigtable."""), except
```

You can have this line as the line after that.

def agg....
   ""Returns Elastic aggregation results .....

   Shim ahead of moving....

   Args: 
 ....
"""

kiddinn (commented):
one tiny comment here

timesketch/lib/aggregators/interface.py
```diff
@@ -236,8 +237,8 @@ def elastic_aggregation(self, aggregation_spec):
         Returns:
             Elasticsearch aggregation result.
         """
-        # pylint: disable=unexpected-keyword-arg
-        aggregation = self.elastic.search(
+        # pylint: disable=unexpected-keyword-arg, no-value-for-parameter
+        aggregation = self.elastic.aggregation_shim(
```
berggren, line 241: Do we need a separate method for this? It is only used in one call from this file.
I suggest you just do:

```
aggregation = self.elastic.client.search( ... )
```

berggren: and remove `aggregation_shim`

timesketch/lib/datastores/elastic.py
berggren: Looking at this PR, this method doesn't add anything. You can simplify it by removing this method and call `self.client.search()` as suggested earlier in this file.

Or am I missing something here? :)

berggren (changes requested):
Hey @joshfrogers Thanks for adding this functionality! I added some suggestions in this review. PTAL

timesketch/lib/aggregators/interface.py
joshfrogers: Thanks for reviewing! The intention of this shim was to move the method from `timesketch.lib.aggregators.interface` into the `timesketch.lib.datastores.elastic.ElasticDatastore` class. 

In line 157 of `timesketch.lib.aggregators.interface` we currently use a direct call to the Elastic Library, rather than using ElasticDatastore. I wanted to standardise this, rather than reimplementing the check for SSL config here, so re-wrote it as ElasticDatastore. Unfortunately (I may be wrong here) the implementation of ElasticDataStore.Search doesn't match the implementation of Elasticsearch.search - hence the shim method. 

I was thinking of moving the method `timesketch.lib.aggregators.interface.elastic_aggregation` into the `timesketch.lib.datastores.elastic.ElasticDatastore` class, but didn't want to break anything (hence leaving this there currently).

timesketch/lib/datastores/elastic.py
joshfrogers: Will Update Docstring as above. @berggren I think i've replied to the wrong comment - the reasoning for the shim is here: https://github.com/google/timesketch/pull/1503#discussion_r536218345 

Though more than happy to change :D

timesketch/lib/aggregators/interface.py
joshfrogers: My apologies - realised my mistake! moving this to the single call

timesketch/lib/datastores/elastic.py
```diff
@@ -870,6 +884,15 @@ def flush_queued_events(self):
         self.import_events = []
         return return_dict
 
+    def get_field_mapping(self, index, fields):
```
berggren, line 887: Same here: Can you call self.client.indices.get_field_mapping directly instead of using this method?

berggren (changes requested):
One more thing, but looks good.

timesketch/lib/datastores/elastic.py
joshfrogers: Confirmed and done 👍

kiddinn (approved):
LGTM

And thank you for the contribution

PR #1504, opened 2020-12-04
Adding picatrix docker container to the development docker config

kiddinn:
This adds a jupyter notebook docker container, based on picatrix to the dev docker config.

This means that once you use the dev docker config a jupyter notebook with access to the Timesketch instance


kiddinn:
Wait with a review, need to add one more change in before

kiddinn:
@berggren I've added the first version of the snippets support, please review

docker/dev/build/timesketch_token
```diff
@@ -0,0 +1,2 @@
+��^_~&l�/�
```
berggren, line 1: Is this token suppose to be checked in?

berggren (commented):
One question, ptal

docker/dev/build/timesketch_token
kiddinn: Yes, this is the dev password

kiddinn (commented):
Back to you

berggren (approved):
LGTM

PR #1508, opened 2020-12-07
Added missing credentials to documentation

abitrolly:
Because I am a kewlhacker \m/ :D


jaegeral (approved):
LGTM, that makes sense to note

PR #1511, opened 2020-12-08
[API] introduce text --> sigma rule

jaegeral:
This is still a draft. The goal of this PR is to open up the different layers (util, API, API client) to accept a raw text that is a Sigma Rule YAML file and turn it into a ES Query based on the Timesketch Sigma config (field mapping) so that it can e.g. be used in the UI have a search field where you just enter a Sigma Rule and it will return the matching events.

This is still far from ready to be merged but WIP.


api_client/python/timesketch_api_client/client.py
```diff
@@ -549,3 +549,24 @@ def get_sigma_rule(self, rule_uuid):
...
+            rule_text: Full Sigma rule text.
+
+        Returns:
+            Instance of a Sigma object.
+        """
+        if not rule_text:
+            raise ValueError('Rule text given.')
+
+        try:
+            sigma_rule = None #sigma_util.get_sigma_rule_by_text(rule_text)
```
kiddinn, line 566: why not

```
sigma_obj = sigma.Sigma(api=self)

sigma_obj.from_text(rule_text)
```

And also with that change the `__init__` to not take in rule_uuid but rather have

```
sigma_obj = sigma.Sigma(api=self)
sigma_obj.from_rule(rule_uuid)
```

that way you can use the same mechanism to accommodate all use cases

kiddinn: one of the reasons is that an API client cannot import sigma_util, the library, since that is not part of the API client codebase but rather part of timesketch, and we don't want to depend on that.

in `from_text` or `from_rule_text` you can then do some textual verifications, or other verifications and then send an API call that does the rest....

kiddinn (commented):
I know this is still a draft, but wanted to send in one comment beforehand....

api_client/python/timesketch_api_client/client.py
jaegeral: That is a very good and valid point, will surely do that!

jaegeral:
Some checks are currently not working, this is kind of expected, waiting for Kiddi to provide feedback to the setup of my api_client sigma object methods.

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
...
+            rule_uuid: Id of the sigma rule.
+
+        """
+        # TODO: Write docstring
+        self.rule_uuid = rule_uuid
+        # TODO: not sure which one is the better one
+        self._resource_uri = f'sigma/{rule_uuid}'
+        self.resource_uri = f'sigma/{rule_uuid}'
+        super().__init__(
+            api=self.api, resource_uri=self._resource_uri)
```
kiddinn, line 115: you should not do this... this is a `from_` function, not an `__init__` function

see how this is done in `search.py` for instance, that is the `__init__` function defines simply what variables are available, that is what you need to have saved...

then in the `from_rule_uuid` you set the resource_uri, and then you fetch the data for it, that is you do the fetching of data, if this is a GET request you can simple use `data = self.data` after you set the `self.resource_uri` and then you can start looking at the data...

it should have a `meta` and `objects` section, which you can then start to open up and assign variables, assign the `self._ATTR` values that you defined in `__init__`, then you've got properties that can fetch these values.

kiddinn: but look at how this is done in `search.py` or `aggregation.py` for example.

you can see how the `from_*` functions operate there.

Essentially you take the input, make a request to teh API, interpret the results and save to variables.... then you also have `from_rule_text` or `from_something_else` and in those cases you may not need to do a request to the API, but just parse/interpret and assign to these same variables.

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
...
+        """Get a Sigma object from a rule text.
+
+        Args:
+            rule_text: Rule text to be parsed.
+
+        """
+        resource_url = '{0:s}/sigma_by_text/'.format(self.api.api_root)
+        data = {'title': 'Get_Sigma_by_text', 'content': rule_text}
+        response = self.api.session.post(resource_url, data=data)
+        print(response.__dict__)
```
kiddinn, line 129: you do here ..

```
data = error.get_response_json(response, logger)
```

and that way you can get the response JSON back, which is a dict with the results.

kiddinn: which I see you do in the next line... so skip this print

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
...
+        response_dict = error.get_response_json(response, logger)
+        #timeline_dict = response_dict['objects'][0]
+        #timeline_obj = sigma.Sigma(
+        #    timeline_id=timeline_dict['id'],
```
kiddinn, line 133: not sure why you do a ` timeline_dict`, since this is not a timeline... 

but here you should do something like

```
objects = response_dict.get('objects')

if not objects:
  # do something
```

and then depending on how the API works, you may only be interested in the first object, or all of the returned ones.

Then the `timeline_obj` should definitely not be another sigma object, you already have that... now you just assign these same variables that you set in `___init__` to the results from here.

kiddinn (changes requested):
leaving few comments here

timesketch/api/v1/routes.py
```diff
@@ -104,6 +105,7 @@
     (VersionResource, '/version/'),
     (SigmaListResource, '/sigma/'),
     (SigmaResource, '/sigma/<string:rule_uuid>/'),
```
jaegeral, line 107: Based on discussion, I will deprecate this and replace with:

```
sigma/text/
```
and 
```
sigma/rule/<string:rule_uuid>
```

timesketch/api/v1/routes.py
```diff
@@ -104,6 +105,7 @@
...
+    (SigmaByTextResource, '/sigma_by_text/'),
```
jaegeral, line 108: remove that and use: ```sigma/text/```

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
                 HTTP_STATUS_CODE_NOT_FOUND, 'No sigma rule found with this ID.')
 
         return return_rule
+
+class SigmaByTextResource(resources.ResourceMixin, Resource):
+    """Resource to get list of Sigma rules."""
```
kiddinn, line 98: not getting a list

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+
+    @login_required
+    def post(self):
+        """Handles POST request to the resource.
+
+        Returns:
+            JSON sigma rule
+        """
+        form = forms.SigmaRuleForm.build(request)
```
kiddinn, line 107: don't add a new form, please use just the request.json ... (see other samples)

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+        # optional: do we want to have a title at this point?
```
kiddinn, line 108: why would you want a title? you are just taking a text here, with a rule content and getting some JSON back

timesketch/lib/forms.py
```diff
@@ -235,3 +235,8 @@ class SearchIndexForm(BaseForm):
     es_index_name = StringField('Index', validators=[DataRequired()])
     public = BooleanField(
         'Public', false_values={False, 'false', ''}, default=False)
+
+class SigmaRuleForm(BaseForm):
```
jaegeral, line 239: Do not use forms, look at: archive.py instead.

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+        title = form.title.data
+        content = form.content.data
```
kiddinn, line 110: I'm missing some input validation here

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+
+        print(f'Content passed: {content}')
+        if not form.validate_on_submit():
```
kiddinn, line 113: don't use this, rather have some sort of valdation of the input

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, 'Unable to validate form data')
+
+        try:
+            sigma_rule = ts_sigma_lib.get_sigma_rule_by_text(content)
+
+        except ValueError:
+            # TODO: make better responses here with different scenarios.
```
kiddinn, line 120: yes... let's do a better error handling here

timesketch/lib/sigma_util.py
```diff
@@ -245,3 +258,61 @@ def get_sigma_rule(filepath, sigma_config=None):
             {'file_relpath':file_relpath})
 
         return rule_return
+
+def get_sigma_rule_by_text(rule_text, sigma_config=None):
```
jaegeral, line 262: input validation needed, because user is uploading text that could contain un-expected content.

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+            logger.error('Parsing error',
+                         exc_info=True)
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND,
+                'Error unable to parse the provided Sigma rule')
+
+        if sigma_rule is None:
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND, 'No sigma was parsed')
```
kiddinn, line 129: is that the case? or that the sigma rule was somehow not properly formed? or would that always raise an exception?

timesketch/api/v1/routes.py
kiddinn: yes

timesketch/lib/forms.py
kiddinn, line 239: remove this

timesketch/lib/sigma_util.py
```diff
@@ -47,9 +45,13 @@ def get_sigma_config_file(config_file=None):
     else:
         config_file_path = current_app.config.get('SIGMA_CONFIG')
 
-    if not config_file:
-        raise ValueError(
-            'SIGMA_CONFIG not found in config file')
+    # this is mostly to be able to test cases via API where the test from
+    # API client
+    if not config_file_path:
+        config_file_path = './data/sigma_config.yaml'
```
kiddinn, line 51: why do this?

you can also do 

```
current_app.config.get('SIGMA_CONFIG', './data/sigma_config.yaml')
```

or simply make sure that you pass in `./data/sigma_config.yaml` in when you do tests.... I would not have this in the production code where you are pointing to a test data set.

timesketch/lib/sigma_util.py
```diff
@@ -66,6 +68,10 @@ def get_sigma_config_file(config_file=None):
 
     sigma_config = sigma_configuration.SigmaConfiguration(sigma_config_file)
 
+    if not sigma_config:
+        raise ValueError(
+            'sigma_config is none - Error')
```
kiddinn, line 73: maybe have a better error message? also no need to break this into several lines for a short message like this

timesketch/lib/sigma_util.py
```diff
@@ -181,6 +182,18 @@ def get_sigma_rule(filepath, sigma_config=None):
 
     sigma_backend = sigma_es.ElasticsearchQuerystringBackend(sigma_conf_obj, {})
 
+    try:
+        sigma_rules_paths = get_sigma_rules_path()
+    except ValueError:
+        sigma_rules_paths = None
+
+    if not filepath.lower().endswith('.yml'):
+        return None
```
kiddinn, line 191: the returns section says that you return a JSON representation... here you are returning none... that's not JSON

timesketch/lib/sigma_util.py
```diff
@@ -245,3 +258,61 @@ def get_sigma_rule(filepath, sigma_config=None):
...
+    if sigma_config is None:
+        sigma_config = get_sigma_config_file()
+
+    sigma_backend = sigma_es.ElasticsearchQuerystringBackend(sigma_config, {})
+
+    rule_return = {}
+
+    try:
+        parser = sigma_collection.SigmaCollectionParser(
+                        rule_text, sigma_config, None)
```
kiddinn, line 281: isn't this a bit weird indentation

timesketch/lib/sigma_util.py
```diff
@@ -245,3 +258,61 @@ def get_sigma_rule(filepath, sigma_config=None):
...
+
+def get_sigma_rule_by_text(rule_text, sigma_config=None):
+    """ Returns a JSON represenation for a rule
+
+    Args:
+        rule_text: Text of the sigma rule to be parsed
+        sigma_config: config file object
+
+    Returns:
+        Json representation of the parsed rule
```
kiddinn, line 270: I see a bunch of `return None`

timesketch/api/v1/resources/sigma.py
```diff
@@ -90,3 +93,39 @@ def get(self, rule_uuid):
...
+
+        return sigma_rule
```
kiddinn, line 131: I'm seeing few things here...

a) I'm not seeing the correct structure returned, a dict with `meta` and `objects`... here should be a structure like that
b) you return back None if there are some issues... then you just check `is None` and have no meaningful error message, I think raising an exception is better, and catching that exception and using that in the return value is better, then the user in the UI/CLI/API will at least be able to see the error message that you get back. Instead of just `no sigma was parsed`, which doesn't tell you much (requiring a Timesketch SRE to look at logs to get some data back)

kiddinn (changes requested):
few more comments

timesketch/lib/sigma_util.py
jaegeral: yeah this is actually not about test data set but where the "default" location is stored, so the proposal with 

```current_app.config.get('SIGMA_CONFIG', './data/sigma_config.yaml')
``` is perfect.

File: timesketch/lib/sigma_util.py (hunk 17 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:73]: Good point, actually a ```try``` block might be even better.

File: timesketch/lib/sigma_util.py (hunk 14 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:262]: there is already ```yaml.safe_load_all```used, so not sure how much more can be done / needed to be done

File: timesketch/lib/sigma_util.py (hunk 20 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:259]: All the ```return None``` have been removed and proper ```Raises``` has been docstringed

File: timesketch/lib/sigma_util.py (hunk 19 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:276]: indeed, fixed

File: timesketch/api/v1/resources/sigma.py (hunk 15 above)
Comment (jaegeral) [reply timesketch/api/v1/resources/sigma.py:137]: now with results other than ```None```  a much more granular response can be given to the user.

Comment (jaegeral) [reply timesketch/api/v1/resources/sigma.py:157]: and this is basically last resort, It should never be reached, but I will keep it in the code

File: timesketch/api/v1/resources/sigma.py (hunk 21 above)
Comment (jaegeral) [reply timesketch/api/v1/resources/sigma.py:159]: ok the ```return None```aspect is fixed. The correct data structure indeed needs to be checked.

File: timesketch/lib/sigma_util.py (hunk 14 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:251]: @kiddinn  WDYT?

File: api_client/python/timesketch_api_client/client.py [hunk 22]
```diff
@@ -573,4 +573,29 @@ def get_sigma_rule(self, rule_uuid):
...
+        """Returns a Sigma Object based on a text POST.
+
+        Args:
+            rule_text: Full Sigma rule text.
+
+        Returns:
+            Instance of a Sigma object.
+        """
+        if not rule_text:
+            raise ValueError('Rule text given.')
```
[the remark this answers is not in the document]
jaegeral: fixed both, good catch

api_client/python/timesketch_api_client/client.py
```diff
@@ -573,4 +573,29 @@ def get_sigma_rule(self, rule_uuid):
...
+
+        try:
+            sigma_obj = sigma.Sigma(api=self)
+            sigma_obj.from_text(rule_text)
+
+        except ValueError:
+            logger.error('OS Error, unable to get the path to the Sigma rules',
```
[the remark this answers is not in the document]
jaegeral: Well this is true, it is the same with ```client.get_sigma_rule(rule_uuid)``` I am fine with both ways...

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
             return ''
 
         return sigma_data.get('file_relpath', '')
+
+    def from_rule(self, rule_uuid):
```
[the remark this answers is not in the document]
jaegeral: renamed

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -72,6 +73,8 @@ def title(self):
         sigma_data = self.data
 
         if not sigma_data:
+            self.lazyload_data()
```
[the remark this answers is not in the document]
jaegeral: does that property now make more sense? (I have not yet changed the other ones.

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
             abort(
                 HTTP_STATUS_CODE_NOT_FOUND, 'No sigma rule found with this ID.')
 
-        return return_rule
+        meta = {'current_user': current_user.username,
+                'rules_count': len(sigma_rules)}
+        return jsonify({'objects': return_rule, 'meta': meta})
+        #return return_rule
```
jaegeral, line 101: will remove that

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,7 +32,7 @@ class Sigma(resource.BaseResource):
     """
 
 
-    def __init__(self, rule_uuid, api, es_query=None,
+    def __init__(self, rule_uuid=None, api=None, es_query=None,
```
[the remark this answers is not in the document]
jaegeral: I changed the init method here.

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -72,6 +73,8 @@ def title(self):
         sigma_data = self.data
 
         if not sigma_data:
+            self.lazyload_data()
+            print("title not set yet")
```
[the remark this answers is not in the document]
jaegeral: done

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
...
         return sigma_data.get('file_relpath', '')
+
+    def from_rule(self, rule_uuid):
+        """Get a Sigma object from a rule uuid.
+
+        Args:
+            rule_uuid: Id of the sigma rule.
+
+        """
+        # TODO: Write docstring
```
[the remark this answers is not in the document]
jaegeral: done

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -95,3 +98,41 @@ def file_relpath(self):
...
+        self.rule_uuid = rule_uuid
+        # TODO: not sure which one is the better one
+        self._resource_uri = f'sigma/{rule_uuid}'
+        self.resource_uri = f'sigma/{rule_uuid}'
```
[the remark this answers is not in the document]
jaegeral: done

api_client/python/timesketch_api_client/sigma.py
jaegeral: ack

timesketch/api/v1/routes.py
jaegeral: implemented

timesketch/lib/forms.py
jaegeral: done imho

timesketch/api/v1/resources/sigma.py
jaegeral: done

timesketch/api/v1/resources/sigma.py
jaegeral: removed it

timesketch/api/v1/resources/sigma.py
jaegeral: I have rewritten that now, so not sure what kind of validation you have in mind

timesketch/api/v1/resources/sigma.py
jaegeral: rewritten it, closing that comment out.

timesketch/lib/sigma_util.py
jaegeral: ack, this now raises an Exception instead.

timesketch/api/v1/resources/sigma.py
jaegeral: done

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,72 @@ def get(self, rule_uuid):
             abort(
                 HTTP_STATUS_CODE_NOT_FOUND, 'No sigma rule found with this ID.')
 
-        return return_rule
+        meta = {'current_user': current_user.username,
+                'rules_count': len(sigma_rules)}
+        return jsonify({'objects': return_rule, 'meta': meta})
```
kiddinn, line 100: so objects here should be `'objects': [return_rule]` as in that should be a list.

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,72 @@ def get(self, rule_uuid):
...
+
```
kiddinn, line 101: are there not two empty lines here?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,72 @@ def get(self, rule_uuid):
...
+        if not form:
+            form = request.data
+
+        action = form.get('action', '')
+        
+        if action != 'post':
+            return abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Action needs to be "post"')
+        if action == 'post':
```
kiddinn, line 123: this `if` is not required... since you already have a check for `action != 'post'` which means that everything that follows has `action == 'post'`

but why are you checking the `action` this is a `def post(` function, and will not be called unless this is a post request. and why does the post request have to have a `action == 'post'` if it has no meaning?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+            content = form.get('content')
+            if not content:
+                return abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    'Missing values from the request.')
+
+        try:
+            sigma_rule = ts_sigma_lib.get_sigma_rule_by_text(content)
```
kiddinn, line 131: does all content verification happen here?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+
+        except ValueError:
+            logger.error('Parsing error',
+                         exc_info=True)
```
kiddinn, line 135: no need to be in two lines

kiddinn: also do we want to be more specific here... `Parsing error` does not tell me much (I know you add in the traceroute) but this is for the sysadmin logs, so perhaps something like `unable to parse a user supplied Sigma rule' or something like that.

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Error unable to parse the provided Sigma rule')
```
kiddinn, line 138: do you want to contain some of the error report here, or is this sufficient?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+        except NotImplementedError as exception:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Error generating rule {0!s}'
+                .format(exception))
```
kiddinn, line 143: too many lines, you can easily ahve the `Error....'.format(exception)` in one line.

don't want to log this down in the `logger`?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+
+        except sigma_exceptions.SigmaParseError as exception:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Sigma parsing error generating rule {0!s}'
```
kiddinn, line 148: `Sigma unable to parse rule, with error: {0!s}` or something like that?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+                .format(exception))
+
+        except yaml.parser.ParserError as exception:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Yaml parsing error generating rule in {0!s}'
```
kiddinn, line 154: what does that mean? is there something you can say here?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+                .format(exception))
+
+
+        if sigma_rule is None:
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND, 'No sigma was parsed')
+        # TODO: check and adjust tests as now meta is given back
```
kiddinn, line 161: remove this and fix that ;)

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+        return_code = HTTP_STATUS_CODE_OK
+        metadata = {'parsed': True}
```
kiddinn, line 163: no other metdata that you want to supply?

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +95,73 @@ def get(self, rule_uuid):
...
+
+        # TODO: rewrite that to use jsonify() instead
+        return self.to_json(
```
kiddinn: yes, use `jsonify` here as we talked about, with `meta` and `objects`

kiddinn (commented):
few more comments

timesketch/api/v1/resources/sigma.py
jaegeral: now there are

timesketch/api/v1/resources/sigma.py
jaegeral: ack

timesketch/api/v1/resources/sigma.py
jaegeral: yes

timesketch/api/v1/resources/sigma.py
jaegeral: ack, made it more accurate

timesketch/api/v1/resources/sigma.py
jaegeral: yes and also added logger outputs to the others.

timesketch/api/v1/resources/sigma.py
jaegeral: yeah a non valid yml file has been provided, clarified the error message

timesketch/api/v1/resources/sigma.py
jaegeral: no cannot think of any useful meta data here

timesketch/api/v1/resources/sigma.py
jaegeral: done

timesketch/api/v1/resources/sigma.py
jaegeral: better, done

timesketch/api/v1/resources/sigma.py
jaegeral: well that is kind of the else case where it is really unclear what was wrong... all the other cases are more granular, this one is the else tree of things

timesketch/api/v1/resources/sigma.py
jaegeral: ack, well I would still keep the action in here, as who knows if we have more later on.

timesketch/api/v1/resources/sigma.py
jaegeral: done

api_client/python/timesketch_api_client/sigma.py
jaegeral: I think that should now be done

api_client/python/timesketch_api_client/sigma.py
jaegeral: IMHo it should now be much better, WDYT?

timesketch/api/v1/resources/sigma.py
jaegeral: done

kiddinn (changes requested):
lets start by fixing the unit tests...


```
======================================================================
FAIL: Testing the Sigma rules list
  File "/home/runner/work/timesketch/timesketch/api_client/python/timesketch_api_client/sigma_test.py", line 100, in test_sigma_rules
    self.assertIn('Alexander', rule.author)
AssertionError: 'Alexander' not found in ''
```

jaegeral:
> lets start by fixing the unit tests...
> 
> ```
> ======================================================================
> FAIL: Testing the Sigma rules list
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/work/timesketch/timesketch/api_client/python/timesketch_api_client/sigma_test.py", line 100, in test_sigma_rules
>     self.assertIn('Alexander', rule.author)
> AssertionError: 'Alexander' not found in ''
> ```

should be fixed: ```------------------------------------------------------------------------------------
TOTAL                                                            11801   6442    45%
Ran 291 tests in 101.022s

OK (SKIP=90)```

api_client/python/timesketch_api_client/client.py
```diff
@@ -550,16 +550,12 @@ def list_sigma_rules(self, as_pandas=False):
...
-            index_obj = sigma.Sigma(
-                rule_uuid, api=self, es_query=es_query, file_name=file_name,
-                title=title, description=description,
-                file_relpath=file_relpath)
+            if not rule_dict:
+                raise ValueError('No rules found.')
+
+            index_obj = sigma.Sigma(api=self)
+            for key, value in rule_dict.items():
+                index_obj.data[key] = value
```
kiddinn, line 558: I would rather have this as 

```python
index_obj.set_value(key, value)
```

with the 

```python
def set_value(self, key, value):
  self._data[key] = value
```

or something like that.... or have this as 

```python
rule_uuid = rule_dict.get(....
....
index_obj = sigma.Sigma(api=self)
index_obj.from_manual(
  rule_uuid=rule_uuid, title=title, ....)
```

api_client/python/timesketch_api_client/client.py
```diff
@@ -573,4 +569,31 @@ def get_sigma_rule(self, rule_uuid):
...
             Instance of a Sigma object.
         """
-        return sigma.Sigma(rule_uuid, api=self)
+        sigma_obj = sigma.Sigma(api=self)
+        sigma_obj.from_rule_uuid(rule_uuid)
+
+        return sigma_obj
+
+    def get_sigma_rule_by_text(self, rule_text):
+        """Returns a Sigma Object based on a sigma rule.
```
kiddinn, line 578: based on a sigma rule or a sigma rule text?

api_client/python/timesketch_api_client/client.py
```diff
@@ -573,4 +569,31 @@ def get_sigma_rule(self, rule_uuid):
...
+            ValueError: No Rule text given or issues parsing it.
+        """
+        if not rule_text:
+            raise ValueError('No rule text given.')
+
+        try:
+            sigma_obj = sigma.Sigma(api=self)
+            sigma_obj.from_text(rule_text)
+        except ValueError:
+            logger.error('Parsing Error, unable to parse the Sigma rule',
```
kiddinn, line 596: super minor nit... do this rather

```python
logger.error(
    'Parsing error, ....', exc_info=True)
```

api_client/python/timesketch_api_client/client.py
kiddinn: also you don't even create `.data` in the object and you don't put that as an attribute... also data would overwrite the `resource.data` which is a property that returns `self.lazyload_data()`

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
 
         """
-        self.rule_uuid = rule_uuid
-        self._description = description
-        self._es_query = es_query
-        self._file_name = file_name
-        self._title = title
-        self._file_relpath = file_relpath
-        self._resource_uri = f'sigma/{self.rule_uuid}'
+        self.api = api
```
kiddinn, line 42: no need to set the `self.api`, since that is done by the parent

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
+        self._falsepositives = None
+        self._file_name = None
+        self._file_relpath = None
+        self._id = None
+        self._level = None
+        self._logsource = None
+        self._modified = None
+        self._rule_uuid = None
+        self._references = None
+        self._resource_uri = 'sigma/' # TODO: clarify: is that okay?
```
kiddinn, line 57: the parent sets `self.resource_uri` so don't be setting the `self._res...` (this should be already removed from all other places that were accidentally setting it)

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
+        self._title = None
+
         super().__init__(
             api=api, resource_uri=self._resource_uri)
 
     @property
     def es_query(self):
         """Returns the elastic search query."""
```
kiddinn, line 65: s/elastic/Elastic/gc

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
-        sigma_data = self.data
+        if self._es_query:
+            return self._es_query
 
-        if not sigma_data:
-            return ''
+        if not self.data or self.data is None:
+            self.lazyload_data(refresh_cache=True)
```
kiddinn, line 70: `self.data` calls `self.lazyload_data`, since that is set by the base resource....

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
 
-        return sigma_data.get('es_query', '')
+        try:
+            self._es_query = self.data.get('es_query', '')
```
kiddinn, line 73: again, let's not overwrite the propery `self.data` and do this as I said before....

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
+        except AttributeError: # in case self.data is Nonetype
```
kiddinn, line 74: this is weird....

if you are going to be using something like `self._data` or `self._resource` or whatever the name is (self._data might be confusing from self.data which is different) then you set it to an empty dict in the init, and then you can use the `set_value` function to set values in this, or individual `set_file_path` to specifically just set the `self._filepath`, etc, etc (or the `from_manual` function)

kiddinn: in line with other API client objects, they either define a specific `set_X` where X is a specific value or a variable or `from_manual` which accepts all the parameters

kiddinn: you can even do `from_manual` and have it accept the parameter of a dict, and then use that dict to populate the values...

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
 
     @property
     def id(self):
         """Returns the sigma rule id."""
+        if self._id:
+            return self._id
         sigma_data = self.data
 
         if not sigma_data:
+            self.lazyload_data()
```
kiddinn, line 101: again, `sigma.data` calls `lazyload_data`

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -31,67 +32,267 @@ class Sigma(resource.BaseResource):
...
+        response = self.api.session.post(self.resource_uri, data=data)
+        response_dict = error.get_response_json(response, logger)
+
+        objects = response_dict.get('objects')
+
+        if not objects:
+            logger.warning(
+                'Unable to parse rule with given text')
+
+        rule_dict = objects[0]
```
kiddinn, line 283: I can see here a list of all these...

so you can re-use this in the `def from_manual` if that function simply accepts a dict... then you can do something like

```python
def _load_rule_dict(self, rule_dict):
   """.....
   """
   self._author = rule_dict.get('....
    ....
```

And then here you can simply call

```python
rule_dict = objects[0]
self._load_rule_dict(rule_dict)
````

and in the `from_manual`

```python
def from_manual(self, rule_dict):
  """"docstring....""""
  self._load_rule_dict(rule_dict)
```

or something like that....


also in here... you do `objects = response_dict.get('objects')` but you never check if `objects` exists, adn to verify it is a list that has at least one element....  a simple

```python
if objects:
  rule_dict = objects[0]
  self._load_rule_dict(rule_dict)
```

would at least be better... or
```python
if not objects:
  # throw an error, and/or log an error
rule_dict = objects[0]
self._load_rule_dict(rule_dict)
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +94,77 @@ def get(self, rule_uuid):
...
+        except ValueError:
+            logger.error('Sigma Parsing error with the user provided rule',
+                         exc_info=True)
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Error unable to parse the provided Sigma rule')
+
+        except NotImplementedError as exception:
+            logger.error(
+                'Sigma Parsing error: Feature in the rule provided is not implemented in this backend', exc_info=True) # pylint: disable=line-too-long
```
kiddinn, line 143: why disable the line too long, wh not just fix that????

```python
logger.error(
    'Sigma parsing error: Feature in the rule provided '
    'is not implemented in this backend.', exc_info=True)
```

and then you don't need to have a too long of a line and don't need to suppress the error message from the linter

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +94,77 @@ def get(self, rule_uuid):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST, 'Error generating rule {0!s}'
```
kiddinn, line 145: nit here...

better to have this as:

```python
abort(
    HTTP_STATUS_CODE_BAD_REQUEST, 
    'Error generating rule {0!s}'.format(exception))
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +94,77 @@ def get(self, rule_uuid):
...
+                .format(exception))
+
+        except sigma_exceptions.SigmaParseError as exception:
+            logger.error(
+                'Sigma Parsing error: unknown error', exc_info=True)
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Sigma parsing error generating rule  with error: {0!s}'
+                .format(exception))
```
kiddinn, line 154: same nit as above

timesketch/api/v1/resources/sigma.py
```diff
@@ -89,4 +94,77 @@ def get(self, rule_uuid):
...
+
+        except yaml.parser.ParserError as exception:
+            logger.error(
+                'Sigma Parsing error: an invalid yml file has been provided',
+                exc_info=True)
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Sigma parsing error: invalid yaml provided {0!s}'
+                .format(exception))
```
kiddinn, line 163: easier to read when the .format is with the string, so 

```python
'Sigma parsing error.....'.format(
    exception))
```

timesketch/lib/sigma_util.py
kiddinn: well, since you use safe load, then you are only parsing standard YAML tags and no arbitrary python objects, so that should suffice in this case.

timesketch/lib/sigma_util.py
```diff
@@ -243,3 +247,66 @@ def get_sigma_rule(filepath, sigma_config=None):
...
+            rule_text, sigma_config, None)
+        parsed_sigma_rules = parser.generate(sigma_backend)
+        rule_yaml_data = yaml.safe_load_all(rule_text)
+        for doc in rule_yaml_data:
+            rule_return.update(doc)
+
+    except NotImplementedError as exception:
+        logger.error(
+            'Error generating rule {0!s}'
+            .format(exception))
```
kiddinn, line 285: nit here nad elsewehere, for consistency in the codebase, keep the `.format` inline with the actual string..

```python
'Error generating rule {0!}'.format(
    exception))
```

that is rather break at the parenthesis instead of at the string end in the case of the format string

kiddinn (changes requested):
few more comments here

api_client/python/timesketch_api_client/client.py
jaegeral: will go with the set_value as the numbers of params can change, so this seems to be the most generic way to do it then.

api_client/python/timesketch_api_client/sigma.py
jaegeral: ACK, removed the line

jaegeral: But to clarify, I still need to set 
```
resource_uri = 'sigma/'
``` 
to able to do: 
```
        super().__init__(
            api=api, resource_uri=resource_uri)
``` correct?

File: timesketch/api/v1/resources/sigma.py (hunk 54 above)
Comment (jaegeral) [reply timesketch/api/v1/resources/sigma.py:143]: TIL

File: timesketch/api/v1/resources/sigma.py (hunk 55 above)
Comment (jaegeral) [reply timesketch/api/v1/resources/sigma.py:146]: done

File: timesketch/lib/sigma_util.py (hunk 58 above)
Comment (jaegeral) [reply timesketch/lib/sigma_util.py:285]: it actually fitted in one line

File: api_client/python/timesketch_api_client/sigma.py (hunk 53 above)
Comment (jaegeral) [reply api_client/python/timesketch_api_client/sigma.py:275]: Well I check just above, but I do not exit here, corect.

Comment (jaegeral) [reply api_client/python/timesketch_api_client/sigma.py:216]: ok, should be better

File: api_client/python/timesketch_api_client/sigma.py [hunk 59]
```diff
@@ -279,20 +211,8 @@ def from_text(self, rule_text):
...
-        self._file_name = rule_dict.get('file_name', '')
-        self._id = rule_dict.get('id', '')
-        self._level = rule_dict.get('level', '')
-        self._logsource = rule_dict.get('logsource', '')
-        self._modified = rule_dict.get('modified', '')
-        self._references = rule_dict.get('references', '')
-        self._rule_uuid = rule_dict.get('id', '')
-        self._title = rule_dict.get('title', '')
+        for key, value in rule_dict.items():
+            self.set_value(key,value)
```
kiddinn, line 218: space between, `key, value`

timesketch/lib/sigma_util.py
```diff
@@ -159,7 +159,8 @@ def get_all_sigma_rules():
 
 
 def get_sigma_rule(filepath, sigma_config=None):
-    """ Returns a JSON represenation for a rule
+    """ Returns a JSON representation for a rule
```
kiddinn: s/""" Returns/"""Returns/

kiddinn (commented):
ok, only tiny nits left... then we are good to go

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -39,214 +39,142 @@ def __init__(self, api):
...
+        return self._attr_dict.get(key, '')
 
-        try:
-            self._es_query = self.data.get('es_query', '')
-        except AttributeError: # in case self.data is Nonetype
-            self._es_query = ''
-        return self._es_query
+    @property
+    def es_query(self):
+        """Returns the ElasticSearch query."""
```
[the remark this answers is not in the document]
jaegeral: ah good point.

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -39,214 +39,142 @@ def __init__(self, api):
...
-        return sigma_data.get('references', '')
+    def set_value(self, key, value):
+        """Sets the value for a given key
 
+        Args:
+            key: key to set the value
+            value: value to set
+
+        """
+        self._attr_dict[key] = value
```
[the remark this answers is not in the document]
jaegeral: Yeah in that case it is fine to just overwrite anything, it is not expected to change the values after they have been parsed.

timesketch/lib/sigma_util.py
jaegeral: ACK, also solved it in other places.

kiddinn (approved):
LGTM

PR #1517, opened 2020-12-09
[docs] How to run a single test file

jaegeral:
There are multiple ways to run single test files, but at least one method for people getting started running a single test file within docker (so we know the environment is controlled and ready to be used)


kiddinn (approved):
LGTM

PR #1520, opened 2020-12-10
Home overview and graph fixes

berggren:
Screenshot 2020-12-10 at 14 13 08


kiddinn (changes requested):
ok, let's start by fixing the linter.. once the linter is finished, I see a lot of changes to the sketch resource...

Can you check to make sure you are not breaking anything in the API client? that is the sketch  API client

berggren:
> ok, let's start by fixing the linter.. once the linter is finished, I see a lot of changes to the sketch resource...
> 
> Can you check to make sure you are not breaking anything in the API client? that is the sketch API client

Fixed: We now fallback to the old behaviour, so the API client should work just like before.

timesketch/api/v1/resources/__init__.py
```diff
@@ -157,13 +157,29 @@ class ResourceMixin(object):
...
+    }
+
     sketch_fields = {
         'id': fields.Integer,
         'name': fields.String,
         'description': fields.String,
         'user': fields.Nested(user_fields),
         'timelines': fields.List(fields.Nested(timeline_fields)),
         'stories': fields.List(fields.Nested(story_fields)),
+        'graphs': fields.List(fields.Nested(graph_fields)),
```
kiddinn, line 182: do we not want to have a list of the cached graphs in here as well?

timesketch/api/v1/resources/sketch.py
```diff
@@ -65,28 +65,86 @@ def get(self):
...
         sketches = []
-        for sketch in filtered_sketches:
-            sketches.append({
+        return_sketches = []
+        num_hits = 0
+
+        if scope == 'recent':
+            views = View.query.filter_by(
+                user=current_user, name='').order_by(
+                View.updated_at.desc()).limit(3)
```
kiddinn, line 95: why limit with 3?

Where is this magical number derived from? And why is that not configurable?

timesketch/api/v1/resources/sketch.py
```diff
@@ -65,28 +65,86 @@ def get(self):
...
+            sketches = [view.sketch for view in views]
+            num_hits = len(sketches)
```
kiddinn, line 97: and why are we basing `recent` on just the last sketches that have saved searches added to it?

is there no other better mechanism to judge recency? (when was a timeline added, when was a story added, event labeled, event starred, comment added, etc, etc)... there are so many other `changes` to a sketch...

is this maybe better to change the API in various places to add an update to the `update` timestamp for the sketch if a change is made... that is in the view resource, we update the sketch if a new view is saved or updated... in the story api, we update the timestamp for the sketch if a story is updated/changed... etc, etc ... and then this check is just to search for last update time of sketches, top X (configurable)... that way you get a miuch more `accurate` representation of recency...

timesketch/api/v1/resources/sketch.py
```diff
@@ -65,28 +65,86 @@ def get(self):
...
+            pagination = filtered_sketches.paginate(page=page, per_page=20)
+            sketches = pagination.items
+            num_hits = pagination.total
+
+        for sketch in sketches:
+            # Last time a user did a query in the sketch, indicating activity.
+            try:
+                last_activity = View.query.filter_by(
+                    sketch=sketch, name='').order_by(
+                    View.updated_at.desc()).first().updated_at
```
kiddinn, line 127: and again, this is not a good indication of a last activity of a sketch... this only tells a very partial story... and if we woudl implement what I talked about in the other comment, then we would have a much more accurate picture, and you could just take the last updated timestamp of the sketch and make use of that...

kiddinn (changes requested):
Few comments

timesketch/api/v1/resources/sketch.py
berggren: This is a smaller UI window at the top, only with the recent sketches (sketces that you activly queried in). There are other paginated windows below in the UI. It can be made configurable, and I considered that but I don't want to add that now until we have tested the feature a bit

timesketch/api/v1/resources/sketch.py
berggren: There are more avenues yes, but not easily queried. This is not just saved searches. This is the special view that is updated anytime the user searches the sketch. So activity here means sketches that you actually did some work on (i.e. searching).

I'm about to add the suggestion that you point out here, but not in this PR :)  I'm aware that there are more actions to take into account, but this is the start of it and we can make it cover more if we think it is useful later.

timesketch/api/v1/resources/sketch.py
berggren: Yes, I agree. In a followup PR though.

berggren (commented):
PTAL

timesketch/api/v1/resources/__init__.py
berggren: No, we don't render them anywhere except in the component and that one is fetching it on demand.

timesketch/api/v1/resources/sketch.py
berggren: And this is the list that the user was active on, not based on what anyone else did on the sketch

timesketch/api/v1/resources/sketch.py
kiddinn: can you then at least add a comment or a TODO here to add an option to configure this number.

timesketch/api/v1/resources/sketch.py
kiddinn: ok, but can you add a TODO here to include other activity, since as you said, this only covers one use case of "activity", that is just the fact that the user searched something in the sketch.. but there are also other ways to interact with the sketch (run an analyzer, bla bla bla)

kiddinn: we might even include a "user visited the sketch" ?

kiddinn (commented):
ok, only minor nits ,adding TODOs

timesketch/api/v1/resources/sketch.py
berggren: done!

timesketch/api/v1/resources/sketch.py
berggren: Done!

berggren (commented):
PTAL

kiddinn (approved):

kiddinn (approved):
LGTM

PR #1521, opened 2020-12-11
Return full NX generated CY graph

berggren:
Instead of just the elements, return the full structure so it is possible to recreate the graph later on.


kiddinn (approved):
LGTM

PR #1529, opened 2020-12-15
[Sigma] improve evtx mapping

jaegeral:
Improve the mappings. Based on https://github.com/Neo23x0/sigma/tree/master/rules/windows/builtin


kiddinn (approved):
LGTM

Issue #1489, opened 2020-11-23
Improve CSV import worker


berggren:
**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

kiddinn:
this is mostly talking about moving away from the `csv` module to something like `pandas.read_csv`

kiddinn:
This will be fixed when #1534 is submitted

PR #1534, opened 2020-12-16
Replace use of the csv library by pandas built-ins for data ingestion

rayanht:
Addresses #1489.

### Summary of changes:

- Extracted the header validation logic into a helper function `validate_csv_fields` to avoid duplicate code between `read_and_validate_csv` and `read_and_validate_redline`
- Replaced calls to `csv.DictReader` by `pandas.read_csv`
- Inverted the `try... except` in `read_and_validate_csv` since `pandas.read_csv` will catch parsing errors during the initial read.
- Updated the `read_and_validate_redline` docstring to list the RuntimeError that might be raised during validation.

### Unimplemented suggestions:

- The consecutive dictionary assignments in `read_and_validate_redline` can be replaced by a literal:

```
row_to_yield = {}
row_to_yield["message"] = summary
row_to_yield["timestamp"] = timestamp
row_to_yield["datetime"] = dt_iso_format
row_to_yield["timestamp_desc"] = timestamp_desc
row_to_yield["alert"] = alert  # Extra field
row_to_yield["tag"] = tags  # Extra field
```
Could simply become:
```
row_to_yield = {"message": summary, "timestamp": timestamp, "datetime": dt_iso_format,
                        "timestamp_desc": timestamp_desc, "alert": alert}
```

- There might be a way of getting rid of the last remaining use of the csv library (`csv.register_dialect(
        'redlineDialect', delimiter=',', quoting=csv.QUOTE_ALL,
        skipinitialspace=True)`) by using pandas built-ins. Need to investigate


rayanht:
For suggestion no.2 it seems like all the dialect parameters can be passed directly to `pandas.read_csv` by using keyword, without having to create a `csv.Dialect` object. The only issue with that approach is with the `quoting` argument which either takes an int or a `csv.QUOTE` object, meaning that we either have to use a magic number there or keep the `csv` dependency.

image
^ Taken from the [pandas documentation](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) of `read_csv`

berggren:
Assigning this review to @kiddinn

timesketch/lib/utils.py
```diff
@@ -83,6 +81,24 @@ def _scrub_special_tags(dict_obj):
...
+        Args:
+            mandatory_fields: a list of fields that must be present.
+            data: a DataFrame built from the ingested file.
+
+        Raises:
+            RuntimeError: if there are missing fields.
+    """
+    parsed_fields = data.columns
+    missing_fields = [field for field in mandatory_fields
+                      if field not in parsed_fields]
```
kiddinn, line 95: you can also do

```
mandatory_set = set(mandatory_fields)
parsed_set = set(data.columns)

if mandatory_set.issubset(parsed_set):
  return

raise RuntimeError('Miss.....', ','.join(list(mandatory_set.difference(parsed_set))))
```

timesketch/lib/utils.py
```diff
@@ -101,79 +117,59 @@ def read_and_validate_csv(file_handle, delimiter=','):
...
-            except ValueError:
-                continue
-
-            yield row
-    except csv.Error as e:
+        reader = pandas.read_csv(file_handle, sep=delimiter)
+    except pandas.errors.ParserError as e:
         error_string = 'Unable to read file, with error: {0!s}'.format(e)
         logger.error(error_string)
         raise errors.DataIngestionError(error_string)
```
kiddinn, line 125: can you do `raise er.... from e` to coincide with py3

timesketch/lib/utils.py
```diff
@@ -101,79 +117,59 @@ def read_and_validate_csv(file_handle, delimiter=','):
...
 
+    validate_csv_fields(mandatory_fields, reader)
+    for row in reader:
+        if not row:
+            continue
+        if not row['datetime']:
```
kiddinn, line 131: I would opt not to do this in the iterator... 

rather do the whole checks for datetime and other fields with the entire data frame, here you can do the checks and parsing on the entire frame instead of a single row.

timesketch/lib/utils.py
```diff
@@ -101,79 +117,59 @@ def read_and_validate_csv(file_handle, delimiter=','):
...
-                if 'tag' in row:
-                    row['tag'] = [x for x in _parse_tag_field(row['tag']) if x]
-
-                _scrub_special_tags(row)
-            except ValueError:
-                continue
-
-            yield row
-    except csv.Error as e:
+        reader = pandas.read_csv(file_handle, sep=delimiter)
```
kiddinn, line 121: I would opt not do it this way..I would rather use

```
try:
   for df_chunk in pd.read_csv(file_handle, sep=delimiter, chunksize=SOME_DEFAULT_VALUE):
      # Do all the validation, etc....
```

That is to read in the file in chunks, that way you are not reading in the entire file in memory at time, preventing us trying to read the entire content of a very large file, rather to split it up in chunks, we are doing a generator here, so it would be better

timesketch/lib/utils.py
```diff
@@ -101,79 +117,59 @@ def read_and_validate_csv(file_handle, delimiter=','):
...
+        if not row['datetime']:
+            logger.warning(
+                'Row missing a datetime object, skipping [{0:s}]'.format(
+                    ','.join([str(x).replace(
+                        '\n', '').strip() for x in row.values()])))
+            continue
+        try:
+            # Normalize datetime to ISO 8601 format if it's not the case.
+            parsed_datetime = parser.parse(row['datetime'])
+            row['datetime'] = parsed_datetime.isoformat()
```
kiddinn, line 140: we can and should probably just use `pd.to_datetime` to work on an entire column

timesketch/lib/utils.py
```diff
@@ -101,79 +117,59 @@ def read_and_validate_csv(file_handle, delimiter=','):
...
-    missing_fields = []
-    # Validate the CSV header
-    for field in mandatory_fields:
-        if field not in csv_header:
-            missing_fields.append(field)
-    if missing_fields:
-        raise RuntimeError(
-            'Missing fields in CSV header: {0:s}'.format(missing_fields))
+    reader = pandas.read_csv(file_handle, delimiter=',',
+                             dialect='redlineDialect')
```
kiddinn, line 170: do we need this special dialect here? this is just indicating quoting, delimiter, etc....

kiddinn (changes requested):
Few comments,b ut thank you for the contribution.

timesketch/lib/utils.py
rayanht: I've talked about this briefly in the PR. The issue is with the quoting parameter which either takes a csv.QUOTE or a magic number

kiddinn: ok, so let's keep the special dialect in here then

kiddinn (commented):
answer to one comment

kiddinn:
let me know when you are ready for another round of review

timesketch/lib/utils.py
```diff
@@ -83,97 +90,99 @@ def _scrub_special_tags(dict_obj):
...
-            # single lines as multiple, causing issues with importing.
-            # TODO: Swap the CSV library for the user of pandas.
-            if not row:
-                continue
-            if not row['datetime']:
+        reader = pandas.read_csv(file_handle, sep=delimiter,
+                                 chunksize=DEFAULT_CHUNK_SIZE)
+        for chunk in reader:
+            skipped_rows = chunk[chunk['datetime'].isnull()]
+            for row in skipped_rows.iterrows():
```
kiddinn, line 143: do we want to do that for all of these rows?

or just say

```
if not skipped_rows.empty:
  logger.warning(f'{len(skipped_rows)} rows skipped since they were missing a datetime field or it was empty')
```

I'm just wondering whether we need all of them to be logged here?

timesketch/lib/utils.py
```diff
@@ -83,97 +90,99 @@ def _scrub_special_tags(dict_obj):
...
-                normalized_timestamp += parsed_datetime.microsecond
-                row['timestamp'] = str(normalized_timestamp)
-                if 'tag' in row:
-                    row['tag'] = [x for x in _parse_tag_field(row['tag']) if x]
 
-                _scrub_special_tags(row)
-            except ValueError:
-                continue
+            # Normalize datetime to ISO 8601 format if it's not the case.
+            chunk['datetime'] = pandas.to_datetime(chunk['datetime'])
```
kiddinn, line 150: maybe better to wrap this around a try/except...

you for instance get `ParserError` if  the string is not parse-able.....

timesketch/lib/utils.py
```diff
@@ -83,97 +90,99 @@ def _scrub_special_tags(dict_obj):
...
 
-            yield row
-    except csv.Error as e:
+            chunk['timestamp'] = chunk['datetime'].apply(_get_unix_time)
```
kiddinn, line 152: you can also just do


```
chunk['timestamp'] = chunk['datetime'].dt.strftime('%s%f').astype(int)
```

and then you don't need an apply function

timesketch/lib/utils.py
```diff
@@ -83,97 +90,99 @@ def _scrub_special_tags(dict_obj):
...
+            chunk['datetime'] = chunk['datetime'].apply(pandas.Timestamp
```
kiddinn, line 153: you can also just do 


```
chunk['datetime'] = chunk['datetime'].dt.strftime('%Y-%m-%dT%H:%M:%S%z')

File: timesketch/lib/utils.py [hunk 11]
```diff
@@ -83,97 +90,99 @@ def _scrub_special_tags(dict_obj):
...
+                                                        .isoformat).astype(str)
+
+            if 'tag' in chunk:
+                chunk['tag'] = chunk['tag'].apply(
+                    lambda tag: [x for x in _parse_tag_field(tag) if x]
+                    if tag else None)
```
kiddinn, line 159: I feel like tiuo don't need to do this, since _parse_tag_field accepts a list, so are you sure you can't just do

```
chunk['tag'] = chunk['tag'].apply(_parse_tag_field)
```

timesketch/lib/utils.py
```diff
@@ -189,9 +198,9 @@ def read_and_validate_redline(file_handle):
         row_to_yield["timestamp"] = timestamp
         row_to_yield["datetime"] = dt_iso_format
         row_to_yield["timestamp_desc"] = timestamp_desc
-        row_to_yield["alert"] = alert #extra field
+        row_to_yield["alert"] = alert  # Extra field
```
kiddinn, line 201: I know thi sis not from this this PR, but can you change all these double quotes to a single one (" vs ') to be consistent with the rest of the codebase

kiddinn (changes requested):
few more comments

kiddinn:
any updates?

rayanht:
> any updates?

Hi, sorry I've been super busy with some uni coursework 😅 Will implement the suggested changes today.

timesketch/lib/utils.py
rayanht: re: `chunk['datetime']`. Isn't it a bit clearer for future maintainers if we use the `Timestamp.isoformat` rather than the magic formatting string? Another concern is that `%z` seem to be [platform-dependent](https://stackoverflow.com/questions/32353015/python-time-strftime-z-is-always-zero-instead-of-timezone-offset).

kiddinn: yes, I'm a fan of using isoformat, and yes I don't know why I added the `%z`, old habit, but I agree

kiddinn (commented):
letme know when ready for another round of review

timesketch/lib/utils.py
rayanht: This already gets caught by the outer try/except, doesn't it?

rayanht: I guess my question boils down to, do we want to handle parsing errors related to dates specifically any differently than parsing errors related to the CSV format as a whole? What's the expected behaviour when a CSV is valid but some row has invalid date? Do we want to cancel ingestion as a whole or do we just want to skip that row? @kiddinn

kiddinn: I don't want to cancel ingestion as a whole, I prefer to skip that row, or at worst skip that chunk, but I also want to know that a row got skipped, we need to know that an error occurred, why and how many, etc.

kiddinn (commented):
.

kiddinn:
what's the status of this?

There is also a conflict that must be resolved when syncing the branch....

rayanht:
@kiddinn I've just pushed the change that catches ValueErrors raised by `pandas.to_datetime`. The behaviour is that the whole chunk gets skipped if we encounter a malformed date and a warning log is emitted to inform the user. The log is pretty generic as of right now, should we maybe include in the log that the chunk that got skipped is the one that spans from row n to row m?

kiddinn:
@rayanht yes, we should include what rows were skipped, as much as we can provide we should do that. Also what the issue was, we need to provide the user as much details as we can here.

kiddinn (approved):
I made minor stylistic changes to the PR, otherwise LGTM

kiddinn:
I might wait with the merging of this PR until Monday, we will need to do some more testing, since this is a pretty significant change.

Thank you though for the PR, much appreciated. And will as I said most likely get merged on Monday after brief testing.

kiddinn:
We will merge this PR in as soon as we make a release, that should happen any time now.

Since this is a large change in how we ingest files, we don't want to do that at the same time as we are changing how we index data. So we opted to wait until after the release to merge this in, giving us some more time to test and tweak the ingestion to make sure it works.

PR #1536, opened 2020-12-16
Bugfix: Set default per_page for pagination

berggren:
Bugfix to set per_page in pagination. Otherwise the UI wont render pagination controls.
Also some smaller cleanups.


kiddinn (approved):
LGTM

PR #1538, opened 2020-12-17
Added last_activity of a sketch to the sketch meta section and into the API client

kiddinn:
This PR adds the last_activity of a sketch as a time string and makes it available in the API Client.

Other things it does is to:

1. Move all use of the special "view" object for each user to common functions inside utils.
2. Add updating the user view in many other places to make sure that `last_activity` is not just bound to exploring a sketch


timesketch/api/v1/resources/sketch.py
```diff
@@ -235,6 +235,16 @@ def _remove_label(self, sketch, label):
         sketch.remove_label(label)
         return True
 
+    def _get_last_activity(self, sketch):
```
berggren, line 238: Can you break this out to a utility function so we can reuse here as well as in the SketchListResource? Now we have this duplicated.

If not in this PR, then add a TODO :)

berggren (changes requested):

timesketch/api/v1/resources/sketch.py
kiddinn: I was just thinking about that, that we should rather do that, same with the updating of the timestamp, so that we can more easily re-use that in other resources, so let's do that now...

kiddinn (commented):
back to you

timesketch/api/v1/resources/explore.py
```diff
@@ -197,12 +197,10 @@ def post(self, sketch_id):
 
         # Update or create user state view. This is used in the UI to let
         # the user get back to the last state in the explore view.
-        view = View.get_or_create(
-            user=current_user, sketch=sketch, name='')
+        view = utils.update_sketch_last_activity(sketch, get_view_back=True)
```
berggren, line 200: I think it is better to remove this, and add back `view = View.get_or_create(...)`

**Reason:**
When we change the implementation of `update_sketch_last_activity` and move away from using the `View` for this this will break. Also seeing that this is the only place this is used.

So, move back to:
`view = View.get_or_create()` and `view.update_modification_time()`

timesketch/api/v1/utils.py
```diff
@@ -73,6 +76,41 @@ def get_sketch_attributes(sketch):
...
+    view = View.get_or_create(
+        user=current_user, sketch=sketch, name='')
+    view.update_modification_time()
+
+    if get_view_back:
+        return view
+
+    db_session.add(view)
+    db_session.commit()
+    return None
```
berggren, line 111:
```suggestion
def update_sketch_last_activity(sketch):
    """Update the last activity date of a sketch.

    Args:
        sketch: a sketch object (sketch.Sketch).
    """
    view = View.get_or_create(
        user=current_user, sketch=sketch, name='')
    view.update_modification_time()

    db_session.add(view)
    db_session.commit()
```

berggren (changes requested):

timesketch/api/v1/resources/explore.py
kiddinn: ack

kiddinn (commented):
done

berggren (approved):

PR #1539, opened 2020-12-17
Bugfix: Wrong count number in timeline list

berggren:
We were getting the wrong number (all shards together).


timesketch/lib/datastores/elastic.py
```diff
@@ -594,18 +594,26 @@ def count(self, indices):
...
+        except NotFoundError:
             es_logger.error(
                 'Unable to count indexes (index not found)',
                 exc_info=True)
-            return 0
-        return result.get('count', 0)
+            es_stats = {}
+
+        doc_count_total = es_stats.get(
+            '_all', {}).get('primaries').get('docs', {}).get('count', 0)
```
kiddinn, line 612: maybe have {} as the default in other places too

.get('primaries', {}).get('docs....

and .get('primaries', {}).get('store', {}).get..... below

kiddinn (changes requested):
CAn you start by fixing all the tests and the linter/

timesketch/lib/datastores/elastic.py
berggren: good catch, fixed.

kiddinn (approved):
LGTM

PR #1544, opened 2020-12-23
Fix remove_timeline_label() permission check

### LINKED ISSUE #1543: Check for proper permission on timeline
In api/v1/resources/timeline.py, remove_timeline_label() checks that the user has permission "delete" on the sketch this timeline is added to:
https://github.com/google/timesketch/blob/b5776d5da1586ea6824e89fd514418a35f4d1908/timesketch/api/v1/resources/timeline.py#L270

This mean a user with "write" pemission on sketch (but not "delete") can add timelines to a sketch  (they need 'write' permissions on the sketch), then add labels to it (currently, no permission check), but not remove the label (which, in turn, might prevent the timeline from being deleted further, if said label is 'noarchive' for example).

rgayon:
Fix #1543


timesketch/api/v1/resources/timeline.py
```diff
@@ -268,7 +268,7 @@ def post(self, sketch_id, timeline_id):
                 changed = any(changes)
             elif label_action == 'remove':
                 if not sketch.has_permission(
-                        user=current_user, permission='delete'):
+                        user=current_user, permission='write'):
```
kiddinn, line 271: well, technically you can just remove this check if you want , since there is a check on line 232 to check for write permissions, so you will not reach this point without write permissions on the sketch.

but I agree, maybe it's best that if yiou allow adding a label with write, then you should be able to remove it if you've got write permissions...


but the other question is...  should the ACL be slightly different

a) if this is a label that you added yourself, then you can remove it using write permissions...
b) if this is a label added by system/another user then you need delete permissions to be able to remove it?

That is we don't have any check here for determining the label user... a label does have user information attached to it, so it can be easily checked.

kiddinn (commented):
one comment

timesketch/api/v1/resources/timeline.py
rgayon: I don't think there is a need for tracking down "per user" permissions for labels. I can't really think of a use case where you'd want people to have write access on timelines and not write access on labels.

Agreed that I can actually just remove the check here

kiddinn (approved):
LGTM

PR #1552, opened 2021-01-07
Move stories to metadata

berggren:
Stories include aggregation data, and this can get a bit verbose on larger sketches. This PR removes the `content` part of the story listing.


kiddinn (approved):
LGTM

PR #1555, opened 2021-01-11
Misc UI fixes

berggren:
This PR fixes one UI bug and one performance optimazation

* Render dropdown that falls out of the screen
* Fixes dropdowns disappearing in story view
* Don't fetch sketch resource on every render of timeline list


kiddinn (approved):
LGTM

PR #1565, opened 2021-01-26
First phase of new way of storing data in TS.

kiddinn:
The first part of #1567

This PR is the first step into changing how newly uploaded data is indexed in Timesketch.

Instead of each new file that is uploaded to have it's own search index object and a timeline there is now only a single (or at most few) actual search indexes created for each sketch, and then all files are uploaded to that index.

This changes quite a few things internally with TS, since indexes are used in the UI to distinguish timelines, it is also used for filtering and other things.

This PR also does not include a UI element to it, making some UI elements potentially not work as well as they should.

There is also a missing piece of plaso file handling. That is, in order for this to work properly a change needs to happen in plaso (psort module) to accommodate the changes here.

Additionally this PR adds in the feature that in search if you add `comment` to the list of return fields the comment is extracted from the database and added to the output.

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -885,13 +885,8 @@ class Sketch(resource.BaseResource):
         searches = []
         for saved_search in sketch['meta'].get('views', []):
             search_obj = search.Search(sketch=self)
-            try:
-                search_obj.from_saved(saved_search.get('id'))
-                searches.append(search_obj)
-            except ValueError:
-                logger.error(
-                    'Unable to load a saved search with ID: {0:d}'.format(
-                        saved_search.get('id',0)), exc_info=True)
+            search_obj.from_saved(saved_search.get('id'))
+            searches.append(search_obj)
 
         return searches
 
@@ -935,10 +930,9 @@ class Sketch(resource.BaseResource):
         # TODO: Deprecate this function.
         logger.warning(
             'This function is about to be deprecated, please use the '
-            'timesketch_import_client instead (this function is not '
-            'guaranteed to work)')
+            'timesketch_import_client instead')
 
-        resource_url = f'{self.api.api_root}/upload/'
+        resource_url = '{0:s}/upload/'.format(self.api.api_root)
         files = {'file': open(file_path, 'rb')}
         _, _, file_ending = file_path.rpartition('.')
         data = {
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -930,12 +930,18 @@ class Sketch(resource.BaseResource):
         # TODO: Deprecate this function.
         logger.warning(
             'This function is about to be deprecated, please use the '
-            'timesketch_import_client instead')
+            'timesketch_import_client instead (this function is not '
+            'guaranteed to work)')
 
-        resource_url = '{0:s}/upload/'.format(self.api.api_root)
+        resource_url = f'{self.api.api_root}/upload/'
         files = {'file': open(file_path, 'rb')}
-        data = {'name': timeline_name, 'sketch_id': self.id,
-                'index_name': index}
+        _, _, file_ending = file_path.rpartition('.')
+        data = {
+            'name': timeline_name,
+            'sketch_id': self.id,
+            'label': file_ending,
+            'index_name': index}
+
         response = self.api.session.post(resource_url, files=files, data=data)
         response_dict = error.get_response_json(response, logger)
         timeline_dict = response_dict['objects'][0]
@@ -945,6 +951,7 @@ class Sketch(resource.BaseResource):
             api=self.api,
             name=timeline_dict['name'],
             searchindex=timeline_dict['searchindex']['index_name'])
+
         return timeline_obj
 
     def add_timeline(self, searchindex):
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -885,8 +885,13 @@ class Sketch(resource.BaseResource):
         searches = []
         for saved_search in sketch['meta'].get('views', []):
             search_obj = search.Search(sketch=self)
-            search_obj.from_saved(saved_search.get('id'))
-            searches.append(search_obj)
+            try:
+                search_obj.from_saved(saved_search.get('id'))
+                searches.append(search_obj)
+            except ValueError:
+                logger.error(
+                    'Unable to load a saved search with ID: {0:d}'.format(
+                        saved_search.get('id',0)), exc_info=True)
 
         return searches
 
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -530,8 +530,8 @@ class Sketch(resource.BaseResource):
         Returns:
             A boolean indicating whether the ACL change was successful.
         """
-        if not user_list and not group_list:
-            return True
+        if not user_list and not group_list and not make_public:
+            return False
 
         resource_url = '{0:s}/sketches/{1:d}/collaborators/'.format(
             self.api.api_root, self.id)
--- a/importer_client/python/timesketch_import_client/importer.py
+++ b/importer_client/python/timesketch_import_client/importer.py
@@ -59,6 +59,7 @@ class ImportStreamer(object):
         self._datetime_field = None
         self._format_string = None
         self._index = uuid.uuid4().hex
+        self._label = ''
         self._last_response = None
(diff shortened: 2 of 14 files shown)


kiddinn:
This is the first part of #1567

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -940,8 +940,13 @@ def upload(self, timeline_name, file_path, index=None):
 
         resource_url = f'{self.api.api_root}/upload/'
         files = {'file': open(file_path, 'rb')}
-        data = {'name': timeline_name, 'sketch_id': self.id,
-                'index_name': index}
+        _, _, file_ending = file_path.rpartition('.')
+        data = {
+            'name': timeline_name,
+            'sketch_id': self.id,
+            'label': file_ending,
```
berggren, line 947: Is this the data_source label? Like plaso/csv?

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -490,12 +494,14 @@ def add_excel_file(self, filepath, **kwargs):
...
         self.add_data_frame(data_frame)
 
-    def add_file(self, filepath, delimiter=','):
+    def add_file(self, filepath, delimiter=',', label=''):
         """Add a CSV, JSONL or a PLASO file to the buffer.
 
         Args:
             filepath: the path to the file to add.
             delimiter: if this is a CSV file then a delimiter can be defined.
+            label: optional label of the timeline, used for determining whether
```
berggren, line 503: Ok, yeah this is the data type label. I think I would like to call this something else.. We use `label` for many things and this could potentially be confusing? Let me think about it a bit more

timesketch/api/v1/resources/explore.py
```diff
@@ -204,6 +214,15 @@ def post(self, sketch_id):
         except KeyError:
             pass
 
+        comments = {}
```
berggren, line 217: Why are you fetching comments here? We do that in a lazy load in the client call based on the __ts_comment special label. Or am I missing some context here?

timesketch/api/v1/resources/upload.py
```diff
@@ -40,9 +41,75 @@
...
+                used or a new one created.
+            extension: optional file extension if a file is being uploaded,
+                if supplied and no label used, then the extension will be
+                used as a label.
+
+        Returns:
+            A SearchIndex object.
+        """
+        if index_name:
+            if not isinstance(index_name, six.text_type):
```
berggren, line 66: Do we still need to use six? We are python3 only

timesketch/api/v1/resources/upload.py
```diff
@@ -40,9 +41,75 @@
...
+            label='', extension=''):
+        """Returns a SearchIndex object to be used for uploads.
+
+        Args:
+            name: the name of the searchindex.
+            description: the description of the searchindex.
+            sketch: sketch object (instance of Sketch).
+            index_name: optional index name, if supplied and if it exists
+                then the index associated with that will be returned.
+            label: optional label of the data, if supplied will be used to
```
berggren, line 55: Yeah, see earlier comment. Maybe call this something like `source_type` or `data_label`? Just want to make sure it is clear what this is.

timesketch/api/v1/resources/upload.py
```diff
@@ -40,9 +41,75 @@
...
+
+            if searchindex and searchindex.has_permission(
+                    permission='write', user=current_user):
+                return searchindex
+
+        if extension and not label:
+            label = extension
+
+        if not label:
+            label = 'log'
```
berggren, line 80: maybe `generic_log`?

timesketch/api/v1/resources/upload.py
```diff
@@ -40,9 +41,75 @@
...
+
+        timelines = Timeline.query.filter_by(
```
berggren, line 82: use sketch.active_timelines here to filter out processing, failed and archived timelines.

timesketch/api/v1/resources/upload.py
```diff
@@ -40,9 +41,75 @@
...
+
+        searchindex.grant_permission(permission='read', user=current_user)
+        searchindex.grant_permission(permission='write', user=current_user)
+        searchindex.grant_permission(permission='delete', user=current_user)
+        searchindex.set_status('processing')
+
+        db_session.add(searchindex)
+        db_session.commit()
+
+        searchindex.add_label(label, user=current_user)
```
berggren, line 106: Is this label used anywhere?

berggren (changes requested):
First round. I'll take another look tomorrow morning

timesketch/api/v1/resources/explore.py
kiddinn: yes, the comment is fetched in a lazy load in the client base on each event, which works fine in the uI, but not in the API client. 

This is to make sure that if you add the `comment` field into the return values, and only then, the comment is fetched, which means you can add it into the return values of the API client.

timesketch/api/v1/resources/upload.py
kiddinn: no... this is old remnants and has no place here anymore ;)

timesketch/api/v1/resources/upload.py
kiddinn: or just generic?

timesketch/api/v1/resources/upload.py
kiddinn: ack

timesketch/api/v1/resources/upload.py
kiddinn: yes... in the code just above... this is how one can figure out whether a `data_label` or whatever we choose to call it matches with another....

timesketch/lib/datastores/elastic.py
```diff
@@ -254,7 +258,44 @@ def build_query(self, sketch_id, query_string, query_filter, query_dsl=None,
                 }
             }]
 
-        if query_string:
+        if timeline_ids and isinstance(timeline_ids, (list, tuple)):
```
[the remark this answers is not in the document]
kiddinn: take as long a look as you can and want

kiddinn (commented):
answered some of the comments

importer_client/python/timesketch_import_client/importer.py
kiddinn: renamed to data_label

api_client/python/timesketch_api_client/sketch.py
kiddinn: data_label

timesketch/api/v1/resources/upload.py
kiddinn: data_label

kiddinn (commented):
done with this round, PTAL

timesketch/api/v1/resources/explore.py
berggren: ok, as discussed offline - this makes sense.

timesketch/api/v1/resources/upload.py
berggren: yeah, generic SGTM

timesketch/api/v1/resources/upload.py
berggren: saw that, 👍

timesketch/lib/datastores/elastic.py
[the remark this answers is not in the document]
berggren: Ok, as dissussed offline - please add a comment explaining the logic here, and the reason for the bool query

berggren (changes requested):

timesketch/api/v1/resources/upload.py
```diff
@@ -62,74 +126,55 @@ def _upload_and_index(
...
+            extension=file_extension)
+        searchindex.set_status('processing')
+
+        timeline = Timeline.get_or_create(
             name=timeline_name,
+            description=timeline_name,
+            sketch=sketch,
             user=current_user,
-            index_name=index_name).first()
+            searchindex=searchindex)
```
berggren, line 143: This means that if the timeline description has changed, a new timeline will be created. And also, if I try to append to an existing timeline that someone else created then a new timeline with the same name will be created but with the current user as creator.

Should we do something like this instead?

```
timeline = Timeline.query.filter_by(name=timeline_name,sketch=sketch,searchindex=searchindex).first()
```

WDYT?

berggren (changes requested):

timesketch/api/v1/resources/upload.py
kiddinn: better?

berggren (approved):
LGTM

PR #1566, opened 2021-01-27
Minor API client changes

kiddinn:
This PR introduces the following API client changes:

1. Adds an `expected_size` property to the search object, so that one can find out how many entries one can expect back from a search query.
2. Added minor error handling into the `list_saved_searches` to account for malformed saved searches.

### DIFF
--- a/api_client/python/timesketch_api_client/search.py
+++ b/api_client/python/timesketch_api_client/search.py
@@ -434,13 +434,16 @@ class Search(resource.SketchResource):
 
             self.add_chip(chip)
 
-    def _execute_query(self, file_name=''):
+    def _execute_query(self, file_name='', count=False):
         """Execute a search request and store the results.
 
         Args:
             file_name (str): optional file path to a filename that
                 all the results will be saved to. If not provided
                 the results will be stored in the search object.
+            count (bool): optional boolean that determines whether
+                we want to execute the query or only count the
+                number of events that the query would produce.
         """
         query_filter = self.query_filter
         if not isinstance(query_filter, dict):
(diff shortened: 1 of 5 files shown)


api_client/python/timesketch_api_client/search.py
```diff
@@ -573,6 +574,28 @@ def description(self, description):
...
+        """Property that returns the expected size of the search query."""
+        if self._total_elastic_size:
+            return self._total_elastic_size
+
+        # To get the total size of the query we first need to send a query
+        # so to minimize the cost we reduce the size to 10 entries.
+        max_entries = self._max_entries
+        scrolling = self.scrolling
+
+        self._max_entries = 10
```
berggren, line 588: There is a better way of doing this. You can pass `count=True` to the search() call and just get back the count of matching documents. This is much cheaper as it doesn't have to retrieve anything.

Ref: https://github.com/google/timesketch/blob/master/timesketch/lib/datastores/elastic.py#L344

api_client/python/timesketch_api_client/search.py
```diff
@@ -573,6 +574,28 @@ def description(self, description):
         self._description = description
         self.commit()
 
+    @property
+    def expected_size(self):
```
berggren, line 578: Why do we need this in the first place?

berggren (changes requested):

api_client/python/timesketch_api_client/search.py
kiddinn: I was attempting to remove the need to make changes in the actual API... made the changes and am now using the `count`

api_client/python/timesketch_api_client/search.py
kiddinn: explained offline, for automation in order to be able to effectively set the max_entries size

kiddinn (commented):
done

berggren (approved):

PR #1569, opened 2021-01-29
Adding the ability to check whether a given index has __timeline_id in it.

kiddinn:
This PR adds the ability to check whether a given index has a `__timeline_id` field in it, which would indicate that the index was created using the newly established method of data upload.


timesketch/api/v1/resources/searchindex.py
```diff
@@ -112,7 +112,17 @@ def get(self, searchindex_id):
...
-        return self.to_json(searchindex)
+
+        mapping = self.datastore.client.indices.get_mapping(
+            searchindex.index_name)
+        fields = list(mapping.get(
+            searchindex.index_name, {}).get('mappings', {}).get(
+                'properties', {}).keys())
+
+        meta = {
+            'contains_timeline_id': bool('__timeline_id' in fields),
```
berggren, line 123:
```suggestion
            'contains_timeline_id': bool('__ts_timeline_id' in fields),
```

berggren (changes requested):

berggren (approved):

PR #1573, opened 2021-02-01
New storage - UI fix [phase one]

berggren:
This PR adds support for mixed new timelines (with compound indices) and old style indices.

It fixes:
* Timeline picker
* Timeline filter on URL param
* Eventlist timeline color


timesketch/frontend/src/views/SketchExplore.vue
```diff
@@ -361,7 +361,7 @@ const defaultQueryFilter = () => {
     'time_end': null,
```
kiddinn, line 361: can you remove `time_start` and `time_end` since this has been removed from teh default in other places (and no longer a thing int he `build_query`)

kiddinn (changes requested):
one comment, also, can you fix the PPA issue?

timesketch/frontend/src/views/SketchExplore.vue
berggren: Good catch, removed now.

kiddinn (approved):
LGTM

PR #1577, opened 2021-02-02
UI: Dedicated analysis tab

berggren:
This PR move analysis controls to its own tab in order to support multi timeline -> multi analyzer runs.

Screenshot 2021-02-02 at 14 37 13


timesketch/api/v1/resources/analysis.py
```diff
@@ -137,25 +179,18 @@ def post(self, sketch_id):
                 HTTP_STATUS_CODE_FORBIDDEN,
                 'Unable to run an analyzer without any data submitted.')
 
-        timeline_id = form.get('timeline_id')
-        if not timeline_id:
+        timeline_ids = form.get('timeline_ids')
```
kiddinn, line 182: if you are changing this, you'll need to change the api_client/python.../timeline.py file as well, there is a function there called `run_analyzers` that needs to changed, line 236 from

```
'timeline_id': self.id,
```

to

```
'timeline_ids': [self.id],
```
otherwise the API client will not work anymore.

timesketch/api/v1/resources/analysis.py
```diff
@@ -62,6 +63,47 @@ def get(self, sketch_id, timeline_id):
         return self.to_json(analysis_history)
 
 
+class AnalyzerSessionActiveListResource(resources.ResourceMixin, Resource):
```
kiddinn, line 66: so in the API client you'll see in `sketch.get_analyzer_status` there is a call to `{api_root}/sketches/{sketch.id}/timelines/{timeline_id}/analysis`, which is defined in this file..  that function will go through all timelines in a sketch and do one API call per timeline. My question to you is:

a) can we not just use that already existing API call?

b) If you prefer this one call over one per timeline, can you update the API client to make use of this new function

timesketch/api/v1/resources/analysis.py
```diff
@@ -62,6 +63,47 @@ def get(self, sketch_id, timeline_id):
...
+        if not sketch.has_permission(current_user, 'read'):
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN,
+                'User does not have read access to sketch')
+
+        active_sessions = sketch.get_active_analysis_sessions()
+
+        counter = collections.Counter(PENDING=0, STARTED=0, ERROR=0, DONE=0)
+        session_ids = set()
+        for session in active_sessions:
```
kiddinn, line 91: tiny nit... can you define `active_sessions` just before it's use in the for loop, makes it easier to read

timesketch/api/v1/resources/analysis.py
```diff
@@ -62,6 +63,47 @@ def get(self, sketch_id, timeline_id):
...
+            session_ids.add(session.id)
+            for analysis in session.analyses:
+                counter[analysis.get_status.status] += 1
+
+        schema = {
+            'objects': [{
+                'total_tasks': sum(counter.values()),
+                'total_sessions': len(active_sessions),
+                'tasks': counter,
```
kiddinn, line 100: do we want to return a counter object back? or just a dict `dict(counter)` or as a list of tuples `counter.most_common()`

timesketch/api/v1/resources/analysis.py
```diff
@@ -137,25 +179,18 @@ def post(self, sketch_id):
...
-        search_index = timeline.searchindex
-
-        if not search_index:
-            return abort(
-                HTTP_STATUS_CODE_BAD_REQUEST, (
-                    'No timeline was found, make sure you\'ve got the correct '
-                    'timeline ID or timeline name.'))
+        for timeline_id in timeline_ids:
+            timeline = Timeline.query.get(timeline_id)
+            if timeline not in sketch.timelines:
```
kiddinn, line 190: There are two approaches here... 

a) just do a list of timelines in a sketch

```
sketch_timelines = [t.id for t in sketch.timelines]
```

And then just compare the timeline_ids to that...

```
timeline_ids = [t for t in timeline_ids if t in sketch_timelines]
```

(or just a check to see if a timeline is not in there)

```
sketch_timelines = set([t.id for t in sketch.timelines])
if not set(timeline_ids).issubset(sketch_timelines):
    abort(....'timeline ID not part of a sketch sentence, bla bla')
```

and then you can get this all done that way.... 

that way you don't have to do a DB query for each of the timelines there, to check if the timeline is not part of the sketch timelines.... 


also if you insist on doing it this way, you would also need a check : 

```
timeline = Timeline.....
if not timeline:
  abort(....
if timeline not in sketch....
```

that is you also assume that `timelien` exists, so if I pass a non-existing timeline ID this will fail, since None object is not in sketch.timelines, etc

timesketch/api/v1/resources/analysis.py
```diff
@@ -186,25 +221,29 @@ def post(self, sketch_id):
...
-        }
-        return jsonify(schema)
+            if analyzer_grp:
+                pipeline = (tasks.run_sketch_init.s(
+                    [searchindex_name]) | analyzer_grp)
+                pipeline.apply_async()
+
+            sessions.append(session)
+
+        return self.to_json(sessions)
```
kiddinn, line 249: This will also mess up the API client, which expects to get the other schema that was previously returned, so you'll have to fix the API client here

timesketch/api/v1/resources/analysis.py
```diff
@@ -186,25 +221,29 @@ def post(self, sketch_id):
...
-                'Unable to build analyzer pipeline, analyzer does not exist. '
-                'Error message: {0!s}'.format(e))
 
-        if analyzer_group:
-            pipeline = (tasks.run_sketch_init.s(
-                [search_index.index_name]) | analyzer_group)
-            pipeline.apply_async()
+        sessions = []
+        for timeline_id in timeline_ids:
+            timeline = Timeline.query.get(timeline_id)
```
kiddinn, line 227: check `if not timeline: continue` or something like that (maybe with a warning log)

timesketch/api/v1/resources/analysis.py
```diff
@@ -186,25 +221,29 @@ def post(self, sketch_id):
...
+            searchindex_id = timeline.searchindex.id
+            searchindex_name = timeline.searchindex.index_name
+
+            try:
+                analyzer_grp, session = tasks.build_sketch_analysis_pipeline(
```
kiddinn, line 232: analyzer_group not grp

timesketch/api/v1/resources/analysis.py
```diff
@@ -186,25 +221,29 @@ def post(self, sketch_id):
...
+                    sketch_id=sketch_id, searchindex_id=searchindex_id,
+                    user_id=current_user.id, analyzer_names=analyzers,
+                    analyzer_kwargs=analyzer_kwargs)
```
kiddinn, line 235: so you are not `fixing` the let's run the analyzer only on the timeline, can you at least add a TODO, and hten I'll fix that in a follow up PR that is already started

timesketch/api/v1/resources/analysis.py
```diff
@@ -186,25 +221,29 @@ def post(self, sketch_id):
...
+            except KeyError as e:
```
kiddinn, line 236: you want to return here? we are saying if I start analyzer on 10 timelines at a given timepoint and the 3rd timeline fails, then we suddenly stop and don't process the rest of the timelines...

or should we continue but indicate to the user that one timeline failed?

timesketch/api/v1/resources/sketch.py
```diff
@@ -423,6 +423,15 @@ def get(self, sketch_id):
...
 
+        analyzers = analyzer_manager.AnalysisManager.get_analyzers()
+        analyzers_detail = []
+        for analyzer_name, analyzer_class in analyzers:
+            analyzers_detail.append({
+                'name': analyzer_name,
+                'display_name': analyzer_class.DISPLAY_NAME,
+                'description': analyzer_class.DESCRIPTION
+            })
+
```
kiddinn, line 434: this is also another API call, '{api_root}/sketches/{sketch.id}/analyzer/` 

it returns ever so slightly different output, but can we not just use that? instead of adding this to all sketch data? which we were trying to reduce the other day

timesketch/models/sketch.py
```diff
@@ -143,6 +145,20 @@ def active_timelines(self):
...
+
+        Returns:
+            List of instances of timesketch.models.sketch.AnalysisSession
+        """
+        active_sessions = []
+        for session in self.analysissessions:
+            for analysis in session.analyses:
+                if analysis.get_status.status in ('PENDING', 'STARTED'):
+                    active_sessions.append(session)
+                    break
```
kiddinn, line 159: why the break here?

kiddinn (changes requested):
few comments

timesketch/api/v1/resources/sketch.py
berggren: Good catch!

In an early iteration of the PR I still used the dropdown component and needed this but I have since removed that. I have refactored this to use the API call you mentioned, extended it with more information. This affects the API client, but on the end user part when listing the analyzers. I have not found anything in picatrix so we should be OK.

timesketch/models/sketch.py
berggren: I break the inner loop on first hit of an active analysis in the session as this is enough for the session to be marked as active. Saves a couple or rounds.

berggren: Added a comment

api_client/python/timesketch_api_client/timeline.py
```diff
@@ -277,16 +276,19 @@ def run_analyzers(
...
             raise error.UnableToRunAnalyzer(
-                'Analyzer may have run, but there is no session ID to '
+                'Analyzer may have run, but there is no sessions to '
                 'verify that it has. Please verify manually.')
 
-        session = analyzer.AnalyzerResult(
-            timeline_id=self.id, session_id=session_id,
-            sketch_id=self._sketch_id, api=self.api)
-        return session
+        return analyzer_results
```
kiddinn, line 291: can you change the docstring as well, the returns section still says a single analyzer result is returned

api_client/python/timesketch_api_client/timeline.py
```diff
@@ -277,16 +276,19 @@ def run_analyzers(
...
+        for session in objects:
+            analyzer_result = analyzer.AnalyzerResult(
+                timeline_id=self.id, session_id=session.id,
+                sketch_id=self._sketch_id, api=self.api)
+            analyzer_results.append(analyzer_result)
+
+        if not analyzer_results:
             raise error.UnableToRunAnalyzer(
-                'Analyzer may have run, but there is no session ID to '
+                'Analyzer may have run, but there is no sessions to '
```
kiddinn, line 288: `but there are no sessions` <nit>

timesketch/api/v1/resources/analysis.py
```diff
@@ -32,6 +34,9 @@
 from timesketch.models.sketch import Timeline
 
 
+logger = logging.getLogger('timesketch.sketch_api')
```
kiddinn, line 37: this shouldn't be the `sketch_api`, this is `analysis_api`

timesketch/api/v1/resources/analysis.py
kiddinn: it's mainly that we now need to serialize and deserialize  a counter object, rather than the more `native` dict, and I Would think the ui could just as easily read a dict object....

timesketch/lib/analyzers/phishy_domains.py
```diff
@@ -23,6 +23,9 @@ class PhishyDomainsSketchPlugin(interface.BaseSketchAnalyzer):
     """Sketch analyzer for phishy domains."""
 
     NAME = 'phishy_domains'
+    DISPLAY_NAME = 'Phishy domains'
+    DESCRIPTION = ('Compare the Jaccard distance between all domains in the '
```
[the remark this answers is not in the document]
kiddinn: still not replied to this

timesketch/models/sketch.py
kiddinn: I read the code again... I thought this was `completed`, but of course this is just that if one analyzer is still in a a pending/started stage then it is active, which is of course right, so yes... this is fine

kiddinn (changes requested):
few comments, mostly minor nits

timesketch/lib/analyzers/phishy_domains.py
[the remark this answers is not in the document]
berggren: oh, missed this one. Changed the text.

timesketch/api/v1/resources/analysis.py
berggren: Ah, I see what you mean now. So the task counter object is being serialised to a dict in the `return jsonify(schema)` so there is not a counter object in the return.

That being said, the name `tasks` is a bit misleading, I rename it to `tasks_status`

api_client/python/timesketch_api_client/timeline.py
berggren: Fixed

api_client/python/timesketch_api_client/timeline.py
berggren: Done

timesketch/api/v1/resources/analysis.py
berggren: done

berggren (commented):
PTAL

kiddinn (approved):
LGTM

Issue #1567, opened 2021-01-27
Change the behavior of uploaded data


kiddinn:
**Problem**

Elastic has a limit on the number of open resources, which means that each deployment only has a limited amount of indices that can be open at any given time.

The current behavior of uploading data is that each and every new data gets assigned a new SearchIndex (SI) and a new Elastic Index. This means that a sketch that has 10 data sources will have 10 open indices.

Elastic has a preference for fewer and bigger indices. Therefore the solution is to change the behavior of file uploads in the following manner:

1. Each sketch will have a SI per `label`
2. Each sketch will have a separate Timeline Object (TO), each one of them pointing to the appropriate SI.

This will mean that instead of each data source having it's own SI, all data sources will share the same SI in the sketch.

This will in part `solve`:  #1200 but it also needs to be added into this design, that is the ability to merge older timelines in a sketch.

This requires multiple changes spread out over the codebase, including a major change in the UI and API client to make sure that searches can still be limited to each data source, or TO, instead of the current behavior that limits to it to hits within each SI.

A list of items to be done:

- [x] Change uploads to add a `__ts_timeline_id` field
- [x] Change psort upload behavior to use `elastic_ts` output module and move TS specific things to lib/tasks.py
- [x] Add a way to identify a timeline, whether it is a `new world` vs an `old world` timeline.
- [x] Change the UI so that it can correctly query timelines
- [x] Change the counts in the UI
- [x] Make sure labels still work
- [x] Make sure analyzers still work
- [x] Make sure aggregations still work
- [x] Make sure graphs still work
- [x] Make sure stories still work
- [x] Make sure saved searches still work
- [x] This also implies that we can't support `tsctl import` or uploading data via `tsctl` anymore, deprecate that.

kiddinn:
First phase is in, which fixes upload REST API, the importer and explore REST api and search api object. Now search works across old sketches, new and mixed.

Next phase will be to fix the UI so that filtering based on individual timelines works and counting of events.

Another parallel stream is to fix psort so that plaso ingestion works

kiddinn:
#1573 submitted as phase II, the first UI fixes for the new indexing.

kiddinn:
The needed change in psort can be found here: https://github.com/log2timeline/plaso/pull/3463

kiddinn:
fixing UI counts: #1576

kiddinn:
Adding aggregation support in #1588

kiddinn:
[...]

PR #1578, opened 2021-02-02
Index Change: Changed plaso file ingestion, added query_dsl support and analyzer support

kiddinn:
This PR is part of #1567

What this does is to change the `run_plaso` task to now use the new `elastic_ts` output module (that is still not yet part of a released psort version).

That removes the inner TS working inside the plaso module, and has psort only do the ES ingestion, and then the Timesketch backend takes care of the actual database changes and other things that are required.

This PR also adds support for timeline IDs into the analyzers.


timesketch/lib/datastores/elastic.py
```diff
@@ -719,27 +783,33 @@ def set_label(self, searchindex_id, event_id, event_type, sketch_id,
 
         return None
 
-    def create_index(self, index_name=uuid4().hex, doc_type='generic_event'):
+    def create_index(
+            self, index_name=uuid4().hex, doc_type='generic_event',
```
berggren, line 787: you can remove the doc_type as this has been deprecated in ES.

timesketch/lib/tasks.py
```diff
@@ -484,12 +500,85 @@ def run_plaso(
...
+        raise RuntimeError(
+            'Plaso version is out of date (version {0:d}, please upgrade to a '
+            'version that is later than {1:d}'.format(
+                plaso_version, PLASO_MINIMUM_VERSION))
+
     if events:
         raise RuntimeError('Plaso uploads needs a file, not events.')
-    # Log information to Celery
+
+    event_type = 'generic_event'  # Document type for Elasticsearch
```
berggren, line 526: remove

timesketch/lib/tasks.py
```diff
@@ -484,12 +500,85 @@ def run_plaso(
...
+        raise RuntimeError(
+            'Unable to connect to Elastic, no port set, unable to '
+            'process plaso file.')
+
+    es = ElasticsearchDataStore(
+        host=elastic_server, port=elastic_port)
+
+    try:
+        es.create_index(
+            index_name=index_name, doc_type=event_type, mappings=mappings)
```
berggren, line 559: remove `doc_type=...`

berggren (changes requested):
just some small things, otherwise LG

timesketch/lib/datastores/elastic.py
berggren: we will do this in a separate PR instead.

berggren (approved):

PR #1583, opened 2021-02-03
UI: Timeline tab  and removing ability to add old indices to sketches

berggren:
This PR removes the old way of adding timlines to a sketch. This is not supported anymore in the compound index world.
It also adds a tab for the timeline management page.


kiddinn (approved):
LGTM

PR #1584, opened 2021-02-03
UI: Move aggregations to its own tab

berggren:
This PR moves the embedded aggregation control out from Explore, and to its own tab.
This new page also displays all saved aggregations.


kiddinn (approved):
LGTM

PR #1585, opened 2021-02-03
Bugfix: Wrong query count for new style timelines

berggren:
Need to provide `'min_doc_count': 0` in order to show buckets with 0 hits.


kiddinn (approved):
LGTM

PR #1587, opened 2021-02-04
UI: Set status on timeline

berggren:
This PR fixes the status check for timelines in the new world of compound indices.


kiddinn (approved):
LGTM

PR #1589, opened 2021-02-04
UI: Add loading spinner while searching

berggren:
This got deactivated at some point. Adding it back in to give the user an indication that searching is running.


kiddinn (approved):
LGTM

PR #1592, opened 2021-02-05
UI: Support index and time range filters in aggregations

berggren:
This PR adds:
* Filter aggregation on index
* Add start_time and end_time filters

Update of some JS libraries.


kiddinn (approved):
LGTM

PR #1594, opened 2021-02-06
Deprecated the option to import timelines in tsctl.

kiddinn:
With the new way of uploading data to Timesketch (#1567) it is hard to support `tsctl` as well as the importer and the upload resources in the API.

Therefore this PR will just deprecate the option of using tsctl to import data in favor of the importer client. Tsctl will rather be used for administrative tasks to Timesketch.


kiddinn:
This is part of #1567

timesketch/tsctl.py
```diff
@@ -431,128 +426,12 @@ class ImportTimeline(Command):
...
-            file_path=file_path, events='', timeline_name=timeline_name,
-            index_name=index_name, file_extension=extension,
-            sketch_id=sketch.id)
-        pipeline.apply_async(task_id=index_name)
-
-        print('Imported {0:s} to sketch: {1:d} ({2:s})'.format(
-            file_path, sketch.id, sketch.name))
+        print(
+            'This function has been deprecated, please use the '
+            'timesketch_importer instead.')
```
berggren, line 434:
```suggestion
            'timesketch_importer instead: https://github.com/google/timesketch/blob/master/docs/UploadData.md')
```

berggren (changes requested):

berggren (approved):

PR #1596, opened 2021-02-08
replace == None with is None

jaegeral:
Replace:

```== None``` with ```is None``` as this should be easier to read and faster execute (and match the rest of the repository)

### DIFF
--- a/timesketch/models/acl.py
+++ b/timesketch/models/acl.py
@@ -133,8 +133,8 @@ class AccessControlMixin(object):
         # pylint: disable=singleton-comparison
         return cls.query.filter(
             or_(cls.AccessControlEntry.user == user,
-                and_(cls.AccessControlEntry.user == None,
-                     cls.AccessControlEntry.group == None),
+                and_(cls.AccessControlEntry.user is None,
+                     cls.AccessControlEntry.group is None),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by kiddinn** [comment]
can you fix all of those unit tests/e2e tests fails

**Review by berggren — CHANGES_REQUESTED** [review]
This will break the ACL system because it depend on SQLalchemy constructs for handling null values. SQLalchamy need this to be `== None` (as part of their implementation of SQL statement generation).

If this is because the linter complains I suggest we suppress the linter warning here instead.

It is technically possible to do something like:
```
cls.AccessControlEntry.user._is(None)
```

But we have to be aware of side effects, and potential bugs because of this. I don't think it is worth it TBH.

jaegeral:
> This will break the ACL system because it depend on SQLalchemy constructs for handling null values. SQLalchamy need this to be `== None` (as part of their implementation of SQL statement generation).
> 
> If this is because the linter complains I suggest we suppress the linter warning here instead.
> 
> It is technically possible to do something like:
> 
> ```
> cls.AccessControlEntry.user._is(None)
> ```
> 
> But we have to be aware of side effects, and potential bugs because of this. I don't think it is worth it TBH.

Uff okay, then it is a good reason to keep it as it is.

PR #1598, opened 2021-02-08
UI: Fixing a bug in the Interval time filters

binglot:
Fixing the bug when the user creates a time filter using an interval and they use a space instead of 'T' to separate the date and time.


timesketch/frontend/src/components/Sketch/TimeFilter.vue
```diff
@@ -130,10 +130,10 @@ export default {
...
       this.radio = 'interval'
       let offset = this.chip.value.split(' ')
-      this.offsetStart = offset[0]
-      this.offsetMinus = offset[1].match(/\d+/)[0]
-      this.offsetPlus = offset[2].match(/\d+/)[0]
-      this.selectedInterval = offset[1].match(/[a-zA-Z]+/)[0]
+      this.selectedInterval = offset[offset.length-1].match(/[a-zA-Z]+/)[0] // interval type (day, minute, etc.)
+      this.offsetPlus = offset.pop().match(/\d+/)[0]
+      this.offsetMinus = offset.pop().match(/\d+/)[0]
+      this.offsetStart = offset.join(' ') // Concatenate in case there was a space inbetween date and time
```
kiddinn, line 136: what about all non-ISO dates?

What if I put in `2021-01` in the interval and then have the interval set to +/- 10 days?

We need to save the chip as a ISO formatted sting, since that's what's expected

binglot: Those are really good points Kiddi! I changed the logic to address it, please take a look in case I misunderstood your requirements. I tested with your scenario and it worked fine.

berggren (approved):

PR #1609, opened 2021-02-15
add bs4 and mock to requirements.txt

jaegeral:
Adding both packages as they are used in multiple tests.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,3 +32,5 @@ xlrd==1.2.0
 tabulate==0.8.6
 markdown==3.2.2
 networkx==2.5
+bs4
+mock


berggren:
These test dependencies are already present in the test_requirements file:
https://github.com/google/timesketch/blob/master/test_requirements.txt

They are automatically installed in the dev and e2e containers.

berggren (changes requested):
These test dependencies are already present in the test_requirements file:
https://github.com/google/timesketch/blob/master/test_requirements.txt

They are automatically installed in the dev and e2e containers.

kiddinn:
yes. please revert this PR, since it is not needed, we don't need testing requirements to be added to the tools requirement file.

jaegeral:
ACK closed

PR #1610, opened 2021-02-15
Fixed few issues: large text file imports, aggregations failing and upgrading PPA in docker files

kiddinn:
This PR fixes few issues with changes made to accommodate new way of indexing:

+ **Upload large text files**: The issue was caused with the importer creating a new `index_name`, which would then differ from the one actually used in the backend, creating an issue when a new data block got uploaded (then there was a mismatch in the index names)
+ **Aggregation Groups didn't work**: There wasn't really an issue with the aggregation groups, it was more that certain aggregator classes (the manual) ones didn't have their `__init__` functions updated after the changes to the interface, that caused few aggregations to fail to load, since they didn't recognize parameters passed to them.
+ **Dockerfiles still using dev**: Changed the PPA track back to stable for all docker files, since plaso has now been released.


berggren (approved):
LGTM

PR #1612, opened 2021-02-15
UI: Fix missing timeline color

berggren:
This PR adds a v-if on the timeline picker, to wait for the current list of indices to be ready. Otherwise some saved searches will fail to render timeline colors in the picker.


kiddinn (approved):
LGTM

PR #1613, opened 2021-02-15
Set correct timeline for analysis pipeline

berggren:
This PR fixes the build_sketch_analysis_pipeline() function to use provided timeline_id if present. Without this new world timelines will use the index instead and create a pipeline for the whole index instead of the timeline_id


kiddinn (approved):
LGTM

PR #1623, opened 2021-02-17
Fixed issues with running analyzers in API client.

kiddinn:
After the new indexing change a bug got introduced into the API client when attempting to run new analyzers.


berggren (approved):
LGTM

PR #1624, opened 2021-02-17
UI: Fix wrong timeline id in aggregations

berggren:
This PR fixes 2 things:

1. Add support for legacy indices in aggregation UI
2. Fix timeline picker where list disappears in the UI when all timelines was deselected.


kiddinn (approved):
LGTM

PR #1625, opened 2021-02-17
Minor bug fixes.

kiddinn:
Two bugs here:

+ File uploads when the `index_name` is not present, it attempts to temporarily create the file in a non-existing file (attempts to open the directory)
+ When a timeline is removed from a sketch it is now archived at the same time.


berggren (approved):
LGTM

PR #1627, opened 2021-02-18
Added more error handling into the API, as well as retry to flushing events queue.

kiddinn:
This PR adds more error handling to survive a search index that is not found in Elastic, as well as adding a retry count to the flushing of the event queue in case there is a socket timeout.

### DIFF
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -105,7 +105,7 @@ class ExploreResource(resources.ResourceMixin, Resource):
             indices = all_indices
 
         # Remove indices that don't exist from search.
-        indices = utils.validate_list_of_indices(indices, self.datastore)
+        indices = utils.validate_indices(indices, self.datastore)
 
         # Make sure that the indices in the filter are part of the sketch.
         # This will also remove any deleted timeline from the search result.
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -104,13 +104,17 @@ class ExploreResource(resources.ResourceMixin, Resource):
         if '_all' in indices:
             indices = all_indices
 
-        # Check if indices are valid.
-        indices = utils.refresh_and_validate_list_of_indices(
-            indices, self.datastore)
+        # Remove indices that don't exist from search.
+        indices = utils.validate_list_of_indices(indices, self.datastore)
 
         # Make sure that the indices in the filter are part of the sketch.
         # This will also remove any deleted timeline from the search result.
         indices, timeline_ids = get_validated_indices(indices, sketch)
+
+        if not indices:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'No valid search indices were found to perform the search on.')
 
         # Make sure we have a query string or star filter
         if not (form.query.data, query_filter.get('star'),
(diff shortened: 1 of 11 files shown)


timesketch/api/v1/resources/upload.py
```diff
@@ -133,6 +133,14 @@ def _upload_and_index(
...
             data_label=data_label,
             extension=file_extension)
+
+        if not searchindex:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'We were unable to acquire a searchindex, and therefore not '
+                'able to upload data, try again? Please create an issue on '
+                'Github.')
+
```
berggren, line 143:
```suggestion
                HTTP_STATUS_CODE_BAD_REQUEST,
                'We were unable to acquire a searchindex and therefore not '
                'able to upload data, please try again. If this error persist please create an issue on '
                'Github: https://github.com/google/timesketch/issues/new/choose')

```

timesketch/lib/datastores/elastic.py
```diff
@@ -912,14 +926,15 @@ def import_event(self, index_name, event_type, event=None, event_id=None,
 
         return self.import_counter['events']
 
-    def flush_queued_events(self):
+    def flush_queued_events(self, retry_count=0):
```
berggren, line 929: Why have retry_count as an argument? Does it get set from a caller somewhere?

timesketch/lib/utils.py
```diff
@@ -83,6 +84,31 @@ def _scrub_special_tags(dict_obj):
...
+    Args:
+        indices (list): List of indices.
+        datastore (ElasticsearchDataStore): a data store object.
+
+    Returns:
+        list of indices that are discovered.
+    """
+    for index in indices:
+        try:
+            datastore.client.indices.refresh(index=index)
```
berggren, line 102: How expensive is this? We are doing this on every search. Is it maybe enough to do this check in the Sketch (GET) resource?

berggren (changes requested):

timesketch/lib/datastores/elastic.py
kiddinn: it is called by the exception handling, only there, to be a `recursive` function

timesketch/lib/utils.py
kiddinn: I measured it, it seems fairly fast, but I also checked the `if index exists` and it seems about the same speed, but then a lot less overhead with the try/except, so I swapped the implementation.

kiddinn (commented):
PTAL

timesketch/lib/utils.py
```diff
@@ -83,6 +84,23 @@ def _scrub_special_tags(dict_obj):
             _ = dict_obj.pop(field)
 
 
+def validate_list_of_indices(indices, datastore):
```
berggren, line 87:
```suggestion
def validate_indices(indices, datastore):
```

timesketch/api/v1/resources/explore.py
```diff
@@ -103,10 +104,18 @@ def post(self, sketch_id):
         if '_all' in indices:
             indices = all_indices
 
+        # Remove indices that don't exist from search.
+        indices = utils.validate_list_of_indices(indices, self.datastore)
```
berggren, line 108:
```suggestion
        indices = utils.validate_indices(indices, self.datastore)
```

berggren (changes requested):

berggren (approved):

Issue #1620, opened 2021-02-16
Re-introduce search templates


berggren:
This will require few changes:

- [x] Port over the search templates to the new UI
- [x] Add support into the search object in the API client
- [ ] Add ability to save searches as templates in UI
- [x] Add ability to save searches as templates in API client (add a parameter to the saved search object to save as a template)
- [ ] Add ability to load YAML/text file in `tsctl` to upload saved templates
- [ ] Add ability to export search templates into a YAML/text file in UI
- [ ] Add ability to export search templates into a YAML/text file in API client (`ts_client.export_search_templates()` or something like that)

kiddinn:
#1643 will add the support into the API client

berggren:
This has been implemented

PR #1643, opened 2021-02-25
Added search templates to API and API client

kiddinn:
This PR does the following:

1. Adds the ability to recall saved search templates in the API client
2. Adds the ability to store saved searches as templates in the REST API
3. Adds `save_as_template` to the API client search object to save searches as templates.
4. Adds a `data_sources` property to the timeline object to display the data sources that are saved to the timeline.

This is part of #1620


api_client/python/timesketch_api_client/search.py
```diff
@@ -944,6 +977,40 @@ def save(self):
...
+        """Save the search as a search template.
+
+        Returns:
+            A string indicating the search template ID.
+
+        Raises:
+            ValueError: if the search hasn't been saved first.
+            RuntimeError: If the search could not be saved as a template.
+        """
+        if not self._resource_id:
```
berggren, line 990: the user is clearly want to save the search here, and then also create a template based on it. Maybe you can just save the search for the user if it isn't already? Or do you prefer to raise here and let the user do the needful?

timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -51,4 +85,100 @@ def get(self):
...
+        """
+        form = request.json
+        if not form:
+            form = request.data
+
+        search_id = form.get('search_id')
+        if not search_id:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'Unable to save the searchtemplate, the search has not '
```
berggren, line 117: maybe a little clearer?

```
Unable to save the template, saved search ID is missing.
```

timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -51,4 +85,100 @@ def get(self):
...
+                'been saved first')
+
+        sketch_id = form.get('sketch_id')
+        if not sketch_id:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                'We need to have a sketch associated with the saved search '
```
berggren, line 124: You should get the sketch_id from the saved search object, right? A saved search isn't possible to save without a sketch attached.

berggren (changes requested):
Some comments.

timesketch/api/v1/resources/searchtemplate.py
kiddinn: yes, that's true... I should just pull it from there instead of requesting it again...

timesketch/api/v1/resources/searchtemplate.py
kiddinn: added

api_client/python/timesketch_api_client/search.py
```diff
@@ -572,6 +572,35 @@ def delete(self):
         response = self.api.session.delete(resource_url)
         return error.check_return_status(response, logger)
 
+    def delete_as_template(self):
```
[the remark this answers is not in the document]
kiddinn: created a `searchtemplate.SearchTemplate` object, changed things quite a bit

api_client/python/timesketch_api_client/search.py
kiddinn: done

kiddinn (commented):
@berggren PTAL

berggren (approved):
LGTM

berggren (approved):

PR #1650, opened 2021-02-27
Docker: e2e / dev Sigma directory was not created before

jaegeral:
There was a glitch in the docker config files that resulted in a Timesketch config not really working as the expected directory was not present


kiddinn (approved):
LGTM

PR #1667, opened 2021-03-05
Added several enhancements to the importer and importer tasks.

kiddinn:
This PR changes the following:

+ fixes an issue in the CSV file ingestion (tasks/utils)
+ adds retries to uploads in the importer
+ adding the ability to check celery task status for specific tasks
+ letting the importer CLI tool wait and check timeline and celery status before exiting
+ adding task IDs to uploads in the REST API 
+ added a fix for timeline deletion issues


berggren (approved):
This PR got me thinking, we should add the option to report back how many documents have been indexed so far. Then we can show more informative info to the user etc. Let's discuss outside of this PR though.

LGTM

PR #1673, opened 2021-03-09
package name typo in the requirements

sa3eed3ed:
package name typo in the requirements, image @ us-docker.pkg.dev/osdfir-registry/timesketch/dev will also need to be rebuilt.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -32,5 +32,5 @@ xlrd==1.2.0
 tabulate==0.8.6
 markdown==3.2.2
 networkx==2.5
-prometheus-client==0.9.0
-prometheus-flask-exporter==0.18.1
+prometheus_client==0.9.0
+prometheus_flask_exporter==0.18.1


kiddinn:
it seems like this works with both... I can do 

```shell
$pip install prometheus-client==0.9.0
```

and it installs just fine, the same with:

```shell
$pip install prometheus_client==0.9.0
```
Pypi even has it `both` ways: https://pypi.org/project/prometheus-client/ that is the client is named with a dash, but the instructions use underscore... seems like both work just as fine.

kiddinn:
see also: https://mail.python.org/pipermail/distutils-sig/2011-August/017936.html

And this (PEP8) https://www.python.org/dev/peps/pep-0008/#package-and-module-names 

```
Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.
```

so underscores are discouraged in package names, they prefer the dash....

sa3eed3ed:
Ok, the reason I opened this is that the image @ us-docker.pkg.dev/osdfir-registry/timesketch/dev errors when run: `import prometheus_client ModuleNotFoundError: No module named 'prometheus_client'`, connecting to the container and installing prometheus_client with `pip3 install prometheus_client==0.9.0` solved this. 

But it looks like unrelated to using a dash. Rebuilding and pushing the image should fix this, I will close this PR then.

kiddinn:
I seem to be able to import that library just fine using my dev docker container...

The image is rebuilt on every new merge into the GH master branch, so there should not be a need to specifically rebuild now.

Have you tried upgrading the dev container?

inside the container I did

```shell
$ python3
>>> import prometheus_client
```

And it seems to load up just fine.

kiddinn:
besides there are several files in the current codebase that do run "import prometheus_client", all of which would fail in the docker container if this didn't work, but they do work on my dev container adn I've not seen any other complaints,

sa3eed3ed:
upgrading solves it, importing the library runs fine. This previous version does have this issue, but latest works fine: us-docker.pkg.dev/osdfir-registry/timesketch/dev@sha256:26d560ece9dc015be5298f32208b3e0ab520fa6929de62e8496ffd4d46545eb8

PR #1697, opened 2021-03-18
Fix docker sigma conf

jaegeral:
the data sigma folder was not linked correctly, resulting in an exception for dev and e2e


kiddinn:
This is an empty PR, I assume you forgot to actually sync in the changes.

kiddinn (changes requested):
Please add in some changes so that i can review them

jaegeral:
Oh wow, is indeed a large PR. ^^

PR #1712, opened 2021-03-24
Changed upload behavior when appending to an already existing timeline

kiddinn:
This PR changes the behavior when new data is appended to an already existing timeline. What changes is:

+ The timeline status is no longer set to `processing` while the new data is being ingested
+ If one data source fails, that does not mean the entire timeline is no longer accessible.

**Checks**

- [x] All tests succeed.

### DIFF
--- a/timesketch/lib/tasks.py
+++ b/timesketch/lib/tasks.py
@@ -176,6 +176,7 @@ def _set_timeline_status(timeline_id, status, error_msg=None):
         logger.warning('Cannot set status: No such timeline')
         return
 
+    # Check if there is at least one data source that hasn't failed.
     multiple_sources = any([not x.error_message for x in timeline.datasources])
     print('Multiple sources: {}'.format(multiple_sources))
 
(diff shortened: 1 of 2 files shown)


timesketch/lib/tasks.py
```diff
@@ -176,14 +176,22 @@ def _set_timeline_status(timeline_id, status, error_msg=None):
         logger.warning('Cannot set status: No such timeline')
         return
 
-    timeline.set_status(status)
-    timeline.searchindex.set_status(status)
+    # Check if there is at least one data source that hasn't failed.
+    multiple_sources = any([not x.error_message for x in timeline.datasources])
+    print('Multiple sources: {}'.format(multiple_sources))
```
berggren, line 181:
```suggestion
    logger.info(('Multiple sources: {}'.format(multiple_sources))
```

berggren (changes requested):

timesketch/lib/tasks.py
kiddinn: this was not supposed to be in there

kiddinn (commented):
.

berggren (approved):

PR #1717, opened 2021-03-24
Allow other OIDC providers for authentication

### LINKED ISSUE #1716: Allow other OIDC providers for authentication
Currently authentication using OIDC is only available to Google identity federation

To be able to use other OIDC providers for authentication

asmirazali:
Allowing other OIDC providers for authentication

Federated identity other can Google can be used, using OIDC protocol for authentication

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**
`closes #1716`


kiddinn (changes requested):
Thank you for the PR, but can you first fix the linter and tests?

berggren:
@asmirazali - Out of curiosity, do you have an example of another OIDC provider? The Google one has some specific implementation details and I wonder if we want to refactor it into a base OIDC implementation and then extend.

But an example here would be great to get a better understanding of what is needed.

Thanks for contributing!

asmirazali:
Hi, it is the same as OIDC standards are the same. I made changes directly on the docker and was able to connect to my provider.

asmirazali:
Google: https://accounts.google.com/.well-known/openid-configuration
Azure: https://login.windows.net/common/.well-known/openid-configuration
Salesforce: https://login.salesforce.com/.well-known/openid-configuration

Most openid-client will pull the configuration from the OIDC dicovery endpoint.

If users are able to change the discovery endpoint and auth_uri, this package will be usable with other providers

asmirazali:
Hmm. I ran the test locally and did not have any errors. Any documentation on how to proceed further?

run_tests.py
```diff
@@ -7,7 +7,7 @@
 import argparse
 
 
-def run_python_tests(coverage=False):
+def run_python_tests(coverage=True):
```
kiddinn, line 10: why are you changing this?

yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
kiddinn, line 1: why are you including this file?

timesketch/lib/google_auth.py
```diff
@@ -38,11 +38,9 @@
 
 from timesketch.lib.definitions import HTTP_STATUS_CODE_OK
 
-
 CSRF_KEY = 'google_oauth2_csrf_token'
-AUTH_URI = 'https://accounts.google.com/o/oauth2/v2/auth'
-DISCOVERY_URL = 'https://accounts.google.com/.well-known/openid-configuration'
-
+AUTH_URI = current_app.config.get('GOOGLE_OIDC_AUTH_URI')
+DISCOVERY_URL = current_app.config.get('GOOGLE_OIDC_DISCOVERY_URL')
```
kiddinn, line 43: this is what is making all the tests fail, what you are doing is to attempt to use the `current_app` in a constant at the module level, which is evaluated before the actual app is setup.

So what you need to do is a different approach...

```python
                                                                                                                                                 
DEFAULT_AUTH_URI = 'https://accounts.google.com/o/oauth2/v2/auth'                                                                                                                                                
DEFAULT_DISCOVERY_URL = (
    'https://accounts.google.com/.well-known/openid-configuration')
```

And then below you can do:

```python
    discovery_url = current_app.config.get(                                                                                                                                                                              
        'GOOGLE_OIDC_DISCOVERY_URL', DEFAULT_DISCOVERY_URL)                                                                                                                                                              
    try:                                                                                                                                                                                                                 
        resp = requests.get(discovery_url)             
```

That way,

a) you do not use the app before it is setup
b) for those that forget to update their config file (that is when you do a docker pull but didn't realize it required a change in the config file, things don't break
c) things just work ;)

kiddinn: and the same applies for the other URL, that is do the `auth_uri = current_app....` instead of using the constant there.

kiddinn (changes requested):
few comments

run_tests.py
asmirazali: I was testing some thing. Will change it back and commit

timesketch/lib/google_auth.py
asmirazali: Noted. Thank you for your advice.

asmirazali:
You Sir, is a genius. I have added some documentation in timesketch.conf as well.

kiddinn (approved):
LGTM

Issue #1687, opened 2021-03-13
Internal server error - issue with 150+ timelines in a sketch.


jacurutu1984:
**Describe the bug**
I imported with timesketch_importer (plaso file) approximatively 150 timelines in one sketch  and now when i want to access the data through the UI, i get a blank page and the message "Internal server error" 

**To Reproduce**
Import 150 timelines et go to the sketch

**Expected behavior**
I expected to access the data and to explore it.

**Screenshots**
![Screenshot from 2021-03-13 16-34-15](https://user-images.githubusercontent.com/48470964/111035283-fc7c8580-8419-11eb-839a-472b6bd3fe85.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Browser Mozilla firefox 85
 - Version latest version of timesketch

**Additional context**
wsgi_error.log : 
 [2021-03-13 15:22:27,298] timesketch.app/ERROR Exception on /api/v1/sketches/1/count/ [GET]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1949, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1935, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 458, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask/views.py", line 89, in view
    return self.dispatch_request(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 573, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask_login/utils.py", line 261, in decorated_view
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/timesketch/api/v1/resources/event.py", line 665, in get
    count, bytes_on_disk = self.datastore.count(indices)
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/datastores/elastic.py", line 767, in count
    es_stats = self.client.indices.stats(
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/client/utils.py", line 84, in _wrapped
    return func(*args, params=params, **kwargs)
[...]

PR #1718, opened 2021-03-24
Always uniq index lists

berggren:
This PR adds safeguards to make sure that the index list is always uniq in order to not balloon the ES HTTP calls.
I have tested a sketch with 500+ timelines without issues.

**Closing issues**
Fixes: #1687


timesketch/api/v1/resources/sketch.py
```diff
@@ -361,39 +364,34 @@ def get(self, sketch_id):
...
 
+        # Get number of events per timeline
         if sketch_indices:
-            # Stats for index. Num docs per shard.
-            for timeline in sketch.active_timelines:
-                index_name = timeline.searchindex.index_name
+            for index_name in sketch_indices:
                 if indices_metadata[index_name].get('is_legacy', False):
                     doc_count, _ = self.datastore.count(indices=index_name)
                     stats_per_timeline[timeline.id] = {'count': doc_count}
```
kiddinn, line 372: `timeline.id` isn't defined here... that is the `timeline` isn't defined.

kiddinn (commented):
early

timesketch/api/v1/resources/sketch.py
berggren: Yeah, that is what I fixed before removing the draft status :)

kiddinn (approved):
LGTM

PR #1719, opened 2021-03-24
UI: Remove old error field

berggren:
Small PR that removes the old error message field in the timeline info modal.


kiddinn (approved):
LGTM

PR #1722, opened 2021-03-25
Added upper memory limits to psort.

kiddinn:
This PR adds configuring the upper limits of the process memory for the psort process.

This can be tuned in case psort runs out of memory while ingesting plaso files.


berggren (approved):
LGTM

PR #1723, opened 2021-03-26
Removed index analyzers and changed API/import client.

### LINKED ISSUE #1711: Deprecate Index Analyzers
With the new way of adding data to TS, there is really nothing that `index` analyzer does.. only sketch analyzers make sense. Therefore the index analyzer needs to be removed/deprecated from the codebase

kiddinn:
This PR removes the notion of an index analyzer and renames the sketch analyzers as an analyzer.

Originally there were two kinds of analyzers, the sketch analyzer and an index one. The idea was that you could have data uploaded into a search index, but it wouldn't be associated with a timeline, and therefore you could run special type of analyzers on top of that dataset, called index analyzers. But with how data is ingested into Timesketch today, all data is associated with a sketch.

Other changes:
+ Incremented API and import client versions
+ Added the ability to indicate sketch name in the import client when a new sketch is created.
+ Added a retry to a search operation in the API client

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #1711


data/timesketch.conf
```diff
@@ -180,7 +180,6 @@ NEO4J_PASSWORD = '<NEO4J_PASSWORD>'
 # Analyzers.
 
 # Which analyzers to run automatically.
-AUTO_INDEX_ANALYZERS = []
```
jaegeral, line 183: Is there a way to tell users who have an older version of the config that the key has been deprecated?

timesketch/lib/analyzers/interface.py
```diff
@@ -762,14 +762,14 @@ class BaseIndexAnalyzer(object):
         timeline_id: The ID of the timeline the analyzer runs on.
         tagged_events: Dict with all events to add tags and those tags.
         emoji_events: Dict with all events to add emojis and those emojis.
+        sketch: A Sketch instance.
+        timeline_id: The ID of the timeline the analyzer runs on.
```
jaegeral, line 766: That seems to be duplicate, is already mentioned in the docstring

data/timesketch.conf
kiddinn: not sure we need to, since there haven't been any INDEX analyzers to pick ... well ever, so this is should be a no-op

timesketch/lib/analyzers/interface.py
kiddinn: done

data/timesketch.conf
jaegeral: SGTM

jaegeral (approved):
LGTM

berggren (approved):

PR #1729, opened 2021-04-01
Sigma api client end2end tests.

jaegeral:
This PR adds Sigma e2e tests for Sigma list and pull a single rule + search for a sigma rule query.

It also fixes an issue in one unit tests while I was at it.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

### DIFF
--- a/data/timesketch.conf
+++ b/data/timesketch.conf
@@ -255,5 +255,5 @@ EXTERNAL_HOST_URL = 'https://localhost'
 #-------------------------------------------------------------------------------
 # Sigma Settings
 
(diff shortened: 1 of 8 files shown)


kiddinn:
ok, so you've got the trifecta of errors... e2e tests fail, linter and unit tests... best to fix those before proceeding ...

jaegeral:
> ok, so you've got the trifecta of errors... e2e tests fail, linter and unit tests... best to fix those before proceeding ...

yes, thats why its still a draft ;-)

Maybe I can produce more errors :-D

PR #1730, opened 2021-04-01
Sigma api e2e

jaegeral:
Sigma 2e2 stuff, some other fixed in the api client and a Sigma Notebook


notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "#!pip install -q timesketch_api_client"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "pip install ../api_client/python"
```
kiddinn, line 44: ok... so why the `#pip install` before and now `pip install` from a path?

Remove this cell, uncomment the other one, we should be wanting to install the pypi version.

kiddinn: you are also not installing `picatrix`, are you making the assumption you are running a picatrix shell? If so, you've already got the timesketch api client installed, since it's a requirement... if not, then I suggest you also install picatrix

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+   "cell_type": "markdown",
+   "metadata": {}
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ts_client = config.get_client(confirm_choices=False)"
```
kiddinn, line 80: `confirm_choices` by default is False, so no need to be explicit here... remove it

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "ts_client.create_sketch(name=\"test\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sketch = ts_client.get_sketch(1)"
```
kiddinn, line 98: ok, so first of all you have 

```python
ts_client.create_sketch(name='test')
```

above... which creates and returns a sketch... the problem here is a) you can't be sure that this newly created sketch is the first sketch ever created in someones installation, so this is not a guarantee of getting the new sketch... BUT since the `create_sketch` already returns the newly created sketch a much better approach is

```python
sketch = ts_client.create_sketch(name='test')
```

that way you create it,b ut also ensure that you get the correct sketch back to work with moving forward.

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+   "source": [
+    "import_helper = helper.ImportHelper() \n",
+    "\n",
+    "with importer.ImportStreamer() as streamer:\n",
+    "  streamer.set_sketch(sketch)\n",
+    "  streamer.set_config_helper(import_helper) \n",
+    "\n",
+    "  streamer.set_timeline_name('sigma_events.jsonl')\n",
+    "\n",
+    "  streamer.add_file('../test_tools/test_events/sigma_events.jsonl')"
```
kiddinn, line 115: there is no guarantee here that this file exists... this depends on how you start your notebook, how it is configured, and the paths you've got there....

you can also fetch the file from GH to ensure it's there and correct

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "for rule in rules:\n",
+    "    rule.__dict__"
```
kiddinn, line 141: do we want to expose the `__dict__` here? Rather not just have a `print(rule.X)` or `print(f'{rule.X} something {rule.Y}')` instead of the dict....

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+    "timelines = sketch.list_timelines()\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "timelines[0].name #sigma_events.jsonl"
```
kiddinn, line 309: well... why is that the sigma_events timeline? what if you had added other timeline there before?

you could do something like

```python
timeline = None
for timeline_ in timelines:
  if timeline_.name == 'sigma_events.jsonl':
    timeline = timeline_
    break
```

and something like that, then you could have a guarantee....

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,340 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "result = sketch.run_analyzer(timeline_name='sigma_events.jsonl', analyzer_name='sigma', ignore_previous=True)"
```
kiddinn, line 318: also, this is not how you run the analyzer, you will get a deprecation warning printed out, and soon this will be removed... better to do

```python
result = timeline.run_analyzer(analyzer_name='sigma', ignore_previous=True)
```

kiddinn (changes requested):
mostly comments on the notebook

kiddinn:
I'm also missing some more text in the notebook, some further explanations on what is happening, why, what you are trying to achieve, etc, etc, etc.

notebooks/Sigma_test_Notebook.ipynb
jaegeral: have not used picatrix in that example

notebooks/Sigma_test_Notebook.ipynb
jaegeral: good point

notebooks/Sigma_test_Notebook.ipynb
jaegeral: will remove the cell, the idea what is in the thing is in the next cell already

notebooks/Sigma_test_Notebook.ipynb
kiddinn: no, I don't see any picatrix references, I was just. wasn't sure what `requirements` would be here,  but if there are no dependencies on either picatrix nor the api client, then just having the `pip install` of hte api client is fine,

notebooks/Sigma_test_Notebook.ipynb
jaegeral: oh that makes it easier then cool

notebooks/Sigma_test_Notebook.ipynb
jaegeral: ack

notebooks/Sigma_test_Notebook.ipynb
jaegeral: GH download it is

jaegeral:
> I'm also missing some more text in the notebook, some further explanations on what is happening, why, what you are trying to achieve, etc, etc, etc.

Good point, added a littel more text of what is going on, hope that helps.

jaegeral:
Not sure why the tests are currently only queued I tried to restart them, but the changes since last run only affected the notebook.

kiddinn:
> Not sure why the tests are currently only queued I tried to restart them, but the changes since last run only affected the notebook.

don't worry about that... there is a build-up queue in the GH actions for the google org, so that's why it's taking some time. I cancelled all your runs, since I'll need to re-sync with master.

I'll do that once the queue has caught up with all the currently running GH actions in the TS repo

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -86,18 +86,84 @@
...
-    "sketch = ts_client.get_sketch(1)"
+    "# Download a file to import\n",
+    "import requests\n",
+    "import io\n",
+    "\n",
+    "url = \"https://raw.githubusercontent.com/google/timesketch/master/test_tools/test_events/sigma_events.csv\" \n",
+    "download = requests.get(url).content"
    ]
   },
+  {
```
kiddinn, line 107: clear the output of all cells before saving it

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -86,18 +86,84 @@
...
+     },
+     "metadata": {},
+     "execution_count": 10
+    }
+   ],
+   "source": [
+    "# Now we convert the CSV to a pandas dataframe\n",
```
kiddinn, line 153: I also prefer to hve these comments as a markdown cell, rather than being a comment inside a code cell

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -86,18 +86,84 @@
...
+    "df = pd.read_csv(io.StringIO(download.decode('utf-8')))\n",
+    "df"
```
kiddinn, line 155: again, clear the output of the cells, but also you can simply do a `df.head(4)` or something like that to show only the first four entries

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -138,7 +205,7 @@
    "outputs": [],
    "source": [
     "for rule in rules:\n",
-    "    rule.__dict__"
+    "    rule.id"
```
kiddinn, line 208: don't want to do something more here?

```python
print(f'ID: {rule.id}')
print('something a bit more than the id?')
```
or somethign like that...

kiddinn (commented):
few more comments

notebooks/Sigma_test_Notebook.ipynb
jaegeral: good point. yes

jaegeral:
ready for next round of review

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,358 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "ts_client = config.get_client()"
+   ]
+  },
+  {
+   "source": [
+    "# We create a new sketch\n",
```
kiddinn, line 85: why is this a header?

notebook_screenshot

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,358 @@
...
+    "\n",
+    "That will be called test and will be filed with some test data from the Github repository.\n",
```
kiddinn, line 87: s/filed/filled/

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,358 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import requests\n",
+    "import io\n",
```
kiddinn, line 110: please keep all imports in the same place, all at the top

notebooks/Sigma_test_Notebook.ipynb
```diff
@@ -0,0 +1,358 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sketch.get_analyzer_status()\n",
+    "\"\"\"This should show something like:\n",
```
kiddinn, line 346: take this comment out of the code cell and put it in a markdown cell instead....

kiddinn (changes requested):
ok, almost there, just small nits and then we are good to go

jaegeral:
> ok, almost there, just small nits and then we are good to go

The nits have been knitted

kiddinn (approved):
LGTM

PR #1732, opened 2021-04-07
Moved the Notebook.md to learn/notebook.md to correspond to where the docs point to it.

kiddinn:
The documentation ATM points to learn/notebook.md but it is actually stored in /Notebook.md . this PR moves it to the correct spot.


jaegeral (approved):
LGTM

PR #1734, opened 2021-04-08
Introduce os filsystem in sigma mapping

jaegeral:
That is especially useful if plaso parsed image timelines are ingested into TS.


kiddinn (approved):

kiddinn (approved):
LGTM

PR #1735, opened 2021-04-08
introduce tags in the sample rule and add tags from the rule to the s…

### LINKED ISSUE #1704: Sigma tags as labels / tags for events
Sigma rules do come with something like:

```
tags:
    - attack.initial_access
    - attack.t1190
   ```

It could be useful to apply those tags to events that are found with the Sigma analyzer that does tag / label events. Instead (or in addition) of just the filename, the tags from the Sigma rule could be applied.

jaegeral:
This PR adds functionality to the Sigma analyzer if a sigma rule has tags, those tags will be added to the matching event as well.

**Checks**

- [x] All tests succeed.
- [x] Documentation updated.

**Closing issues**
https://github.com/google/timesketch/issues/1704
Put `closes #1704` in your comment to auto-close the issue that your PR fixes
(if such).


kiddinn (approved):
LGTM

PR #1738, opened 2021-04-11
Set timeline ID in the GCS importer

berggren:
Quick fix for the GCS importer.

### DIFF
--- a/contrib/gcs_importer.py
+++ b/contrib/gcs_importer.py
@@ -77,7 +77,7 @@ def setup_sketch(timeline_name, index_name, username, sketch_id=None):
         sketch_id: (str) Optional sketch_id to add timeline to
 
     Returns:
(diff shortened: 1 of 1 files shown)


contrib/gcs_importer.py
```diff
@@ -77,7 +77,7 @@ def setup_sketch(timeline_name, index_name, username, sketch_id=None):
         sketch_id: (str) Optional sketch_id to add timeline to
 
     Returns:
-        (str) Sketch ID
+        (str) Sketch ID and Timeline ID
```
kiddinn, line 80: technically it doesn't return a string, but a tuple with the sketch ID and the timeline ID (aren't both of them integers?)

kiddinn (commented):
one tiny tiny comment

kiddinn (approved):
LGTM

Issue #1754, opened 2021-04-30
Jsonl content too long


lprat:
**Is your feature request related to a problem? Please describe.**
When I upload jsonl, in some case, the length of line is very too long. When API upload flush events to elastic then it get 'error 413' (line too long) and this error break timeline.
```
[2021-04-29 19:19:14,491] timesketch.tasks/ERROR Error: TransportError(413, '')
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/tasks.py", line 681, in run_csv_jsonl
    es.import_event(
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/datastores/elastic.py", line 983, in import_event
    _ = self.flush_queued_events()
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/datastores/elastic.py", line 1012, in flush_queued_events
    results = self.client.bulk(
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/client/utils.py", line 84, in _wrapped
    return func(*args, params=params, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/client/__init__.py", line 448, in bulk
    return self.transport.perform_request(
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/transport.py", line 351, in perform_request
    status, headers_response, data = connection.perform_request(
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/connection/http_urllib3.py", line 261, in perform_request
    self._raise_error(response.status, raw_data)
  File "/usr/local/lib/python3.8/dist-packages/elasticsearch/connection/base.py", line 181, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
elasticsearch.exceptions.TransportError: TransportError(413, '')
[2021-04-29 19:19:14,719] celery.app.trace/INFO Task timesketch.lib.tasks.run_csv_jsonl[ed224b9758ec43c18862f28350eafbff] succeeded in 51.03816904872656s: None
[2021-04-29 19:19:37,105] timesketch.tasks/INFO Index timeline [spyre_pole3] to index [1619722543.3322954spyre_pole3] (source: jsonl)
[2021-04-29 19:19:38,409] timesketch.elasticsearch/ERROR Errors while attempting to upload events.
[2021-04-29 19:19:38,410] timesketch.elasticsearch/ERROR Unable to upload document, and unable to log the error itself.
[...]

PR #1755, opened 2021-04-30
Fixed an issue with too long lines in JSON ingestion

lprat:
fix issue #1754


timesketch/lib/utils.py
```diff
@@ -247,9 +247,16 @@ def read_and_validate_jsonl(file_handle):
...
     lineno = 0
+    max_length_line_jsonl = current_app.config.get('MAX_LENGTH_LINE_JSONL')
     for line in file_handle:
         lineno += 1
         try:
+            current_len = len(line)
+            if current_len > max_length_line_jsonl:
+                raise RuntimeError(
+                    'Line too long at line {0:n}: {1:s}'.format(
+                        lineno, ','.join(current_len)))
```
kiddinn, line 258: do we want to raise a runtime error here, or simply log this and move on? I mean if you've got a file with 100k events in them, but 2 lines are too long, don't you want all the rest of the lines to be ingested? and just those two not? and then have a way to notify the user that not everythogn got ingested?

kiddinn (commented):
one question here

timesketch/lib/utils.py
lprat: Yes, you are right! It's best way.
I modify...

timesketch/lib/utils.py
```diff
@@ -247,9 +247,18 @@ def read_and_validate_jsonl(file_handle):
...
     for line in file_handle:
         lineno += 1
         try:
+            current_len = len(line)
+            if current_len > max_length_line_jsonl:
+                warning_string = (
+                    'Line too long at line {0}: {1}')
+                logger.warning(warning_string.format(
+                    lineno, current_len))
+                continue
```
kiddinn, line 260: ok, few things here... sorry for the nitpicking...

a) skip the `warning_string` creation, just do

```python
logger.warning(
    'Line {0:d} too long ({1:d} chars which is over the maximum supported value of {2:d}'.format(
        line_number, current_len, max_length_line_jsonl))
```

b) so here the behavior is to skip the line... another option would be to actually ingest it, but instead of skipping it, reducing the line size.. would mean a bit more logic here... that means we need to put a flag here for instance, then load the thing as a JSON, to get a dict, and then figure out what parameter is the `long` one and then shorten it... maybe we can put a TODO here to add that in, not do that now, just have the current behavior of skipping teh line as it is, but it might be better for a future follow up to have the behavior to log the issue, but at the same time ingest the line with a shortened field(s) to make sure it still fits within the size boundaries...

c) instead of `lineno` can you rename that to a more descriptive `line_number`

kiddinn (changes requested):
one minor comment, PTAL

timesketch/lib/utils.py
```diff
@@ -246,12 +246,46 @@ def read_and_validate_jsonl(file_handle):
...
+        line_number += 1
         try:
             linedict = json.loads(line)
             ld_keys = linedict.keys()
+
+            current_len = len(line)
+            if current_len > max_len_line_jsonl:
+                field_len={}
+                total_len=0
+                for x in ld_keys:
```
kiddinn, line 262:
```suggestion
                for key in ld_keys:
```

timesketch/lib/utils.py
```diff
@@ -246,12 +246,46 @@ def read_and_validate_jsonl(file_handle):
...
+                    if isinstance(linedict[x], str):
```
kiddinn, line 263:
```suggestion
                    if isinstance(linedict[key], str):
```

timesketch/lib/utils.py
```diff
@@ -246,12 +246,46 @@ def read_and_validate_jsonl(file_handle):
...
+                        field_len[x]=len(linedict[x])
+                        total_len+=field_len[x]
+                for x in ld_keys:
```
kiddinn, line 266:
```suggestion
                for key in ld_keys:
```

kiddinn: why loop over this twice? you can also just say 

```python
for key, key_length in field_len.items():
```

and not needing to call up `field_len[key]` all the time

timesketch/lib/utils.py
```diff
@@ -246,12 +246,46 @@ def read_and_validate_jsonl(file_handle):
...
+                    if field_len[x] > max_len_line_jsonl:
+                        if (total_len-field_len[x]) > max_len_line_jsonl:
```
kiddinn, line 268: ok, let's imagine this... total length possible is 50 characters, and you've got a line that is 80 characters long, ok, so it is too long, but you go over each of the fields and all of them are just 40 characters long (2 columns in the line), so you'll have here
```
if (80 - 40) > 50:
```
so `if 40 > 50`, all of these will not trigger, however the total length is still longer than the maximum allowed

the question is how is it best to handle that?

a) you can do something `advanced` or you can start by doing the simple thing

1. check if length of any field is greater than the max, and then reduce it by the amount needed to have the total line length below the threshold
  --> this  would not work in the above example, so you'll move on

2. go through the fields and as soon as you start to get close to the limit you just cut all of the remaining fields shorter...
    --> first column is 40 chars, and then the second one is only 7 chars plus '...' to indicate it was cut short

3. or you calculate what fields are the longest and then you cut each one of them by enough characters to fit 
    --> This is the alternative, you can have then field A as 22 chars + '...' and then field B as 22 chars and '...', combining total to 50, but still keeping a bit more of field B than in the other approach

kiddinn (changes requested):
some comments

timesketch/lib/utils.py
lprat: Ok thanks for your idea, i will work it

timesketch/lib/utils.py
lprat: fixed

timesketch/lib/utils.py
lprat: fixed

berggren:
Hi @lprat What is the status of this PR?

lprat:
Hi,
I waiting for validation. But maybe @kiddinn has fix it?

Thanks.

timesketch/lib/utils.py
```diff
@@ -246,12 +246,75 @@ def read_and_validate_jsonl(file_handle):
...
+                    elif isinstance(linedict[key], list):
+                        field_len[key]=len(str(linedict[key]))
+                        total_len+=field_len[key]
+                field_len = dict(sorted(
+                        field_len.items(),
+                        key=lambda item: item[1],
+                        reverse=True)
+                    )
+                cur_len_fields = 0
+                reduc_field = []
```
kiddinn: OK, I would like to see a comment here to explain a bit the logic here, since it seems to go into a bit more complicated calculations here... would be good for potential debugging sessions...

why are you dividing by 100 and multiplying by 10? why these subtractions and divisions, etc... can you explain the logic  better?

kiddinn (changes requested):
sorry for the late review, I did a minor change, so make sure you sync your branch and then can you answer the comment I put in there?

timesketch/lib/utils.py
lprat: Hi,
I tried to remember my idea!

https://github.com/lprat/timesketch/blob/proposal_issue_1754/timesketch/lib/utils.py#L290 : I search fields with long data (sorted), i put key in reduc_field, when lenght of the remaining fields  is smaller than max_lent_line_jsonl, I found all keys to reduce.

https://github.com/lprat/timesketch/blob/proposal_issue_1754/timesketch/lib/utils.py#L292 :
current_max: I take 90% of max_len_line_jsonl (for json structure not calculated in length). It may be too much ... Or unnecessary ...

https://github.com/lprat/timesketch/blob/proposal_issue_1754/timesketch/lib/utils.py#L295 : I am looking for the size that remains free (allocated to field to reduce)

https://github.com/lprat/timesketch/blob/proposal_issue_1754/timesketch/lib/utils.py#L296 : I get this size for each field to reduce.

After, I try to reduce fields in reduc_field[]

jaegeral:
@lprat  could you check this PR? It is outstanding for quite a while with open conversations and conflicts that make it impossible to merge, could you please have a look?

Thx a lot

lprat:
@jaegeral sorry for the old PR. I changed fix to simply log error when line json is too long.

timesketch/lib/utils.py
```diff
@@ -259,11 +259,17 @@ def read_and_validate_jsonl(file_handle):
     # Fields that must be present in each entry of the JSONL file.
     mandatory_fields = ['message', 'datetime', 'timestamp_desc']
     lineno = 0
+    max_len_line_jsonl = current_app.config.get('MAX_LENGTH_LINE_JSONL')
```
jaegeral, line 262: Can you add a default value here in case the config variable is not set (which will happen in a lot of cases where people do not update their config file to match the recent version)?

jaegeral (commented):
Small change PTAL

timesketch/lib/utils.py
lprat: Done

jaegeral:
Looks like we have not merged that in over a year and as far as I can tell, we have not seen any issues from other people, so I am going to close this out, thx lprat still for the contribution, if one would run in the issue again we can re-visit the PR.

PR #1757, opened 2021-04-30
Search history

### LINKED ISSUE #653: Query history stack
I personally find the back button behavior very tricky when working on Timesketch.

As with many analysis tools, one tends to start from a query and "narrow down from there". I find myself often wanting to "go back" to a previous level of narrowing-down, which I intuitively do with my browser's back button. This takes me to the actual previous page which is usually the timeline view, losing track of the view I had before.

Ideas that spring to mind:

* Store a stack history of search queries.
  * When the "back" button is pressed, you go back one level on the stack.
  * This allows for `Ctrl+r`-style searching as well for "queries that worked"
* From that stack, allow the user to select which queries brought the most significant results or findings.
  * This could be very useful for better understanding an analyst's workflow when they hand over an investigation.

### LINKED ISSUE #1639: Search history
Implement Search history per sketch/user to enable navigate between searches. The feature will be implemented as a new DB model. Idea is to implement as an adjacent list in SQL:

```
class SearchHistory(BaseMode;):
    id = Column(Integer, primary_key=True)
    parent_id = Column(Integer, ForeignKey('node.id'))
    query_string = Column(String(50))
    query_filter = Column(String(50))
    query_dsl = Column(String(50))
    children = relationship("SearchHistory",
         backref=backref('parent', remote_side=[id])
     )
```

With this we can represent the history as a tree, with branches off searches.

```
root --+---> child1
       +---> child2 --+--> subchild1
       |              +--> subchild2
       +---> child3
```

berggren:
Implement Search history per sketch/user to enable navigate between searches. The feature will be implemented as a new DB model with an adjacent list in SQL.

* New DB model `SearchHistory` as a self referencing adjency list
* New UI component to display an interactive tree of your searches
* Annotation of Search history, i.e. record all user action based on a query such as labeling, star, comment on events.
* Ability to travel back in the tree and branch off.
* Automatic scroll to the current search node
* Zoom in/out for large search trees.

**Screenshot**
Screenshot 2021-05-07 at 09 48 13

**Closing issues**
Closes #653 
Closes #1639


timesketch/models/sketch.py
```diff
@@ -732,3 +738,74 @@ def __init__(self, timeline, user, provider, context, file_on_disk,
...
         self.data_label = data_label
         self.error_message = error_message
+
+# Association table for the many-to-many relationship between SearchHistory
+# and Event.
+association_table = Table('searchhistory_event', BaseModel.metadata,
+    Column('searchhistory_id', Integer, ForeignKey('searchhistory.id')),
+    Column('event_id', Integer, ForeignKey('event.id'))
+)
+
```
kiddinn, line 748: nit of the day... two empty lines between top level class/methods

timesketch/models/sketch.py
```diff
@@ -732,3 +738,74 @@ def __init__(self, timeline, user, provider, context, file_on_disk,
...
+    query_string = Column(UnicodeText())
+    query_filter = Column(UnicodeText())
+    query_dsl = Column(UnicodeText())
+    query_result_count = Column(BigInteger())
+    query_time = Column(BigInteger())
+    children = relationship(
+        'SearchHistory',
+        # Cascade deletions
+        cascade='all, delete-orphan',
+        backref=backref("parent", remote_side=id),
```
kiddinn, line 766: another nit... singe vs double quote

timesketch/models/sketch.py
```diff
@@ -732,3 +738,74 @@ def __init__(self, timeline, user, provider, context, file_on_disk,
...
+        """Recursive function to generate full search history tree.
+        Args:
+            node: SearchHistory object as root node.
+            node_dict: Dictionary to use for recursion.
+            recurse: Boolean indicating if the function whould recurse.
+
+        Returns:
+            Dictionary with a Search History tree.
+        """
+        node_dict["id"] = node.id
```
kiddinn, line 790: again, same as before.... single vs double quote...

I would also argue that it would be good to check if `node_dict` is a dict ;)

kiddinn (changes requested):
few minor comments

timesketch/api/v1/resources/explore.py
```diff
@@ -273,6 +279,46 @@ def post(self, sketch_id):
         db_session.add(view)
         db_session.commit()
 
+        # Search History
+        search_node = None
+        new_search_node = SearchHistory(user=current_user, sketch=sketch)
```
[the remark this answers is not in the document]
berggren: No, the object is only _comitted_ to the DB if a few conditions are met. See L315

timesketch/api/v1/resources/event.py
```diff
@@ -623,8 +650,12 @@ def post(self, sketch_id):
                     sketch.id,
                     current_user.id,
                     form.annotation.data,
-                    toggle=toggle)
+                    toggle=toggle
+                )
 
+                if current_search_node:
+                    current_search_node.add_label(form.annotation.data)
```
kiddinn, line 657: why adding the label itself as well, not just the notion that a label has been added? (which is done in the line below)... similar to what is done in comments

timesketch/models/sketch.py
```diff
@@ -732,3 +738,90 @@ def __init__(self, timeline, user, provider, context, file_on_disk,
...
+        self.sketch = sketch
+        self.description = description
+        self.query_string = query_string
+        self.query_filter = query_filter
+        self.query_dsl = query_dsl
+        self.parent = parent
+
+    def build_tree(self, node, node_dict, recurse=True):
+        """Recursive function to generate full search history tree.
+        Args:
```
kiddinn, line 795: nit, an empty line before Args section

timesketch/models/sketch.py
```diff
@@ -732,3 +738,90 @@ def __init__(self, timeline, user, provider, context, file_on_disk,
...
+        node_dict['created_at'] = node.created_at
+        node_dict['parent'] = node.parent_id
+        node_dict['query_string'] = node.query_string
+        node_dict['query_filter'] = node.query_filter
+        node_dict['query_dsl'] = node.query_dsl
+        node_dict['children'] = []
+
+        children = node.children.values()
+        if children and recurse:
+            for _, child in enumerate(children):
```
kiddinn, line 820: why are you using enumerate if you are not taking advantage of hte index number here?

why not just `for child in children` ?

kiddinn (commented):
minor comments, otherwise LGTM

timesketch/api/v1/resources/event.py
berggren: Yeah, I have been going back and forth on this one. But I think you are right, we should skip it. The idea was to do some UI work to show what labels was added but I think this is too early. We can always add in later if needed. 

I'll remove it for now.

timesketch/models/sketch.py
berggren: Done

timesketch/models/sketch.py
berggren: Good catch - this was a leftover from a depth-control hook I was exploring but abandoned. I have removed it now.

berggren (commented):
@kiddinn PTAL

kiddinn (commented):
can you upgrade the version in the API `timesketch/version.py` before I approve?

kiddinn (approved):
LGTM

Issue #1733, opened 2021-04-08
Improve Exception text of Attributes cannot overwrite values already set


jaegeral:
In the following line:

https://github.com/google/timesketch/blob/beaba0be9f7a213b945d82fed09d6b233c246864/api_client/python/timesketch_api_client/sketch.py#L1520

It would be easer to tell, which of the passed attributes are duplicate. Granted the current way is easier to code, but from a user perspective it would be good to know which of the attributes caused the issue esp. if you have more than one.

PR #1758, opened 2021-05-03
Changed how duplicate attributes are displayed when adding manual eve…

warrinot:
Changed how duplicate attributes are displayed when adding manual events in the sketch API client. Issue #1733


kiddinn (approved):
LGTM

PR #1764, opened 2021-05-05
[DRAFT] Add ability to search for defanged indicators

itsmvd:
Gives users the ability to search for defanged indicators without having to unescape them, which the UI will now do for them. A user can now directly search for a domain such as timesketch(.)org, an IP like 8[.]8[.]8[.]8 or a URL like hxxps://timesketch.org.


timesketch/frontend/src/views/SketchExplore.vue
```diff
@@ -459,6 +459,11 @@ export default {
       // Update with selected fields
       this.currentQueryFilter.fields = this.selectedFields
 
+      // Escape defanged indicators (such as timesketch(.)org or 8.8.8[.]8)
+      this.currentQueryString = this.currentQueryString.replace(/\((?:\.|dot)\)/g, '.')
+      this.currentQueryString = this.currentQueryString.replace(/\[(?:\.|dot)\]/g, '.')
+      this.currentQueryString = this.currentQueryString.replace(/hxxp(s?):\/\//g, 'http$1://')
```
kiddinn, line 465: what if that's what you are searching for? I mean what if a string contains hxxp://foobar.com, but now you silently replace it without really telling me, isn't that confusing for the end user?

kiddinn (commented):
one question

berggren:
Closing this as discussed offline. We will approach this in the UI and suggesting new query instead.

Issue #1762, opened 2021-05-04
Break up Tags added by Sigma analyzer


jaegeral:
Right now the Sigma analyzer adds two types of Tags to an event:
- sigma_$FILENAME
- Tags provided by the Sigma rule itself

Depending on the rule, this can clutter the Events list quite a bit.

To clean it up, the suggestion is:
- Add the Filename as an attribute (list of strings) to the matching event as ```sigma_ts_filename:$FILENAME```
- Add all tags provided by sigma that match ```attack.*``` to a new attribute (list of strings) called ```ttp_ts:attack.foobar```
- rest of the tags (if any) would go to the normal ```tags```attribute.

tomchop:
I think this is a good idea, however I would either:

* Pick a more generic prefix than `attack.*`, as some matching TTPs won't always match those in the ATT&CK framework (I assume this what the `attack` is about)
* Alternatively, make the list of prefixes that add tags to attributes easy to extend (e.g. if I want to add another prefix such as `yeti.*` or `misp.*`

jaegeral:
yes, it will be easily extendable, right now as far as I can tell, the Sigma rules only have attack.* in them. But extending them is a good idea, I will add that.

PR #1766, opened 2021-05-05
Updated Sigma tagging, few improvements

jaegeral:
There are a lot of improvements along the Sigma tagger functionality.

Besides others, it fixes: https://github.com/google/timesketch/issues/1762


test_tools/sigma_verify_rules.py
```diff
@@ -104,9 +159,12 @@ def move_problematic_rule(filepath, move_to_path, reason=None):
             file_objec.write(f'{filepath}\n{reason}\n\n')
 
         base_path = os.path.basename(filepath)
+        logging.info('Moving the rule: {0:s} to {1:s}'.format(
+        filepath, f'{move_to_path}{base_path}'))
```
kiddinn, line 163: indentation is weird here

test_tools/sigma_verify_rules.py
```diff
@@ -104,9 +159,12 @@ def move_problematic_rule(filepath, move_to_path, reason=None):
...
         os.rename(filepath, f'{move_to_path}{base_path}')
```
kiddinn, line 164: use the os.path.join instead of `move_path}{base_path}`

test_tools/sigma_verify_rules.py
```diff
@@ -104,9 +159,12 @@ def move_problematic_rule(filepath, move_to_path, reason=None):
...
-    except OSError:
-        logger.error('OS Error - no rules moved')
+    except OSError as e:
```
kiddinn, line 165:
```suggestion
    except OSError:
        logger.error('OS Error - rule not moved', exc_info=True)
```

then you don'\t need the second logger command and don't need to specify a variable for the exception

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,22 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
         tagged_events_counter = 0
+        # TODO: tune the below line to be more effective
         events = self.event_stream(
             query_string=query, return_fields=return_fields)
         for event in events:
-            event.add_tags(['sigma_{0:s}'.format(rule_name)])
-            event.add_tags(tag_list)
+            ts_sigma_rule = event.source.get('ts_sigma_rule',[])
+            ts_sigma_rule.append(rule_name)
+            event.add_attributes({'ts_sigma_rule': list(set(ts_sigma_rule))})
```
kiddinn, line 44: what about if the rule is already there? I know that you deal with it using the `list(set(....` but if it's already there, then there is no need to call the `add_attributes` call, that is you can save yourself an API Call if it's there already... why not

```python
if rule_name not in ts_sigma_rule:
    ts_sigma_rule.append...
    event.add_attri....
```

otherwise you just don't call the `add_attributes` and then save yourself an API call

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,22 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
+            ts_ttp = event.source.get('ts_ttp',[])
+            for tag in tag_list:
+                print(tag)
```
kiddinn, line 47: debugging statemen t

jaegeral:
P.s. die functionality to check the blocklist csv at runtime is not yet implemented, I plan to do that in the next itteration, but that comes with new config variable and such, so I rather have smaller PRs.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,22 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
+                if tag.startswith(('attack.','car.')):
+                    ts_ttp.append(tag)
+                else:
+                    event.add_tags([tag])
```
kiddinn, line 51: why do this, that is for each tag you call the `add_tags` function, instead of just calling it once with the list of tags you want to add, that way you can save yourself multiple API Calls

kiddinn (changes requested):
few early comments

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: Yes that is a valid question, however, rules might change, esp more tags added to things, so I think "spending" the API call here is okay.

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: Good point, hope my new code is clever enough, I basically remove elements that will be added to the new attribute from the taglist and then take what is left.

test_tools/sigma_verify_rules.py
```diff
@@ -18,21 +18,58 @@
...
 
 from timesketch.lib import sigma_util# pylint: disable=no-name-in-module
 
 logger = logging.getLogger('timesketch.test_tool.sigma-verify')
 logging.basicConfig(level=os.environ.get('LOGLEVEL', 'INFO'))
 
 
+
+def get_sigma_blocklist(blocklist_path=None):
+    """Get a List of paths to ignore.
```
kiddinn, line 40: paths or sigma rules?

test_tools/sigma_verify_rules.py
```diff
@@ -18,21 +18,58 @@
...
+
+    Args:
+        blocklist_path: Optional path to a blocklist file
+    Returns:
```
kiddinn, line 44: empty line between args/returns/raises section

test_tools/sigma_verify_rules.py
```diff
@@ -18,21 +18,58 @@
...
+                config_file_path))
+
+    if not os.access(config_file_path, os.R_OK):
+        raise ValueError(
+            'Unable to open file: [{0:s}], cannot open it for '
+            'read, please check permissions.'.format(config_file_path))
+
+    ignore = pd.read_csv(config_file_path)
+
+    return ignore
```
kiddinn, line 71: why not just 

```python
return pd.read_csv(config_file_path)
```

test_tools/sigma_verify_rules.py
```diff
@@ -65,6 +102,8 @@ def run_verifier(rules_path, config_file_path):
     return_verified_rules = []
     return_rules_with_problems = []
 
+    ignore = get_sigma_blocklist()
```
kiddinn, line 105: no path supplied? are you sure this file will then exist in all installations?

test_tools/sigma_verify_rules.py
```diff
@@ -76,8 +115,24 @@ def run_verifier(rules_path, config_file_path):
...
+                        logging.info('{0:s} Ignoring because {1:s}'.format(
+                            rule_file_path,item))
+
+                if block_because_csv:
+                    continue
+
+                try:
+                    parsed_rule = sigma_util.get_sigma_rule(
+                        rule_file_path, sigma_config)
+                except Exception:# pylint: disable=broad-except
```
kiddinn, line 133: why the broad exception here? I really really dislike broad exceptions, and if it's absolutely necessary then please document why that is ht ecase? I have a feeling that there is a finite amount of exceptions that are raised and it woudl be better to handle those rather than have this broad catch all here

test_tools/sigma_verify_rules.py
kiddinn: but you didn't fix the indentation it's still not correct

```python
logging.info('Moving the rule: {0:s} to {1:s}'.format(
    filepath, ...
```

I'm still seeing:
```python
logging.info('Moving the rule: {0:s} to {1:s}'.format(
filepath, ...
```

test_tools/sigma_verify_rules.py
```diff
@@ -104,9 +159,11 @@ def move_problematic_rule(filepath, move_to_path, reason=None):
             file_objec.write(f'{filepath}\n{reason}\n\n')
 
         base_path = os.path.basename(filepath)
-        os.rename(filepath, f'{move_to_path}{base_path}')
+        logging.info('Moving the rule: {0:s} to {1:s}'.format(
+        filepath, f'{move_to_path}{base_path}'))
+        os.rename(filepath, os.path.join(move_to_path,base_path))
```
kiddinn, line 164: nit, space between the comma (`move_to_path, base_path`)

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,21 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
         """
         return_fields = []
         tagged_events_counter = 0
+        # TODO: tune the below line to be more effective
```
kiddinn, line 38: can you elaborate? what's ineffective, why not fix it now?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,21 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
         events = self.event_stream(
             query_string=query, return_fields=return_fields)
         for event in events:
-            event.add_tags(['sigma_{0:s}'.format(rule_name)])
+            ts_sigma_rule = event.source.get('ts_sigma_rule',[])
```
kiddinn, line 42: nit, space after the comma

kiddinn: another nit... this is a list, so a better name would be `ts_sigma_rules` (plural)

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,21 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
+            ts_sigma_rule.append(rule_name)
+            event.add_attributes({'ts_sigma_rule': list(set(ts_sigma_rule))})
+            ts_ttp = event.source.get('ts_ttp',[])
```
kiddinn, line 45: nit, space after comma

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -35,11 +35,21 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
...
+            for tag in tag_list:
+                if tag.startswith(('attack.','car.')):
```
kiddinn, line 47: nit, space after comma

kiddinn: can you add a comment to explain why these two? why attack. and car.?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -69,14 +77,22 @@ def run(self):
                     rule.get('es_query'), rule.get('file_name'),
                     tag_list=rule.get('tags'))
                 tags_applied[rule.get('file_name')] += tagged_events_counter
+                time.sleep(0.5)
```
kiddinn, line 80: why this sleep? is that really necessary, if a sleep is necessary then please document why, and why the 0.5?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -69,14 +77,22 @@ def run(self):
...
+                if sigma_rule_counter % 10 == 0:
+                    logger.debug('Rule {0:d}/{1:d}'.format(
+                        sigma_rule_counter,len(sigma_rules)))
```
kiddinn, line 83: nit, space after comma

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -69,14 +77,22 @@ def run(self):
...
             except elasticsearch.TransportError as e:
                 logger.error(
                     'Timeout executing search for {0:s}: '
                     '{1!s} waiting for 10 seconds'.format(
                         rule.get('file_name'), e), exc_info=True)
                 # this is caused by to many ES queries in short time range
                 # thus waiting for 10 seconds before sending the next one.
-                time.sleep(10)
+                time.sleep(15)
```
kiddinn, line 91: same with this sleep here (as above)

timesketch/lib/sigma_util.py
```diff
@@ -230,6 +230,10 @@ def get_sigma_rule(filepath, sigma_config=None):
         sigma_es_query = ''
 
         for sigma_rule in parsed_sigma_rules:
+            # TODO Investigate how to handle .keyword
+            # fields in Sigma.
+            # https://github.com/google/timesketch/issues/1199#issuecomment-639475885
+            sigma_rule = sigma_rule.replace(".keyword:", ":")
```
kiddinn, line 236: ' vs "

timesketch/lib/sigma_util.py
```diff
@@ -279,14 +283,13 @@ def get_sigma_rule_by_text(rule_text, sigma_config=None):
...
-
     # TODO check if input validation is needed / useful.
     try:
-        parser = sigma_collection.SigmaCollectionParser(
-            rule_text, sigma_conf_obj, None)
-        parsed_sigma_rules = parser.generate(sigma_backend)
         rule_yaml_data = yaml.safe_load_all(rule_text)
         for doc in rule_yaml_data:
+            parser = sigma_collection.SigmaCollectionParser(
+                str(doc), sigma_conf_obj, None)
```
kiddinn, line 291: why the `str` casting?

timesketch/lib/sigma_util.py
```diff
@@ -308,6 +311,7 @@ def get_sigma_rule_by_text(rule_text, sigma_config=None):
     sigma_es_query = ''
 
     for sigma_rule in parsed_sigma_rules:
+        sigma_rule = sigma_rule.replace(".keyword:", ":")
```
kiddinn, line 314: ' vs "

kiddinn (changes requested):
few more comments

test_tools/sigma_verify_rules.py
jaegeral: I will make a comment, I am afraid as this method in the file is made to catch which rule is causing issues, I really do not know what exceptions could be caused, and the idea of that method is catch those and make a useful list for people, so raising the exception and ending the execution would be against the purpose of this method / tool but I will add more comment to it

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: nit is gone. attack. and car. are the ones that are currently present in Sigma rules that IMHO qualify as TTPs.

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: would you rather have that a config variable to be set and then default values here if no config variable is set?

test_tools/sigma_verify_rules.py
jaegeral: took care of that.

timesketch/lib/analyzers/sigma_tagger.py
kiddinn: at least I would like this number, since it is a constant to be defined as such, and have a comment explaining the reasoning for this, and I'm really curious in why there is a sleep here, what is the actual reason for it?

timesketch/lib/sigma_util.py
```diff
@@ -230,6 +230,10 @@ def get_sigma_rule(filepath, sigma_config=None):
         sigma_es_query = ''
 
         for sigma_rule in parsed_sigma_rules:
+            # TODO Investigate how to handle .keyword
```
jaegeral, line 233: That was in previously already and needs to be added back

jaegeral:
Ok added the sleep timer to a config variable, and added documentation why this value exists as well as what recommended values for more powerful installations.

jaegeral:
The flask error seems to be not related to this PR, so I hope it is fine to ignore.

test_tools/sigma_verify_rules.py
```diff
@@ -76,8 +116,27 @@ def run_verifier(rules_path, config_file_path):
                     continue
 
                 rule_file_path = os.path.join(dirpath, rule_filename)
-                parsed_rule = sigma_util.get_sigma_rule(
-                    rule_file_path, sigma_config)
+
+                block_because_csv = False
+                for item in ignore['path']:
```
kiddinn, line 121: ignore is a dataframe, so I would just do

```python
ignore_list = list(ignore['path'].unique())
```

and then you can iterate over that, so you'll only need to do that once, not once for every filepath and for every rule

kiddinn: ok, so what you are doing is iterating over all the ignore paths and checking each one of the ignore paths if they are in the rule file path... is not possible to do something like `if rule_file_path in ignore_list`? or normalize the rule path first and check if it is there?

test_tools/sigma_verify_rules.py
```diff
@@ -76,8 +116,27 @@ def run_verifier(rules_path, config_file_path):
...
+
+                if block_because_csv:
+                    continue
+
+                try:
+                    parsed_rule = sigma_util.get_sigma_rule(
+                        rule_file_path, sigma_config)
+                # This except is to keep the unknown exceptions
+                # this function is made to catch them and document
+                # them the broad exception is needed
```
kiddinn, line 136: can you reword this comment?

something like
```
# Wide exception handling since there are multiple exceptions that can be raised by the underlying sigma library.
```

OR something like that, this documentation is just confusing...

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -38,8 +40,17 @@ def run_sigma_rule(self, query, rule_name, tag_list = None):
         events = self.event_stream(
             query_string=query, return_fields=return_fields)
         for event in events:
-            event.add_tags(['sigma_{0:s}'.format(rule_name)])
+            ts_sigma_rules = event.source.get('ts_sigma_rule', [])
+            ts_sigma_rules.append(rule_name)
+            event.add_attributes({'ts_sigma_rule': list(set(ts_sigma_rules))})
+            ts_ttp = event.source.get('ts_ttp', [])
+            for tag in tag_list:
+                if tag.startswith(('attack.', 'car.')):
```
kiddinn, line 48: can you add a comment about why we hav `attack.` and `car.` handled in a special way?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -69,14 +78,22 @@ def run(self):
...
                 tags_applied[rule.get('file_name')] += tagged_events_counter
+                if sigma_rule_counter % 10 == 0:
+                    logger.debug('Rule {0:d}/{1:d}'.format(
+                        sigma_rule_counter, len(sigma_rules)))
             except elasticsearch.TransportError as e:
                 logger.error(
                     'Timeout executing search for {0:s}: '
                     '{1!s} waiting for 10 seconds'.format(
                         rule.get('file_name'), e), exc_info=True)
                 # this is caused by to many ES queries in short time range
```
kiddinn, line 89: I know this comment is old, but can you have it start with a capital and `s/to/too/`

kiddinn: also, why have a sleep here? It seems like a duck tape solution to a leak... maybe just fix the underlying issue here, or wait until rules are done before running a new one, can't we do something else to limit the searches (buffer, wait until others are done, etc, etc, etc) rather than have an arbitrary sleep command?

kiddinn (changes requested):
few more comments

test_tools/sigma_verify_rules.py
jaegeral: Oh yes, your suggestion is spot on

test_tools/sigma_verify_rules.py
```diff
@@ -18,27 +18,65 @@
...
 logging.basicConfig(level=os.environ.get('LOGLEVEL', 'INFO'))
 
 
-def run_verifier(rules_path, config_file_path):
+def get_sigma_blocklist(blocklist_path=None):
+    """Get a dataframe of sigma rules to ignore.
+    This includes filenames, paths, ids.
+
+    Args:
+        blocklist_path: Optional path to a blocklist file
```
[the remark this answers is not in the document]
jaegeral: done

test_tools/sigma_verify_rules.py
jaegeral: yes, much better

test_tools/sigma_verify_rules.py
```diff
@@ -18,27 +18,65 @@
...
 from timesketch.lib import sigma_util# pylint: disable=no-name-in-module
 
 logger = logging.getLogger('timesketch.test_tool.sigma-verify')
 logging.basicConfig(level=os.environ.get('LOGLEVEL', 'INFO'))
 
 
-def run_verifier(rules_path, config_file_path):
+def get_sigma_blocklist(blocklist_path=None):
+    """Get a dataframe of sigma rules to ignore.
+    This includes filenames, paths, ids.
```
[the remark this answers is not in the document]
kiddinn: it's resolved but I still don't see the new line

```python
"""Get a dataframe of sigma rules to ignore.

    This includes filenames, paths, ids.
```

test_tools/sigma_verify_rules.py
```diff
@@ -76,8 +118,24 @@ def run_verifier(rules_path, config_file_path):
...
+                    continue
+
+                try:
+                    parsed_rule = sigma_util.get_sigma_rule(
+                        rule_file_path, sigma_config)
+                # This except is to keep the unknown exceptions
+                # this function is made to catch them and document
+                # them the broad exception is needed
+                except Exception:# pylint: disable=broad-except
+                    return_rules_with_problems.append(rule_file_path)
```
kiddinn, line 137: do we not want to log this as well?

timesketch/lib/sigma_util.py
kiddinn: wait, so is this sometimes a dict and sometimes a string?

I fell like that should be handled better than just by string casting

kiddinn: the upstream function does this:
```
self.yamls = yaml.safe_load_all(content)
```

So it expects this to be a string, more specifically a YAML string.... so if we have it as a dict, we should dump it as a YAML string, rather than string cast it

kiddinn (changes requested):
only minor comments

test_tools/sigma_verify_rules.py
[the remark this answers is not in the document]
jaegeral: ah stupid me added a blank line after the whole comment block instead of the docstring.

test_tools/sigma_verify_rules.py
jaegeral: yes we can, I opt we keep it as debug because otherwise it might get really noisy and by default, the problematic rules are printed at the end of the run anyway, but if then someone enables verbose, they would get the full info, which is actually a good idea.

jaegeral:
hm i cant respond to the latest comments, but this is now again ready to review

kiddinn (approved):
LGTM

PR #1768, opened 2021-05-06
Added regular expressions into the tagging analyzer to provide more flexibility.

kiddinn:
This PR introduces the concept of a regular expression into the tagger analyzer.

+ Tagger analyzer now supports indicating a regular expression and an attribute to run the expression against
+ For each event that the original query finds a regular expression can be run in addition to the query to further narrow down if an event should be tagged or not.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [x] Documentation updated.


timesketch/lib/analyzers/tagger.py
```diff
@@ -58,26 +64,62 @@ def tagger(self, name, config):
...
             event_counter += 1
             event.add_tags(tags)
             event.add_emojis(emojis_to_add)
 
             # Commit the event to the datastore.
             event.commit()
 
-        if create_view and event_counter:
+        if save_search and event_counter:
             self.sketch.add_view(
```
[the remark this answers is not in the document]
kiddinn: you mean the `add_view`? I was thinking about that, that's a change in the interface and then a change in most of the analyzers, I think I'll do that in a follow up PR,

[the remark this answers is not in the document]
berggren: Yeah, definitely do that in a follow up. 👍

data/tags.yaml
```diff
@@ -2,46 +2,60 @@
...
-#        view_name: 'TEST view name'
+#        save_search: true
+#        search_name: 'TEST the tag'
+#
+# test_tagger_with_re:
+#        query_string: 'test'
+#        tags: ['test-retag']
+#        emojis: ['FISHING_POLE']
+#        save_search: true
+#        regular_expression: '[a-z][A-C]?://foobarX{0,4}'
```
berggren, line 31: Do we want to add re_flags as well? In features.yaml we do support flags:
```
re: '([a-zA-Z0-9_\.+\-]+@[a-zA-Z0-9\-]+\.[a-zA-Z0-9\-\.]+)'
re_flags: []
```

timesketch/lib/analyzers/tagger.py
```diff
@@ -58,26 +66,62 @@ def tagger(self, name, config):
...
+                attributes = [attribute]
+
         event_counter = 0
-        events = self.event_stream(query_string=query, query_dsl=query_dsl)
+        events = self.event_stream(
+            query_string=query, query_dsl=query_dsl, return_fields=attributes)
 
         for event in events:
+            if expression:
+                value = event.source.get(attributes[0])
```
berggren, line 107: Whats the idea with attributes[0] here - do you plan to support multiple attributes in the future?

berggren (changes requested):
Looks good - a couple of questions

data/tags.yaml
kiddinn: sure, SGTM, I'll add that

timesketch/lib/analyzers/tagger.py
kiddinn: yes, that is the idea, to keep that open, since we might do that

kiddinn (commented):
done, PTAL

berggren (approved):

PR #1769, opened 2021-05-07
refactor and bugfix

berggren:
Refactor Search node tree components (changing filename and location).

+ small bugfix where the search history entry was labeled wrong when adding a star.


kiddinn (approved):
Lgtm

PR #1770, opened 2021-05-11
Add metrics to SearchHistory actions

berggren:
Add a couple of metrics for searchhistory actions.
Also small bugfix in the frontend tree calculation.


kiddinn (approved):
LGTM

PR #1773, opened 2021-05-17
Bugfix: Ensure correct root history node is fetched

berggren:
Bug in how we get the root history node for a user. This fix adds the ordering on the ID of the node. This ensures that we always get the first entry.


berggren:
The PPA unit tests are running a version of Flask that has known issues with flask-restful. There is a fix in the works upstream so this will be resolved in the near future. Let's ignore these errors for now from the TS side.

jaegeral (approved):
LGTM

PR #1776, opened 2021-05-18
Fixed some links pointing to non-existing docs

warrinot:
I fixed links pointing to:
 - Create timeline from JSON/JSONL/CSV file
 - Upload data via the importer/API

Link to 
- Enable Plaso upload via HTTP

is also broken, but i couldn't find corresponding file.


kiddinn (approved):
LGTM

PR #1777, opened 2021-05-18
Fixed image paths for sketch-overview.md

warrinot:


berggren:
These pages are meant to be generated and presented here:
https://timesketch.org/learn/sketch-overview/

So the links are correct as they are. I will close this PR, but thanks for your contribution!

PR #1778, opened 2021-05-18
Bugfix: Navigation links

berggren:
This PR fixes some broken navigation links that got introduced in a recent refactor.


kiddinn (approved):
Lgtm

PR #1784, opened 2021-05-20
Bugfix: Not able to star events

berggren:
Bugfix for empty search history that resulted in annotations (start, comment etc) to fail on some situations.
This PR fixes this, and also adds scrolling to the correct search node when the history window is opened.


jaegeral (approved):
LGTM

PR #1789, opened 2021-05-26
Changed how sketch attributes are stored and read from datastore.

kiddinn:
Making the sketch attributes more modular by defining a specific class for handling encode and decode operation as well as to add a new type (dict) into the mix with a new ontology (intelligence)

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

### DIFF
--- a/timesketch/lib/ontology.py
+++ b/timesketch/lib/ontology.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Ontology for attributes as well as related functions."""
+import json
 
 from timesketch.lib.analyzers import interface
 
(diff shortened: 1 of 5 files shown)


timesketch/lib/ontology.py
```diff
@@ -21,37 +22,175 @@ def ontology():
...
+        if data:
+            return 'true'
+        return 'false'
+
+    @staticmethod
+    def decode(data):
+        """Returns a bool value from the stored string."""
+        if data == 'true':
+            return True
+        return False
```
tomchop, line 133:
```suggestion
        return data == 'true'
```

tomchop (approved):

tomchop (approved):

PR #1791, opened 2021-05-26
Added a data analyzer

kiddinn:
This PR adds the concept of a data analyzer into Timesketch.

This is one of the steps into having the ability to add investigative questions into Timesketch (link to an issue).

What this PR does is to add a YAML file that defines how to look for certain data sets, and then the corresponding code and a task that reads that YAML file and searches for particular data within a certain timeframe and comes back with a True/False verdict whether any data exists that fulfills the requirements of the data analyzer.

In the future scenarios can be constructed that require certain datasets to be present in order for them to be able to run, imagine an investigation with the scenario of `did user X visit website foo.com`, then the pre-cursor to that question can be whether any web history records are in the dataset in the given timeframe of the scenario, as well as maybe logon information, or paths suggesting the history came from a particular person. Then a data analyzer is run on that timeframe that can at least give an indication that the timeframe contains data of the correct type (does not look for the presence of foo.com, but for generic web history data)... then if all the dataset is in place, then a more detailed analyzes can take place, such as to determine whether a visit to foo.com took place... if foo.com did not take place but there is web history, then some conclusion is possible to be drawn (possibly, but not necessarily)


data/data_analyzer.yaml
```diff
@@ -0,0 +1,74 @@
...
+#       attribute: username
+#       regular_expression: "^Administrator"
+
+# ------------------------------------------------------------------------
+windows_administrator_login:
+      description: Login information for Administrator in Windows EVTX files.
+      notes: Collect the SECURITY.EVTX log from a Windows machine.
+      query_string: data_type:"windows:evtx:record" AND event_identifier:4624
+      attribute: username
+      regular_expression: "^Administrator"
```
berggren, line 62: do we want to support re_flags as well?

berggren (changes requested):

data/data_analyzer.yaml
kiddinn: done

timesketch/lib/datafinder.py
```diff
@@ -0,0 +1,185 @@
...
+from timesketch.lib.analyzers import utils
+from timesketch.lib.datastores.elastic import ElasticsearchDataStore
+
+
+logger = logging.getLogger('timesketch.data_finder')
+
+
+class DataFinder:
+    """The data finder class."""
+
```
berggren, line 29: add attributes docstring

timesketch/lib/datafinder.py
```diff
@@ -0,0 +1,185 @@
...
+        """Sets the start date of the time period the data finder uses."""
+        # TODO: Implement a check if this is a valid ISO formatted date.
+        self._start_date = start_date
+
+    def set_timeline_ids(self, timeline_ids):
+        """Sets the timeline identifiers."""
+        self._timeline_ids = timeline_ids
+
+    def find_data(self):
+        """Returns a tuple with a boolen on whether data was found and text.
```
berggren, line 123: text of what?

timesketch/lib/datafinder.py
```diff
@@ -0,0 +1,185 @@
...
+            source = event.get('_source', {})
+            value = source.get(attribute)
+            if not value:
+                logger.warning('Attribute: [{0:s}] is empty'.format(attribute))
+
+            result = expression.findall(value)
+            if not result:
+                continue
+
+            return True, 'Hits Discovered using Regular Expression'
```
berggren, line 183:
```suggestion
            return True, 'Data discovered using Regular Expression'
```

timesketch/lib/datafinder.py
```diff
@@ -0,0 +1,185 @@
...
+            query_filter=query_filter,
+            indices=self._indices,
+            return_fields=attribute,
+            enable_scroll=True,
+            timeline_ids=self._timeline_ids
+        )
+
+        for event in event_generator:
+            if not expression:
+                return True, 'Hits Discovered'
```
berggren, line 172:
```suggestion
                return True, 'Data discovered'
```

timesketch/lib/tasks.py
```diff
@@ -730,3 +736,107 @@ def run_csv_jsonl(
...
+
+    if not data_finder.can_run():
+        results[rule_name] = (False, 'Unable to run the data finder.')
+        return results
+
+    results[rule_name] = data_finder.find_data()
+    return results
+
+
+def find_if_data_exists(
```
berggren, line 817: May call it `run_data_finder` is a bit clearer?

timesketch/lib/tasks.py
```diff
@@ -730,3 +736,107 @@ def run_csv_jsonl(
...
+            logger.error(
+                'Unable to read in YAML config file', exc_info=True)
+            return results
+
+    if rule_name not in data_finder_dict:
+        results[rule_name] = (False, 'Rule not defined')
+        return results
+
+    data_finder = datafinder.DataFinder()
+    data_finder.set_start_date(start_date)
```
berggren, line 794: Any particular reason for using `setters` here instead of just have these as arguments to `data_finder.find_data()`?

E.g:
```
data_finder.find_data(start_date=foo, end_date=bar, parameters= ...)
```

timesketch/lib/tasks.py
```diff
@@ -730,3 +736,107 @@ def run_csv_jsonl(
...
+    sketch = Sketch.query.get(sketch_id)
+    indices = set()
+    for timeline in sketch.active_timelines:
+        if timeline.id not in timeline_ids:
+            continue
+        indices.add(timeline.searchindex.index_name)
+
+    data_finder.set_indices(list(indices))
+
+    if not data_finder.can_run():
```
berggren, line 809: Does this need to be its own function? Why not just do the checks in `data_finder.find_data()` and return `False` from there?

timesketch/api/v1/resources/data.py
```diff
@@ -0,0 +1,122 @@
+# Copyright 2021 Google Inc. All rights reserved.
```
berggren, line 1: `data.py` is a bit generic. Maybe rename to `data_finder.py`?

timesketch/lib/datafinder.py
```diff
@@ -0,0 +1,185 @@
...
+            query_dsl=query_dsl,
+            query_filter=query_filter,
+            indices=self._indices,
+            return_fields=attribute,
+            enable_scroll=True,
+            timeline_ids=self._timeline_ids
+        )
+
+        for event in event_generator:
+            if not expression:
```
berggren, line 171:
```suggestion
            # TODO: Save the result to the Investigation object when that exist in the future.
            if not expression:
```

berggren (changes requested):

timesketch/lib/tasks.py
kiddinn: done

kiddinn (commented):
done with this round

berggren (approved):
LGTM

PR #1792, opened 2021-05-27
Changed how sketch attributes are handled.

kiddinn:
This PR changes few ways on how sketch attributes are handled:

1. Attributes are returned as a dict as an opposed to a list of tuples
2. Attributes can be modified, instead of just added/deleted
3. REST API now uses HTTP methods to determine whether you are adding/modifying or deleting, vs. relying on a `action` variable

### DIFF
--- a/timesketch/api/v1/utils.py
+++ b/timesketch/api/v1/utils.py
@@ -47,8 +47,8 @@ def bad_request(message):
 
 
 def get_sketch_attributes(sketch):
-    """Returns a list of attributes of a sketch."""
-    attributes = []
+    """Returns a dict with all attributes of a sketch."""
+    attributes = {}
     ontology_def = ontology.ONTOLOGY
     for attribute in sketch.attributes:
         if attribute.sketch_id != sketch.id:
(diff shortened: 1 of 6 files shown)


timesketch/api/v1/utils.py
```diff
@@ -71,11 +71,9 @@ def get_sketch_attributes(sketch):
             attribute_values.append(value)
 
         if len(attribute_values) == 1:
-            attributes.append(
-                (name, attribute_values[0], ontology_string))
+            attributes[name] = (attribute_values[0], ontology_string)
```
tomchop, line 74: don't we want something like

```python
attributes.append({
  'name': name,
  'value': attribute_values[0],
  'ontology': ontology_string
})
```

or

```python
attributes[name] = {
  'value': attribute_values[0],
  'ontology': ontology_string
}
```

tomchop (changes requested):

kiddinn:
this is ready for a review, @tomchop PTAL

tomchop (approved):

PR #1793, opened 2021-05-27
UI: Expose sketch attributes

tomchop:
WIP - waiting on https://github.com/google/timesketch/pull/1792 to be merged

-----

Sketch attributes are not visible / discoverable in the web UI.

View a sketch's attributes in table form.

This PR adds a Vue component to display attributes from a sketch. The navbar is adjusted to show the total number of attributes next to the corresponding tab, through a `sketchMeta` prop.

This PR also adds two functions to the `RestApi` client, to get and add attributes.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Screenshot**

image

### DIFF
--- a/timesketch/frontend/src/components/AppNavbarSecondary.vue
+++ b/timesketch/frontend/src/components/AppNavbarSecondary.vue
@@ -64,6 +64,12 @@ limitations under the License.
                   <span>Stories</span>
                 </router-link>
               </li>
+              <li v-bind:class="{'is-active': currentPage === 'attributes'}">
+                <router-link :to="{ name: 'Attributes' }">
+                  <span class="icon is-small"><i class="fas fa-table" aria-hidden="true"></i></span>
+                  <span>Attributes <b-tag type="is-light">{{sketchMeta.attributes.length}}</b-tag></span>
+                </router-link>
+              </li>
             </ul>
           </div>
         </div>
@@ -81,7 +87,12 @@ limitations under the License.
 <script>
 export default {
   name: 'ts-navbar-secondary',
-  props: ['currentAppContext', 'currentPage']
+  props: ['currentAppContext', 'currentPage', 'sketchMeta'],
+  methods: {
+    hasAttributeOntology: function (ontologyName) {
+      return this.sketchMeta.attributes.filter(item => item[2] === ontologyName).length > 0
+    }
+  }
 }
 </script>
 
--- a/timesketch/frontend/src/components/AppNavbarSecondary.vue
+++ b/timesketch/frontend/src/components/AppNavbarSecondary.vue
@@ -64,10 +64,10 @@ limitations under the License.
                   <span>Stories</span>
                 </router-link>
               </li>
-              <li v-bind:class="{'is-active': currentPage === 'attributes'}">
+              <li v-if="meta" v-bind:class="{'is-active': currentPage === 'attributes'}">
                 <router-link :to="{ name: 'Attributes' }">
                   <span class="icon is-small"><i class="fas fa-table" aria-hidden="true"></i></span>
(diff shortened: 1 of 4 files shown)


timesketch/frontend/src/views/Attributes.vue
```diff
@@ -0,0 +1,82 @@
...
+        {{ sketch.name }}
+      </template>
+    </ts-navbar-main>
+
+    <ts-navbar-secondary
+      currentAppContext="sketch"
+      currentPage="attributes"
+    ></ts-navbar-secondary>
+
+    <section v-if="meta.permissions.read" class="section">
```
berggren, line 29: You will not be able to get this far if you don't have read access to the sketch, so this check should be safe to remove.

timesketch/frontend/src/utils/RestApiClient.js
```diff
@@ -101,6 +101,18 @@ export default {
     }
     return RestApiBlobClient.post('/sketches/' + sketchId + '/archive/', formData)
   },
+  getSketchAttributes (sketchId) {
+    return RestApiClient.get('/sketches/' + sketchId + '/attribute/')
```
berggren, line 105: I would prefer to make the naming consistent with other API resources. Make this `/attributes/` instead. Let's not do that in this PR though, but lets followup with this. :)

timesketch/frontend/src/components/AppNavbarSecondary.vue
```diff
@@ -81,7 +87,20 @@ limitations under the License.
 <script>
 export default {
   name: 'ts-navbar-secondary',
-  props: ['currentAppContext', 'currentPage']
+  props: {
+    currentAppContext: String,
+    currentPage: String
+  },
+  methods: {
+    hasAttributeOntology: function (ontologyName) {
```
berggren, line 95: Hmm, you could to this as a computed instead, directly on the @store.meta.attributes. This works to though. Up to you :)

berggren (changes requested):
@tomchop PTAL

timesketch/frontend/src/utils/RestApiClient.js
tomchop: Yeah, let's do the API changes in another PR. I'll update this client at the same time.

timesketch/frontend/src/components/AppNavbarSecondary.vue
tomchop: I'm not sure I can do this with a computer property, as this one is taking a parameter. The idea is to call `hasAttributeOntology('blah')` so that bits of the menubar are displayed only if the sketch has an attribute called `blah`. Can you do this with a computed property?

berggren: Ah, I missed that you were passing a parameter. In that case, you need to use a method. 👍

timesketch/frontend/src/utils/RestApiClient.js
berggren: SG

timesketch/frontend/src/views/Attributes.vue
```diff
@@ -0,0 +1,82 @@
...
+  },
+  computed: {
+    sketch () {
+      return this.$store.state.sketch
+    },
+    meta () {
+      return this.$store.state.meta
+    },
+    sketchAttributes () {
+      // return this.$store.state.meta.attributes
```
berggren, line 73:
```suggestion

```

berggren (changes requested):
Just a small fix, remove the commented line. Otherwise LGTM

berggren (approved):
LGTM

berggren:
LGTM

PR #1795, opened 2021-05-31
Fix broken attribute count in navbar

tomchop:
Attribute count was broken (attributes is no longer an array but an object). Countin entries instead fixes the bug.


berggren (approved):
LGTM

PR #1804, opened 2021-06-03
Sigma verify tool improv

jaegeral:
Logical error in the verify tool that was not handling the ignore_list as it should be.


kiddinn (approved):
LGTM

PR #1809, opened 2021-06-09
Add additional regexes to extract IP addresses from sshd messages

itsmvd:
The following types of sshd messages are not being picked up by the feature extraction as is (and also not by the ssh_sessionizer analyzer, but there will be another PR for that one):

[sshd, pid: 19774] Accepted password for admin from 1.1.1.1 port 62867 ssh2
[sshd, pid: 19774] Failed password for admin from 1.1.1.1 port 62867 ssh2
[sshd, pid: 23794] Connection reset by 1.1.1.1 port 10854 [preauth]
[sshd, pid: 27039] Connection closed by 1.1.1.1 port 55752 [preauth]


binglot (approved):
LGTM, I don't know why the unit test fails, I think it's completely unrelated

PR #1831, opened 2021-06-16
TsDynamicTable now has links pointing back to searches in sketch.

### LINKED ISSUE #1830: TsDynamicTable.vue should have links to search queries
It would be nice if an attribute could be clicked on and searched for in a skethc.

tomchop:
Fixes #1830 

Attributes are static and there's very little we can do with them.

Some attributes can now be made clickable, and they will run a search against the sketch with the attribute value.

Some structural changes in the metadata required for a complex attribute:

```javascript
meta: {
          sections: [
            {
              key: 'vt_summary',
              label: 'Virustotal Summary',
              columns: [
                { field: 'hash', label: 'SHA256 hash', type: 'text' },
                { field: 'first_seen', label: 'First seen', type: 'timestamp'},
                { field: 'last_seen', label: 'Last seen', type: 'timestamp'},
                { field: 'last_analysis', label: 'Last analysis', type: 'timestamp'},
                { field: 'ratio', label: 'Detection ratio', type: 'number', searchable: false },
                { field: 'tag', label: 'Tags', type: 'list', searchable: true, searchKey: 'vt_tag' },
              ],
            },
          ],
        }
```

`searchable` attributes will generate a query on the sketch with `"<attribute_value>"`. Attributes with a `searchKey`  will be prepended with the key. In the example above, Tags would link back to a query like `vt_tag:"<attribute_value>"`

image


timesketch/frontend/src/assets/main.scss
```diff
@@ -30,7 +30,7 @@
 [data-theme='light'] {
   --background-color: linear-gradient(90deg, rgba(249, 249, 249, 1) 11%, rgba(241, 241, 241, 1) 100%);
   --default-font-color: var(--font-color-dark);
-  --default-link-color: var(--font-color-dark);
+  --default-link-color: #3273dc;
```
berggren, line 33: Where does this color come from? :) It is not consistent with the rest of the UI and this will change all links which is a substantial change.  If you really want blue links here, could you scope them to this component only for now?

berggren (changes requested):
Looks good - just one comment.

timesketch/frontend/src/assets/main.scss
berggren: Ping @tomchop - PTAL

tomchop: Sounds good, I've removed it for now and moved to the local colors

berggren: 👍  The new link color is till in the PR though :) (line 33)

timesketch/frontend/src/components/Common/TsDynamicTable.vue
```diff
@@ -33,11 +33,44 @@ limitations under the License.
...
+                    :to="{ name: 'Explore', query: generateQuery(props.row[column.field], column) }"
+                  >
+                    {{ props.row[column.field] }}
+                    <i class="fas fa-search" aria-hidden="true"></i>
+                  </router-link>
+                  <span v-else>{{ props.row[column.field] }}</span>
+                </div>
+
+                <!-- column is a timestamp -->
+                <div v-else-if="column.type === 'timestamp'">
```
berggren, line 49: is this a timestamp (seconds from epoc?) or is this a datatime?
If it is a datetime, do we want to format it somehow or just show as is?

berggren (changes requested):

timesketch/frontend/src/components/Common/TsDynamicTable.vue
tomchop: It depends what the provider sends, so it's hard to know in advance. We can do some more checks on the data format and attempt to translate it here, but I would leave as-is for now.

berggren: Ack, SG.

berggren (approved):
LGTM

PR #1836, opened 2021-06-17
UI: Interactive histogram

### LINKED ISSUE #1835: Add interactive histogram
It would be useful to have an interactive histogram that when clicked will add time rage filter for the query.

Investigate if there is an open source library to help with the viz, and make sure it integrated with the filter system.

berggren:
Adds an interactive histogram chart to the explore view.

Screenshot 2021-06-17 at 16 33 56

When bars (buckets) are clicked the chart will update the current filter with the new time range.

* Uses open source Apexcharts library
* Using Elasticsearch auto-date-histogram aggregation to automatically set aggregation resolution.

**Closing issues**
closes #1835


binglot (approved):
LGTM

binglot:
Btw - my Dev Console shows an error but I traced it back to the ApexCharts code, and I don't think it has anything to do with your code.

```
apexcharts.min.js?3d1e:formatted:10573 Uncaught TypeError: Cannot read property 'classList' of null
    at s.value (apexcharts.min.js?3d1e:formatted:10573)
```

berggren:
> Btw - my Dev Console shows an error but I traced it back to the ApexCharts code, and I don't think it has anything to do with your code.
> 
> ```
> apexcharts.min.js?3d1e:formatted:10573 Uncaught TypeError: Cannot read property 'classList' of null
>     at s.value (apexcharts.min.js?3d1e:formatted:10573)
> ```

Ah, yeah. This is from the library and not us.

PR #1842, opened 2021-06-21
UI: Vertical 3-dots on Timeline Chips

### LINKED ISSUE #1841: UI: Vertical 3-dots on Timeline Chips
The Timeline tab seems redundant, it'd be a better UX to provide the options to view/change the timeline within the Explore view.

Create a vertical 3-dots on Timeline Chips.

I haven't come up with any alternatives.

**Additional context**
I'd like the 3-dots button to also provide an option for filtering by Data Type but at the moment I'm unsure how to best implement it so that it's intuitive.

binglot:
Is your feature request related to a problem? Please describe.
The Timeline tab seems redundant, it'd be a better UX to provide the options to view/change the timeline within the Explore view.

Describe the solution you'd like
Create a vertical 3-dots on Timeline Chips.

Describe alternatives you've considered
I haven't come up with any alternatives.

Additional context
I'd like the 3-dots button to also provide an option for filtering by Data Type but at the moment I'm unsure how to best implement it so that it's intuitive.

Closes #1841


timesketch/frontend/src/components/Explore/TimelineChip.vue
```diff
@@ -14,28 +14,320 @@ See the License for the specific language governing permissions and
...
+            <span>Delete</span>
+          </b-dropdown-item>
+        </b-dropdown>
+      </span>
+    </span>
   </div>
 </template>
 
 <script>
+import Vue from 'vue'
```
jaegeral, line 199: Shouldn't the Vue package already be imported?

timesketch/frontend/src/components/Explore/TimelineChip.vue
```diff
@@ -14,28 +14,320 @@ See the License for the specific language governing permissions and
...
+      style="cursor: pointer; margin-right: 7px;margin-bottom:7px;padding-right: 6px;"
+      :style="timelineColor(timeline)"
+      @click="toggleTimeline(timeline)"
+    >
+      {{ timeline.name }}
+      <!-- Show a warning icon, if there were import errors. -->
+      <span
+        v-if="datasourceErrors.length"
+        class="b-tooltips"
+        style="padding-right: 6px; padding-left: 12px;"
```
tomchop, line 127: Inline styling is usually tricky to debug (it's sprinkled throughout the codebase) - it's better to add a class here and move styling directives to the `<style>` section at the bottom.

timesketch/frontend/src/components/Explore/TimelineChip.vue
```diff
@@ -14,28 +14,320 @@ See the License for the specific language governing permissions and
...
+        @click.stop
+      >
+        <b-tooltip :label="failedImportsMessage" :type="isDarkTheme ? 'is-dark' : 'is-light'">
```
tomchop, line 130: What about

```suggestion
        <b-tooltip :label="datasourceErrors.length + ' failed imports'" :type="isDarkTheme ? 'is-dark' : 'is-light'">
```
 
and get rid of the computed property below?

timesketch/frontend/src/components/Explore/TimelineChip.vue
```diff
@@ -14,28 +14,320 @@ See the License for the specific language governing permissions and
...
+      if (this.newColor.startsWith('#')) {
+        this.newColor = this.newColor.substring(1)
+      }
+      Vue.set(this.timeline, 'color', this.newColor)
+      this.$emit('save', this.timeline)
+    }, 300),
+    toggleTheme: function() {
+      this.isDarkTheme = !this.isDarkTheme
+    },
+    timelineColor(timeline) {
```
tomchop, line 261: May I suggest renaming this to `getTimelineStyle`? You're returning more than just the color

timesketch/frontend/src/components/Explore/TimelineChip.vue
```diff
@@ -14,28 +14,320 @@ See the License for the specific language governing permissions and
...
+        opacity: opacity,
       }
     },
+    toggleTimeline: function(timeline) {
+      this.$emit('toggle', timeline)
+    },
+  },
+  mounted() {
+    // Hide color picker when clicked outside.
+    let self = this
```
tomchop, line 297: What's the reason we're assigning `this` to `self`? Can't we just do these operations on `this`?

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: Thanks Alex for pointing this out! I copied this part of code from one of the Timelines components so I didn't question it too much. It turns out you were right, removing that import doesn't cause any issues.

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: Good catch! Will make the change right away :)

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: Yup, completely agree, was planning to move it but it escaped me. Will make the change.

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: Funny because I converted the label into a computed property thinking that it would be more readable. I'm fine either way, and since you suggested it, I'll change it back.

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: To be completely honest, I copied this function from a previous component, so I didn't inspect the purpose of this variable. At first glance, it looks completely unnecessary, let me check if it's still working fine without it.

binglot: As suspected, it was there on purpose, as "this" becomes something else when the event is triggered compared to when the page is "mounted".

```
TimelineChip.vue?f907:291 Uncaught TypeError: Cannot read property 'contains' of undefined
    at eval (TimelineChip.vue?f907:291)
```

binglot: I've added a comment though, to make it clear that this code is there for a good reason. Thanks for challenging it!

timesketch/frontend/src/components/Explore/TimelineChip.vue
binglot: I also searched for `Vue.set()` and `Vue.use()` across the code and it seems that this is the only time that we use `Vue.set()` outside of the `store.js` file, which is probably why we have to import Vue. There might be a better way of calling `Vue.set()` that doesn't require this file to import Vue, but that would probably need to wait for Johan to be back :)

timesketch/frontend/src/components/Explore/TimelineChip.vue
tomchop: My guess is that `this` is referencing the window object since that's where the function is defined? Ideally, you could add a method in the Vue object and call this method:

```
window.addEventListener('click', this.hideColorPicker(e))
```

(there's already a `showColorPicker`, I suspect for the same reason)

tomchop (changes requested):

berggren (changes requested):
Hey @binglot - looks good, and this is almost ready to merge. Can you add a little bit more details into the PR description (for posterity) and a couple of screenshots on how the feature looks like?

binglot:
> Hey @binglot - looks good, and this is almost ready to merge. Can you add a little bit more details into the PR description (for posterity) and a couple of screenshots on how the feature looks like?

Thanks @berggren ! I've added screenshots to the issue referenced in this PR, and it already had a short description. Let me know if it's sufficient.

https://github.com/google/timesketch/issues/1841

berggren (approved):
LGTM

PR #1845, opened 2021-06-22
Copy to clipboard for EventListRow and EvetlistRowDetail

### LINKED ISSUE #1334: Add a copy button next to datetime / message
If an analyst wants to copy data from Timesketch to any other system, e.g. an email or a ticketing system and for whatever reason can not use the API or other export functions we have, it might come handy to have a little copy button.

I would suggest (but happy to hear other opinions) to put a button next to the datetime and the message field.

Of course we could also add a copy button  in every row for the event details.

For the question "what to copy" several options to be discussed:
- key
- value
- key : value

I would opt to only copy the value, as this would reflect 80 % of the use cases I would have. E.g. I am writing an email and need certain datetimes from the sketch.

jaegeral:
This PR will add Copy buttons to EventListRow and EventListRowDetails

Currently it will alert if it was successful, happy to remove that if desired.

Also the position of the buttons is debatable, this is just a first shot.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.
- [x] Make buttons hover away
- [x] Make a nicer "copied" message that will disappear after a few seconds
Copy_Buttons_ts

closes #1334

### DIFF
--- a/docs/learn/sketch-overview.md
+++ b/docs/learn/sketch-overview.md
@@ -37,6 +37,14 @@ You can share the sketch with users, groups of users, make it available to all u
 
 **Filter** Add a filter, e.g. to show only starred events
 
+For each event, you can apply the following actions by clicking small icons in each event row:
+
+- Star the event
+- Label the event
+- Comment the event
+- Start a context query to find related events
+- Copy an event attributes
+
 ## More
 
 If you click on the "More" Button in the Sketch Overview, you get the following three options.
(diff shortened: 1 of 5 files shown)


jaegeral:
It will show the following "Copied" message on top right for a few seconds:

Screenshot 2021-06-24 at 10 17 13

jaegeral:
@berggren PTAL

berggren (approved):
LGTM

PR #1862, opened 2021-06-30
UX local dev doc updates

jaegeral:
Only minor changes to the UX local dev doc that had some code formatting problems.


berggren (approved):
LGTM

Issue #1872, opened 2021-07-09
Wrong error message in timesketch_importer


jaegeral:
When an upload filed it gives out the timesketch timeline description instead of the status.

https://github.com/google/timesketch/blob/3907a5982bcd2b97ee03cdf816adeddceaead34f/importer_client/python/tools/timesketch_importer.py#L478

PR #1873, opened 2021-07-09
Update timesketch_importer.py

jaegeral:
Fixes https://github.com/google/timesketch/issues/1872


kiddinn (commented):
Why not have both the description and the status?

jaegeral:
> Why not have both the description and the status?

That is certainly also an option, I was not sure if the timeline name here was just a copy paste error, because with the lack of status it was very hard to tell why it failed so I made it the status, but happy to at the description as well.

kiddinn (approved):
LGTM

but please send another PR in to upgrade the version number, so that a new build can be pushed

berggren:
@jaegeral Can you update the version number in this PR (version.py) so we can push a new release as well? After that this is good to merge.

jaegeral:
> @jaegeral Can you update the version number in this PR (version.py) so we can push a new release as well? After that this is good to merge.

Sure that, done

Issue #1772, opened 2021-05-13
Aggregation calender widget using browser time instead of UTC


honorthecode:
**Description**
In the aggregations screen, when selecting times for  the `ISO formatted timestamp for the start time of the aggregated data` and `ISO formatted end time for the aggregation` fields the UI give the user a calendar and time selection widget. The date and time for that calendar appear to be in my browser's timezone, rather than UTC time (see the image below). This means that the time I select then gets instantly converted into the UTC/ISO time for the field. This makes selecting start and end times a little more tricky since you will have to account for the timezone your browser is in. I get this behavior in both a locally installed Timesketch v20210507 and https://demo.timesketch.org/, as of May 13, 2021.

![image](https://user-images.githubusercontent.com/69635449/118192231-d40e0900-b413-11eb-84ef-181bf29f6524.png)

**Steps to reproduce**
1. Ensure your web browser is set to time timezone other than UTC. My browser was set to EST5EDT (UTC-04).
2. Log into Timesketch v20210507 or https://demo.timesketch.org/
3. Go into a sketch, e.g Case of the stolen szechuan sauce.
4. Go to Aggregate 
5. Select all  timelines
6. Select Terms aggregations
7. Select a table chart to render
8. Select any field to aggregate on, e.g. parser
9. Click on the  `ISO formatted timestamp for the start time of the aggregated data`  field and in the calendar widget select 2020-09-19 03:22:00. If your browser was set to UTC-04 then the ISO time will be converted to 2020-09-19T07:22:00.
10. Click on the `ISO formatted end time for the aggregation` field and in the calendar widget select 2020-09-19 03:24:00. If your browser was set to UTC-04 then the ISO time will be converted to 2020-09-19T07:24:00.

**Expected Behavior**
The time selected in the calendar widget should be UTC time, meaning it will be the same as what eventually gets written to the start/end time fields.

**Desktop**
OS- this issue was observed on Windows 10 and Ubuntu 20.04 hosts
Browser- this issue was observed using Firefox on both Windows 10 and Ubuntu 20.04 hosts

berggren:
Good catch! I'll take a look at the calendar widget we use here.

berggren:
Ok, I found the issue. We are converting back the input from the widget to UTC, which in this case is confusing. I have a PR ready with a fix.

jaegeral:
This has been closed

PR #1889, opened 2021-08-03
Don't convert datetime string in output (datetime picker widget)

berggren:
The datestring on aggregation datetime picker output was converted to UTC. This is confusing as the widget itself is local time. This PR fixes this.

**Closing issues**
#1772


timesketch/frontend/src/components/Common/DynamicFormDatetimeInput.vue
```diff
@@ -28,7 +28,7 @@ export default {
   props: ['placeholder', 'label', 'name', 'value', 'display'],
   methods: {
     dateFormatter(dt) {
-      let dateString = dt.toISOString().replace('.000Z', '')
+      let dateString = new Date(dt.getTime() - dt.getTimezoneOffset() * 60000).toISOString().replace('.000Z', '')
```
tomchop, line 31: Can we just add a comment about what calculation is going on here? Otherwise LGTM

tomchop (approved):

timesketch/frontend/src/components/Common/DynamicFormDatetimeInput.vue
berggren: Done.

PR #1896, opened 2021-08-12
UI consistency

berggren:
This PR fixes some UI (especially in dark mode) consistency issues. Dropdowns are now rendered correctly, and the color switcher has changed to be more intuitive. 

Screenshot 2021-08-12 at 14 31 55


timesketch/frontend/src/assets/main.scss
```diff
@@ -158,19 +187,25 @@ $input-placeholder-color: var(--input-placeholder-color);
...
 $navbar-item-color: var(--navbar-item-link-color);
+
 $label-color: var(--form-label-font-color);
+
 $message-header-background-color: var(--message-header-background-color);
 $message-header-color: var(--message-header-font-color);
 $message-background-color: var(--message-background-color);
 $message-body-color: var(--message-body-color);
 
+$dropdown-background-color: #333;
```
jaegeral, line 207: Shouldn't this also be similar to ```--dropdown-background-color: var(--background-color-3);``` and not a fixed color here?

timesketch/frontend/src/assets/main.scss
```diff
@@ -194,14 +229,18 @@ body {
...
+
 .login-page {
   background: var(--login-page-background-color);
   color: var(--default-font-color);
 }
 
 .card {
   border-radius: 5px;
-  background-color: var(--card-background-color);
+  background: var(--card-background-color);
```
jaegeral, line 243: not that familiar with CSS, but I kind a expected this to stay `-color`

PR #1897, opened 2021-08-13
Sigma: Rule detail inline with table rows

berggren:
Move Sigma rule details to inline table row. Also change the table columns to show Title and Description only.

Screenshot 2021-08-13 at 11 59 22


jaegeral (approved):
LGTM

Issue #1898, opened 2021-08-14
Recursion Error in GeoIP analyzer when configuration values not set


sydp:
A RecursionError occurs in the Client Adapter when configuration values are not set.

**To Reproduce**
1. Run the GeoIP analyzer

```
timesketch.tasks/INFO [geo_ip_maxmind_db] result: Traceback (most recent call last):
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/interface.py", line 995, in run_wrapper
    result = self.run()
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/geoip.py", line 282, in run
    response = client.ip2geo(ip_address)
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/geoip.py", line 95, in __exit__
    return self.__exit__(exc_type, exc_value, traceback)
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/geoip.py", line 95, in __exit__
    return self.__exit__(exc_type, exc_value, traceback)
  File "/usr/local/src/timesketch/timesketch/lib/analyzers/geoip.py", line 95, in __exit__
    return self.__exit__(exc_type, exc_value, traceback)
  [Previous line repeated 957 more times]
RecursionError: maximum recursion depth exceeded 
```

**Expected behavior**
Should report "MaxMind Database configuration not set."

jaegeral:
According to the PR, this issue is fixed so I am closing this issue

PR #1899, opened 2021-08-14
Geoip updates

sydp:
Updates:

* Concrete client adapter subclasses should return super() instead of self - fixes recursion error
* Improvements to error handling

**Closing issues**

#1898


timesketch/lib/analyzers/geoip.py
```diff
@@ -274,8 +280,11 @@ def run(self):
                     ip_addresses[ip_addr][ip_address_field].append(event)
 
         for ip_address, ip_address_fields in ip_addresses.items():
-            with self.GEOIP_CLIENT() as client:  # pylint: disable=E1102
-                response = client.ip2geo(ip_address)
+            try:
+                with self.GEOIP_CLIENT() as client:  # pylint: disable=E1102
+                    response = client.ip2geo(ip_address)
+            except GeoIPClientError as error:
+                return f'GeoIP Client error - {error}'
```
berggren, line 287: This will stop the entire run, is that what you want here? Maybe just log the error, and add it to a `fail` counter and report # fails in the end?

berggren (changes requested):

berggren (approved):

PR #1906, opened 2021-08-16
1895 bugfix copy buttons

### LINKED ISSUE #1895: Copy icons issues
1. The copy icons while on the explore page are not consistent (different colors)

Screenshot 2021-08-11 at 11 10 22

Can we make these icons appear on hover instead of always showing them?

2. The `copy message string` icon does not work. It will only expand the event, not copy the content of the message string. Also, do we need this one? You get the copy icon when the event is exapanded.

Screenshot 2021-08-11 at 11 15 41

jaegeral:
Addressing the issue with several copy buttons:

- remove the copy button from the event row
- make the copy value button from the ```EventRowDetail```dark
- move the copy key button next to the key value in ```EventRowDetail```
- make both copy buttons in ```EventRowDetail``` to only display when hovered.
Screenshot 2021-08-16 at 16 55 08

**Closing issues**

closes #1895


berggren (approved):
LGTM

Issue #1908, opened 2021-08-17
SSH analyzer creates saved search and story even with 0 results


berggren:
The SSH analyzer will create a saved search even if there are no results from the analyzer.

Fix: Only create saved search if there was any sessions created.

PR #1909, opened 2021-08-17
ssh analyzer: Only create saved search if there are any sessions created

berggren:
Saved search is created even if there are no SSH sessions detected. This PR fixes that.

Closes: #1908


jaegeral (approved):
LGTM

Issue #1840, opened 2021-06-19
Deploy script should start containers


berggren:
It would be convenient if deplay_timesketch.sh also started the containers at the end of the run.

jaegeral:
https://github.com/google/timesketch/blob/master/contrib/deploy_timesketch.sh for reference which script is meant here.

avitomar12:
@jaegeral @berggren  Just wanted to know this issue is still open for contribution. After going through the issue I have developed the understanding which is that Deployment script should also start containers whereas deployment commands are in echo statements. to achive the objective of deployment of containers following code snippet should be introduced. Let me know following code snippet achive the objective.

```
echo "Start the system:"

echo "1. Performing <cd timesketch>"
cd timesketch
echo "2. Running <docker-compose up -d>"
docker-compose up -d
echo "3. docker-compose exec timesketch-web tsctl add_user --username <USERNAME>"
read -p 'USERNAME : ' usrname
docker-compose exec timesketch-web tsctl add_user --username $usrname

```

jaegeral:
> @jaegeral @berggren Just wanted to know this issue is still open for contribution. After going through the issue I have developed the understanding which is that Deployment script should also start containers whereas deployment commands are in echo statements. to achive the objective of deployment of containers following code snippet should be introduced. Let me know following code snippet achive the objective.
> 
> ```
> echo "Start the system:"
> 
> echo "1. Performing <cd timesketch>"
> cd timesketch
> echo "2. Running <docker-compose up -d>"
> docker-compose up -d
> echo "3. docker-compose exec timesketch-web tsctl add_user --username <USERNAME>"
> read -p 'USERNAME : ' usrname
> docker-compose exec timesketch-web tsctl add_user --username $usrname
> ```

Sure thing this is open for contributions, feel free to make a Pull request for your changes.

harshalchaudhari35:
@jaegeral @berggren is this issue still up for grabs? I can raise a PR if so

jaegeral:
> @jaegeral @berggren is this issue still up for grabs? I can raise a PR if so

Hey, I think you already did in https://github.com/google/timesketch/pull/1925 and it was closed based on lack of response. If you are able to work on the PR again, just re-open the PR and we take it from there.

harshalchaudhari35:
@jaegeral I don't think I was the original author on that but sure will re-open and move forward. Cheers

harshalchaudhari35:
[...]

PR #1925, opened 2021-08-20
Added deploy script to start containers at the end of run

avitomar12:
This PR resolve the issue #1840 .  Added code to start the container.

### DIFF
--- a/contrib/deploy_timesketch.sh
+++ b/contrib/deploy_timesketch.sh
@@ -114,9 +114,15 @@ echo "* Installation done."
 
 echo
 echo "Start the system:"
+
 echo "1. cd timesketch"
+cd timesketch
 echo "2. docker-compose up -d"
+docker-compose up -d
 echo "3. docker-compose exec timesketch-web tsctl add_user --username <USERNAME>"
+#input the username
+read -p 'USERNAME : ' USRNAME
+docker-compose exec timesketch-web tsctl add_user --username $USRNAME
 echo
 echo "WARNING: The server is running without encryption."
 echo "Follow the instructions to enable SSL to secure the communications:"


contrib/deploy_timesketch.sh
```diff
@@ -114,9 +114,15 @@ echo "* Installation done."
 
 echo
 echo "Start the system:"
+
 echo "1. cd timesketch"
+cd timesketch
 echo "2. docker-compose up -d"
+docker-compose up -d
```
berggren, line 121: I'm not sure everyone wants to automatically start the containers. Could we put this behind a flag instead?

`--start-containers`

berggren (changes requested):
Thanks for contributing. Please take a look at my comment and let me know what you think.

contrib/deploy_timesketch.sh
avitomar12: Where do input this flag? Can we ask user whether to start container after installation is done?

jaegeral: I think both would be fine:
```sh deploy_timesketch.sh —start-container```

Or if that is not set at the end of the script ask
```Would you like start your container (y/n) [default:no]```

What do you think @avitomar12 ?

File: contrib/deploy_timesketch.sh [hunk 2]
```diff
@@ -114,9 +114,15 @@ echo "* Installation done."
...
 echo "3. docker-compose exec timesketch-web tsctl add_user --username <USERNAME>"
+#input the username
+read -p 'USERNAME : ' USRNAME
```
jaegeral, line 124: Should that be s/USRNAME/USERNAME?

jaegeral (changes requested):
Some comments inline

berggren:
Friendly ping @avitomar12 - please let us know the status of this PR and if you plan to update it.

contrib/deploy_timesketch.sh
avitomar12: Yes. I also think its better to provide both options to the user.

jaegeral:
@avitomar12 Friendly ping

berggren:
@avitomar12 Ping again - please see the comments in the review. If we don't hear back we need to close this PR for the time being (we can open it again if needed later).

avitomar12:
> @avitomar12 Ping again - please see the comments in the review. If we don't hear back we need to close this PR for the time being (we can open it again if needed later).

@berggren I am working on it. I will make change within two days.

berggren:
> > @avitomar12 Ping again - please see the comments in the review. If we don't hear back we need to close this PR for the time being (we can open it again if needed later).
> 
> @berggren I am working on it. I will make change within two days.

Hello again, any progress on this PR?

PR #1936, opened 2021-09-07
Scenarios: Data Model

berggren:
This is a DRAFT PR to share the ongoing process of adding the data model for the upcoming Scenarios feature.


jaegeral:
@googlebot I consent

jaegeral:
@googlebot I consent

data/scenarios/questions.yaml
```diff
@@ -0,0 +1,108 @@
...
+#                           used to define things that should be present in the
+#                           dataset or should not be there in order to give
+#                           an answer to the investigative question.
+#
+#       answer_analyzer     If the answer cannot be derived from a simple
+#                           search a name of an analyzer can be provided
+#                           that is specifically designed to answer this
+#                           particular investigative question.
+# ------------------------------------------------------------------------
+user_logged_in_win:
```
jaegeral, line 78: Would the answer to that question be bool or a a set of n events?

data/scenarios/questions.yaml
```diff
@@ -0,0 +1,108 @@
...
+    applies to Windows machines.
+  data_sources:
+    - windows_login
+  analyzers:
+    - login
+  answer_simple:
+    must:
+      - windows_login
+
+user_downloaded_file:
```
jaegeral, line 93: Would it make sense to first answer the question "Did someone download files" and if that question is answered, we move on the more scoped one?

data/scenarios/scenarios.yaml
```diff
@@ -0,0 +1,27 @@
...
+#                           be used in an UI to describe the question.
+#
+#       description         Further description of the question, what the
+#                           question attempts to answer, whether it will
+#                           require further manual analysis, etc.
+#
+# ------------------------------------------------------------------------
+compromise_assessment:
+  display_name: Compromise Assessment
+  description: Maecenas tristique sapien lorem, et vehicula ipsum vestibulum
```
jaegeral, line 20: We should have a good description here or add an TODO

data/scenarios/questions.yaml
berggren: Yes, good question. We could have a depend on [list of other questions]. This can be handled in code though, so no need to change the model.

data/scenarios/scenarios.yaml
berggren: Yes, these placeholders need to be updated in a future PR :)

berggren:
@jaegeral PTAL

jaegeral (approved):
@berggren and myself did a review session today, there are many things still to be polished which will happen in future PRs. As this new feature is not meant to be used soon by anyone we agreed to get it in and work from there as it is a rather large change on database and stuff.

So from my side: LGTM

PR #1938, opened 2021-09-08
fix(sigma-util): do not fail on bad rules

gaelmuller:
Fix an issue with Sigma integration.

/api/v1/sigma is currently failing with Internal Server Error if a problematic rule exists in the rules directory. This is especially problematic because the official Sigma repository contains problematic rules.

This fix prevents the Sigma integration from failing while still logging problematic rules.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

### DIFF
--- a/timesketch/lib/sigma_util.py
+++ b/timesketch/lib/sigma_util.py
@@ -213,13 +213,13 @@ def get_sigma_rule(filepath, sigma_config=None):
             logger.error(
                 'Error generating rule in file {0:s}: {1!s}'
                 .format(abs_path, exception))
-            raise
+            return None
 
         except sigma_exceptions.SigmaParseError as exception:
             logger.error(
                 'Sigma parsing error generating rule in file {0:s}: {1!s}'
                 .format(abs_path, exception))
-            raise
+            return None
 
         except yaml.parser.ParserError as exception:
             logger.error(


berggren:
@jaegeral I'm assigning this to you.

jaegeral:
Also note that https://github.com/google/timesketch/blob/master/data/sigma_blocklist.csv keeps track of rules that are considered problematic. So it is strongly recommended, to not deploy them.

jaegeral:
> IMHO the user should never see an "Internal Server Error" and this behavior doesn't do the Sigma support justice.
> 
> Right now, I think this may be a common user flow:
> 
> 1. Cool, timesketch has Sigma support now !
> 2. Let's try it, I will just download the public Sigma repository
> 3. Go to the Sigma tab, blank page, Internal Server Error

Yeah that sounds like a very common flow. With your suggested PR I assume the flow could look like:

> 1. Cool, timesketch has Sigma support now !
> 2. Let's try it, I will just download the public Sigma repository
> 3. Go to the Sigma tab, some rules show up some not
> 4. Go to https://github.com/google/timesketch open an issue
> 5. We need to help troubleshoot which rule caused the issue and there will be less info in the logs

PR #1941, opened 2021-09-12
Update plaso.mappings

### LINKED ISSUE #1935: Timesketch doesn't understand when http_response_bytes is "-"
**Describe the bug**
When you upload data from Plaso's `apache_access` parser, Timesketch complains and doesn't display log events that have `http_response_bytes` as "-". The following log line can be seen on the worker's log:

```
[400] <mapper_parsing_exception> failed to parse field [http_response_bytes] of type [long] in document with id 'xyz'. Preview of field's value: '-' [illegal_argument_exception/For input string: "-"]
```
[Apache's documentation](https://httpd.apache.org/docs/2.4/mod/mod_log_config.html#logformat) states that: "%b - Size of response in bytes, excluding HTTP headers. In CLF format, i.e. a '-' rather than a 0 when no bytes are sent.".

**To Reproduce**
1. Get Apache logs that have `http_response_bytes` as "-" (see [here](https://github.com/log2timeline/plaso/blob/5198673e2312eac0bbf35fb200cb8bfedea88423/test_data/access.log#L7))
2. Process with Plaso's `apache_access` parser
3. Upload to Timesketch
4. Observe events missing

**Expected behavior**
See all the events in Timesketch.

berggren:
the attribute from the apache log plaso parser sometimes set the `http_response_bytes` to `-` and this doesn't work for the default mapping of `long`. This PR fixes that.

Closes #1935

### DIFF
--- a/data/plaso.mappings
+++ b/data/plaso.mappings
@@ -79,7 +79,7 @@
       "type": "text",
       "fields": {"keyword": {"type": "keyword"}}
     },
-    "http_response_bytes" {
+    "http_response_bytes": {
       "type": "text",
       "fields": {"keyword": {"type": "keyword"}}
     }
--- a/data/plaso.mappings
+++ b/data/plaso.mappings
@@ -78,6 +78,10 @@
     "version": {
       "type": "text",
       "fields": {"keyword": {"type": "keyword"}}
+    },
+    "http_response_bytes" {
+      "type": "text",
+      "fields": {"keyword": {"type": "keyword"}}
     }
   }
 }


itsmvd (approved):

data/plaso.mappings
```diff
@@ -78,6 +78,10 @@
     "version": {
       "type": "text",
       "fields": {"keyword": {"type": "keyword"}}
+    },
+    "http_response_bytes" {
```
berggren, line 82:
```suggestion
    "http_response_bytes": {
```

Issue #1942, opened 2021-09-13
Menubar broken when intelligence attributes are malformed


tomchop:
**Describe the bug**
The menubar disappears when intelligence attributes can't be counted correctly.

**To Reproduce**
Add a single `intelligence_local` attribute (e.g. from the explore view) to a new sketch, menubar disappears.

**Expected behavior**
Menu bar stays!

PR #1943, opened 2021-09-13
Bugfix for disappearing menubar

tomchop:
Fixes https://github.com/google/timesketch/issues/1942


tomchop:
Should we be using something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining to check whether objects have specific attributes?

berggren (approved):
LGTM

berggren:
> Should we be using something like https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining to check whether objects have specific attributes?

That sounds like a good idea for these type of cases, I agree.

PR #1952, opened 2021-09-16
Properly split file extension

### LINKED ISSUE #1951: Properly parse file extensions on web uploads
**Describe the bug**
There is a bug in timesketch/frontend/src/components/Common/UploadForm.vue on line 126 when grabbing the fileExtension. If the file name has multiple '.' in it, the code will not pull the correct file extension as it's searching for the first item after a period. This won't always be the last item in the split. Plaso currently outputs a filename like plaso.name.name2.plaso. If you try to upload this plaso file you will get an extension error. The code should be [-1] instead of [1]. 

**To Reproduce**
Try to upload a plaso file that has multiple periods in the file name such as: plaso.computername.domain.com.plaso.

**Expected behavior**
The file extension should be pulled from the end of the filename. Not the first item after splitting on the periods.

**Screenshots**
Not applicable.

 - OS: linux (any)
 - Browser any
 - Version latestt

**Additional context**
Creating a pull request to fix:

backcountryinfosec:
TLDR: This PR fixes a small bug in the way that the fileExtension is grabbed. Current implementation will not grab the real file extension if there are multiple '.' in the filename. Since this is a minor bug fix I didn't creatte a problem but if you'd like I can still do thatt.

Error in code on line 126 where the file extension is grabbed. Currently splitting and grabbing [1] will only grab you the first item after the file name. This may not always be the extension. Plaso currently outputs the .plaso file as filename.something.something2.somethin3.plaso. To properly grab the file extension and prevent issues when uploading due to unknown file extension it should be grabbing the last item in the split [-1] which would always be the real extension.

IE:
let fileExtension = fileName.split('.')[-1]

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

closes #1951

### DIFF
--- a/timesketch/frontend/src/components/Common/UploadForm.vue
+++ b/timesketch/frontend/src/components/Common/UploadForm.vue
@@ -123,7 +123,7 @@ export default {
     },
     setFileName: function(fileList) {
       let fileName = fileList[0].name
-      let fileExtension = fileName.split('.')[1]
+      let fileExtension = fileName.split('.')[-1]
       this.form.file = fileList[0]
       this.form.name = fileName
         .split('.')


backcountryinfosec:
@googlebot I signed it!

joachimmetz:
What about extensions like `.tar.gz` ?

backcountryinfosec:
Code was incorrect, I was in a hurry and put [-1] (python) when it needs to be .pop() for vue JS. As to joachimmetz's comment above. The only allowed extensions are csv, json, jsonl, and plaso. Tar.gz wouldn't need to be worried about since it's an invalid extension.

joachimmetz:
if with invalid you mean unsupported then your argument makes sense, I'll leave this to the Timesketch devs, but I would recommend putting a comment there that only single level extension are support, in case in the future there might want to add `.json.gz` or equiv

itsmvd:
I concur with @joachimmetz's last comment. Add a comment, and feel free to check if the file name indeed ends with one of these 4 possible extensions. Looking forward to your next PR!

PR #1960, opened 2021-10-04
Multi analyzer - tagger and feature extractor

berggren:
Refactor tagger and feature extractor analyzers to use multi task runtime. This enable the searches to be run in parallell when Celery is setup for that.


jaegeral (approved):
LGTM

Issue #1761, opened 2021-05-04
Sigma rules with backslashes causing wrong queries


jaegeral:
A rule e.g.:

```
title: RedMimicry Winnti Playbook Dropped File
id: 130c9e58-28ac-4f83-8574-0a4cc913b97e
description: Detects actions caused by the RedMimicry Winnti playbook
references:
    - https://redmimicry.com
author: Alexander Rausch
date: 2020/06/24
tags:
    - attack.defense_evasion
    - attack.t1027
logsource:
    product: windows
    category: file_event
detection:
    selection:
        TargetFilename|contains:
            - gthread-3.6.dll
            - sigcmm-2.4.dll
            - \Windows\Temp\tmp.bat
    condition: selection
falsepositives:
    - Unknown
level: high
```

Causing issues because of the backslashes. Need to investigate how to best parse and carry them without crazy string replaces.

jaegeral:
Ok this seems toonly affect the get_rule_by_text api, not the read from file.

jaegeral:
was fixed with #1968 it now creates:
```(data_type:"windows:evtx:record" AND TargetFilename:(*gthread-3.6.dll* OR *sigcmm-2.4.dll* OR *Windows\\Temp\\tmp.bat*))```

PR #1968, opened 2021-10-07
Fix Sigma backslash issue

jaegeral:
This is a very hacky way to solve: 

https://github.com/google/timesketch/issues/1761

But at the moment it would break a lot of cases, so better hacky then wrong.


jaegeral:
Ok the last failing test is not me, so @berggren PTAL. Thx

PR #2026, opened 2021-11-11
More efficient IOC searching

tomchop:
This PR changes the way IOCs are searched for in event details, by introducing a `match_fields` key in the regexes object, that binds the search to only a specific event field (to induce less false positives and to avoid searching in fields that aren't interesting)

Additionally, this PR introduces a few more IOC types:

* `hostname` (only search in the `hostname` field)
* `fs_path` (only search in the `message` field)

All other IOCs are only searched for in the `message` field, reducing the number of TextHighlight components that are rendered in the page with nothing to highlight.


berggren (approved):
LGTM

PR #2027, opened 2021-11-12
Generate graph for a specific timeline

berggren:
This PR add support for per timeline graph generation. There are times where it is desirable to be able to choose which timeline to generate the graph for.

* Add URL parameter support (?timeline=<ID>)
* Dropdown to choose timeline
* Warning to indicate that the cached graph is generated from a single timeline
* Keeps backwards compatibility with old graph caches

Screenshot 2021-11-12 at 12 18 06

We can add support for multiple timelines in the future.

### DIFF
--- a/timesketch/lib/graphs/interface.py
+++ b/timesketch/lib/graphs/interface.py
@@ -142,6 +142,7 @@ class BaseGraphElement:
         Returns:
             MD5 hash (str): MD5 hash of the provided label.
         """
+
         id_string = self.attributes.get('id', self.label)
         return hashlib.md5(id_string.encode('utf-8')).hexdigest()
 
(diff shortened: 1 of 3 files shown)


timesketch/lib/graphs/interface.py
```diff
@@ -142,6 +142,7 @@ def _generate_id(self):
         Returns:
             MD5 hash (str): MD5 hash of the provided label.
         """
+
```
jaegeral, line 145: Lint, this newline seems not needed.

jaegeral:
Please also add a blob to the documentation and maybe include your screenshot to it.

jaegeral (approved):
Dedicated issue for Graph documentation will be filed.

LGTM

berggren:
Doc: https://github.com/google/timesketch/issues/2028

PR #2038, opened 2021-11-19
Add incompatible Sigma rules to the blocklist

### LINKED ISSUE #2037: Move Sigma rules that cause problems to a dedicated folder that is ignored by TS
Often times people just git clone the upstream Sigma project and expect the rules to work despite documented on multiple places.

One interim solution could be to just move the troubling filed to a dedicated folder and add an entry in ```https://github.com/google/timesketch/blob/master/data/sigma_blocklist.csv``` (note we can not do that by rule uuid because to get the rule uuid the rule must be parse, which in those cases would / could fail.

E.g. move the files to a deprecated sub folder.

jaegeral:
From time to time, people try to copy the whole Sigma repository into Timesketch and it will cause side effects and frustration.

With this PR two things will happen:

a) Rules / folders that are known not good and are mentioned in the `sigma_blocklist.csv` will now be ignored when returing rules over the API.

b) if a new rules is added to the Filesystem and is causing problems, it will add a line to the `sigma_blocklist.csv` and therefor ignore it for future iterations.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**
closes #2037

### DIFF
--- a/docs/guides/user/sigma.md
+++ b/docs/guides/user/sigma.md
@@ -63,6 +63,24 @@ The rules then will be under
 timesketch/data/sigma
 ```
 
+### Sigma Rules Blocklist file
+
+The `data/sigma_blocklist.csv` is the central point where Timesketch tries to maintain which rules are considered good or bad. By default each rule is considered good, but it is good practice to add them in that file if they are tested.
+
+Each method that reads Sigma rules from the a folder is checking, if part of the full path of a rule is mentioned in the `data/sigma_blocklist.csv` file.
+
+For examle a file at `/etc/timesketch/data/sigma/rules-unsupported/foo/bar.yml` would not be parsed as a line in `data/sigma_blocklist.csv` mentions:
+
+```
+/rules-unsupported/,bad,Sigma internal folder name,2021-11-19,
+```
+
+The header for that file are:
+
+```
+path,bad,reason,last_ckecked,rule_id
+```
+
 ### Sigma Rules
 
 The windows rules are stored in
@@ -87,6 +105,7 @@ There are multiple sigma related config variables in ```timesketch.conf```.
 SIGMA_RULES_FOLDERS = ['/etc/timesketch/sigma/rules/']
 SIGMA_CONFIG = '/etc/timesketch/sigma_config.yaml'
 SIGMA_TAG_DELAY = 5
+SIGMA_BLOCKLIST_CSV = '/etc/timesketch/sigma_blocklist.csv'
 ```
 
 The ```SIGMA_RULES_FOLDERS``` points to the folder(s) where Sigma rules are stored. The folder is the local folder of the Timesketch server (celery worker and webserver). For a distributed system, mounting network shares is possible.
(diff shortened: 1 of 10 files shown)


docs/guides/user/sigma.md
```diff
@@ -63,6 +63,24 @@ The rules then will be under
 timesketch/data/sigma
 ```
 
+### Sigma Rules Blocklist file
+
+The `data/sigma_blocklist.csv` is the central point where Timesketch tries to maintain which rules are considered good or bad. By default each rule is considered good, but it is good practice to add them in that file if they are tested.
```
berggren, line 68:
```suggestion
The `data/sigma_blocklist.csv` is where Timesketch maintains a list of incompatible rules. By default each rule is considered compatible, but it is good practice to add them in this file if they are tested and verified to not be compatible.
```

timesketch/lib/sigma_util.py
```diff
@@ -133,6 +141,14 @@ def get_sigma_rules(rule_folder, sigma_config=None):
                     continue
 
                 rule_file_path = os.path.join(dirpath, rule_filename)
+                block_because_csv = False
+
+                if any(x in rule_file_path for x in ignore_list):
+                    block_because_csv = True
+
+                if block_because_csv:
```
berggren, line 149: You can get rid of this block, and just `continue` in the `if any()` above. block_because_csv is not used anywhere else it seems.

timesketch/lib/sigma_util.py
```diff
@@ -273,6 +295,91 @@ def _sanatize_sigma_rule(sigma_rule_query: str) -> str:
...
+
+    Returns:
+        Sigma Blocklist path
+
+    Raises:
+        ValueError: Sigma blocklist file is not readabale.
+    """
+    logger.error(blocklist_path)
+
+    if blocklist_path is None or blocklist_path == '':
```
berggren, line 334:
```suggestion
    if not blocklist_path or blocklist_path == '':
```

berggren (changes requested):

timesketch/lib/sigma_util.py
jaegeral: done

docs/guides/user/sigma.md
```diff
@@ -63,6 +63,24 @@ The rules then will be under
 timesketch/data/sigma
 ```
 
+### Sigma Rules Blocklist file
+
+The `data/sigma_blocklist.csv` is where Timesketch maintains a list of incompatible rules. By default each rule is considered compatible, but it is good practice to add them in this file if they are tested and verified to not be compatible.
+
+Each method that reads Sigma rules from the a folder is checking, if part of the full path of a rule is mentioned in the `data/sigma_blocklist.csv` file.
```
berggren, line 70:
```suggestion
Each method that reads Sigma rules from the a folder is checking if part of the full path of a rule is mentioned in the `data/sigma_blocklist.csv` file.
```

berggren (approved):

PR #2045, opened 2021-11-23
Better intelligence view

tomchop:
## This PR suggests changes for the Intelligence view.

There is no more external vs. internal intelligence. All intelligence are a series of objects added to the `intelligence` attribute of a sketch.

### Intelligence now expects a specific type of entries

Intelligence entries have the following, simple format;

```json
{
  "externalURI" :"yeti/127.0.0.1"
  "ioc" :"127.0.0.1"
  "tags": ["evil", "c2", "random"]
  "type": "ip"
}
```

### Intelligence view has been revamped

image

From the IOC view, one can:

* Copy IOC to clipboard
* Search a sketch for a given IOC
* Open the external reference in a new browser tab, if it looks like a link.

It also lists a sketch's tags (right side), as well as tags that have been associated with added IOCs (left side)

* IOC tags: list and search for all IOCs with a specific tag in a sketch's events.
* Sketch tags: list and search for all events that have a specific tag.

### IOCs can be edited

IOCs can be edited and deleted individually. Tag views are automatically updated upon save.

image


jaegeral:
I wonder if we want to include the Sigma table to the Intel view as well and list the tags that are present in active Sigma rules, WDYT?

tomchop:
That sounds like a good idea! I suggest we get this checked in first and add it later, the PR is big enough as it is.

berggren:
Looks good! A few UI related things:
* The title and subtitle are a bit too big IMO, and aren't consistent with the other components throughout the UI. Can you adjust to match?
* The Tags are blue? We use grey in the rest of the UI and we should stay consistent.
* The gap between the "cards" are a bit off (too much space between the indicator of compromise and the Tag list), maybe lay them out with the same padding/margin?
* Can you tone down the red color on the trash can? It really pops right now :)

Let me know if you want me to sketch this out in Figma.

timesketch/frontend/src/views/Intelligence.vue
```diff
@@ -22,34 +22,161 @@ limitations under the License.
...
+            <b-table-column field="edit" label="" v-slot="props">
+              <span class="icon is-small" style="cursor:pointer;" title="Edit IOC" @click="startIOCEdit(props.row)"
+                ><i class="fas fa-edit"></i>
+              </span>
+            </b-table-column>
+
+            <b-table-column field="delete" label="" v-slot="props">
+              <span
+                class="icon is-small"
+                style="cursor:pointer;color:red"
```
berggren, line 118: Tone this down, may something like #7C0D0E ?

timesketch/frontend/src/views/Intelligence.vue
```diff
@@ -22,34 +22,161 @@ limitations under the License.
...
+                </b-table-column>
+              </b-table>
+              <span v-else>No IOCs have been tagged yet.</span>
+            </div>
+          </div>
+
+          <!-- labels column -->
+          <div class="column">
+            <div class="box">
+              <h1 class="title">Event tags</h1>
```
berggren, line 163: Make this a bit smaller, maybe a h2

timesketch/frontend/src/views/Intelligence.vue
```diff
@@ -59,53 +186,116 @@ limitations under the License.
...
-        deletable: true,
-      },
+      sketchTags: [],
+      tagInfo: {},
+      editingIoc: {},
+      showEditModal: false,
+      IOCTypes: [
+        { regex: /^(\/[\S]+)+$/i, type: 'fs_path' },
+        { regex: /^([-\w]+\.)+[a-z]{2,}$/i, type: 'hostname' },
+        { regex: /^((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)$/g, type: 'ip' },
```
berggren, line 201: Seeing this is v4 only, maybe name it `ipv4`?

timesketch/frontend/src/views/Intelligence.vue
```diff
@@ -59,53 +186,116 @@ limitations under the License.
...
+              tagName: tag,
+            }
+          }
+          this.tagInfo[tag].count++
+          this.tagInfo[tag].iocs.push(ioc.ioc)
         }
-      })
+      }
+    },
+    generateOrElasticQuery(valueList) {
```
berggren, line 263:
```suggestion
        // TODO: Use filter chips instead
        generateOrElasticQuery(valueList) {
```

berggren (changes requested):
Looks good. Some smaller UI related things.

tomchop:
> Looks good! A few UI related things:
> 
> * The title and subtitle are a bit too big IMO, and aren't consistent with the other components throughout the UI. Can you adjust to match?

Sure, will do.

> * The Tags are blue? We use grey in the rest of the UI and we should stay consistent.

Yeah, blue looks nicer. Maybe we should change them elsewhere too? 😉 More seriously, what would you think about sticking to grey for the tags that come from events, and staying with blue (or any other color) for the ones that are associated with IOCs? Happy to stick with grey everwhere too, if that's what you want to go with. However, I do think it would be good to start thinking ahead about differentiating thats that have been added by users, by analyzers, by intelligence... food for thought.

> * The gap between the "cards" are a bit off (too much space between the indicator of compromise and the Tag list), maybe lay them out with the same padding/margin?

Ah yes, good catch. Bulma scaffolding is Hard.

> * Can you tone down the red color on the trash can? It really pops right now :)

At least you know to keep your mouse away from it 😄 Sure, I'll done it down.

Thanks for the comments!

jaegeral:
One additional thing, once you settled with all the fine tweaking, please also adjust / adopt https://github.com/google/timesketch/blob/master/docs/guides/user/intelligence.md with the new screenshots, functionalities and such so it is up2date.

Thx

timesketch/frontend/src/views/Intelligence.vue
tomchop: Done - moved from boxes to cards, as changing the header type didn't change the font size.

timesketch/frontend/src/views/Intelligence.vue
tomchop: Done

timesketch/frontend/src/views/Intelligence.vue
tomchop: I went for a slightly redder color, so that it matches the `<code>` blocks that I have in the IOC list.

tomchop:
PR comments done. @jaegeral @berggren can you PTAL?

jaegeral (approved):
LGTM, I leave it to @berggren to LGTM his requested changes and merge the PR.

berggren (approved):
LGTM

berggren:
> > Looks good! A few UI related things:
> > 
> > * The title and subtitle are a bit too big IMO, and aren't consistent with the other components throughout the UI. Can you adjust to match?
> 
> Sure, will do.
> 
> > * The Tags are blue? We use grey in the rest of the UI and we should stay consistent.
> 
> Yeah, blue looks nicer. Maybe we should change them elsewhere too? 😉 More seriously, what would you think about sticking to grey for the tags that come from events, and staying with blue (or any other color) for the ones that are associated with IOCs? Happy to stick with grey everwhere too, if that's what you want to go with. However, I do think it would be good to start thinking ahead about differentiating thats that have been added by users, by analyzers, by intelligence... food for thought.
> 

I agree here, and I'm looking into the color and differentiation in a larger visual overhaul in 2022.

> > * The gap between the "cards" are a bit off (too much space between the indicator of compromise and the Tag list), maybe lay them out with the same padding/margin?
> 
> Ah yes, good catch. Bulma scaffolding is Hard.
> 
> > * Can you tone down the red color on the trash can? It really pops right now :)
> 
> At least you know to keep your mouse away from it 😄 Sure, I'll done it down.
> 
> Thanks for the comments!

PR #2105, opened 2022-01-12
Fix broken link to notebooks

jaegeral:
The current notebook link only works if you visit it on Github, but opening it on timesketch.org (where it should be).


binglot (changes requested):
The new link (https://github.com/google/timesketch/notebooks) doesn't work.

PR #2107, opened 2022-01-12
Mention two blog post as reading recommendation

jaegeral:
I think it will be beneficial for starters to read up on those two articles before getting started.

Only update to documentation


binglot (changes requested):
Please fix the spelling.

binglot (approved):

PR #2110, opened 2022-01-14
Added permissions change to opensearch data directory

### LINKED ISSUE #2109: Incorrect permissions on directory data/opensearch
**Describe the bug**
The permissions on directory `data/opensearch` are `root:root 0755` by default. Opensearch in the opensearch container will fail if it cannot write to this directory, which as user `1000` it cannot do.

**To Reproduce**
Steps to reproduce the behaviour:
1. Follow instructions at https://timesketch.org/guides/admin/install/ to start the containers
2. Create a sketch, navigate to sketch page
3. Observe `Internal Server Error`. 

I verified by listing listening ports inside the container. On timesketch host: 
```
$ docker exec -u root -it opensearch /bin/bash
bash-4.2# cat /proc/net/tcp
  sl  local_address rem_address   st tx_queue rx_queue tr tm->when retrnsmt   uid  timeout inode
   0: 00000000:25B2 00000000:0000 0A <snip>
   1: 0B00007F:A859 00000000:0000 0A <snip> 
   2: 00000000:2580 00000000:0000 0A <snip>
```

(Those ports need to be converted from hex, but 9200 is not listed)

**Expected behaviour**
Opensearch starts and is listening on port 9200.

 - OS: Ubuntu 20.04.3 LTS
 - Browser Firefox
 - Version 91.4.1esr (64-bit)

Adding `chown 1000 timesketch/data/opensearch` to the deploy script fixed the issue for me. PR with the same incoming.

ramo-j:
Fix for permissions on opensearch's data directory, `data/opensearch` that prevents opensearch from starting properly. 

Closes #2109

### DIFF
--- a/contrib/deploy_timesketch.sh
+++ b/contrib/deploy_timesketch.sh
@@ -55,6 +55,7 @@ fi
 
 # Create dirs
 mkdir -p timesketch/{data/postgresql,data/opensearch,logs,etc,etc/timesketch,etc/timesketch/sigma/rules,upload}
+# TODO: Switch to named volumes instead of host volumes.
 chown 1000 timesketch/data/opensearch
 
 echo -n "* Setting default config parameters.."
--- a/contrib/deploy_timesketch.sh
+++ b/contrib/deploy_timesketch.sh
@@ -55,6 +55,7 @@ fi
 
 # Create dirs
 mkdir -p timesketch/{data/postgresql,data/opensearch,logs,etc,etc/timesketch,etc/timesketch/sigma/rules,upload}
+chown 1000 timesketch/data/opensearch
 
 echo -n "* Setting default config parameters.."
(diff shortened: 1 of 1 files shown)


contrib/deploy_timesketch.sh
```diff
@@ -55,6 +55,7 @@ fi
 
 # Create dirs
 mkdir -p timesketch/{data/postgresql,data/opensearch,logs,etc,etc/timesketch,etc/timesketch/sigma/rules,upload}
+chown 1000 timesketch/data/opensearch
```
berggren, line 58:
```suggestion
# TODO: Switch to named volumes instead of host volumes.
chown 1000 timesketch/data/opensearch
```

berggren (approved):

Issue #2130, opened 2022-02-03
BigQuery Analyzer Plugin


anttitikkanen:
**Is your feature request related to a problem? Please describe.**
I have threat indicators stored in a BigQuery table. I'd like to match these indicators (file hashes, IP's, domains) to Timesketch events to identify known badness.

**Describe the solution you'd like**
An analyzer plugin that matches selected fields from Timesketch events against data in a BigQuery table. The user can configure this using a new configuration file with entries like:

```
matcher_rule_name:
     event_field_name: sha256_hash # Timesketch events with this field will be matched
     bq_project: 'project_name' # Use this Google Cloud project to run the BQ job
     bq_query: 'SELECT DISTINCT hash FROM project.dataset.tablename WHERE hash IN UNNEST(@sha256_hash)' # The query to use for matching
     tags: ['bigquery-sha256-match'] # Tags to apply if matching data is found in the BQ table
     emojis: ['SKULL'] # Emojis to apply if matching data is found in the BQ table
```

I'll be happy to contribute this feature myself!

PR #2131, opened 2022-02-03
Added a new BigQuery Matcher Analyzer plugin and related documentation.

anttitikkanen:
https://github.com/google/timesketch/issues/2130

It allows matching Timesketch event metadata against indicators stored in a BigQuery table.

A new BigQuery Analyzer plugin.

The PR adds a single analyzer plugin, along with tests.

**Checks**

- [X] All tests succeed.
- [X] Unit tests added.
- [ ] e2e tests added.
- [X ] Documentation updated.

I didn't add e2e tests, as it looked tricky to test with BQ in the current e2e test setup? Happy to do so if you give me a tip how.

### DIFF
--- a/timesketch/lib/analyzers/bigquery_matcher.py
+++ b/timesketch/lib/analyzers/bigquery_matcher.py
@@ -1,0 +1,124 @@
+"""Index analyzer plugin for matching against data in BigQuery tables."""
+from __future__ import unicode_literals
+
+import itertools
+import logging
+
+from google.cloud import bigquery
+from google.auth import exceptions as google_auth_exceptions
+
+from timesketch.lib import emojis
+from timesketch.lib.analyzers import interface
+from timesketch.lib.analyzers import manager
+
+logger = logging.getLogger('timesketch.analyzers.bigquery_matcher')
+
+
+class BigQueryMatcherPlugin(interface.BaseAnalyzer):
+    """Analyzer for matching events to BigQuery data."""
+
+    NAME = 'bigquery_matcher'
+    DISPLAY_NAME = 'BigQuery matcher'
+    DESCRIPTION = 'Match pre-defined event fields to data in BigQuery tables'
+
+    _BQ_BATCH_SIZE = 10000  # Number of entries per BQ query
+
+    def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
+        """Initialize the BQ Matcher Analyzer.
+
+        Args:
+            index_name: OpenSearch index name
+            sketch_id: Sketch ID
+            timeline_id: The ID of the timeline
+        """
+        self.index_name = index_name
+        self._matcher_name = kwargs.get('matcher_name')
+        self._matcher_config = kwargs.get('matcher_config')
+        super().__init__(index_name, sketch_id, timeline_id=timeline_id)
+
+    @staticmethod
+    def get_kwargs():
+        """Get kwargs for the analyzer.
+
+        Returns:
+            List of matchers.
+        """
+        bq_config = interface.get_yaml_config('bigquery_matcher.yaml')
+        if not bq_config:
+            logger.error('BigQuery Matcher could not load configuration file.')
+            return []
+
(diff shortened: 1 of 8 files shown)


berggren:
Looks good! I'll review in detail as soon as possible.

timesketch/lib/analyzers/bigquery_matcher_test.py
```diff
@@ -0,0 +1,94 @@
+"""Tests for BigQueryMatcher Plugin."""
+from __future__ import unicode_literals
```
berggren, line 2: You can skip this line, we are python 3 only :)
```suggestion

```

timesketch/lib/analyzers/bigquery_matcher.py
```diff
@@ -0,0 +1,124 @@
+"""Index analyzer plugin for matching against data in BigQuery tables."""
+from __future__ import unicode_literals
```
berggren, line 2:
```suggestion

```

berggren:
Look good! I don't have many comments on the code. This prompted me to finally add support for contrib analyzers :) Seeing you need to add the dependency `google.cloud bigquery` I went and created a PR so we can move this to the contrib directory instead. This way we only load the plugin if the dependency can be imported. Then we don't need to add the dep to the Core Timesketch requirements.

Let's chat about this on Slack when you are around. Should be a simple change for you, and I'll show you what you need to do.

Ref: #2141

timesketch/lib/analyzers/contrib/bigquery_matcher_test.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class TestBigQueryMatcherPlugin(BaseTest):
+    """Tests the functionality of the analyzer."""
+
+    _TEST_EMOJI = 'SKULL'
+    _TEST_TAG = 'test-tag'
+    
+    # Skip the tests if bigquery is not imported.
+    __test__ = True if 'google.cloud.bigquery' in sys.modules else False
```
anttitikkanen, line 20: Note that this seemed to be the most convenient way to skip nosetests. I also tried https://pypi.org/project/nose-exclude/ and modifying run_tests.py but it did not seem to work with the coverage options.

timesketch/lib/analyzers/contrib/__init__.py
```diff
@@ -11,3 +11,5 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
```
berggren, line 14:
```suggestion
"""Contrib Analyzer module."""
```

timesketch/lib/analyzers/contrib/bigquery_matcher_test.py
```diff
@@ -0,0 +1,96 @@
...
+from timesketch.lib.testlib import BaseTest
+
+
+class TestBigQueryMatcherPlugin(BaseTest):
+    """Tests the functionality of the analyzer."""
+
+    _TEST_EMOJI = 'SKULL'
+    _TEST_TAG = 'test-tag'
+    
+    # Skip the tests if bigquery is not imported.
```
berggren, line 19:
```suggestion
    # Skip the tests if bigquery is not imported.
    # pylint: disable=simplifiable-if-expression
```

timesketch/lib/analyzers/contrib/bigquery_matcher_test.py
```diff
@@ -0,0 +1,96 @@
...
+from timesketch.lib.analyzers.contrib import bigquery_matcher
+from timesketch.lib.testlib import BaseTest
+
+
+class TestBigQueryMatcherPlugin(BaseTest):
+    """Tests the functionality of the analyzer."""
+
+    _TEST_EMOJI = 'SKULL'
+    _TEST_TAG = 'test-tag'
+
```
berggren, line 18:
```suggestion

```

PR #2132, opened 2022-02-05
Fix formating for nanosecond timestamps

### LINKED ISSUE #2093: Update FormatTimestamp filter to handle nanosecond precision timestamps
**Describe the bug**
Timestamps in nanosecond precision don't appear to be interpreted correctly in the explore view.

Looking at this filter used in the UI explore view, it seems the heuristics don't account if a timestamp is in nanoseconds (only seconds, milliseconds and microseconds are handled)

https://github.com/google/timesketch/blob/fce0f54e20d9385b264f2401079dbeb5ad773a38/timesketch/frontend/src/filters/FormatTimestamp.js

sydp:
Fixes #2093.


berggren (approved):
LGTM

PR #2140, opened 2022-02-08
Serve Intelligence tag metadata via the API

tomchop:


data/intelligence_tag_metadata.yaml
```diff
@@ -0,0 +1,15 @@
+malware:
```
berggren, line 1: Can we add a description of the format i this file? Explain what weight and class means for example.

berggren (commented):
Looks good! One thing, we want to add support for this resource to the Python API client as well. Fine to do that in another PR though.

data/intelligence_tag_metadata.yaml
tomchop: Done.

berggren (approved):
LGTM

PR #2141, opened 2022-02-09
Add contrib directory for analyzers

berggren:
This PR sets up the scaffolding for community contributed analyzers. The idea is that these analyzers are not maintained by the core Timesketch team, but they are hosted in the repository.

The main reason for this is to set the expectation on maintenance but also to not add more dependencies to the core system. Instead, if an analyzer depends on a package it will add that to the requirements file in the contrib directory, and only load the analyzer if the dependency has been installed. 

Next step is to support completely external analyzers, hosted elsewhere, but that is out of scope for this PR.


timesketch/lib/analyzers/contrib/README.md
```diff
@@ -0,0 +1,26 @@
+## Community contributed analyzers
```
jaegeral, line 1: I opt that we move some of that to https://github.com/google/timesketch/blob/master/docs/developers/analyzer-development.md and then link from here to the docs folder.

berggren: 👍

PR #2147, opened 2022-02-15
CSV with wrong extension stopped on UI side

### LINKED ISSUE #2064: CSV with wrong extension is not stopped on UI side
https://github.com/google/timesketch/blob/fce0f54e20d9385b264f2401079dbeb5ad773a38/timesketch/frontend/src/components/Common/UploadForm.vue#L124

Uploading a foobar.txt file that has CSV content would still get uploaded and errored out on the backend side, even if the Client JS should error out already.

```Request URL: https://timesketch/api/v1/upload/
Request Method: POST
Status Code: 500 ```

### PR DESCRIPTION
**Pull request by rishav-karanjit**
closes #2064 

Checks:
- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [ ] Documentation updated.

### CODE REVIEW

File: timesketch/frontend/src/components/Common/UploadForm.vue [hunk 1]
```diff
@@ -97,29 +97,31 @@ export default {
...
-      }
-      ApiClient.uploadTimeline(formData, config)
-        .then(response => {
-          this.$store.dispatch('updateSketch', this.$store.state.sketch.id)
-          this.$emit('toggleModal')
-          this.clearFormData()
-          this.percentCompleted = 0
-        })
-        .catch(e => {})
+      if (this.error != 'Please select a file with a valid extension' ) {
```
jaegeral, line 100: could we instead check if this.error is the given error and then exit the function instead?

rishav-karanjit: Hey @jaegeral, I have send a new commit. Would you like to review it?

berggren:
ping @jaegeral can you PTAL?

jaegeral (approved):

PR #2155, opened 2022-03-29
Scenarios: Adding system_information, wlan_history, persistence, credential_access

jaegeral:
A lot more scenario based stuff to be discussed with jbn

- system_information
- wlan_history
- persistence
- credential_access


berggren:
Something is wrong with the data_finder.yml file. Take a look at the e2e tests.

jaegeral:
@berggren  I tried to fix the data finder, but I do not see what is causing the problem, not sure if you can easily spot it, otherwise I might have a deeper look and see if we can introduce something that makes it more verbose if failing.

PR #2166, opened 2022-04-13
Sigma use in analyzer

### LINKED ISSUE #2165: Sigma new field "use in analyzer"
Having a field in the Sigma api that provides a "use in analyzer" bool.

That way we could expose Sigma rules that are in review etc to the user without risking the rule would case a havoc with tags and so on.

### LINKED ISSUE #2171: Sigma Explore page: Display the author
Would be nice to display the author in the explore page to give credit to those who created the rule.

### LINKED ISSUE #2177: Sigma ts_comment should not have the line number
Remove the line number in `ts_comment` in https://github.com/google/timesketch/pull/2166 as it makes it hard to write unit tests

jaegeral:
Introduce a new field in the Sigma API that indicates if a rule should be used in the analyzer or not.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #2165
closes #2171
closes #2177


timesketch/lib/sigma_util.py
```diff
@@ -129,7 +133,13 @@ def get_sigma_rules(rule_folder, sigma_config=None):
 
     blocklist_path = None
     ignore = get_sigma_blocklist(blocklist_path)
-    ignore_list = list(ignore["path"].unique())
+    ignore_list = list(ignore.loc[ignore['bad'] == 'bad']["path"].unique())
+    experimental_list = list(
+        ignore.loc[ignore['bad'] == 'experimental']["path"].unique()
+    )
+    to_be_used_in_analyzer = list(
```
berggren, line 140: Is the attribute `bad` part of the Sigma spec? It feels like we are overloading this term, and IMO it is getting a bit difficult to reason about what `bad` vs `good` vs `experimental` means.

Maybe change this from `bad` --> `status` instead? And document what the different statuses means.

bad = Not able to parse to TS supported query
good = TS supported query 
experimental = Enabled in TS but not run automatically

berggren (changes requested):

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -159,7 +166,12 @@ def get_kwargs():
         Returns:
             sigma_rules All Sigma rules
         """
-        sigma_rules = [{"rule": rule} for rule in ts_sigma_lib.get_all_sigma_rules()]
+        sigma_rules = []
+        for rule in ts_sigma_lib.get_all_sigma_rules():
+            # Only rules that are meant to run in the analyzer will be added
```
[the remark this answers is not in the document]
jaegeral: ack done

timesketch/lib/sigma_util.py
jaegeral: yes that make sense, the only reason I would avoid the experimental word is that it is used as a status in the Sigma rules itself and could be confusing

jaegeral: ok changed it to status and introduced the term "exploratory" which seems more appropriate

berggren (approved):

PR #2170, opened 2022-04-14
Reduce load time for the CLI client

berggren:
This PR refactors how we do imports in the CLI client. The current implementation  has some slow imports (the API client with cloud dependencies etc). The way Click works is it imports all commands at startup.

By moving the imports in the commands to inline with the function we lazy import when needed. This speeds up the UI significantly.

I also refactored the state object, moving initializing the API client and config assistant to properties instead, with the ability to cache the objects (for commands that access them multiple times).

These changes reduced time to load the tool from ~1.4s -> ~0.2s. Making the tool much snappier to use.

* Further more, some cleanup of help messages, and the renaming of `import` to `upload` to make it more intuitive.

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/config.py
+++ b/cli_client/python/timesketch_cli_client/commands/config.py
@@ -17,8 +17,6 @@ import sys
 
 import click
 
-from timesketch_cli_client.definitions import SUPPORTED_OUTPUT_FORMATS
-
 
 @click.group("config")
 def config_group():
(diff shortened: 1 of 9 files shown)


cli_client/python/timesketch_cli_client/commands/config.py
```diff
@@ -34,12 +32,7 @@ def set_group():
 @click.argument("sketch_id")
 @click.pass_context
 def set_sketch(ctx, sketch_id):
-    """Set the active sketch.
-
-    Args:
-        ctx: Click CLI context object.
```
jaegeral, line 40: just stumbled across this PR, any reason you removed the docstring here?

cli_client/python/timesketch_cli_client/commands/upload.py
```diff
@@ -17,23 +17,17 @@
 import time
 
 import click
-from timesketch_import_client import importer as import_client
 
 
-@click.command("import")
+@click.command("upload")
```
jaegeral, line 22: I like that suggestion

jaegeral (commented):
I know it is only a draft, but I think it would be worth to merge those proposals, Added some comments

justzh:
Can I pull this to my local machine (macOS)?

jaegeral:
since this PR is several years old, I am gonna close it, if we think it is worth we could re-open at any time.

Issue #2201, opened 2022-06-02
Add rule hits for sigma rules that are considered "bad" in analyzer


jaegeral:
Adding functionality in the analyzer code for Sigma to use rules that are considered "bad" in the status csv but limiting the attributes it would add to `ts_sigma_filename`.

Right now we struggle to find out which rule is how noisy, to do that, we would need to run every rule query over a sketch every time it is visited, which could result in 1000 queries over all timelines just to get a count.

To preserve that count, the analyzer would use all the rules installed on a system, but only add User visible tags / labels for those that are considered "good".

By reverse referencing the rule name with the status csv we could in the future run a cleanup job to remove noisy entries from events if we choose to.

Alternative would be to store the count results in a database somewhere.

jaegeral:
That is now implemented. Closing it out

PR #2202, opened 2022-06-02
Use all rules in the analyzer

jaegeral:
Adding functionality in the analyzer code for Sigma to use rules that are considered "bad" in the status csv but limiting the attributes it would add to ts_sigma_filename.

Right now we struggle to find out which rule is how noisy, to do that, we would need to run every rule query over a sketch every time it is visited, which could result in 1000 queries over all timelines just to get a count.

To preserve that count, the analyzer would use all the rules installed on a system, but only add User visible tags / labels for those that are considered "good".

By reverse referencing the rule name with the status csv we could in the future run a cleanup job to remove noisy entries from events if we choose to.

Alternative would be to store the count results in a database somewhere.

https://github.com/google/timesketch/issues/2201

### DIFF
--- a/timesketch/frontend/src/components/Common/UploadForm.vue
+++ b/timesketch/frontend/src/components/Common/UploadForm.vue
@@ -97,6 +97,9 @@ export default {
       this.fileName = ''
     },
     submitForm: function() {
+      if (this.error == 'Please select a file with a valid extension' ) {
+        return;
+      }
       let formData = new FormData()
       formData.append('file', this.form.file)
       formData.append('name', this.form.name)
@@ -118,8 +121,8 @@ export default {
           this.$emit('toggleModal')
           this.clearFormData()
           this.percentCompleted = 0
-        })
-        .catch(e => {})
+       })
+       .catch(e => {}) 
     },
     setFileName: function(fileList) {
       let fileName = fileList[0].name


timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -168,8 +174,8 @@ def get_kwargs():
         """
         sigma_rules = []
         for rule in ts_sigma_lib.get_all_sigma_rules():
-            if rule.get('ts_use_in_analyzer') is True:
-                sigma_rules.append({"rule": rule})
+            # if rule.get('ts_use_in_analyzer') is True:
```
tomchop, line 177: Please remove this commented-out code. (or add an actual comment)

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -54,19 +58,20 @@ def run_sigma_rule(self, query, rule_name, tag_list=None):
...
+            if status_good:
+                ts_ttp = event.source.get("ts_ttp", [])
+                special_tags = []
+                for tag in tag_list:
+                    # Special handling for sigma tags that TS considers TTPS
+                    # https://car.mitre.org and https://attack.mitre.org
+                    if tag.startswith(("attack.", "car.")):
+                        ts_ttp.append(tag)
+                        special_tags.append(tag)
+                # ad the remaining tags as plain tags
```
tomchop, line 70:
```suggestion
                # add the remaining tags as plain tags
```

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -54,19 +58,20 @@ def run_sigma_rule(self, query, rule_name, tag_list=None):
...
-            # ad the remaining tags as plain tags
-            tags_to_add = list(set(tag_list) - set(special_tags))
-            event.add_tags(tags_to_add)
-            if len(ts_ttp) > 0:
-                event.add_attributes({"ts_ttp": list(set(ts_ttp))})
+            if status_good:
+                ts_ttp = event.source.get("ts_ttp", [])
+                special_tags = []
+                for tag in tag_list:
+                    # Special handling for sigma tags that TS considers TTPS
```
tomchop, line 65:
```suggestion
                    # Special handling for sigma tags that TS considers TTPs
```

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -32,13 +32,17 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
...
+    def run_sigma_rule(
+        self, query, rule_name, tag_list=None, status_good=True
+    ):
         """Runs a sigma rule and applies the appropriate tags.
 
         Args:
             query: OpenSearch search query for events to tag.
             rule_name: rule_name to apply to matching events.
             tag_list: a list of additional tags to be added to the event(s)
+            status_good: value from sigma_status.csv good / bad / exploratory
```
tomchop, line 44: This makes it sound as if we expect a string, but we're using a boolean here, so the comparison has to be made by the caller.

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -32,13 +32,17 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
         self._rule = kwargs.get("rule")
         super().__init__(index_name, sketch_id, timeline_id=timeline_id)
 
-    def run_sigma_rule(self, query, rule_name, tag_list=None):
+    def run_sigma_rule(
+        self, query, rule_name, tag_list=None, status_good=True
+    ):
```
tomchop, line 37:
```suggestion
        self, query, rule_name, tag_list=None, status_good=True):
```

tomchop (changes requested):

jaegeral (commented):
feedback

jaegeral (commented):
feedback

tomchop (approved):

tconqueror:
Can i get a sample of sigma_rule_status.csv, it cause http status 404 when access sketch if sigma_rule_status.csv not exist

tconqueror:
> > Can i get a sample of sigma_rule_status.csv, it cause http status 404 when access sketch if sigma_rule_status.csv not exist
> 
> Hey hey, the file in question is part of the PR: https://github.com/google/timesketch/blob/master/data/sigma_rule_status.csv
> 
> You need to add the new config value as well. Hope that helps

Yeah, thank you very much!

PR #2205, opened 2022-06-03
Preview search results directly from intelligence tab

tomchop:
This provides a way for the analyst to quickly spot which intelligence findings are relevant to the sketch.

The main functionality is provided in an `ExplorePreview` component:

* The component takes in a query as a prop to explore the sketch with
* As soon as it's loaded, it runs the query, and displays an eye icon with the number of results
* When a user hovers over the icon, a popup is displayed with the results of the search query
* Users can then proceed to annotate, star, comment on events that are displayed there

Search nodes are propagated, meaning that these explore searches will be recorded in the search history, and any star / annotation events occuring from that view will also be recorded.


timesketch/frontend/src/components/Common/ExplorePreview.vue
```diff
@@ -0,0 +1,102 @@
...
+    this.refreshPreview(this.searchQuery)
+  },
+}
+</script>
+
+<style scoped>
+.preview-box {
+  z-index: 100;
+  position: fixed;
+  background: white;
```
berggren, line 92: How does this look when in dark mode?

berggren (changes requested):
Looks good. One question, but it is not a blocker.

timesketch/frontend/src/components/Common/ExplorePreview.vue
tomchop: Much better now: [fc76f08](https://github.com/google/timesketch/pull/2205/commits/fc76f083fe385de93f19e9b4fcbcf959f706fb69)

PR #2207, opened 2022-06-03
Escape backslashes and spaces in generated OpenSearch queries

### LINKED ISSUE #2206: Intelligence not escaping certain characters correctly
Marking a path that value is:

`C:\Documents and Settings\dfir`
Will create the following intel entry:

`"C:\Documents and Settings\dfir"`
But should be:

`"C:\\Documents\ and\ Settings\\dfir"`

tomchop:
This fixes #2206, a bug that would not properly escape backslashes and spaces in queries generated from the Intelligence view.


jaegeral (approved):
LGTM

PR #2211, opened 2022-06-07
Additions to the Sigma List view

jaegeral:
Some additions to the Sigma List view:

- List the Rule Names that have hits in the sketch (with number of events)
- List TTPs added by Sigma rules in the sketch (with number of hits)


berggren:
First round of feedback :)

1. Can you split this into 2 PRs, one for the model/api and one for the frontend?
2. Why add this to the bulma based UI and not the new Vuetify one?

jaegeral:
This PR will be sliced into smaller ones to make it more digestible for reviewers.

PR #2214, opened 2022-06-09
[Sigma] Adding fs:bodyfile:entry to keep up with plaso output

jaegeral:
To be prepared for: https://github.com/log2timeline/plaso/pull/4132

We adding another option to be checked if one uses filesystem as type. That should make the data mapping compatible with both the legacy and the new format that plaso will prepare compatible.


sydp (approved):
LGTM

Issue #2224, opened 2022-06-19
Stories seem to be blank...


kiddinn:
**Describe the bug**
I just deployed a new TS instance, uploaded some files/timelines and ran some analyzers, some of which generated stories, the stories are visible in the UI, but once I click on them, the page is completely empty...

Then I go into notebook, and I type...

```python
%timesketch_set_active_sketch 1
stories = timesketch_list_stories_func()
a_story = stories.get('Browser Artifacts - <NAME OF TIMELINE>')
```

Then I can easily do:

```python
a_story.title
```
And the title gets printed, BUT if I do

```python
a_story.to_markdown()
```

An error appears...

```
[2022-06-19 00:35:30,603] timesketch.app/ERROR Exception on /api/v1/sketches/1/stories/1/ [POST]
Traceback (most recent call last):
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1516, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.8/dist-packages/flask/app.py", line 1502, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**req.view_args)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 467, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask/views.py", line 84, in view
    return current_app.ensure_sync(self.dispatch_request)(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask_restful/__init__.py", line 582, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/flask_login/utils.py", line 261, in decorated_view
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/timesketch/api/v1/resources/story.py", line 227, in post
    story=self._export_story(
  File "/usr/local/lib/python3.8/dist-packages/timesketch/api/v1/resources/story.py", line 137, in _export_story
    exporter.from_string(story.content)
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/stories/interface.py", line 61, in from_string
    self.from_block_dict(block)
  File "/usr/local/lib/python3.8/dist-packages/timesketch/lib/stories/interface.py", line 97, in from_block_dict
    "value": self._data_fetcher.get_aggregation_group(
[...]

PR #2225, opened 2022-06-19
Fixed a bug in the stories

kiddinn:
This fixes at least partially #2224 

This fixes issues that came up during the transition to open search, at least in the notebook.

There is also a fix for saved searches inside the API client, where the indices were not correctly restored/read from the backend.

Before one could not get the markdown content of a story... now...

image

as you can see, a markdown version of the story can be produced inside a notebook.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [ ] Documentation updated.


api_client/python/timesketch_api_client/search.py
```diff
@@ -249,6 +249,11 @@ def add_end_time(self, end_time):
         """
         if end_time.endswith("Z"):
             end_time = end_time[:-1]
+
+        # Check for a whole day, YYYY-MM-DD.
+        if len(end_time) == 10:
```
berggren, line 254: Do we want to check for the format here as well, and not just the length of the string?
`YYYY-MM-DD`

timesketch/lib/stories/api_fetcher.py
```diff
@@ -165,6 +173,31 @@ def get_aggregation_group(self, agg_dict):
...
+            A tuple with two items, a list of indices and a list of
+            timeline IDs.
+        """
+        indices = []
+        timeline_ids = []
+
+        if isinstance(index_list, str):
+            index_list = index_list.split(',')
+
+        for item in index_list:
```
berggren, line 193: `item` is a bit generic here, can you rename to `index` ?

berggren (changes requested):
A coupe of things, I know this is in draft already.

timesketch/lib/stories/api_fetcher.py
kiddinn: yes, changed it

api_client/python/timesketch_api_client/search.py
kiddinn: using re

kiddinn (commented):
made some changes

PR #2261, opened 2022-07-21
Add feature to allow a user to map missing CSV headers

### LINKED ISSUE #2237: Check the columns of csv in Javascript in Browser
First iteration would be to check if the four columns
`message,datetime,timestamp,timestamp_desc`are missing.

If one of them is missing, give an error message.

marcobrotto:
**Headers Selection in Timesketch**

This PR aims to solve the issue #2237. Moreover, this solution extends Timesketch's functionalities in the following ways: 
1. It gives the user the possibility to upload a CSV file and map an existing CSV's header to a mandatory one that is not present in the file.
2. New characters can be used as the delimiter in the CSV file, i.e., semicolon and pipe. 

**Current State of the Art**
Currently, when the CSV is uploaded, it is not checked on the client and it is sent directly to the server. The file is processed and then, if the mandatory headers are not present in the CSV, an error is sent back to the client. To upload correctly the file on the server, the user needs to modify the CSV locally. Moreover, the CSV cannot be parsed correctly by the server if the delimiter is different from the comma (",").

The main contribution of this PR are the following:
* When the user uploads a CSV file, the headers are processed client-side. If one or more of the three mandatory headers are missing, then the file cannot be uploaded on the server. This solution has the following benefit: avoid sending to the TS server (e.g., over the network) the entire file when it has wrong headers, such as, the message header is missing.
* The user can choose a different type of CSV delimiter that will be used to parse the file. The default one is the comma, but, as an alternative, the user can select the pipe or the semicolon. When the CSV delimiter changes, also the interpretation of the CSV is updated.
* When the uploaded CSV file has incorrect headers, the UI allows the user to instruct the server to map the original headers with a new configuration. For example, if there is a header named “msg”, the user can instruct the server to map it as “message”. To this end, a CSV file that does not contain the three mandatory headers can be uploaded on the server. 

**Design idea (in short terms)**

This paragraph explains the design ideas of the mapping between the existing headers and the mandatory ones.

* For each mandatory header, the UI shows a dropdown menu that is filled with the existing headers in the CSV. The user can map each mandatory header with an existing column’s name. As an alternative, if the users do not find any appropriate column’s name, then they can insert a new one and fill each row associated with it with a default value. For example, if there is not any header that can be mapped with the timestamp_desc, the user can add a new column with this name (i.e., timestamp_desc) and fill it with a chosen default value. The image below shows an example of the UI.

dropDownMenu02

* When the user clicks on the Upload button, the file, the CSV delimiter and the new mapping is sent in the POST request. The server will interpret the CSV headers according to the CSV separator and mapping previously sent.

**Overview of changes to existing functions (mainly)**
* Client side: `UploadForm.vue` in `timesketch/frontend/src/components/Common/`
* Server side: `tasks.py` and `utils.py` in `timesketch/lib/`, and `upload.py` in `timesketch/api/v1/resources`

**Checks**
We added 4 _unit tests_ that check the consistency of CSV files and the headers mapping.

The consistency of the CSV file is checked on the server side. If the file cannot be opened, then the server will trigger an error.

The headers’ mapping is a Python dictionary. The key is the name of the header that we want to substitute in the CSV, while the value is an array whose length is 2. The first value represents the name of the header in the CSV to be substituted. If this value is equal to “New Header”, then we are not mapping the mandatory field with any existing column, but we create a new one. In this scenario, we use the second entry as a default value for each row of that column.

Although the consistency of the headers mapping is done also on the client-side, we tested it also on the server-side. 
The sanity check of the headers mapping can trigger the following errrors:
1. The mapping is done only if the mandatory header is missing.
2. When a new column is created, a mandatory default value must be set.
4. The headers to be substituted and specified in the dictionary must be present in the CSV.
5. Two or more mandatory headers cannot be mapped to the same exisiting header

The table below shows some of the tested dictionary (TE = Trigger Error?).

|Dictionary   	|TE   	|Error type   	| Comment |
|---	|---	|---	|--- |
|  `{}`	|   No	|   N/a	|  Empty dictionary is considered as None, i.e., as if nothing had passed |
|   `{"datetime":["DT",""],"timestamp_desc":["No.",""], "message":["Source",""]}`	|  Yes 	|   1	| -   |
|   `{"datetime":["DT",""],"timestamp_desc":["NotExisitingColumn",""]}`	|   Yes	|  3 	| -   |
|   `{"datetime":["DT",""],"timestamp_desc":["DT",""]}`	|   Yes	|   4	|  -  |
|  `{"datetime":["DT",""],"timestamp_desc":["No.","test"]}` |   No	|   N/a |  Deafult value is not considered at all  |
|  `{"datetime":["New header","foo"],"timestamp_desc":["New header",""]}`	|   Yes	|  2 	|  -  |
|  `{"datetime":["New header","foo"],"timestamp_desc":["New header","baz"]}`	|   No	|   N/a	|  -  |

All errors are triggered correctly.
All the unit tests added passed correctly.

**Closing issues**
`closes #2237`


timesketch/api/v1/resources/upload.py
```diff
@@ -440,6 +450,14 @@ def post(self):
         if not form:
             form = request.form
 
+        # map existing headers with the mandatory ones
+        res = form.get('headersMapping')
```
jaegeral, line 454: How about `res = form.get('headersMapping',None)` so that it is clear if the param does not exist it will be None?

timesketch/api/v1/resources/upload.py
```diff
@@ -440,6 +450,14 @@ def post(self):
...
+        if res:
+            headersMapping = json.loads(res)
+        else:
+            headersMapping = None
+        print(headersMapping)
```
jaegeral, line 459: Please remove the print statement, if output is needed, use logger instead

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -36,10 +36,34 @@ limitations under the License.
...
+        <div v-if="error">
+          <article class="message is-danger mb-0">
+            <div class="message-body">
+              {{ error }}
+            </div>
+          </article>
+        </div>
+      </div>
+
+      <!---
```
jaegeral, line 48: Only — not three -

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -36,10 +36,34 @@ limitations under the License.
...
+        </div>
+      </div>
+
+      <!--- 
+      
+      Next lines of code represent the dropdwon menu
+      It is dynamically generated (with an extra option: New header) to allow
+        the user to map the missing header with an exsting one
+
+      --->
```
jaegeral, line 54: Same here, only two times -

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -36,10 +36,34 @@ limitations under the License.
...
+      <div v-for="(value, key) in hMissing">
```
jaegeral, line 55: HMissing - is that a typo? If it is for headerMissing I suggest to write it out.

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -74,27 +98,84 @@ limitations under the License.
...
     return {
+      headers: [], // headers in the CSV 
+      hMissing : {}, 
+      headersMapping : {},
       form: {
         name: '',
         file: '',
       },
-      fileName: '',
+      fileName: '',
```
jaegeral, line 114: Remove trailing whitespace

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -74,27 +98,84 @@ limitations under the License.
...
+       *      it can be "created" a new one and a default value is associated to each row for that header
+       */
+
+        var key = e.target.name; // key = datetime | message | timestamp_desc
+        var value = e.target.options[e.target.options.selectedIndex].text; // value = existing CSV header
+        this.headersMapping[key][0] = value
+        if(value === "New header"){ // ask to the user the default row's value
+          var newHVal = ""
+          while(!newHVal){
+            newHVal = prompt("Insert the default value for this header")
```
jaegeral, line 137: NewHeaderValue maybe clearer

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -133,13 +237,78 @@ export default {
...
+        // this is done to prevent JS reading a large CSV file (GBs) 
+        var vueJS = this
+        reader.readAsText(file.slice(0, 1000))
+        reader.onloadend = function(e){
+            if (e.target.readyState == FileReader.DONE){  // DONE == 2
+              
+              /* 3a. Extract the headers from the CSV*/ 
+              
+              var data = e.target.result
+              var headers = data.split("\n")[0] //<--- is there a better way to split columns?
```
jaegeral, line 266: Comment after // seems double

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -133,13 +237,78 @@ export default {
...
+              headers = headers.split(",") // all  headers of CSV uploaded
```
jaegeral, line 267: Is there a way to find out a separator other then comma

timesketch/lib/utils.py
```diff
@@ -130,13 +129,55 @@ def validate_indices(indices, datastore):
...
-def read_and_validate_csv(file_handle, delimiter=",", mandatory_fields=None):
+def headers_mapping_sanity_check(headers, headersMapping):
+    """Sanity check for headers mapping
+
+    Args:
+        headers: list of headers found in the CSV file
+        headersMapping: mapping of the mandatory headers with the exsting one.
+                            This feature is useful only for CSV file
+    
+    Returns: error message if any of the sanity checks fails, otherwise, OK
```
jaegeral, line 140: Hm wouldn’t you better return error messages if it is not okay?

timesketch/lib/utils.py
```diff
@@ -149,6 +190,24 @@ def read_and_validate_csv(file_handle, delimiter=",", mandatory_fields=None):
         delimiter = codecs.decode(delimiter, "utf8")
 
     header_reader = pandas.read_csv(file_handle, sep=delimiter, nrows=0)
+
+    if headersMapping:
+        # sanity check of headersMapping 
+        # e.g., 2 or more mandatory headers are mapped with the same exisiting CSV header
+        res = headers_mapping_sanity_check(header_reader, headersMapping)
+        if res != "OK":
```
jaegeral, line 198: I would rather prefer if the method returns a true or false if it was okay or not. Comparing strings is not trivial, e.g. someone could use the method and then try to compare agains „ok“ and not „OK“.

jaegeral (changes requested):
First round of comments. Overall I really like the coding style and comments, keep that up. There seem to be some linter issues in the PR.

timesketch/lib/utils.py
marcobrotto: That's a good consideration. I thought to return an error to allow the user to understand which is the problem with her headers mapping. What about: returning an error (i.e., e msg string) ONLY when any of the sanity checks fail? In this case the if condition would become: `if res:`, thus, no string comparison. What do you think?

tomchop (changes requested):

timesketch/lib/utils.py
```diff
@@ -249,7 +338,7 @@ def read_and_validate_redline(file_handle):
         yield row_to_yield
 
 
-def read_and_validate_jsonl(file_handle):
+def read_and_validate_jsonl(file_handle, **_):
```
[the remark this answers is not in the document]
tomchop: `# pylint: disable=error-message`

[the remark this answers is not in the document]
tomchop: Maybe rename this to `**kwargs` and explain why in the docstring.

timesketch/lib/utils.py
```diff
@@ -130,13 +130,66 @@ def validate_indices(indices, datastore):
...
+    headersMapping=None
+):
     """Generator for reading a CSV file.
 
     Args:
         file_handle: a file-like object containing the CSV content.
         delimiter: character used as a field separator, default: ','
-        mandatory_fields: list of fields that must be present in the CSV header.
+        mandatory_fields: list of fields that must be present in the CSV header
+        headersMapping: mapping the mandatory headers with the exsting one.
```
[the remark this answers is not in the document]
marcobrotto: I left a verbose description where this arg is actually used. On the other parts I changed it in this way: 
`Python dictionary representing the mapping between mandatory (key) and existing CSV headers (value).`

tomchop (changes requested):

timesketch/lib/utils.py
```diff
@@ -130,14 +148,93 @@ def validate_indices(indices, datastore):
...
+
+    Returns: error message if any of the sanity checks fails
+    """
+
+    # 1. The mapping is done only if the mandatory header is missing, and
+    # 2. When a new column is created, a mandatory default value must be set
+    for mand_header in headers_mapping:
+        if mand_header in headers:
+            return "Mapping done only if the mandatory header is missing"
+        if (headers_mapping[mand_header][0] == "New header" and
```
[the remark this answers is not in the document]
marcobrotto: here you raise an exception if the condition is true. In the other if-statements I return the error message. Did you mean something like that:
```
if new_header_name == "New header" and not new_header_value:  # equivalent to new_header_value == ""
  error = "..."
  return error
```
?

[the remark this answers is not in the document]
tomchop: Yes, sorry - I meant raise or return or whatever the behavior of your function is - although as previously stated if you only check for a result, you might want to raise the error in your function direclty instead of return-and-check

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -107,6 +238,12 @@ export default {
       formData.append('context', this.fileName)
       formData.append('total_file_size', this.form.file.size)
       formData.append('sketch_id', this.$store.state.sketch.id)
+      let exstension = this.fileName.split('.')[1]
+      if(exstension === 'csv'){
```
berggren, line 242:
```suggestion
      let extension = this.fileName.split('.')[1]
      if(extension === 'csv'){
```

berggren (changes requested):

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -74,30 +114,121 @@ limitations under the License.
...
   methods: {
+    checkMandatoryHeader: function(){
+      this.mandatoryHeaders.forEach(mandatoryHeader => {
+        if( this.headers.indexOf(mandatoryHeader) < 0 ){
+          this.missingHeaders.push(mandatoryHeader)
+        }
+        else{
+          this.missingHeaders = this.missingHeaders.filter(header => header !== mandatoryHeader)
+        }
+      })
```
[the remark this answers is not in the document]
tomchop: if so, you can make missingHeaders a reactive property as well

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -74,30 +114,121 @@ limitations under the License.
...
+        this.headersMapping = this.headersMapping.filter(mapping => mapping["target"] !== target)
+        this.headersMapping.push({
+          target : target,
+          source : source,
+          default_value : defaultValue // leave snake case for python server code
+        })
+
+        // 1. check if mapping is completed, i.e., if all mandatory headers have been set
+        if(this.headersMapping.length !== this.missingHeaders.length){
+          this.error = "Finish to map mandatory headers"
```
tomchop, line 208:
```suggestion
          this.error = `All mandatory headers need to be mapped (missing: ${this.missingHeaders}`
```

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -74,30 +114,121 @@ limitations under the License.
...
+        }
+        else{
+          // 2. check for duplicate headers (except from 'New header')
+          let duplicates = this.headersMapping.filter(mapping => mapping["source"] !== "New header").map(e => e.source)
+          if(duplicates.length > new Set(duplicates).size){
+            this.error = "New headers mapping contains duplicates"
```
tomchop, line 214:
```suggestion
            this.error = `New headers mapping contains duplicates (${duplicates})`
```

tomchop (changes requested):

timesketch/lib/utils.py
```diff
@@ -249,11 +368,13 @@ def read_and_validate_redline(file_handle):
         yield row_to_yield
 
 
-def read_and_validate_jsonl(file_handle):
+def read_and_validate_jsonl(file_handle, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
marcobrotto: It would make more sense. However, since read_and_validate is called here:
```
for event in read_and_validate(
            file_handle=file_handle,
            headers_mapping=headers_mapping,
            delimiter=delimiter
        ):
```

With an if/else statement it would be something like that:
```
if exstension == "csv":
for event in read_and_validate(
            file_handle=file_handle,
            headers_mapping=headers_mapping,
            delimiter=delimiter
        ):
            opensearch.import_event(
                index_name, event_type, event, timeline_id=timeline_id
            )
            final_counter += 1
else:
for event in read_and_validate(file_handle=file_handle):
            opensearch.import_event(
                index_name, event_type, event, timeline_id=timeline_id
            )
            final_counter += 1
```

What do you think?

tomchop (changes requested):

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -36,19 +36,70 @@ limitations under the License.
...
+      -->
+      <div class="field" v-if="fileName">
+        <label class="label">Name</label>
+        <div class="control">
+          <input v-model="form.name" class="input" type="text" required placeholder="Name your timeline" />
+        </div>
+        <div v-if="extension === 'csv'">
+          <label class="label">CSV Separator</label>
+          <div class="control" v-for="(v, key) in delimitersList" :key="key">
+              <input type="radio" name="CSVDelimiter" :value="v" v-on:click="changeCSVDelimiter($event)" >
```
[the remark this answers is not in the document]
marcobrotto: Gotcha!
```
<input type="radio" name="CSVDelimiter" :value="v" v-model="CSVDelimiter" @change="changeCSVDelimiter"/>
```
Still, I need :value to assign the value at this radio button. However, changeCSVDelimiter is not update anymore the CSVDelimiter variable

berggren (approved):
LGTM

PR #2262, opened 2022-07-21
Allow complex tags in the Intelligence attribute

tomchop:
Currently, IOC tags in the intelligence attribute are just strings, and they get coloring and weight from default values set in a server-side config file. This isn't very flexible, as tags can be quite arbitrary, especially if they are imported from a third-party TIP.

This change allows the Intelligence view to interpret tags that are already rich - if a third-party application wants to send data to the intelligence view, it can manually set the weight and class that will be used for display.


jaegeral (changes requested):
/home/runner/work/timesketch/timesketch/timesketch/frontend/src/views/Intelligence.vue
Error:   300:[17](https://github.com/google/timesketch/runs/7453855761?check_suite_focus=true#step:5:18)  error  'tagKey' is already defined  no-redeclare

itsmvd:
Still same error `error 'tagKey' is already defined no-redeclare`.

jaegeral (approved):

PR #2267, opened 2022-07-29
Update dev README

Nexidian:
- Updated dev README container commands to match the updated container name

---
The dev guide for starting the development webserver is using commands that are looking for a container named `dev_timesketch` however the `docker-compose.yml` has this container named as `timesketch-dev`. 

This causes sets `$CONTAINER_ID` to an empty string which breaks the startup command

```yml
services:
  timesketch:
    container_name: timesketch-dev
```

Please let me know if you would like an issue raised for this, I was unsure as this counts as documentation changes

### DIFF
--- a/docker/dev/README.md
+++ b/docker/dev/README.md
@@ -21,13 +21,13 @@ Timesketch development server is ready!
 ### Find out container ID for the timesketch container
 
 ```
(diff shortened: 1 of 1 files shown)


berggren (approved):
LGTM. Thanks for contributing!

PR #2288, opened 2022-08-19
JSONL MAPPING HEADERS

### LINKED ISSUE #2284: JSON and JSONL headers mapping
# Headers mapping for JSON and JSONL file

We aim to extend the headers mapping selection feature for JSON and JSONL files (see PRs #2257 and #2273).

**Solution / Approach**
Adopt the same approach used for CSVs headers mapping

**Note for JSON file**

A JSON file differs from a CSV because it is not a "flat" data structure. Our purpose is to allow Timesketch to elaborate JSON file as well. How?
For example:
```
{"datetime" : 2022-11-11,
 "timestamp_desc" : "creation_time",
 "message" : {
                         "field1" : "foo",
                         "field2" : "baf",
                     }
"extra_field" : {
                         "field1" : "foo",
                         "field2" : "baf",
                      } 

}
```
We aim to make this structure flat. How? In the following way:
| datetime   	| timestamp_desc 	| message                             	| extra_field                         	|
|------------	|----------------	|-------------------------------------	|-------------------------------------	|
| 2022-11-11 	| creation_time  	| {"field1": "foo", "field2" : "baf"} 	| {"field1": "foo", "field2" : "baf"} 	|

marcobrotto:
# JSONL MAPPING HEADERS

`closes #2284`

This PR closes the issue #2284 and it  extends the PR #2261 and #2287 regarding the headers mapping feature.

## Main contribution and accomplishment

This PR allows a user to use the headers mapping feature also for JSONL file.

Currently a JSONL file that does not have the required headers, i.e., `message`, `datetime` and `timestamp_desc`, can not be imported into Timesketch.

With this feature we overcome this limitation by introducing the headers mapping functionality. In particolar, a user can now map an existing field in the JSONL with a missing required header.

### How are the headers extracted from the JSONL?

It is easier to extract the headers of a CSV because we can assume that they are on the first line. However, a JSONL file is a different structure where each row corresponds to a dictionary. For this reason, each row could have different headers.

```
{"time" : "2022-11-11", "fileData" : {"path": "/home/luke/Desktop", "filename": "helloBR8", "size" : "22"}}

{"time" : "2022-11-12", "fileData" : {"path": "/home/luke/Desktop", "filename": "na", "size" : "22"}, "logsource" : "sysmon", "" }
```

We can observe that the second line of the example contain a new field that is: "logsource"

**What can't we say?**

All headers are present in the first line of the file. How can we overcome these limitations? Making assumptions and *failing safe* (see below).

**What can we assume?** 

* [Each entry of a JSONL terminates with a new line](https://dbconvert.com/blog/json-lines-data-stream/). To this end, parsing a JSONL doesn’t require custom parsers. Just read a line, parse as JSON, read a line, parse as JSON… and so on.
* We assume that all the headers are in the first line of the JSONL. When we perform the mapping on the server-side, we check first if the `target` field (e.g., `datetime`) is already in the JSONL line. If so, there is no need for mapping. Otherwise, we proceed with the mapping: if any of the fields specified in the `source` value is missing in the JSONL line, we throw a Runtime exception.

```
ld_keys = jsonl_line.keys()
for mapping in headers_mapping:
    if mapping["target"] not in ld_keys:
        # proceed with the mapping
        if mapping["source"]:
            for source in mapping["source"]:
                if source in ld_keys:
                    # proceed with the mapping
                else:
                    # throw a Runtime Error
        else:
            # create the new column with a default value
    else:
        continue
```

## Frontend major modifications:

Wrt the previous request, we did not modify the structure of the HTML.

We added the following methods:

* `extractJSONLHeader()` that extracts the first 
lines of the JSONL file and converts them into JSON objects.

We added around the code of `UploadForm.vue` the various conditions to display the headers mapping functionality (previously displayed only for `CSV` files).

## Backend major modifications:

We added the function `rename_jsonl_headers(linedict, headers_mapping, lineno)` in the file `timesketch/lib/utils.py` to map the missing headers according to the headers mapping

## CHECKS
- [x] All tests succeed.
- [x] Unit tests added 
    - 2 test to verify the new function `rename_jsonl_headers(linedict, headers_mapping, lineno)` in the file `timesketch/lib/utils.py`
- [x] Documentation updated: see `import-from-json-csv.md` file.

## Final note on JSON file

This PR does not take into consideration importing JSON files. A JSON file has a similar structure of a JSONL but tricky to be considered by Timesketch compared to a JSONL.

In fact, a JSONL file is a flat structure. On the contrary, a JSON file has a hierarchy structure, for example:

```
{
    "event-01" : {data-01},
    "event-02" : {data-02},
    ...
    "event-N" : {data-N},
}
```
Even if the structure is not too deep, we need to extract the values from `event-i`. After careful consideration, extracting this information from a JSON file seems not worthy because the structure of the file could change and not be processed by the headers mapping (e.g., `data-i` could be in a deeper layer of the JSON object).

### DIFF
--- a/timesketch/api/v1/resources/upload.py
+++ b/timesketch/api/v1/resources/upload.py
@@ -289,7 +289,7 @@ class UploadFileResource(resources.ResourceMixin, Resource):
         meta["task_id"] = task_id
         return self.to_json(timeline, status_code=HTTP_STATUS_CODE_CREATED, meta=meta)
 
-    def _upload_events(self, events, form, sketch, index_name):
+    def _upload_events(self, events, form, sketch, index_name, headers_mapping=None):
         """Upload a file like object.
 
         Args:
@@ -315,6 +315,7 @@ class UploadFileResource(resources.ResourceMixin, Resource):
             form=form,
             data_label=data_label,
             enable_stream=form.get("enable_stream", False),
+            headers_mapping=headers_mapping,
         )
 
     def _upload_file(
@@ -475,6 +476,8 @@ class UploadFileResource(resources.ResourceMixin, Resource):
 
         # headers mapping: map between mandatory headers and new ones
         headers_mapping = json.loads(form.get("headersMapping", "{}")) or None
+        logger.info(headers_mapping)
+
         delimiter = form.get("delimiter", ",")
 
         sketch_id = form.get("sketch_id", None)
@@ -529,5 +532,9 @@ class UploadFileResource(resources.ResourceMixin, Resource):
             )
 
         return self._upload_events(
-            events=events, form=form, sketch=sketch, index_name=index_name
+            events=events,
+            form=form,
+            sketch=sketch,
+            index_name=index_name,
+            headers_mapping=headers_mapping,
         )
--- a/timesketch/frontend/src/components/Common/UploadForm.vue
+++ b/timesketch/frontend/src/components/Common/UploadForm.vue
@@ -41,7 +41,7 @@ limitations under the License.
       </div>
 
       <div class="field">
-        <div v-if="extension === 'csv'">
+        <div v-if="['csv', 'jsonl'].includes(extension)">
           <hr />
 
           <!-- List of button: showHelper, showPreview, addColumnsToPreview -->
@@ -263,17 +263,24 @@ limitations under the License.
             <hr />
           </div>
 
-          <!-- CSV delimiter selection: the program will parse the file according to this choice -->
-          <label class="label">CSV Separator</label>
-          <div class="control" v-for="(v, key) in delimitersList" :key="key">
-            <input
-              type="radio"
-              name="CSVDelimiter"
-              :value="v"
-              v-model="CSVDelimiter"
-              @change="changeCSVDelimiter"
-            />
-            {{ key }} ({{ v }})
+          <!--
+
+            CSV delimiter selection: the program will parse 
+            the file according tothis choice
+          
+          -->
+          <div v-if="extension === 'csv'">
+            <label class="label">CSV Separator</label>
+            <div class="control" v-for="(v, key) in delimitersList" :key="key">
+              <input
+                type="radio"
+                name="CSVDelimiter"
+                :value="v"
+                v-model="CSVDelimiter"
+                @change="changeCSVDelimiter"
+              />
+              {{ key }} ({{ v }})
+            </div>
           </div>
         </div>
       </div>
@@ -378,7 +385,13 @@ export default {
   },
   computed: {
     headers() {
-      return this.headersString.split(this.CSVDelimiter);
+      let headers = [];
+      if (this.extension === "csv") {
+        headers = this.headersString.split(this.CSVDelimiter);
+      } else if (this.extension === "jsonl") {
+        headers = Object.keys(this.headersString);
+      }
+      return headers;
     },
     missingHeaders() {
       return this.mandatoryHeaders.filter(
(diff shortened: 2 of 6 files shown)


timesketch/api/v1/resources/upload.py
```diff
@@ -475,6 +476,8 @@ def post(self):
 
         # headers mapping: map between mandatory headers and new ones
         headers_mapping = json.loads(form.get("headersMapping", "{}")) or None
+        logger.info(headers_mapping)
```
jaegeral, line 479: do we need this moving forward or is this more a debug message, if we leave it, I opt to add more text around what it is.

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -263,17 +263,24 @@ limitations under the License.
...
-              name="CSVDelimiter"
-              :value="v"
-              v-model="CSVDelimiter"
-              @change="changeCSVDelimiter"
-            />
-            {{ key }} ({{ v }})
+          <!--
+
+            CSV delimiter selection: the program will parse 
+            the file according tothis choice
```
jaegeral, line 269: add a space

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -263,17 +263,24 @@ limitations under the License.
...
-            <input
-              type="radio"
-              name="CSVDelimiter"
-              :value="v"
-              v-model="CSVDelimiter"
-              @change="changeCSVDelimiter"
-            />
-            {{ key }} ({{ v }})
+          <!--
+
```
jaegeral, line 267: remove the empty line

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -423,29 +445,36 @@ export default {
...
+              valuesAndHeaders[header].push(this.valuesString[i][header]);
+            } else {
+              valuesAndHeaders[header] = [this.valuesString[i][header]];
+            }
+          }
         }
       } else {
-        console.log("JSONL not supported (yet) for this feature");
+        console.error(
+          this.extension + " extension not supported for this feature"
```
jaegeral, line 467: maybe "File extension not supported for this feature"?

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -638,11 +663,14 @@ export default {
       if (this.form.file.size <= 0) {
         this.error.push("Please select a non empty file");
       }
-      let allowedExtensions = ["csv", "json", "jsonl", "plaso"];
```
jaegeral, line 641: hm I am not sure if we should take away json as a filetype. Yes in theory everyone should use jsonl as extension but I am not sure it might break stuff for people

jaegeral (commented):
First round of comments

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -41,7 +41,7 @@ limitations under the License.
       </div>
 
       <div class="field">
-        <div v-if="extension === 'csv'">
+        <div v-if="['csv', 'jsonl'].includes(extension)">
```
jaegeral, line 44: please add `.json` as a valid file extension (also in other places of that PR).

jaegeral (changes requested):

marcobrotto:
I close this PR because I port the functionality on a new PR (only frontend-ng)

PR #2290, opened 2022-08-22
Fix redirect for OIDC logins

### LINKED ISSUE #2289: Redirect URL is missing after OIDC authn
**Describe the bug**
Redirect URL is missing after OIDC authn. The URL parameter is lost when the redirect happen. The solution is to store the parameter in the session, and redirect after the OIDC has been verified (state with CSRF token etc).

berggren:
Redirect URL is missing after OIDC authn. The URL parameter is lost when the redirect happen. The solution is to store the parameter in the session, and redirect after the OIDC has been verified (state with CSRF token etc).

**Closing issues**
closes #2289


jaegeral (approved):
LGTM, thx

Issue #1621, opened 2021-02-16
Rename master -> main


berggren:
Rename master branch -> main branch.
https://github.com/github/renaming

jaegeral:
Maybe we should start to track which side effects this might have if we start the journey.

It will be more or less a ```s/master/main``` on https://github.com/google/timesketch/search?q=master

WDYT @kiddinn  / @berggren

jaegeral:
```
 1. Renamed your local master branch to main.
 2. Pushed the new main branch to the remote repository and set it to track origin/main.

  The next critical step, which you need to perform manually on GitHub, is:

  Step 3: Update the Default Branch on GitHub

   1. Go to your repository on GitHub.
   2. Click on the Settings tab.
   3. In the left sidebar, click on Branches.
   4. You will see the "Default branch" section. Click the switch branches icon (two arrows).
   5. Select main from the dropdown and click Update. Read the warning and confim
  ```

jaegeral:
I asked Gemini for a comprehensive guide and here is it:

# Part 1: For Main Repository Maintainers (`google/timesketch`)

These steps should be performed by a maintainer with administrative rights on the main
  repository.

## Step 1: Rename the Local `master` Branch

  In a local clone of the main repository, rename the branch.

   1 # Make sure you have the latest version of the master branch
   2 git checkout master
   3 git pull origin master
   4
   5 # Rename the branch to main
   6 git branch -m master main

## Step 2: Push the New `main` Branch to Remote

  Push the newly named main branch to the remote repository on GitHub. The -u flag sets the
  upstream tracking branch.

   1 git push -u origin main

 ## Step 3: Update the Default Branch on GitHub (Critical Step)

  This must be done on the GitHub website.

   1. Navigate to the main repository's page on GitHub.
   2. Go to Settings > Branches.
   3. Next to the "Default branch" name, click the switch icon (two arrows).
   4. In the dialog, select main as the new default branch and click Update.
   5. Read the warning and confirm. GitHub will now show main as the default branch.

##  Step 4: Update Branch Protection Rules

  If you have branch protection rules for master, you must update them to apply to main.

   1. Go to Settings > Branches.
   2. In the "Branch protection rules" section, you can either edit the existing rule for master and
      change the branch name to main, or create a new rule for main that mirrors the old settings.

 ## Step 5: Update Internal Repository References

  Search the repository for any hardcoded references to master and update them to main. This is
[...]

PR #2293, opened 2022-08-22
Replace master in most places and replace with main

jaegeral:
This issue will most likely fail all the tests.

This is an attempt to replace the term master with main.

#1621

When merged, the main branch also needs to be renamed according to https://github.com/github/renaming.


berggren:
I will close this out for now. We will get to this, but the longer this PR sits still the conflicts are gonna be more difficult to manage. We will do this as soon as time permits.

Issue #2301, opened 2022-08-24
Migrate Sigma away from file based to SQL based


jaegeral:
Migrate the Sigma rules into the Database to be able to improve various aspects

Background:

At the moment, Sigma rules are stored and pulled into Timesketch from disk with os.walk every time an analyst does actions with Sigma rules (e.g. run an analyzer, list Sigma rules). That takes some time but was optimized by enabling caching.
To get a new rule into a Timesketch instance, the raw file needs to be copied via SSH or other means to a Timesketch server and the rule needs to have a row in the Sigma status csv file.

- [x] Database model
- [x] Database model tests
- [x] API methods
   - [x] add
   - [x] modify
   - [x] list
   - [x] delete
 - [x] API methods tests
 - [x] API client methods
 - [x] API client method tests
 - [x] UX update
 - [x] refactor `es_query` to `query_string` - https://github.com/google/timesketch/pull/2333
 - [x] update JS REST API Client - https://github.com/google/timesketch/pull/2333
 - [x] remove references to `folder` and `rel_folder` - https://github.com/google/timesketch/pull/2333
 - [x] Refactor `sigma` to `sigmarule` - https://github.com/google/timesketch/pull/2333
 - [x] Modify analyzers to use the database - https://github.com/google/timesketch/pull/2347
- [x] Remove File bases stuff
   - [x] Remove Sigma rule folder config value - https://github.com/google/timesketch/pull/2509
   - [x] Remove Sigma rule status csv file from config - https://github.com/google/timesketch/pull/2509
   - [x] Remove deprecated methods from JS REST client - https://github.com/google/timesketch/pull/2509
   - [x] Remove deprecated methods from API client (and tests) - https://github.com/google/timesketch/pull/2509
   - [x] Remove deprecated methods from API v1 (and tests) - https://github.com/google/timesketch/pull/2509
- [x] Documentation

jaegeral:
https://github.com/google/timesketch/pull/2296

PR #2302, opened 2022-08-24
Add Sigma Database to the API

jaegeral:
Next step on https://github.com/google/timesketch/issues/2301 - migrating the API

This PR will not be merged until API, API Client and UI are merged

The PR will introduce new temporary new routes for the database interactions so that the API client can also be developed based on that and once that all is in and merged, we will deprecate the new routes and redirect the old routes to the new routes so it will not break anything for users.

- [x] rename `es_query` to `query_string`


timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
         sigma_rules = []
 
         try:
-            sigma_rules = ts_sigma_lib.get_all_sigma_rules()
+            sigma_rules_results = SigmaRule.query.filter_by()
```
berggren, line 61: You want to get all rules, correct? Then you can do
`SigmaRule.query.all()`

timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
...
-        meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
-        return jsonify({"objects": [return_rule], "meta": meta})
+    @login_required
+    def delete(self, rule_uuid):
+        """Handles DELETE request to the resource.
+        Args:
+            rule_uuid: uuid of the rule
+        """
+
+        all_rules = SigmaRule.query.all()
```
berggren, line 149: Isn't it enough to get the rules for that specific UUID?

timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
...
+
+        rule_yaml = form.get("rule_yaml", "")
+
+        if not rule_yaml:
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND,
+                "No rule_yaml supplied.",
+            )
+
+        # not sure if that is needed
```
berggren, line 216: ? please find out and remove :)

timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
...
+                abort(HTTP_STATUS_CODE_CONFLICT, "Rule already exist")
+
+        sigma_rule = SigmaRule.get_or_create(
+            rule_yaml=form.get("rule_yaml", ""),
+            description = parsed_rule.get("description", ""),
+            title = parsed_rule.get("title", ""),
+            user = current_user
+            )
+
+        sigma_rule.query_string = parsed_rule.get("es_query", "")
```
berggren, line 240: can we call this opensearch_query?

timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
...
+            parsed_rule["rule_uuid"] = rule_uuid
+        parsed_rule["id"] = rule.id #this is the database id (1,2,3...)
+        parsed_rule["created_at"] = str(rule.created_at)
+        parsed_rule["updated_at"] = str(rule.updated_at)
+        parsed_rule["status"] = rule.get_status.status
+        parsed_rule["title"] = rule.title
+        parsed_rule["description"] = rule.description
+        parsed_rule["rule_yaml"] = rule.rule_yaml
+        return_rules.append(
+            parsed_rule)
```
berggren, line 134: Seeing we are doing this in two resources, I opt to create a private util function at the top of this file to do this

timesketch/api/v1/resources/sigma.py
jaegeral: done

timesketch/api/v1/resources/sigma.py
jaegeral: true, will change

timesketch/api/v1/resources/sigma.py
jaegeral: found out :-)

timesketch/api/v1/resources/sigma.py
jaegeral: refactored it and called it `_enrich_sigma_rule_object`WDYT?

timesketch/api/v1/resources/sigma.py
```diff
@@ -47,53 +58,201 @@ def get(self):
         sigma_rules = []
 
         try:
-            sigma_rules = ts_sigma_lib.get_all_sigma_rules()
+            sigma_rules_results = SigmaRule.query.filter_by()
+            # Return a subset of the Sigma objects to reduce the amount of
+            # data sent to the client.
+            for rule in sigma_rules_results:
+                parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule.rule_yaml)
+                sigma_rules.append({"id": rule.id,
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/api/v1/resources/sigma.py
jaegeral: as per conversation, we will name this query_string moving forward

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+            logger.error(
+                "Error, unable to get Sigma rules", exc_info=True
+            )
+            abort(HTTP_STATUS_CODE_NOT_FOUND, f"Value Error, {e}")
+        meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
+        return jsonify({"objects": sigma_rules, "meta": meta})
+
+
+class SigmaRuleResource(resources.ResourceMixin, Resource):
+    """Resource to get a Sigma rule."""
```
jkppr, line 221: This docstring implies the class only offers a method to get a Sigma rule. But, it also offers methods for deleting and changing or creating a sigma rule. This should be reflected in the docstring.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+        sigma_rule.rule_uuid = parsed_rule.get("id", None)
+
+        try:
+            db_session.add(sigma_rule)
+            db_session.commit()
+        except IntegrityError:
+            abort(HTTP_STATUS_CODE_CONFLICT, "Rule already exist")
+
+        if sigma_rule is None:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sigma was parsed")
```
jkppr, line 362: Shouldn't this be checked before you try to insert the content into the database?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+            )
+
+        sigma_rule.query_string = parsed_rule.get("query_string", "")
+        sigma_rule.rule_uuid = parsed_rule.get("id", None)
+
+        try:
+            db_session.add(sigma_rule)
+            db_session.commit()
+        except IntegrityError:
+            abort(HTTP_STATUS_CODE_CONFLICT, "Rule already exist")
```
jkppr, line 359: Does failing with an IntegrityError really means the Rule already exist? You have checked for existence of the rule already. To my understanding, an IntegrityError would also be thrown if any of the input data does not match the required field etc. WDYT?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+                abort(HTTP_STATUS_CODE_CONFLICT, "Rule already exist")
+
+        sigma_rule = SigmaRule.get_or_create(
+            rule_yaml=form.get("rule_yaml", ""),
+            description = parsed_rule.get("description", ""),
+            title = parsed_rule.get("title", ""),
+            user = current_user
+            )
+
+        sigma_rule.query_string = parsed_rule.get("query_string", "")
```
jkppr, line 352: `sigma_rule.query_string`: Just to be sure, since I do not see this field existing in the SigmaRule class. Does it need to be defined there?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+        if not rule_uuid:
+            rule_uuid = parsed_rule.get("id")
+
+        if not isinstance(rule_yaml, str):
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN, "rule_yaml needs to be a string."
+            )
+
+        # Query rules to see if it already exist
+        rule = SigmaRule.query.filter_by(rule_uuid=rule_uuid).first()
```
jkppr, line 340: Since you only use the `rule_uuid` to check if a rule with that ID already exist, you could save some data by just querying what you need from the db: `rule = SigmaRule.query(SigmaRule.rule_uuid).filter_by(rule_uuid=rule_uuid).first()`

Or you can make the whole check a bit easier:
```
if SigmaRule.query(SigmaRule.rule_uuid).filter_by(rule_uuid=parsed_rule.get("rule_uuid")).first() is None:
            abort(HTTP_STATUS_CODE_CONFLICT, "Rule already exist")
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+
+        parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule_yaml)
+
+        if not rule_uuid:
+            rule_uuid = parsed_rule.get("id")
+
+        if not isinstance(rule_yaml, str):
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN, "rule_yaml needs to be a string."
+            )
```
jkppr, line 337: Shouldn't the check for `rule_yaml` being a str/text happen before you try to convert it via `ts_sigma_lib.get_sigma_rule_by_text(rule_yaml)`? If it is not text I guess it will already fail during that call.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,253 @@ def get(self, rule_uuid):
...
+            # data sent to the client.
+            for rule in sigma_rules_results:
+                sigma_rules.append(_enrich_sigma_rule_object(rule=rule))
+
+        except ValueError as e:
+            logger.error(
+                "Error, unable to get Sigma rules", exc_info=True
+            )
+            abort(HTTP_STATUS_CODE_NOT_FOUND, f"Value Error, {e}")
+        meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
```
jkppr, line 216: What is the `current_user.username` in the metadata used for?

jkppr (commented):
some comments

timesketch/api/v1/resources/sigma.py
jaegeral: yes it was `es_query` before and we decided to use the opportunity to rename it to a more general term

timesketch/api/v1/resources/sigma.py
jaegeral: good point, changed

timesketch/api/v1/resources/sigma.py
jaegeral: true, changed it

timesketch/api/v1/resources/sigma.py
jaegeral: at the moment the api client does not use it, but it was in there before: https://github.com/google/timesketch/blob/master/timesketch/api/v1/resources/sigma.py#L58 and I do not intend to remove things that where in in this PR, so if that is questionable, let me make a issue for this

jaegeral: https://github.com/google/timesketch/issues/2323

timesketch/api/v1/resources/sigma.py
```diff
@@ -58,17 +88,15 @@ def get(self):
         meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
         return jsonify({"objects": sigma_rules, "meta": meta})
 
-
+# TODO(jaegeral): deprecate this class
```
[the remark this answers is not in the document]
jaegeral: Well that will not be deprecated in this PR, so I rather have this TODO. It will be deprecated once the API client is also ported over to the newly created API methis in here.

timesketch/api/v1/resources/sigma.py
jaegeral: I like that

timesketch/api/v1/resources/sigma.py
jaegeral: true good point, moved

timesketch/api/v1/resources/sigma.py
jaegeral: made the error more general.

timesketch/api/v1/resources/sigma.py
jaegeral: your query would fail with an ACL Error ```(Pdb) SigmaRule.query(SigmaRule.rule_uuid).filter_by(rule_uuid=rule_uuid).first()
*** TypeError: 'AclBaseQuery' object is not callable```

If the rule does not exist, so all the remaining code would be in a try except block, so I actually would like to exit only if the rule was found so I reverted it back to what it was before, lmk what you think

timesketch/api/v1/resources/sigma.py
```diff
@@ -25,22 +25,52 @@
...
+    """Helper function: Returns an enriched Sigma object givin a SigmaRule.
+
+    Args:
+        SigmaRule: uuid of the rule
+
+    Returns:
+        enriched Sigma dict
+    """
+    parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule.rule_yaml)
+    parsed_rule["rule_uuid"] = parsed_rule.get("id",rule.rule_uuid)
```
tomchop, line 57:
```suggestion
    parsed_rule["rule_uuid"] = parsed_rule.get("id", rule.rule_uuid)
```

Didn't your linter catch this? There are other place in the code where this seems to reoccur.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+        try:
+            sigma_rule = ts_sigma_lib.get_sigma_rule_by_text(content)
+
+        except ValueError:
+            logger.error(
+                "Sigma Parsing error with the user provided rule", exc_info=True
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Error unable to parse the provided Sigma rule",
```
tomchop, line 153: Consider storing the error message in a variable so as to be consistent in what gets logged by `logger.error` and what gets sent back to the user. Same goes for all cases below.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+
+        except NotImplementedError as exception:
+            logger.error(
+                "Sigma Parsing error: Feature in the rule provided "
+                " is not implemented in this backend",
+                exc_info=True,
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Error generating rule {0!s}".format(exception),
```
tomchop, line 164: In line with my comment above, I think the error message that gets sent back to the user could be a bit more verbose, don't you think? "Error generating rule: NotImplementedError" is a very obscure error message to present to a client.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+            # Return a subset of the Sigma objects to reduce the amount of
+            # data sent to the client.
+            for rule in sigma_rules_results:
+                sigma_rules.append(_enrich_sigma_rule_object(rule=rule))
+
+        except ValueError as e:
+            logger.error(
+                "Error, unable to get Sigma rules", exc_info=True
+            )
+            abort(HTTP_STATUS_CODE_NOT_FOUND, f"Value Error, {e}")
```
tomchop, line 215: `NOT_FOUND` for a ValueError doesn't seem very REST-y? This seems a bit weird. What in your codeflow raises `ValueError`? Maybe we can be a bit more specific in how errors are handled here, and what message is displayed to the user?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+        meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
+        return jsonify({"objects": sigma_rules, "meta": meta})
+
+
+class SigmaRuleResource(resources.ResourceMixin, Resource):
+    """Resource to get / delete / post / put a Sigma rule."""
+
+    @login_required
+    def get(self, rule_uuid):
+        """Handles GET request to the resource.
```
tomchop, line 225: It would be good to be a bit more specific here - from the function name it's obvious that this handles a GET request, but it doesn't tell me much about what exactly that handling is. e.g. "Fetches a single sigma rule from the database and returns enriched contents" would be better.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+
+        Returns:
+            Sigma rule object and HTTP status code indicating
+            whether operation was sucessful.
+        """
+        form = request.json
+        if not form:
+            form = request.data
+
+        rule_yaml = form.get("rule_yaml", "")
```
tomchop, line 321:
```suggestion
        rule_yaml = form.get("rule_yaml")
```

`get` returns `None` by default, so your test in L323 will still be False.

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+                HTTP_STATUS_CODE_NOT_FOUND,
+                "No rule_yaml supplied.",
+            )
+
+        if not isinstance(rule_yaml, str):
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN, "rule_yaml needs to be a string."
+            )
+
+        parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule_yaml)
```
tomchop, line 334: The name `get_sigma_rule_by_text` really makes me think that this is gonna do a DB query and return sigma rules that match `rule_yaml` (i.e. exactly what you're doing in L340). Is that the case? If the function just parses yaml to return a rule, I would then not use `get` in the function name, but rather something like `parse_sigma_yaml(rule_yaml)`

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+        if not form.validate_on_submit():
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, "Unable to validate form data.")
+
+        abort(HTTP_STATUS_CODE_NOT_FOUND, "Method not implemented yet")
+
+        # TODO(jaegeral): complete this method
+        rule_yaml = form.get("rule_yaml", "")
+        parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule_yaml)
+
+        logger.debug(rule_yaml + parsed_rule + rule)
```
tomchop, line 303: This function looks like it's still under development, I would leave it out of this PR

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+            rule_yaml=form.get("rule_yaml", ""),
+            description = parsed_rule.get("description", ""),
+            title = parsed_rule.get("title", ""),
+            user = current_user
+            )
+
+        sigma_rule.query_string = parsed_rule.get("query_string", "")
+        sigma_rule.rule_uuid = parsed_rule.get("id", None)
+
+        if sigma_rule is None:
```
berggren, line 355:
```suggestion
        if not sigma_rule:
```

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +487,132 @@ def test_get_sigma_rule_list(self):
...
+        self.assertIsNotNone(response)
+
+        response = self.client.get(self.resource_url + "foobar/")
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_NOT_FOUND)
+
+
+class SigmaRuleListResourceTest(BaseTest):
+    """Test Sigma resource."""
+
+    resource_url = "/api/v1/sigmarule/"
```
tomchop, line 603: see comment above about DRY in tests

tomchop (changes requested):

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+                exc_info=True,
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Sigma parsing error: invalid yaml provided {0!s}".format(exception),
+            )
+
+        if sigma_rule is None:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sigma was parsed")
+        metadata = {"parsed": True}
```
berggren, line 188: Where is this used? In the UI? Do we need it?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+            Dict of sigma rules
+        """
+        sigma_rules = []
+
+        try:
+            sigma_rules_results = SigmaRule.query.all()
+            # Return a subset of the Sigma objects to reduce the amount of
+            # data sent to the client.
+            for rule in sigma_rules_results:
+                sigma_rules.append(_enrich_sigma_rule_object(rule=rule))
```
berggren, line 209:
```suggestion
            all_sigma_rules = SigmaRule.query.all()
            # Return a subset of the Sigma objects to reduce the amount of
            # data sent to the client.
            for rule in all_sigma_rules:
                sigma_rules.append(_enrich_sigma_rule_object(rule=rule))
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+
+        except ValueError as e:
+            logger.error(
+                "Error, unable to get Sigma rules", exc_info=True
+            )
+            abort(HTTP_STATUS_CODE_NOT_FOUND, f"Value Error, {e}")
+        meta = {"current_user": current_user.username, "rules_count": len(sigma_rules)}
```
berggren, line 216: Why are you sending the username in the metadata?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+
+        form = request.json
+        if not form:
+            form = request.data
+        if not form.validate_on_submit():
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, "Unable to validate form data.")
+
+        abort(HTTP_STATUS_CODE_NOT_FOUND, "Method not implemented yet")
+
+        # TODO(jaegeral): complete this method
```
berggren, line 299: Can we complete it now, before merge? :)

berggren (changes requested):
PTAL

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No rule found with this ID.")
+        return_rules = []
+
+        assert isinstance(rule, SigmaRule)
+
+        return_rules.append(
+            _enrich_sigma_rule_object(rule=rule))
+
+        meta = {
+            "current_user": current_user.username,
```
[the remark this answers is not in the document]
jaegeral: @jkppr mentioned the same, it is what is currently implemented in the API as well, so I opt to migrate that as well and make a different effort and deprecate the current username, it was requested back in the day by @kiddinn  to include it. Removing it now might break code for people (but I am not sure who would use that)

timesketch/api/v1/resources/sigma.py
jaegeral: see below

timesketch/api/v1/resources/sigma.py
```diff
@@ -25,22 +25,52 @@
...
+        SigmaRule: uuid of the rule
+
+    Returns:
+        enriched Sigma dict
+    """
+    parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule.rule_yaml)
+    parsed_rule["rule_uuid"] = parsed_rule.get("id",rule.rule_uuid)
+    parsed_rule["created_at"] = str(rule.created_at)
+    parsed_rule["updated_at"] = str(rule.updated_at)
+    parsed_rule["status"] = rule.get_status.status
```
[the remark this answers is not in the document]
jaegeral: yep this is StatusMixin, I added a comment

timesketch/api/v1/resources/sigma.py
jaegeral: no, but trying to correct it now in this file

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+            logger.error(
+                "Sigma Parsing error: an invalid yml file has been provided",
+                exc_info=True,
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Sigma parsing error: invalid yaml provided {0!s}".format(exception),
+            )
+
+        if sigma_rule is None:
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/api/v1/resources/sigma.py
jaegeral: removed

timesketch/api/v1/resources/sigma.py
jaegeral: hm that is a valid point indeed

timesketch/api/v1/resources/sigma.py
jkppr: The effort to change/debug this is probably not worth the potential benefits of only query the uuid instead of the full entry. Will mark it resolved.

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +487,132 @@ def test_get_sigma_rule_list(self):
...
+            response.json['objects'][0]["description"],
+        )
+        self.assertIn(
+            'shell',
+            response.json['objects'][0]["rule_yaml"],
+        )
+
+        # Attempt to add the same thing again
+        response = self.client.post(
+            self.resource_url + "5266a592-b793-11ea-b3de-bbbbbb/",
```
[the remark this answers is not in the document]
jaegeral: so I see your point here, and while I agree, it is different then the rest of the code style in this test file.
Maybe @berggren has thoughts on it?

timesketch/api/v1/resources/sigma.py
```diff
@@ -92,10 +120,255 @@ def get(self, rule_uuid):
...
+        rule_yaml = form.get("rule_yaml", "")
+        parsed_rule = ts_sigma_lib.get_sigma_rule_by_text(rule_yaml)
+
+        logger.debug(rule_yaml + parsed_rule + rule)
+
+
+
+    @login_required
+    def post(self, rule_uuid=None):
+        """Handles POST request to the resource.
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
 
+# TODO(jaegeral): deprecate this class
 class SigmaByTextResource(resources.ResourceMixin, Resource):
     """Resource to get a Sigma rule by text."""
 
+    @login_required
+    def post(self):
+        """Handles POST request to the resource.
+        Returns:
+            JSON sigma rule
```
tomchop, line 147: I see the docstrings for these endpoints are still a bit vague. Can we make them more explicit?

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Error unable to parse the provided Sigma rule",
+            )
+
+        except NotImplementedError as exception:
+            logger.error(
+                "Sigma Parsing error: Feature in the rule provided "
+                " is not implemented in this backend",
```
tomchop, line 177: Storing these error messages in an intermediary variable would reduce deduplication and inconsistencies in error messages (e.g. L166 / L171)

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+
+class SigmaRuleResource(resources.ResourceMixin, Resource):
+    """Resource to get / delete / post / put a Sigma rule."""
+
+    @login_required
+    def get(self, rule_uuid):
+        """Handles GET request to the Sigma Rule resource.
+
+        Handels GET API calls to /sigmarule/<string:rule_uuid>/ where the
+        rule_uuid is the primaray way to identify the rule.
```
tomchop, line 243: I think we're looking for a docstring describing **the effect** of the query rather than the process, e.g. "Fetches a single SigmaRule from the database by filtering on rule_uuid and returns a JSON representation of it"

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+        return_rules.append(_enrich_sigma_rule_object(rule=rule))
+
+        return jsonify({"objects": return_rules, "meta": {}})
+
+    @login_required
+    def delete(self, rule_uuid):
+        """Handles DELETE request to the resource.
+
+        Handels DELETE API calls to /sigmarule/<string:rule_uuid>/ where the
+        rule_uuid is the primaray way to identify the rule.
```
tomchop, line 279: "Deletes the SigmaRule matching the specified UUID from the database"

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+        db_session.commit()
+
+        return HTTP_STATUS_CODE_OK
+
+    @login_required
+    def post(self, rule_uuid=None):
+        """Handles POST request to the resource.
+        Handels POST API calls to /sigmarule/<string:rule_uuid>/ where the
+        rule_uuid is the primaray way to identify the rule.
+        The remaining attributes of the rule are provided in request itself.
```
tomchop, line 307: This is still unclear. Does POST create a new SigmaRule? Does it update an existing one? What happens if rule_uuid is not provided? What happens when it's different than the rule_uuid that's provided in the body of the POST request?

As a developer somehow familiar with flask, it's pretty clear from the code that this handles post requests, but what's not clear the "business logic" behind it.

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+
+        if not rule_uuid:
+            rule_uuid = parsed_rule.get("id")
+
+        # Query rules to see if it already exist and exit if found
+        sigma_rule_from_db = SigmaRule.query.filter_by(
+            rule_uuid=rule_uuid
+        ).first()
+        if sigma_rule_from_db:
+            if sigma_rule_from_db.rule_uuid == parsed_rule.get("rule_uuid"):
```
tomchop, line 346: You're using `parsed_rule.get("rule_uuid")` here, why not use the variable you declared in L 338?

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+        """
+        form = request.json
+        if not form:
+            form = request.data
+
+        rule_yaml = form.get("rule_yaml")
+
+        if not rule_yaml:
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND,
```
tomchop, line 327: I don't think this should return a 404, rather a 400 https://en.wikipedia.org/wiki/List_of_HTTP_status_codes or maybe a 403. Whatever you settle on, it should still be the same error than on L333. Let's try to be consistent in the error code we return when the request is valid but the sigma rule isn't.

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Missing value: 'title' from the request.",
+            )
+
+        description = parsed_rule.get("description")
+        if not description:
+            return abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Missing value: 'description' from the request.",
+            )
```
tomchop, line 363: All these integrity checks on the parsed sigma rule should happen before you query the database (L342)

timesketch/api/v1/resources/sigma.py
tomchop, line 346: If I filter the database by uuid, and get a result - is it possible that the uuid in the result is different from the uuid that I initially provided to the DB query? I wonder if this check is necessary at all.

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+            description=description,
+            title=title,
+            user=current_user,
+        )
+
+        if not sigma_rule:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sigma rule object created")
+
+        sigma_rule.query_string = parsed_rule.get("es_query")
+        sigma_rule.rule_uuid = parsed_rule.get("id")
```
tomchop, line 377: Are `id` and `uuid` the same in the database? If so, why not use `uuid` here? It might be less confusing.

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+            rule_yaml=form.get("rule_yaml"),
+            description=description,
+            title=title,
+            user=current_user,
+        )
+
+        if not sigma_rule:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sigma rule object created")
+
+        sigma_rule.query_string = parsed_rule.get("es_query")
```
tomchop, line 376: Can't we add `es_query` as part of the `get_or_create` call before? Same questio for `status=experimental` you're doing in L378; this would avoid yet another query to the database.

timesketch/api/v1/resources/sigma.py
```diff
@@ -87,15 +125,274 @@ def get(self, rule_uuid):
...
+            )
+
+        return self.to_json(sigma_rule, status_code=HTTP_STATUS_CODE_CREATED)
+
+
+class SigmaRuleByTextResource(resources.ResourceMixin, Resource):
+    """Resource to get a Sigma rule by text."""
+
     @login_required
     def post(self):
```
tomchop, line 397: What does this function do? It returns a parsed Sigma Rule based on the body of the POST request? Please make the docstring explicit.

timesketch/api/v1/routes.py
```diff
@@ -139,9 +142,12 @@
...
     (VersionResource, "/version/"),
-    (SigmaListResource, "/sigma/"),
-    (SigmaResource, "/sigma/rule/<string:rule_uuid>/"),
-    (SigmaByTextResource, "/sigma/text/"),
+    (SigmaListResource, "/sigma/"), # TODO(jaegeral): deprecate this
+    (SigmaResource, "/sigma/rule/<string:rule_uuid>/"), # TODO: deprecate
+    (SigmaByTextResource, "/sigma/text/"), # TODO(jaegeral): deprecate this
+    (SigmaRuleListResource, "/sigmarule/"),
+    (SigmaRuleResource, "/sigmarule/<string:rule_uuid>/"),
+    (SigmaRuleByTextResource, "/sigmarule/text/"),
```
tomchop, line 150: No tests for this resource? :(

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +483,166 @@ def test_get_sigma_rule_list(self):
...
+        )
+        self.assertIn(
+            'shell',
+            response.json['objects'][0]["rule_yaml"],
+        )
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_CREATED)
+        # Now GET the ressources
+        response = self.client.get(
+            "/api/v1/sigmarule/5266a592-b793-11ea-b3de-bbbbbb/"
+        )
```
tomchop, line 554: 1 request should roughly equal one new test. If this tests fails, you'll have no idea why :)

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +483,166 @@ def test_get_sigma_rule_list(self):
...
+
+        self.assertIsNotNone(response)
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_OK)
+        self.assertIn('bbbbbb', response.json['objects'][0]["rule_uuid"])
+        # unclear why the db object is not updated here
+        # self.assertIn(
+        #    'cccccc',
+        #    response.json['objects'][0]["description"],
+        # )
```
tomchop, line 616: What is this?

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +483,166 @@ def test_get_sigma_rule_list(self):
...
+        """Authenticated request to get an sigma rule."""
+        self.login()
+        response = self.client.get(
+            "/api/v1/sigmarule/5266a592-b793-11ea-b3de-0242ac130004"
+        )
+        self.assertIsNotNone(response)
+
+        # Search a rule that does not exist
+        response = self.client.get("/api/v1/sigmarule/foobar/")
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_NOT_FOUND)
```
tomchop, line 628: Same - new test please.

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +483,166 @@ def test_get_sigma_rule_list(self):
...
+    keywords:
+        # Generic suspicious commands
+        - '*apt-get install cccccc*'
+    condition: keywords
+falsepositives:
+    - Unknown
+level: high
+"""
+
+        response = self.client.post(
```
tomchop, line 591: No need to store the response to this `post` call if you're not going to do anything on it.

timesketch/api/v1/resources_test.py
```diff
@@ -486,6 +483,166 @@ def test_get_sigma_rule_list(self):
...
+
+        # check if the previous POST updated the element
+        response = self.client.get(
+            "/api/v1/sigmarule/5266a592-b793-11ea-b3de-bbbbbb/"
+        )
+
+        self.assertIsNotNone(response)
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_OK)
+        self.assertIn('bbbbbb', response.json['objects'][0]["rule_uuid"])
+        # unclear why the db object is not updated here
```
tomchop, line 612: I'm assuming because there is no actual stateful database behind it and it's all mocks? You should probably be testing whether the backend function that updates the rule gets called with the appropriate parameters.

tomchop (changes requested):

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+        return_rule.append(_enrich_sigma_rule_object(rule=rule))
+
+        return jsonify({"objects": return_rule, "meta": {}})
+
+    @login_required
+    def delete(self, rule_uuid):
+        """Handles DELETE request to the resource.
+
+        Handels DELETE API calls to /sigmarule/<string:rule_uuid>/ where the
+        rule_uuid is the primaray way to identify the rule.
```
tomchop: Plz see previous examples of how this docstring should be formatted.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+            abort(
+                HTTP_STATUS_CODE_FORBIDDEN, "rule_yaml needs to be a string."
+            )
+        try:
+            parsed_rule = ts_sigma_lib.parse_sigma_rule_by_text(rule_yaml)
+        except ValueError as e:
+            error_msg = "Sigma Rule Parsing error: {0!s}".format(e)
+            logger.debug(error_msg)
+            return abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
```
tomchop, line 374: Isn't this the same situation than in L 371? Essentially, you just want to make sure that the rule is parseable. SO maybe the check in 369 is not necessary here?

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+                error_msg,
+            )
+        rule_uuid = parsed_rule.get("id")
+
+        title = parsed_rule.get("title")
+        description = parsed_rule.get("description")
```
tomchop, line 380: Move this two to right above L401 - in fact you might not even need them given that the only place you're using them is L405 and L404

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+
+        # Query rules to see if it already exist and exit if found
+        sigma_rule_from_db = SigmaRule.query.filter_by(
+            rule_uuid=rule_uuid
+        ).first()
+        if sigma_rule_from_db:
+            if sigma_rule_from_db.rule_uuid == rule_uuid:
```
tomchop, line 387: Is this check really needed? if you filter by rule_uuid, you will never get a rule that doesn't have this uuid.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+                error_msg = (
+                    "Rule {0!s} was already found in the database".format(
+                        rule_uuid
+                    )
+                )
+                logger.debug(error_msg)
+                abort(HTTP_STATUS_CODE_BAD_REQUEST, error_msg)
```
tomchop, line 394: According to https://en.wikipedia.org/wiki/List_of_HTTP_status_codes this should probably be 403

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+            rule_yaml=rule_yaml,
+            description=description,
+            title=title,
+            user=current_user,
+        )
+
+        if not sigma_rule:
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, "No sigma rule object created")
+
+        sigma_rule.set_status(parsed_rule.get("status", "experimental"))
```
tomchop, line 407: Can't this be set at create time in L401?

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+        rule_uuid is the primaray way to identify the rule.
+        The remaining attributes of the rule are provided in request itself.
+        If no `rule_yaml` is found in the reuqest, the method will fail as this
+        is required to parse the rule.
+
+        Args:
+            rule_uuid: uuid of the rule
+        Returns:
+            The updated sigma object in JSON (instance of
+            flask.wrappers.Response)
```
tomchop, line 442: This function returns a JSON object, please adjust docstring.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Sigma parsing error: no yaml provided",
+            )
+        try:
+            parsed_rule = ts_sigma_lib.parse_sigma_rule_by_text(rule_yaml)
+        except ValueError as e:
+            error_msg = "Error parsing Sigma rule {0!s}: {1!s}".format(
+                rule_uuid, e
+            )
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, error_msg)
```
tomchop, line 457: You're doing the same check in L 376, and returning a different error message: error_msg = "Sigma Rule Parsing error: {0!s}".format(e). Can we be consistent in the error messages that are returned in these situations across the changes in this PR?

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+        Returns:
+            The updated sigma object in JSON (instance of
+            flask.wrappers.Response)
+        """
+
+        rule_yaml = request.json.get("rule_yaml")
+        if not rule_yaml:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Sigma parsing error: no yaml provided",
```
tomchop, line 449: The error message in L366 is different, it's best to be consistent here given that the problem is exactly the same.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+        ).first()
+
+        if not sigma_rule_from_db:
+            error_msg = "Sigma rule with uuid: {0!s} not found".format(
+                rule_uuid
+            )
+            logger.error(error_msg)
+            abort(HTTP_STATUS_CODE_NOT_FOUND, error_msg)
+
+        sigma_rule_from_db.rule_uuid = rule_uuid
```
tomchop, line 476: This should never have any effect, and can probably be removed.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+        sigma_rule_from_db.rule_yaml = rule_yaml
+        sigma_rule_from_db.title = parsed_rule.get("title")
+        sigma_rule_from_db.description = parsed_rule.get("description")
+
+        try:
+            db_session.add(sigma_rule_from_db)
+            db_session.commit()
+        except IntegrityError as e:
+            error_msg = "Problem adding Sigma rule {0!s}".format(e)
```
tomchop, line 485: Favor "error" over "problem" (same in other places I've seen this)

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                error_msg,
+            )
+
+        except NotImplementedError as e:
+            error_msg = (
+                "Sigma Parsing error: Feature in the rule provided"
+                " is not implemented in this backend{0!s}".format(e)
+            )
```
tomchop, line 547:
```suggestion
            error_msg = (
                "Sigma Parsing error: Feature in the rule provided"
                " is not implemented in this backend: {0!s}".format(e)
            )
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+            " with error: {0!s}".format(e)
+            logger.error(error_msg, exc_info=True)
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                error_msg,
+            )
+
+        except yaml.parser.ParserError as e:
+            error_msg = (
+                "Sigma parsing error: invalid yaml provided {0!s}".format(e)
```
tomchop, line 568:
```suggestion
                "Sigma parsing error: invalid yaml provided: {0!s}".format(e)
```

timesketch/api/v1/resources_test.py
```diff
@@ -486,10 +483,215 @@ def test_get_sigma_rule_list(self):
...
+        """Authenticated request to get an sigma rule."""
+        self.login()
+        response = self.client.get(
+            "/api/v1/sigmarule/5266a592-b793-11ea-b3de-0242ac130004"
+        )
+        self.assertIsNotNone(response)
+
+        # Search a rule that does not exist
+        response = self.client.get("/api/v1/sigmarule/foobar/")
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_NOT_FOUND)
```
tomchop, line 578: Split in new test.

timesketch/api/v1/resources_test.py
```diff
@@ -486,10 +483,215 @@ def test_get_sigma_rule_list(self):
...
+            ),
+            content_type="application/json",
+        )
+        self.assertIsNotNone(response)
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_OK)
+        self.assertIn('bbbbbb', response.json['objects'][0]["rule_uuid"])
+        self.assertIn(
+            'cccccc',
+            response.json['objects'][0]["rule_yaml"],
+        )
```
tomchop, line 665: Why are you not using `assertEqual` here? The check is deterministic.

timesketch/api/v1/resources_test.py
```diff
@@ -555,10 +757,112 @@ def test_get_sigma_rule(self):
...
+        self.assertEqual(response.status_code, HTTP_STATUS_CODE_BAD_REQUEST)
+
+    def test_get_rule_by_text_no_form_data(self):
+        """Authenticated request to get an sigma rule by text with no form
+        data"""
+        self.login()
+        data = dict(action="post")
+        response = self.client.post(
+            "/api/v1/sigmarule/text/",
+            data=json.dumps(data, ensure_ascii=False),
```
tomchop, line 782:
```suggestion
            data={'action': 'post'},
```

Wouldn't this work?

timesketch/api/v1/resources_test.py
```diff
@@ -555,10 +757,112 @@ def test_get_sigma_rule(self):
...
+    }
+
+    def test_get_sigma_rule(self):
+        """DEPRECATED: Authenticated request to get an sigma rule by text."""
+        self.login()
+
+        data = dict(content=self.correct_rule)
+        response = self.client.post(
+            "/api/v1/sigma/text/",
+            data=json.dumps(data, ensure_ascii=False),
```
tomchop, line 851: Do we really need `ensure_ascii` here? (and in other places?) given that the test strings we're using don't have non-ASCII characters in them.

timesketch/lib/sigma_util_test.py
```diff
@@ -26,7 +26,93 @@
...
+
+        self.assertEqual(
+            test_3, r'(xml_string:"\foobar.exe" AND GrantedAccess:"10")'
+        )
+
+        test_4 = sigma_util._sanitize_query(
+            '(xml_string:C:\\Program Files\\WindowsApps\\\" AND xml_string: "GamingServices.exe)'  # pylint: disable=line-too-long
+        )
+        self.assertIsNotNone(test_4)
+        # pylint: enable=protected-access
```
tomchop, line 97: What's the difference between test_2, test_3, test_4? Any chance we can actually split these into different tests, or parametrize them?

timesketch/api/v1/resources/sigma.py
```diff
@@ -25,22 +25,68 @@
...
 
 
+def _enrich_sigma_rule_object(rule: SigmaRule):
+    """Helper function: Returns an enriched Sigma object given a SigmaRule.
+
+    It will extract the `status`, `created_at` and `updated_at` and make them
+    a field.
+
+    Args:
+        SigmaRule: uuid of the rule
```
tomchop, line 56: This seems to be an object according to the type hinting, not a string.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        Fetches fetch all Sigma rules stored in the database on the system
+        and returns a list of JSON representations of the rules.
+
+
+        Returns:
+            List of sigma rules represented in JSON e.g.
+                {"objects": [sigma_rules], "meta": {"rules_count": 42}
+        """
+        sigma_rules = []
```
tomchop, line 246:
```suggestion
        """Fetches Sigma rules from the database.

        Fetches fetch all Sigma rules stored in the database on the system
        and returns a list of JSON representations of the rules.

        Returns:
            List of sigma rules represented in JSON e.g.
                {"objects": [sigma_rules], "meta": {"rules_count": 42}
        """
        sigma_rules = []
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+
+class SigmaRuleResource(resources.ResourceMixin, Resource):
+    """Resource to read / delete / create / update a Sigma rule."""
+
+    @login_required
+    def get(self, rule_uuid):
+        """Fetches a single Sigma rule from the databse.
+
+        Fetches a single Sigma rule selected by the `uuid` in
```
tomchop, line 263:
```suggestion
        Fetches a single Sigma rule selected by the `UUID` in
```

Please capitalize UUID everywhere in docstrings and comments.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        return jsonify({"objects": return_rule, "meta": {}})
+
+    @login_required
+    def delete(self, rule_uuid):
+        """Deletes a Sigma rule from the database,.
+
+        Deletes a single Sigma rule selected by the `uuid` in
+        `/sigmarule/<string:rule_uuid>/` and returns a HTTP Code 200 if
+        the rule is deleted.
```
tomchop, line 304:
```suggestion
        `/sigmarule/<string:rule_uuid>/`.
```

RESTful APIs make it implicit that 200 is returned upon success of the request.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        Args:
+            rule_uuid: uuid of the rule to be deleted
+
+        Returns:
+            HTTP_STATUS_CODE_NOT_FOUND if rule not found
+            HTTP_STATUS_CODE_OK if rule deleted
```
tomchop, line 311:
```suggestion
            HTTP_STATUS_CODE_NOT_FOUND if rule not found.
            HTTP_STATUS_CODE_OK if rule deleted.
```

Please go over the docstrings and add periods at the end of sentences.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+        request will fail.
+
+        Remark: To update a rule, use `PUT`instead.
+
+        Args:
+            rule_uuid: optional field that is not needed
+
+        Returns:
+            Sigma rule object and HTTP status 401 code indicating
+            whether operation was sucessful.
```
tomchop, line 352:
```suggestion
            Sigma rule object and HTTP status 200 code indicating
            whether operation was sucessful.
```

You mean 200 here, right? 401 is UNAUTHORIZED

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+            HTTP Error code 400 if no `rule_yaml` is provided
+            HTTP Error code 400 if rule_uuid does not match id in the yaml.
```
tomchop, line 354: You're also returning HTTP_STATUS_CODE_CONFLICT, so I'd encourage you to add a description here if you're also mentioning the other scenarios. Maybe also just something like:

```suggestion
            HTTP Error code 400 if no `rule_yaml` is provided or a parsing error occurs.
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+            abort(
+                HTTP_STATUS_CODE_CONFLICT,
+                error_msg,
+            )
+
+        return self.to_json(sigma_rule, status_code=HTTP_STATUS_CODE_CREATED)
+
+    @login_required
+    def put(self, rule_uuid):
+        """Update existing Sigma rule and its values in the database.
```
tomchop, line 424:
```suggestion
        """Update an existing Sigma rule in the database.
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        Calls to `/sigmarule/<string:rule_uuid>/` where the
+        rule_uuid is way to identify the rule.
```
tomchop, line 427:
```suggestion
        Handles calls to `/sigmarule/<string:rule_uuid>/`, where
        `rule_uuid` is used to identify the rule.
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        If the parameter `rule_uuid`and the id provided in `rule_yaml` does
+        not match, the method will abort.
+
+        Args:
+            rule_uuid: uuid of the rule
+
+        Returns:
+            The updated sigma object in JSON
+        """
```
tomchop, line 441:
```suggestion
        The remaining attributes of the rule are provided in request itself
        through the `rule_yaml` key.
        If no `rule_yaml` is found in the request, the request will fail as this
        is required to parse the rule.

        If `rule_uuid` doesn't match the UUI in `rule_yaml`, the request will fail.

        Args:
            rule_uuid: uuid of the rule

        Returns:
            The updated Sigma object in JSON
        """
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        rule_yaml = request.json.get("rule_yaml")
+        if not rule_yaml:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Error parsing Sigma rule {0!s}: no yaml provided".format(
```
tomchop, line 447:
```suggestion
                "Error parsing Sigma rule {0!s}: no YAML provided".format(
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+        except ValueError as e:
+            error_msg = "Error parsing Sigma rule {0!s}: {1!s}".format(
+                rule_uuid, e
+            )
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, error_msg)
+
+        if rule_uuid != parsed_rule.get("id"):
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Rule ID mismatch parameter:{0!s} and yaml content:{1!s}".format(
```
tomchop, line 462:
```suggestion
                "Rule ID mismatch parameter: {0:s} and yaml content:\
                n{1:s}".format(
```

These are already strings, so no need to coerce them into strings (using `!` instead of `:`)

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        Will parse a provided text `rule_yaml`, parse it and return as JSON
+
+
+        If no form content is given, the method will abort with HTTP Error 400.
+
+        Returns:
+            JSON sigma rule e.g.
+                {"objects": [sigma_rule], "meta": {"parsed": True}}
+        """
```
tomchop, line 513:
```suggestion
        """Obtain a parsed Sigma rule by providing text.

        Will parse a provided text `rule_yaml`, parse it and return as JSON.
        If no form content is given, the method will abort with HTTP error 400.

        Returns:
            JSON sigma rule e.g.
                {"objects": [sigma_rule], "meta": {"parsed": True}}
        """
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+
+        form = request.json
+        if not form:
+            form = request.data
```
tomchop, line 517: I thought we said we'd assume all data is contained in JSON now?

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+        if not content:
+            return abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Missing value in the request. Nothing to parse",
+            )
+
+        try:
+            sigma_rule = ts_sigma_lib.parse_sigma_rule_by_text(content)
+
+        except ValueError as e:
```
tomchop, line 529:
```suggestion
        try:
            sigma_rule = ts_sigma_lib.parse_sigma_rule_by_text(content)
        except ValueError as e:
```

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+            error_msg = (
+                "Sigma rule Parsing error with provided rule {0!s}".format(e)
```
tomchop, line 531:
```suggestion
                "Sigma rule Parsing error with provided rule {0:s}".format(e)
```

Don't coerce to string if you don't need to - this implies that the variable may be something else and would be confusing for someone reading the code. Please have a look at other places in the code where you do this, I've spotted it in many places.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+                exc_info=True,
+            )
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                error_msg,
+            )
+
+        except sigma_exceptions.SigmaParseError as e:
+            error_msg = "Sigma parsing error generating rule"
+            " with error: {0!s}".format(e)
```
tomchop, line 558:
```suggestion
            error_msg = "Sigma parsing error: {0!s}".format(e)
```

`e` needs to be coerced as it's an Exception and not a string.

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +217,371 @@ def post(self):
...
+                error_msg,
+            )
+
+        if not sigma_rule:
+            error_msg = "Sigma parsing error: No sigma rule was parsed"
+            logger.error(
+                error_msg,
+                exc_info=True,
+            )
+            abort(HTTP_STATUS_CODE_NOT_FOUND, error_msg)
```
tomchop, line 584: In which scenarios can this ever happen? It feels like with all the checks you're doing before, there should be a rule here. 

In any case, I wouldn't raise a 404 because it's not about a resource not being found, but about a request not being properly formatted. So something like 400 would be a bit better.

tomchop (changes requested):

jaegeral:
some comments

jaegeral (commented):
some comments

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+            abort(
+                HTTP_STATUS_CODE_NOT_FOUND,
+                error_msg,
+            )
+
+        if not rule:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No rule found with this ID.")
+        return_rule = []
+
+        assert isinstance(rule, SigmaRule)
```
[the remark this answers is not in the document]
jaegeral: removed

timesketch/api/v1/resources/sigma.py
```diff
@@ -153,11 +213,374 @@ def post(self):
...
+
+        Args:
+            rule_uuid: uuid of the rule
+
+        Returns:
+            JSON sigma rule representation
+        """
+        try:
+            rule = SigmaRule.query.filter_by(rule_uuid=rule_uuid).first()
+
```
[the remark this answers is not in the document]
tomchop: Done.

timesketch/api/v1/resources_test.py
tomchop: But here you have control over the text - so you know these characters are not gonna happen.

tomchop (changes requested):
L G T M

tomchop (approved):
L G T M

berggren (approved):
LGTM

PR #2304, opened 2022-08-25
Add manual event to timelines

marcobrotto:
# Add manual events to timelines

This PR is relative only to _frontend-ng_.

With the following PR, the user can insert a new manual event.

The manual event can be independent or associated with a timeline's event. The only difference between the 2 cases is that in the second one, the default datetime is the one of the timeline's event.

The next image shows an example of the first case: the user needs to specify the datetime.

Screenshot 2022-08-25 1 12 31 PM

The next image shows an example of the second case: the default datetime is the one associated to the 

Screenshot 2022-08-25 2 13 25 PM

With the following feature the user can also enrich the event with attributes (button *Add attribute*).

Screenshot 2022-08-25 2 15 31 PM

## Main modifications

New file created: `src/components/Explore/AddmanualEvent.vue`

New RestAPI added in `src/components/utils/RestAPIClient.js`: `createEvent`

## Open problems: 
1. The web page needs to be reloaded when the user inserts a manual event.
2. The menu is not displayed at the top of the page when the manual event is inserted from one timeline's event at the bottom of the list. (style problem?)
3. [General problem] When I insert a manual event using the python API sketch.add_event(...) the server creates a new timeline (said T1). However, if I delete T1, and then I insert another manual event (i.e., again sketch.add_event(...)), the server crashes and I cannot use anymore the sketch where T1 was previously created.

## Checks

- [x] All tests succeed.

## Closing Issue
 `closes 1940`


timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
```diff
@@ -0,0 +1,182 @@
...
+      set(val) {
+        if (val !== null) this.datetime = dayjs.utc(val).millisecond(0).toISOString()
+      },
+      get() {
+        return this.datetime
+      },
+    },
+  },
+  methods: {
+    setAddAttributeBtn: function (i) {
```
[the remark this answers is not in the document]
tomchop: As a second though, if the sole purpose of this function is to determine whether the button should be enabled (through the isDisabled attribute), I think a better way to reflect this would be to make isDisabled a computed property that does the check you do on L139 and returns true or false.

The only caveat that I can think of is whether the computed property will register changes on the attributes in objects contained in your list, but according to [this](https://vuejs.org/guide/essentials/reactivity-fundamentals.html#deep-reactivity) it should.

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
```diff
@@ -0,0 +1,182 @@
...
+      if (this.attributes[i].name === '' || this.attributes[i].value === '') {
+        this.isDisabled = true
+      } else {
+        this.isDisabled = false
+      }
+    },
+    addAttribute: function () {
+      this.attributes.push({ name: '', value: '' })
+      this.isDisabled = true
```
tomchop, line 147: You wouldn't need to set isDisabled if you go with the computed property. (same for L154)

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
```diff
@@ -0,0 +1,182 @@
...
+      let config = {
+        headers: {
+          'Content-Type': 'application/json',
+        },
+      }
+      let attributes = {}
+      this.attributes.forEach((attribute) => {
+        if (attribute.name === '' || attribute.value === '') return
+        attributes[attribute.name] = attribute.value
+      })
```
tomchop, line 171: It's probably fine to leave this this way, but I'd change this to use `filter` and `reduce` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce).

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -627,9 +656,11 @@ export default {
       leftDrawer: true,
       expandedRows: [],
       timeFilterMenu: false,
+      addManualEvent: false,
       saveSearchMenu: false,
       saveSearchFormName: '',
       fromSavedSearch: false,
+      datetime: '',
```
tomchop, line 663: Could we be a bit more specific in the naming of this variable? i.e. someone stumbling upon this component would have a hard time figuring out that this is to be used exclusively for adding new events.

tomchop (dismissed):

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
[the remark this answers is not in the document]
marcobrotto: Make sense!
What about something like this:
```
isDisabled() {
      return this.attributes.some((attribute) => {
        return attribute.name === '' || attribute.value === ''
      })
    },
```
In this way, also the previous input text will contribute to the final value of isDisabled

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
marcobrotto:
```
<v-btn text color="primary" @click="attributes.push({ name: '', value: '' })" :disabled="isDisabled"> Add Attribute </v-btn>

```

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
marcobrotto:
```
<v-btn text color="primary" @click="attributes.push({ name: '', value: '' })" :disabled="isDisabled"> 
    Add Attribute 
</v-btn>
```

timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue
marcobrotto: I am not sure I understood: do you mean something like this?
```
let attributes = {}
this.attributes.reduce((_, attribute) => {
  if (!(attribute.name === '' || attribute.value === '')) attributes[attribute.name] = attribute.value
}, attributes)
```

jaegeral (approved):

PR #2313, opened 2022-08-31
Upload json/jsonl/plaso/csv button in frontend-ng

marcobrotto:
# UPLOAD BUTTON IN THE FRONTEND-NG

In this PR we implemented the button to upload a new timeline in the new frontend of Timesketch. The button works but the post-upload functionalities are still missing.

## Client side updates
In `Overview.vue` there is a new button `Upload Timeline`. This shows a dialog to upload a timeline in the format `Plaso`, `CSV` and `JSONL` (for `JSON` see side notes).

The dialog is designed in the file `UploadForm.vue` (NEW file). Its main structure is the following:

```
<div v-if=errors>
 SHOW ERRORS, e.g., HEADERS MISSING
</div>
<div v-if=(extension is csv, json or jsonl)>
 SHOW THE TABLE TO PERFORM THE HEADER MAPPING OR SIMPLY VIEW THE PREVIEW
</div>
```

The following figure shows a situation where the uploaded file can be directly uploaded

Screenshot 2022-08-31 3 38 51 PM

The following figure shows a situation where the uploaded file needs to be mapped

Screenshot 2022-08-31 3 39 51 PM

The following figure shows a situation where the uploaded file has been mapped

image

When there is no need to map the headers, because the required fields are already in the CSV, then the preview is shown but the user cannot perform the mapping (see following figure).

image

There are 3 button at the bottom of the form:
1. Close: it closes the dialog.
2. Clear Form: it clears all the fields of the form, the user can upload a new file.
3. Submit: it is enabled only when the required headers are present or the user did the mapping.

## Server side updates
See PR #2288.

## Missing functionalities
Handle the file "post upload". As in the preview frontend, it would be nice to notify the user when the file has been successfully uploaded/failed.

## Side notes
This PR embeds the logic of the PR #2288 - JSONL Mapping Header. With respect #2288, this PR adds the possibility to upload a JSON file formatted as a JSONL.


timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -704,6 +730,30 @@ export default {
...
       };
     },
+    extractJSONLHeader: function () {
+      let reader = new FileReader();
+      let file = document.getElementById("datafile").files[0];
+      let vueJS = this;
+      reader.readAsText(file.slice(0, 10000));
+      reader.onloadend = function (e) {
+        if (e.target.readyState === FileReader.DONE) {
+          /* 3a. Extract the headers from the CSV */
```
berggren, line 740:
```suggestion
          /* 3a. Extract the headers from the JSONL */
```

timesketch/lib/tasks.py
```diff
@@ -717,6 +717,8 @@ def run_csv_jsonl(
     Returns:
         Name (str) of the index.
     """
+    logger.info("run_csv_jsonl-tasks.py")
+    logger.info(headers_mapping)
```
berggren, line 721: Do we need to log this?

timesketch/lib/utils.py
```diff
@@ -409,6 +472,13 @@ def read_and_validate_jsonl(file_handle, **kwargs):  # pylint: disable=unused-ar
                     linedict["timestamp"] = int(
                         parser.parse(linedict["datetime"]).timestamp() * 1000000
                     )
+                except TypeError:
```
berggren, line 475:
```suggestion
                // TODO: REcord this somewhere else and make available to the user.
                except TypeError:
```

timesketch/frontend-ng/src/components/UploadForm.vue
```diff
@@ -0,0 +1,552 @@
...
+        headers: {
+          'Content-Type': 'multipart/form-data',
+        },
+        onUploadProgress: function (progressEvent) {
+          this.percentCompleted = Math.round((progressEvent.loaded * 100) / progressEvent.total)
+        }.bind(this),
+      }
+      ApiClient.uploadTimeline(formData, config)
+        .then((response) => {
+          this.$store.dispatch('updateSketch', this.$store.state.sketch.id)
```
berggren, line 442: You can set sketch in computed to have access to the sketch object in this component

```
  computed: {
    sketch() {
      return this.$store.state.sketch
    }
}
```

timesketch/frontend-ng/src/components/UploadForm.vue
```diff
@@ -0,0 +1,552 @@
...
+          this.percentCompleted = 0
+        })
+        .catch((e) => {})
+    },
+    validateFile: function () {
+      this.error = []
+      if (this.form.file.size <= 0) {
+        this.error.push('Please select a non empty file')
+      }
+      let allowedExtensions = ['csv', 'json', 'jsonl', 'plaso']
```
berggren, line 453: Isn't json a bit misleading here? We treat any file with .json as JSONL, correct?
Maybe consider dropping json?

berggren (dismissed):
Overall, looks good. A few comments, PTAL

timesketch/frontend-ng/src/components/UploadForm.vue
marcobrotto: That's a good point, @jaegeral and I discussed about it last week.

The answer for your first 2 questions is: YES. To this end, in the message for the client (when we ask for uploading a new file) we only specify JSONL/CSV/Plaso files. As you said, we treat (on purporse) any .JSON file as .JSONL because we assumed that there could be some users that upload a JSONL file with the JSON extension (e.g., OpenSOC challenge, all file are JSONL but with JSON extension).

If the file is actually a JSON then the parsing fails either client and server side. We could actually add a message saying the conversion failed, something like: "JSONL parsing failed: please add a JSONL/CSV/Plaso file instead."

What do you think?

timesketch/frontend/src/components/Common/UploadForm.vue
```diff
@@ -423,29 +441,36 @@ export default {
...
+              valuesAndHeaders[header].push(this.valuesString[i][header]);
+            } else {
+              valuesAndHeaders[header] = [this.valuesString[i][header]];
+            }
+          }
         }
       } else {
-        console.log("JSONL not supported (yet) for this feature");
+        console.error(
+          this.extension + " file extension not supported for this feature"
```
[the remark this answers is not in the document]
marcobrotto: Actually we cannot arrive in this part of the code. I.e., if the file is not Plaso/JSONL/CSV then this method/computed prop is not computed at all. To this end, probably it is better to remove this "else" branch.

PR #2314, opened 2022-08-31
Initial triage scenario

jaegeral:
An initial commit for the initial triage scenario.

It can be used in an empty state investigation.

It is very minimal and only asks the analyst for the OS that is expected to show up in the data (e.g. Windows, Linux, MacOS) and what time window the analyst is interested in.

On top of it, the tool asks for external references of the investigation, e.g. a link to a ticketing system.

Besides the data model it also adds a field for uuid for every aspect, which we might use moving forward


data/scenarios/facets.yaml
```diff
@@ -33,3 +33,24 @@ interference_with_protective_measures:
...
+		- operating_system
+
+time_window_of_interest:
+	uuid: 26ef99ad-fcd8-47c6-9aa7-566dfb2a7e87
+  display_name: Time window of interest facet
+	description: What is the time window that is relevant for this investigation.
+	questions:
+		- time_window_of_interest
+
+external_investigation_reference:
```
berggren, line 51: This will be referenced as an attribute on the sketch and not as a facet.

data/scenarios/facets.yaml
```diff
@@ -33,3 +33,24 @@ interference_with_protective_measures:
...
     - user_logged_in_win
+
+operating_system:
+	uuid: f131c057-0c06-421d-b606-f3f410927d94
+	display_name: What are the involved operating systems?
+	description: What OperatingSystem(s) is in scope or relevant for this case.
+	questions:
+		- operating_system
+
+time_window_of_interest:
```
berggren, line 44: Instead of a facet, the timeframe will be saved as a FacetTimeFrame DB object (it is part of the model) if on the facet level. Otherwise we will use the GenericAttribute mixin table to record this.

Take a look at the model for reference: https://github.com/google/timesketch/blob/88750ad09c295535f6664654ca630568bdb18c5f/timesketch/models/sketch.py#L952

data/scenarios/questions.yaml
```diff
@@ -106,3 +106,18 @@ user_downloaded_file:
...
+
+time_window_of_interest:
+	uuid: 7026f193-f34e-46bb-9308-76aa46ad2495
+  display_name: Time window of interest
+  description: What is the potential time span for this investigation?
+
+external_investigation_reference:
+	uuid: 7004aa29-07a6-4ce0-ba18-aaef424defb5
+  display_name: External reference of the investigation
+	description: Provide a link to a ticket or trigger that started the investigation.
```
berggren, line 123:
```suggestion
```

berggren (changes requested):

data/scenarios/facets.yaml
jaegeral: ack will remove

data/scenarios/facets.yaml
jaegeral: ack will remove

berggren:
Closing this for now. The format has been reworked and we need to redo this. We can reopen this if we need to take a look later.

PR #2316, opened 2022-09-01
Proposed fix for Export Sketch feature which was not working.

### LINKED ISSUE #2315: Export Sketch feature returns HTTP 500 error
**Describe the bug**
When attempting to export a Sketch the backend responds with a HTTP 500 Error

**To Reproduce**
1. Open any sketch
2. Click on 'More' in the top menu then click on 'Export'
3. Open the developers tools in Chrome browser and the error will be there. 

**Expected behavior**
A zip file with all the timelines in the Sketch downloaded to your machine.

 - OS: Mac 
 - Browser Chrome
 - Version 104

hkhalifa:

This PR fixes a bug in the Sketch Export feature. 

**Checks**

- [x] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**
closes #2315

### DIFF
--- a/timesketch/api/v1/resources/archive.py
+++ b/timesketch/api/v1/resources/archive.py
@@ -343,13 +343,13 @@ class SketchArchiveResource(resources.ResourceMixin, Resource):
         """
         name = "{0:04d}_{1:s}".format(view.id, view.name)
 
-        query_filter = json.loads(view.query_filter)
+        if view.query_filter:
+            query_filter = json.loads(view.query_filter)
+
         if not query_filter:
             query_filter = self.DEFAULT_QUERY_FILTER
 
-        indices = query_filter.get("indices", self.sketch_indices)
-        if not indices or "_all" in indices:
-            indices = self.sketch_indices
+        indices = self.sketch_indices
 
         # Ignoring the size limits in views to reduce the amount of queries
         # needed to get all the data.
--- a/timesketch/api/v1/utils.py
+++ b/timesketch/api/v1/utils.py
@@ -109,7 +109,7 @@ def update_sketch_last_activity(sketch):
     db_session.commit()
 
 
-def run_aggregator(sketch_id, aggregator_name, aggregator_parameters=None, index=None):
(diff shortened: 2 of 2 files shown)


hkhalifa:
Note that
```
query_filter = json.loads(view.query_filter)
indices = query_filter.get("indices", self.sketch_indices)
```
returned the wrong index. It was just a 1 digit int. So i replaced it with 
```
indices = self.sketch_indices
```
Not sure if this is the best way to do it, and also could find the origin of the wrong index number in the query_filter.

jaegeral (approved):
Thx for the fix

timesketch/api/v1/resources/archive.py
```diff
@@ -343,13 +343,13 @@ def _export_view(self, view, sketch, zip_file):
...
+        if view.query_filter:
+            query_filter = json.loads(view.query_filter)
+
         if not query_filter:
             query_filter = self.DEFAULT_QUERY_FILTER
 
-        indices = query_filter.get("indices", self.sketch_indices)
-        if not indices or "_all" in indices:
-            indices = self.sketch_indices
+        indices = self.sketch_indices
```
berggren, line 352: I know that this is already merged, but this block means that the saved search will be wrong in the case that the filter has specific indices selected, and will potentially return unexpected result, which is probably not what we want?

berggren: ping @jaegeral who reviewed this PR

hkhalifa: Ya, i mentioned this in the comment above. 

"Not sure if this is the best way to do it, and also could find the origin of the wrong index number in the query_filter. hence this is just a Proposal, and will be waiting for input regarding this."

hkhalifa: There is something wrong with 

```
query_filter = json.loads(view.query_filter)
indices = query_filter.get("indices", self.sketch_indices)
```

Not sure where is the index in the query filter is set. it returned the wrong index. It was just a 1 digit int

PR #2320, opened 2022-09-06
Search history tree generation

berggren:
The History tree used a naive approach to find a reasonable root node to generate the tree from. In some edge cases, the last node was not present in the tree and the UI failed to render the history.

This PR changes so we always start from the last node, and then traverse backwards in the tree 10 nodes. After that generate the full sub tree with all children etc.


jkppr (approved):
LGTM

PR #2325, opened 2022-09-13
XML Viewer for xml_string in (mostly) WINEVTX timelines

marcobrotto:
# XML Viewer for xml_string attribute

Plaso can parse WINEVTX logs. It creates an attribute `xml_string` that is the "dump" of the log itself. The next figure shows how it is represented in Timesketch.

image

This valuable information might be hard to interpret since it is written in a compat format. With this PR we aim to create an *icon* to visualize better this attribute.

## Frontend files modified:
1. `EventDetail.vue`: added the icon (`</>`) for the attribute `xml_string`

Screenshot 2022-09-13 1 42 24 PM

2. [NEW!] `FormatXMLString.vue`: new component to display the XML string. It receives as input (~ property) the xml_string of the event's attribute `xml_string`. It outputs a formatted version of `xml_string`

image

### Areas of improvements
We add the xml button to view the XML only when the attribute in the timeline event is equal to `xml_string`. This assumption works fine for the majority of the winevtx file parsed with Plaso. We could add the same button for whatever attribute that contains some XML.

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
@@ -28,7 +28,28 @@ limitations under the License.
               </thead>
               <tbody>
                 <tr v-for="(value, key) in fullEventFiltered" :key="key">
-                  <td>{{ key }}</td>
+                  <td>
+                    {{ key }}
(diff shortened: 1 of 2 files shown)


marcobrotto:
See PR #2327

PR #2333, opened 2022-09-19
Updated Sigma support in the API client

jaegeral:
- Update the JS API client to add recent additions of the File based API
- replace the term `es_query` with `query_string` to be consistent with the rest of the codebase
- make API client methods deprecated that will soon be gone away as they link to the file based API
- add methods in the API client that talk to the databased Sigma Rule API

This is a next step to: https://github.com/google/timesketch/issues/2301


api_client/python/timesketch_api_client/client.py
```diff
@@ -573,7 +573,7 @@ def refresh_oauth_token(self):
         self.credentials.credential.refresh(request)
 
     def list_sigma_rules(self, as_pandas=False):
-        """Get a list of sigma objects.
+        """DEPRECATED: Get a list of sigma objects.
```
berggren, line 576: Can you add a comment that this is being replaced by the new functions.

api_client/python/timesketch_api_client/client.py
```diff
@@ -605,8 +605,87 @@ def list_sigma_rules(self, as_pandas=False):
             rules.append(index_obj)
         return rules
 
+    def list_sigmarules_rules(self, as_pandas=False):
```
berggren, line 608: Can we call this list_sigmarules instead. So many `rules` in the name otherwise.. :)

api_client/python/timesketch_api_client/client.py
```diff
@@ -605,8 +605,87 @@ def list_sigma_rules(self, as_pandas=False):
...
+            if not rule_dict:
+                raise ValueError("No rules found.")
+
+            index_obj = sigma.Sigma(api=self)
+            for key, value in rule_dict.items():
+                index_obj.set_value(key, value)
+            rules.append(index_obj)
+        return rules
+
+    def get_sigmarule_rule(self, rule_uuid):
```
berggren, line 644: same here, please rename. 

Also, can we add a logger here telling the user that they are using a deprecated method?

api_client/python/timesketch_api_client/sigma.py
```diff
@@ -22,8 +22,172 @@
...
+            raise ValueError(error_msg)
+
+        rule_dict = objects[0]
+        for key, value in rule_dict.items():
+            self.set_value(key, value)
+
+
 class Sigma(resource.BaseResource):
-    """Timesketch sigma object.
+    """DEPRECATED Timesketch sigma object.
```
berggren, line 190: Same here, add logger telling the user that the method is deprecated.

berggren (changes requested):
PTAL

api_client/python/timesketch_api_client/client.py
jkppr: Would it make sense to use a name that includes where the data is coming from? e.g. `list_sigma_rules_from_db`?

api_client/python/timesketch_api_client/client.py
jaegeral: done

api_client/python/timesketch_api_client/client.py
jaegeral: done
@jkppr we could but as we do not have any other option at the moment it would just add clutter

api_client/python/timesketch_api_client/sigma.py
jaegeral: done

api_client/python/timesketch_api_client/client.py
jaegeral: done

jkppr (approved):
lgtm

berggren (approved):
LGTM

PR #2343, opened 2022-09-27
DataSource error handling for Plaso files and upload bugfix with extension handling

marcobrotto:
# Solving some bugs related to PR 2326

1. Extension of file with "multiple dots": the extension in `uploadForm.vue` is computed as the last substring of the array `extensionArray = filename.split('.')`.
2. When `pinfo.py` cannot run on the server (see `tasks.py`) the server set the datasource status as `fail` and an error message is returned to the client.

image

4. Minor changes: a timeline with a fail status does not trigger the command `this.$emit('toggle', response.data.objects[0])` (see `Timelinechip.vue`)


timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -516,6 +516,9 @@ limitations under the License.
           </v-btn>
           <!-- Tag menu -->
           <ts-event-tag-menu :event="item"></ts-event-tag-menu>
+          <v-btn small icon @click="addEventBtn(item._source.datetime)">
```
marcobrotto, line 519: This is the button to insert a manual event related to an existing event. With the previous PRs it was accidentally removed.

timesketch/lib/tasks.py
```diff
@@ -657,19 +657,35 @@ def run_plaso(file_path, events, timeline_name, index_name, source_type, timelin
...
-    except subprocess.CalledProcessError:
-        pass
+            total_file_events = total_events_json["total"]
+    except subprocess.CalledProcessError as e:
+        error_msg = e.stderr.decode("utf-8")
+        _set_datasource_total_events(timeline_id, file_path, total_file_events=0)
+        _set_datasource_status(timeline_id, file_path, "fail", error_message=error_msg)
+        _close_index(
+            index_name=index_name, data_store=opensearch, timeline_id=timeline_id
+        )
```
berggren, line 676: This will close the whole index, making all other datasources that uses this index non-functional. Is that what we want here?

timesketch/lib/tasks.py
```diff
@@ -657,19 +657,35 @@ def run_plaso(file_path, events, timeline_name, index_name, source_type, timelin
...
+        raise
+    except Exception as e:  # pylint: disable=broad-except
+        # Mark the searchindex and timelines as failed and exit the task
+        error_msg = traceback.format_exc()
+        _set_datasource_status(timeline_id, file_path, "fail", error_message=error_msg)
+        logger.error("Error: {0!s}\n{1:s}".format(e, error_msg))
+        _close_index(
+            index_name=index_name, data_store=opensearch, timeline_id=timeline_id
+        )
```
berggren, line 685: Same here - I guess we only want to mark the timeline as fail, but keep the index open?

berggren (changes requested):

berggren (changes requested):

berggren (approved):
LGTM

Zawadidone:
Since this merge a few JSONL timelines imported by the Timesketch importer keep haning on "Indexing in progress.." while the worker tasks are already finished. Is it possible this PR caused a bug?

image

PR #2347, opened 2022-09-28
Modify Sigma analyzer / API Client / UX to use Database

jaegeral:
Based on https://github.com/google/timesketch/issues/2301 following changes have ben made:

# the Analyzer has been modified.

The new Analyzer takes the rule status. If the rule status is not `stable` it is not picked up by the analyzer.

Screenshot 2022-09-28 at 13 52 34

# Modifications to the Sigma util class

Some modifications to the SigmaUtil class where needed

# API client

- Several additions to the API client in the Sigma context.
- deprecation of several methods
- increase of version number for api_client

# UX update

Now using Database instead of file based rules.
The compose view comes with pre-defined templates based on the Sigma wiki.
When modifying the Sigma rule in the Text Editor, the rule is parsed on a regular base to give the analyst a near real time feedback on what `search_query` the rule would produce and how many hits it would cause in the selected sketch.

Screenshot 2022-09-29 at 15 55 50

It is also possible to modify and delete rules. When clicking the lens icon near to the rule, the `search` query is executed.

# Notebook

To make the transition from file-based rules to Database stored rules easier, a Notebook has been added in `/notebooks` that gives a simple example on how to read a single SigmaRule from disk, parse it to see if it works and then adds it to the database.

SigmaRuleUX_2347

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.


timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -32,16 +33,16 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
         self._rule = kwargs.get("rule")
         super().__init__(index_name, sketch_id, timeline_id=timeline_id)
 
-    def run_sigma_rule(
-        self, query, rule_name, tag_list=None, status_good=False
```
jkppr, line 36: Are there situations where you want to run a sigma rule against the timeline but not tag anything? `tag_list=None` seems to imply that it is an option but the description says the analyser will tag events. Also, there would be only the `ts_ttp` attributes added if the tag variable stays None. What should happen with stable sigma rules that have no `tag` field but hits in the events?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -94,31 +100,31 @@ def run(self):
...
             sigma_rule_counter += 1
             tagged_events_counter = self.run_sigma_rule(
                 rule.get("search_query"),
-                rule.get("file_name"),
+                rule.get("title"),
                 tag_list=rule.get("tags"),
-                status_good=rule.get('ts_use_in_analyzer'),
             )
-            tags_applied[rule.get("file_name")] += tagged_events_counter
+            tags_applied[rule.get("title")] += tagged_events_counter
```
jkppr, line 111: If I understand the usage of the `def get_kwargs():` function correct, each rule will spawn it's own `run()` instance. If this is true, the `tags_applied` dict would only contain one entry and the value equals `tagged_events_counter`. This is also nowhere used later on, since the output uses `tagged_events_counter` directly. So this dict can probably be removed. wdyt?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -94,31 +100,31 @@ def run(self):
...
+            )
             logger.error(
-                "Problem with rule in file {0:s}: ".format(
-                    rule.get("file_name")
-                ),
+                error_msg,
                 exc_info=True,
             )
-            problem_strings.append("* {0:s}".format(rule.get("file_name")))
+            problem_strings.append(error_msg)
```
jkppr, line 120: Since each rule has their own `run()` instance there is only one problem in this list if it occurs. So why not return the `error_msg` directly?

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -94,31 +100,31 @@ def run(self):
...
         output_strings.append(
             f"{tagged_events_counter} events tagged for rule [{rule_name}]"
         )
 
         if len(problem_strings) > 0:
-            output_strings.append("Problematic rules:")
+            output_strings.append("Problematic rule:")
             output_strings.extend(problem_strings)
 
         return "\n".join(output_strings)
```
jkppr, line 130: As discussed, the function `add_sigma_match_view` is not used. Please remove it from the analyzer. We can add it later again when the whole Sigma integration is done.

jkppr (changes requested):
Looks good. Some comments on the `sigma_tagger.py`.

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: done

timesketch/lib/analyzers/sigma_tagger.py
jaegeral: so the tag list can be empty, in that case indeed only the file title would be applied to the event

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -91,79 +98,31 @@ def run(self):
             logger.error("No  Sigma rule given.")
             return "Unable to run, no rule given to the analyzer"
         rule_name = rule.get("title", "N/A")
-        problem_strings = []
         output_strings = []
```
jkppr, line 101: Similar to the `problem_strings` you only add one line to this list. You could therefore also just return the string directly without appending it to the list first.

jkppr (dismissed):
One more comment.

timesketch/lib/analyzers/sigma_tagger.py
jkppr: Talking about the `output_strings`. here.

jaegeral: done

jaegeral:
- Display a preview of events right in the edit view
- auto generate / parse the rule in the compose view
- remove the `data_type`thing in the compose view
- fix the creation of rules
- show the error message right
- disable `create / save` button when rule is not parseable
- show warnings for to broad queries and errors when parsing right in the compose view
- more space between the TEmplate selection and the edit sigma rule field
- info mouseover or something that indicates if a status is not stable, it will be ignored by the analyzer

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,105 +15,182 @@ limitations under the License.
...
-      text: `Place your Sigma rule here and press parse`,
+      editingRule: {},
+      showEditModal: false,
+      save_button_text: "Update", // false: update rule. true: create new in DB
+      sketchTags: [],
+      sketchTTP: [],
+      analyses: [],
+      SigmaTemplates: SigmaTemplates,
+      RuleStatus: ["stable", "test", "experimental", "deprecated", "unsupported"],
+      text: ``,
```
tomchop, line 191:
```suggestion
      text: '',
```

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+    }
+  },
+  mounted() {
+    this.loadSketchSigmaTags()
+    this.loadSketchTTP()
+  },
   methods: {
-    parseSigma: function(event) {
-      ApiClient.getSigmaByText(this.text)
+    problem_detector: function (search_query) {// eslint-disable-line
```
tomchop, line 231:
```suggestion
    problemDetector: function (searchQuery) {// eslint-disable-line
```

tomchop: why are you asking eslint to not format this? Can we come up with a more descriptive name for this function? e.g. `extractWarningsFromRule`

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+        reason = ("No data_type or source_name defined in the rule consider add field mappings in the sigma config file")
+      }
+      if (search_query.length < 10) {
+        reason.concat("Query seems very short. It might return a lot of events and be to broad.")
+      }
+      return reason
+    },
+    templateSelected: function () {
+      // I am actually not sure what this is for, but it works.
+    },
```
tomchop, line 243: Please remove, or figure out what this is for :)

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+    parseSigma: function (rule_yaml) { // eslint-disable-line
+      this.parsing_issues = []
+      ApiClient.getSigmaRuleByText(rule_yaml)
         .then(response => {
           let SigmaRule = response.data.objects[0]
           this.parsed = SigmaRule
+          this.data_type_present = (SigmaRule['search_query'].includes("data_type") || SigmaRule['search_query'].includes("source_name"))
+
+        })
```
tomchop, line 252:
```suggestion
          this.data_type_present = (SigmaRule['search_query'].includes("data_type") || SigmaRule['search_query'].includes("source_name"))
        })
```

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+            message: 'Sigma rule parsing failed. See Browser console for more',
+            type: 'is-danger',
+            position: 'is-top',
+            indefinite: false,
+          })
         })
-        .catch(e => {})
+    },
+    addRule: function (event) {
+      this.parseSigma(this.editingRule.rule_yaml)
```
tomchop, line 263: Given that this is called all the time, do we need to call it again here?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+        ApiClient.getSigmaRuleByText(this.editingRule.rule_yaml)
+          .then(response => {
+            let SigmaRule = response.data.objects[0]
+            this.parsed = SigmaRule
+            ApiClient.updateSigmaRule(this.editingRule.id, this.editingRule.rule_yaml)
+              .then(response => { })
+              .catch(e => {
+                console.error(e)
+              })
+            location.reload();
```
tomchop, line 293: This will break the whole flow of the progressive webapp. Why do you need to reload here?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+      if (confirm('Delete Rule?')) {
+        ApiClient.deleteSigmaRule(ioc.rule_uuid)
+          .then(response => { })
+          .catch(e => {
+            console.error(e)
+          })
+        // TODO: remove the element from the array
+      }
+      // instead of removing the element from the array gonna relad the page
+      location.reload();
```
tomchop, line 326: Why? It's much more expensive and slow to reload the page than to remove the element from the array.

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+    loadSketchTTP() {
+      ApiClient.runAggregator(this.sketch.id, {
+        aggregator_name: 'field_bucket',
+        aggregator_parameters: { field: 'ts_ttp' },
+      }).then((response) => {
+        this.sketchTTP = response.data.objects[0].field_bucket.buckets
+        // of the form [{count: 0, tag: 'foo'}]
+      })
+    },
+    getColor(status) {
```
tomchop, line 356:
```suggestion
    getColorFromStatus(status) {
```

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+      if (status === 'false') return "red"
+      return "green";
+    },
+    getRuleByUUID(RuleUuid) {
```
tomchop, line 360:
```suggestion
    getRuleByUUID(ruleUuid) {
```

camelCase starts with lowercase

tomchop (changes requested):

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: done

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -32,16 +31,20 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
         self._rule = kwargs.get("rule")
         super().__init__(index_name, sketch_id, timeline_id=timeline_id)
 
-    def run_sigma_rule(
-        self, query, rule_name, tag_list=None, status_good=False
-    ):
+    def run_sigma_rule(self, query, rule_title, tag_list=None):
         """Runs a sigma rule and applies the appropriate tags.
```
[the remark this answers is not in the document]
jaegeral: changed the docstring

timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -91,78 +98,24 @@ def run(self):
...
+            error_msg = "* {0:s} {1:s}".format(
+                rule.get("title"), rule.get("id")
+            )
             logger.error(
-                "Problem with rule in file {0:s}: ".format(
-                    rule.get("file_name")
-                ),
+                error_msg,
                 exc_info=True,
             )
```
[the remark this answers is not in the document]
jaegeral: auto formatter keeps putting it in the same line 🤷

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: true

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,105 +15,182 @@ limitations under the License.
...
+    <span class="icon is-small" style="cursor: pointer" title="Add a rule Rule"
+      @click="composeRule()"><i class="fas fa-plus"></i>Add
+    </span>
+
+    <b-table v-if="sigmaRuleList" :data="sigmaRuleList"
+      :current-page.sync="currentPage" :per-page="perPage" paginated
+      pagination-simple pagination-position="bottom"
+      default-sort-direction="desc" sort-icon="arrow-down"
+      sort-icon-size="is-small" icon-pack="fas" icon-prev="chevron-left"
+      icon-next="chevron-right" default-sort="title" key="props.row.id">
```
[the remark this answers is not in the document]
jaegeral: hm it does not auto format that for me

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: fixed

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: fixed, now directly done with the array

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,105 +15,182 @@ limitations under the License.
...
+            v-model="editingRule.rule_yaml"
+            @input="parseSigma(editingRule.rule_yaml)"></b-input>
+        </b-field>
+        <b-field grouped>
+          <b-field grouped expanded position="is-right">
+            <p class="control">
+              <b-button type="is-primary"
+                @click="parseSigma(editingRule.rule_yaml)">
+                Parse
+              </b-button>
```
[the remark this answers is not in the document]
jaegeral: removed

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,105 +15,182 @@ limitations under the License.
...
+          label="Templates" label-position="on-border"
+          @change="templateSelected">
+          <option v-for="template in SigmaTemplates" :value="template.text"
+            :key="template.os">
+            {{ template.os }}
+          </option>
+        </b-select>
+        Parsed rule Search query:
+        <b><code>{{ parsed['search_query'] }}</code></b>
+        <explore-preview style="margin-left: 10px"
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/frontend/src/components/Sigma/SigmaList.vue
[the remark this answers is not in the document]
jaegeral: done button is gone

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +204,165 @@ export default {
...
+      }
+      if (search_query.length < 10) {
+        reason.concat("Query seems very short. It might return a lot of events and be to broad.")
+      }
+      return reason
+    },
+    templateSelected: function () {
+      // I am actually not sure what this is for, but it works.
+    },
+    parseSigma: function (rule_yaml) { // eslint-disable-line
```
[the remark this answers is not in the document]
jaegeral: it is debounced now

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: I removed the whole error message stuff.... It does not feel good yet I rather do it good then half baked

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,105 +15,182 @@ limitations under the License.
...
         </div>
       </b-table-column>
 
-      <b-table-column field="ts_use_in_analyzer" label="Use in Analyzer" v-slot="props" sortable>
-        <div @click="props.toggleDetails(props.row)" style="margin-top:5px;cursor:pointer;">
-          {{ props.row.ts_use_in_analyzer }}
+      <b-table-column field="author" label="Author" v-slot="props" searchable
+        sortable>
+        <div @click="startRuleEdit(props.row)"
+          style="margin-top:5px;cursor:pointer;">
```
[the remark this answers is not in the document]
jaegeral: done, it now has a custom class

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,104 +15,163 @@ limitations under the License.
...
 
-      <b-table-column field="description" label="Description" v-slot="props" searchable>
-        <div @click="props.toggleDetails(props.row)" style="margin-top:5px;cursor:pointer;">
-          {{ props.row.description }}
+      </b-table-column>
+      <b-table-column field="status" label="Status" v-slot="props" sortable
+        searchable>
+        <div @click="startRuleEdit(props.row)" custom-class="margintop-pointer">
+          {{ props.row.status }} <span class="icon is-small"
+            style="cursor: pointer"
```
tomchop, line 87: superfluous? maybe a class called "clickable" could be added here, and re-used in L113  and L116

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -15,104 +15,163 @@ limitations under the License.
...
     return {
       currentPage: 1,
       ascending: false,
       sortColumn: '',
       perPage: 10,
-      isComposed: false,
-      text: `Place your Sigma rule here and press parse`,
+      editingRule: {},
+      showEditModal: false,
+      save_button_text: "Update", // false: update rule. true: create new in DB
```
tomchop, line 169: Why no camelCase on `save_button_text`? The comment on this line also doesn't seem to make sense here, remnant of previous code?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+        .catch(e => { })
+    }
+    // If no timeline was specified then loop over all of them
+    else {
+      this.sketch.timelines.forEach(timeline => {
+        ApiClient.getSketchTimelineAnalysis(this.sketch.id, timeline.id)
+          .then(response => {
+            this.analyses = this.analyses.concat(response.data.objects[0])
+          })
+          .catch(e => { })
```
tomchop, line 204: Given that you're already catching the error, might as well log / display something?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+            position: 'is-top',
+            indefinite: false,
+          })
+        })
+    }, 300),
+
+
+
+
+
```
tomchop, line 234:
```suggestion
```

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+      this.parsing_issues = []
+      ApiClient.getSigmaRuleByText(rule_yaml)
         .then(response => {
           let SigmaRule = response.data.objects[0]
           this.parsed = SigmaRule
         })
-        .catch(e => {})
+        .catch(e => {
+          Snackbar.open({
+            message: 'Sigma rule parsing failed. See Browser console for more',
```
tomchop, line 223:
```suggestion
            message: 'Sigma rule parsing failed. See Browser console for more.',
```

Didn't we say we'd display the error mesasge somewhere in the Sigma editor?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+            let SigmaRule = response.data.objects[0]
+            this.parsed = SigmaRule
+            ApiClient.createSigmaRule(this.editingRule.rule_yaml).then(response => {
+              this.$buefy.notification.open({ message: 'Succesfully added Sigma rule!', type: 'is-success' })
+              this.showEditModal = false
+              this.sigmaRuleList.push(response.data.objects[0])
+            })
+              .catch(e => {
+                console.error(e)
+              })
```
tomchop, line 248: Can we make this error message more apparent to the user?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+            ApiClient.updateSigmaRule(this.editingRule.id, this.editingRule.rule_yaml)
+              .then(response => {
+                this.$store.state.sigmaRuleList = this.sigmaRuleList.filter(obj => {
+                  return obj.rule_uuid !== this.editingRule.rule_uuid
+                })
+                this.sigmaRuleList.push(response.data.objects[0])
+              })
+              .catch(e => {
+                console.error(e)
+              })
```
tomchop, line 268: Can we make this error message more apparent to the user?

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +186,167 @@ export default {
...
+
+
+
+
+    addRule: function (event) {
+      if (this.save_button_text === "Create") {
+        ApiClient.getSigmaRuleByText(this.editingRule.rule_yaml)
+          .then(response => {
+            let SigmaRule = response.data.objects[0]
+            this.parsed = SigmaRule
```
tomchop, line 240: This is a repeat of what's happening in lines 255->259, maybe we can check for the state of save_button_text in the callback of this sucessful function?

tomchop: Actually, do we really need this call to `getSigmaRuleByText` in the first palce? isn't `createSigmaRule` also going to return an object that we can assign to `this.parsed`?

tomchop (changes requested):

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: actually gonna rename the button to `ActionButton` since Update is different action to Save

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: done

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: https://github.com/google/timesketch/issues/2357

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: done

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: https://github.com/google/timesketch/issues/2357

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: I refactored that a bit, PTAL

timesketch/frontend/src/components/Sigma/SigmaList.vue
jaegeral: https://github.com/google/timesketch/issues/2357

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +182,148 @@ export default {
...
+            type: 'is-danger',
+            position: 'is-top',
+            indefinite: false,
+          })
+        })
+    }, 300),
+    addOrUpdateRule: function (event) {
+      if (this.save_button_text === "Create") {
+        // use parseSigma() instead of the direct call here
+        this.parseSigma(this.editingRule.rule_yaml)
```
tomchop, line 214: Remove

tomchop (changes requested):
Few nits, otherwise LGTM!

timesketch/frontend/src/components/Sigma/SigmaList.vue
```diff
@@ -127,14 +182,148 @@ export default {
...
+              message: this.problemString,
+              type: 'is-danger',
+              position: 'is-top',
+              indefinite: false,
+            })
+          })
+      }
+      if (this.save_button_text === "Update") {
+        // Only update the rule if the parsing was positive.
+        this.parseSigma(this.editingRule.rule_yaml)
```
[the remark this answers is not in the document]
jaegeral: done

tomchop (approved):

PR #2350, opened 2022-09-30
Remove doc_type from opensearch client code

berggren:
doc_type has been deprecated for a while by opensearch, and finally removed in recent python clients. This break TS and need to be removed. This PR removes the use of doc_type.

Additionally there is a small bugfix for invalid timestamp rendering in the UI.


jaegeral (changes requested):
```************* Module timesketch.lib.datastores.opensearch
timesketch/lib/datastores/opensearch.py:817:8: W0613: Unused argument 'event_type' (unused-argument)
```

please fix

berggren:
@jaegeral Ready for review, PTAL

jaegeral (approved):
Lgtm

PR #2355, opened 2022-10-02
Search template UI

berggren:
This PR add UI controls to run search templates. Support for both templates with parameters and without.

Screen Shot 2022-10-10 at 11 52 55


timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -100,21 +100,19 @@ def get(self, searchtemplate_id):
...
+        form = request.json or {}
         searchtemplate = SearchTemplate.query.get(searchtemplate_id)
         if not searchtemplate:
             abort(HTTP_STATUS_CODE_NOT_FOUND, "Search template was not found")
 
         try:
             template = jinja2.Template(searchtemplate.query_string)
-            parsed_query_string = template.render(parameters)
+            parsed_query_string = template.render(form)
+            parsed_query_string = parsed_query_string.replace("\\/", "\\/")
```
jaegeral, line 111: Is that something that might end up required in multiple places? If so maybe we should add a util function?
Also I would suggest we add some tests to unit tests that those replacements provide what they are suppose do.

timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -84,7 +84,7 @@ class SearchTemplateParseResource(resources.ResourceMixin, Resource):
     """Resource to parse a search template query string using Jinja2 template."""
 
     @login_required
-    def get(self, searchtemplate_id):
+    def post(self, searchtemplate_id):
         """Parse the query string template with Jinja2.
 
         This resource take a form with parameters. These will be sent to the Jinja2
```
jaegeral, line 90: there is a `will` missing

timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -100,21 +100,19 @@ def get(self, searchtemplate_id):
         Returns:
             Search template in JSON (instance of flask.wrappers.Response)
```
jaegeral, line 101: Can you make that maybe with an example on what is returned? E.g. `{"objects": [{"query_string": "*foobar*"}], "meta": {}}`

timesketch/api/v1/resources/searchtemplate.py
```diff
@@ -100,21 +100,19 @@ def get(self, searchtemplate_id):
...
             template = jinja2.Template(searchtemplate.query_string)
-            parsed_query_string = template.render(parameters)
+            parsed_query_string = template.render(form)
+            parsed_query_string = parsed_query_string.replace("\\/", "\\/")
+            parsed_query_string = parsed_query_string.replace(".", "\\.")
         except jinja2.exceptions.TemplateSyntaxError as e:
             abort(HTTP_STATUS_CODE_BAD_REQUEST, f"Search template syntax error: {e}")
+
         result_dict = {"query_string": parsed_query_string}
         return jsonify({"objects": [result_dict], "meta": {}})
```
jaegeral, line 117: here you could simplify to:

`return jsonify({"objects": [{"query_string": parsed_query_string}], "meta": {}})`

jaegeral (commented):
Some comments

timesketch/api/v1/resources/searchtemplate.py
berggren: I opt to keep it as i find it more readable.

timesketch/api/v1/resources/searchtemplate.py
berggren: Yeah, it actually returns a search query string. Changed the wording.

timesketch/api/v1/resources/searchtemplate.py
berggren: I don't understand what you mean with `will missing` :) But I rewording this sentence to make it more accurate.

timesketch/api/v1/resources/searchtemplate.py
berggren: I agree, let's do it now rather than later. I refactored this into a utility function and added tests.

jaegeral (approved):

PR #2366, opened 2022-10-11
Sigma UI and import command with tsctl

berggren:
* Initial UI for Sigma rules in explore/search view.
* Added command to import rules from tsctl

Screen Shot 2022-10-11 at 13 20 11

### DIFF
--- a/timesketch/tsctl.py
+++ b/timesketch/tsctl.py
@@ -24,12 +24,14 @@ from sqlalchemy.exc import IntegrityError
 
 from timesketch import version
 from timesketch.app import create_app
+from timesketch.lib import sigma_util
 from timesketch.models import db_session
 from timesketch.models import drop_all
 from timesketch.models.user import Group
 from timesketch.models.user import User
 from timesketch.models.sketch import Sketch
 from timesketch.models.sketch import SearchTemplate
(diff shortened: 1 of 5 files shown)


timesketch/tsctl.py
```diff
@@ -309,3 +311,58 @@ def import_search_templates(path):
...
+            )
+            db_session.add(sigma_db_rule)
+            db_session.commit()
+
+            sigma_db_rule.set_status(sigma_rule.get("status", "experimental"))
+            # query string is not stored in the database but we attach it to
+            # the JSON result here as it is added in the GET methods
+            sigma_db_rule.query_string = sigma_rule.get("search_query")
+        else:
+            print(f"Rule already imported: ", sigma_rule.get("title"))
```
jaegeral, line 368: timesketch/tsctl.py:368:18: W1309: Using an f-string that does not have any interpolated variables (f-string-without-interpolation)

jaegeral (changes requested):
otherwise LGTM

timesketch/tsctl.py
jaegeral: I fixed that

jaegeral (approved):

PR #2372, opened 2022-10-14
Update docker images and CI tests to Ubuntu 22.04

berggren:
This PR updates our Docker images and CI tests to use Ubuntu 22.04. We still suport 20.04, but with the caveat that 20.04 won't have Plaso versions > 20220724.

* Changed Ubuntu base version to 22.04 on all Docker images, and made it the default.
* Added 22.04 to the OS matrix in GH actions to run all tests on that version
* Cleaned up actions and removed legacy workarounds
* Temporarily removed PPA tests as this process need to be updated. This work is tracked in #2373  
* Migrated away from `nosetests` -> `pytest` - nosetest has been deprecated and don't work on Python 3.10, hence can't run on our CI infrastructure.
* Use ubuntu-latest as our host OS on GH actions.
* Bumped celery, pandas, numpy and redis to new versions supporting Python 3.10

### DIFF
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   PyLint:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
         os: [ubuntu-20.04, ubuntu-22.04]
@@ -36,7 +36,7 @@ jobs:
           for FILE in `git --no-pager diff origin/master --name-only --diff-filter=ACMR | grep \.py$`; do echo "Running pylint against ${FILE}"; pylint --rcfile=.pylintrc ${FILE}; done
 
   ESLint-frontend:
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
     defaults:
       run:
         working-directory: ./timesketch/frontend
--- a/.github/workflows/linters.yml
+++ b/.github/workflows/linters.yml
@@ -22,9 +22,6 @@ jobs:
         # Pin pip to pip==22.0.3 due to upstream bug in pip.
         # TODO: Revisit and remove this when fixed upstream.
         run: |
-          pip install -U pip==22.0.3
-          pip install astroid==2.4.0
-          pip install pylint==2.6.0
           pip install -r test_requirements.txt
           pip install -r requirements.txt
           pip install -e .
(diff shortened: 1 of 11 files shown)


.github/workflows/linters.yml
```diff
@@ -6,11 +6,11 @@ on:
 
 jobs:
   PyLint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
```
joachimmetz, line 9: since you're using docker why not `ubuntu-latest` ?

berggren: Good catch, I changed it to `latest` 👍

berggren:
Tests passes now. @joachimmetz PTAL

.github/workflows/linters.yml
```diff
@@ -6,11 +6,11 @@ on:
...
 jobs:
   PyLint:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-20.04]
-        python-version: [3.7, 3.8]
+        os: [ubuntu-20.04, ubuntu-22.04]
+        python-version: [3.8, 3.10.7]
```
joachimmetz, line 13: i opt: `python-version: ['3.8', '3.10']`

.github/workflows/unit-tests.yml
```diff
@@ -5,13 +5,13 @@ on:
...
-    runs-on: ubuntu-20.04
+  # Backend tests (Python/Flask)
+  Python:
+    runs-on: ubuntu-latest
     strategy:
       matrix:
-        os: [ubuntu-20.04]
-        python-version: [3.7, 3.8]
+        os: [ubuntu-20.04, ubuntu-22.04]
+        python-version: [3.8, 3.10.7]
```
joachimmetz, line 14: same as previous comment

joachimmetz (approved):
LGTM, left 1 suggestion

.github/workflows/linters.yml
berggren: Thanks, good catch! I got `unknown version` for 3.10 earlier, but I forgot that I hadn't stringified the version. Fixed now, and running on 3.10.

.github/workflows/unit-tests.yml
berggren: done

PR #2377, opened 2022-10-19
Sigma rules to be deployed by default

### LINKED ISSUE #2376: Ship top 10 Sigma rules with Timesketch
When installing a fresh Timesketch instance, users are left out which Sigma rules to deploy best, so it might be a good idea to include some rules in the installation scripts when the docker container is started.

jaegeral:
Add a small set of default rules to Sigma that will be installed when someone starts a new dev instance.

Options considered:
* looping over a plain text file with one line per sigma rule to be deployed
* create a dedicated full blown bash script that contains the tsctl command for each sigma rule.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #2376


jaegeral:
Not sure why black is not accepting my formatting, did you change anything @berggren ?

timesketch/tsctl.py
```diff
@@ -439,6 +455,6 @@ def export_sigma_rules(path):
             continue
 
         with open(file_path, "wb") as fw:
-            fw.write(rule.rule_yaml.encode("utf-8"))
+            fw.write(rule.rule_yaml.encode('utf-8'))
```
berggren, line 458: Please setup Black for formatting. Singel quotes are not consistent with the rest of the codebase.

docker/dev/build/docker-entrypoint.sh
```diff
@@ -71,6 +71,42 @@ if [ "$1" = 'timesketch' ]; then
   # Add web user
   tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}"
 
+  # Add Sigma rules
+  git clone https://github.com/SigmaHQ/sigma /usr/local/src/sigma
```
berggren, line 75: I opt to create a file of the filenames that we want to import and loop over them while doing an os.walk. This way we have a text file that we can easily edit. We should also consider making this part of the release image as well.

berggren (changes requested):

docker/dev/build/docker-entrypoint.sh
jaegeral: This would be a little larger effort to get done, not sure if you would be okay to still accept this here, if so: https://github.com/google/timesketch/issues/2415 to track it to be rewritten

berggren: moving the paths to a file should be simple. Then we can store and update that in the data directory instead of hardcoding this in code.

berggren (changes requested):
PTAL - moving the paths to a file should be simple.

PR #2384, opened 2022-10-21
[API client] Add get_intelligence_attribute to sketch resource

itsmvd:
This allows users of the API to retrieve the indicators stored in the Sketch's intelligence attribute.


tomchop:
This looks good - but how about we also include an API endpoint that returns just this attribute? That way we skip sending all other attributes from the sketch when the user only wants intelligence. @berggren thoughts?

tomchop (approved):

berggren:
> This looks good - but how about we also include an API endpoint that returns just this attribute? That way we skip sending all other attributes from the sketch when the user only wants intelligence. @berggren thoughts?

Isn't that what this resource do? It only returns the intelligence attribute. Or am I missing something? 

In any case - this LGTM to me as well. I'll merge this tomorrow depending on the response to the above.

itsmvd:
> > This looks good - but how about we also include an API endpoint that returns just this attribute? That way we skip sending all other attributes from the sketch when the user only wants intelligence. @berggren thoughts?
> 
> Isn't that what this resource do? It only returns the intelligence attribute. Or am I missing something?
> 
> In any case - this LGTM to me as well. I'll merge this tomorrow depending on the response to the above.

Forgot to reply yesterday but was going to say the same, this method only returns the intelligence attribute.

tomchop:
To clarify my comment - this is a function in the python API client. Whenever the intelligence attribute is needed, the client needs to first load all attributes, which might be unnecessary and potentially bandwidth consuming depending on the size of the other attributes.

PR #2391, opened 2022-10-21
Update tasks.py

Zawadidone:
Use paramater value if exists otherwise use Plaso default settings.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


berggren (approved):
Good catch, LGTM

PR #2379, opened 2022-10-19
Migrate OOB flow to localhost flow for OAuth2 auth

jleaniz:
Please provide enough information so that others can review your pull request:

This PR addresses issue #2378.

**Checks**

- [X] All tests succeed.
- [X] Unit tests added.
- [X] e2e tests added.
- [X] Documentation updated.

**Closing issues**
closes #2378

PR #2409, opened 2022-11-02
Remove oob

berggren:
#2379 Switched to use the local server flow. This PR changes the redirect URL as well as removing the last reference of the deprecated OOB flow.


jleaniz (approved):
LGTM. Sorry I missed the redirect_uri in the other PR.

PR #2414, opened 2022-11-07
[DRAFT] SigmaStudio

jaegeral:
This PR does migrate the Sigma rule maintenance over to frontend-ng

- [x] Have a Auto complete dropdown for the template lookup
- [ ] Have a second view that is not a raw text editor for the rule but a more guided form
- [x] When pressing **cancel** the previous URL should be pushed to the route explorer to go back
- [x] Have a button in the Sigma rule list that says **"create a new rule"**
- [x] Make the Parsed search query in a box like the parsing error
- [x] Load a new rule when pressing edit button
- [ ] Display parsing error messages

# Example UI

2022-11-11_Sigma_rule_creation

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


jaegeral:
https://github.com/google/timesketch/pull/2444 is the canonical one

PR #2419, opened 2022-11-09
Update SearchTemplates.vue to check if no searchTemplates are on a system

### LINKED ISSUE #2395: frontend-ng 0 Searchtemplates leads to Type Error
Having no Search templates leads to the following error in Chrome Console:

```
TypeError: Cannot read properties of undefined (reading 'length')
    at o.ci (SearchTemplates.vue?9468:1:452)
    at t._render (vue.runtime.esm.js:3569:22)
    at o.i (vue.runtime.esm.js:4081:21)
    at ii.get (vue.runtime.esm.js:4495:25)
    at ii.run (vue.runtime.esm.js:4570:22)
    at Kn (vue.runtime.esm.js:4326:13)
    at Array.<anonymous> (vue.runtime.esm.js:1989:12)
    at ue (vue.runtime.esm.js:1915:12)


```

jaegeral:
Fixes: https://github.com/google/timesketch/issues/2395
Currently if no SearchTemplates are installed on a system, the check for lengths in the UI does fail because the API returns undefined.

This fix addresses that.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #2395

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
@@ -80,6 +80,9 @@ export default {
     ApiClient.getSearchTemplates()
       .then((response) => {
         this.searchtemplates = response.data.objects[0]
+        if (typeof(this.searchtemplates) == 'undefined') {
+          this.searchtemplates = []
+        }
       })
       .catch((e) => {})
   },
--- a/timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
@@ -80,7 +80,7 @@ export default {
     ApiClient.getSearchTemplates()
       .then((response) => {
         this.searchtemplates = response.data.objects[0]
-        if (typeof(this.searchtemplates) == 'undefined') {
+        if (typeof(this.searchtemplates) === 'undefined') {
           this.searchtemplates = []
         }
       })


timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
```diff
@@ -80,6 +80,9 @@ export default {
     ApiClient.getSearchTemplates()
       .then((response) => {
         this.searchtemplates = response.data.objects[0]
+        if (typeof(this.searchtemplates) === 'undefined') {
+          this.searchtemplates = []
+        }
```
berggren, line 85: Should we return an empty list from the backend API instead?

berggren (changes requested):

timesketch/frontend-ng/src/components/LeftPanel/SearchTemplates.vue
jaegeral: image

jaegeral: So I just looked at https://github.com/jaegeral/timesketch/blob/2022-11-09_2395/timesketch/api/v1/resources/scenarios.py#L73 and we do return an undefined in many more places, so I suggest we accept this change and make a bigger effort to make all the API endpoints behave similar. WDYT?

berggren: I'm working on that code elsewhere, and it is being revamped. So, yeah SG, I'll fix it later.

berggren (approved):

PR #2430, opened 2022-11-17
[Sigma] ship some basic but broad rules with Timesketch

jaegeral:
While we do not want to maintain and curate a broad Set up Sigma rules, some basic rules that can be imported to get quick wins might be nice.

Inspired by https://github.com/SigmaHQ/sigma/blob/master/other/godmode_sigma_rule.yml and https://twitter.com/cyb3rops/status/1592914258974879746


jaegeral:
> @jaegeral this is a great idea. Would it possible to add the rule in a separate repository? This could allow the reuse and the sharing with other tools.

Would you prefer those rules to also be in github.com/sigmahq/ or somewhere else?

So far I tried to stay away from creating another repo with Sigma rules that requires / expects to be maintained, but if we have more rules that make only sense in a forensic context, I might get convinced to start something...

jkppr (approved):

Issue #2439, opened 2022-12-01
Create a context lookup mechanism for field values


jkppr:
### Feature Idea

Provide a mechanism to lookup individual event values within Timesketch by linking it to other tools. The analyst will see some indicator in the event detail view and can manually trigger that mechanism to open the lookup for that specific value in another tab.

### Background

Sometimes we analysts want to get additional information on a specific value in an event from external tools. For example looking up the hash value of a suspicious file in VirusTotal or checking a domain against an internal CTI tool.
With the current state of Timesketch, an analyst needs to copy/paste the value into a lookup tool of their choice. This also requires that each analyst knows the best and approved lookup tools for the job.

This feature would benefit the UX workflow for analysts by reducing the copy/paste activity and spreading the knowledge about lookup options in the team.

**Use-Case ideas:**

- Link IoC fields like IP, Hash, Domain, URL to a lookup at a favorite CTI platform
- Quickly decode a URL using unfurl
- Sent data to your cyberchef instance for decoding/converting
- Link documentation for Windows event IDs for a quick lookup
- Submit URLs to your favorite sandbox tool

### Basic Requirements

- A configuration schema for the Timesketch administrator to define which fields should be linked to a lookup in their instance.
- An UI element with each configured field in the event detail view that clearly indicates a call to action for analysts and provides the selection of prepared links.
- A redirect warning feature, that prevents analysts from accidental sharing investigation data with external platforms.
- The feature needs to be documented in the Timesketch docs.

### Development steps

- [x] Add a configuration schema/file
- [x] API endpoint
- [x] Add tests for the API endpoint
- [x] Update API client  in the front-end
- [x]  Update the `EventDetail.vue` component
- [x] Add documentation on how to use the feature
- [x] Add the feature to the change log `docs/changelog/2022-12.md`
- [x] Create GitHub issues for future development and new ideas of the context links.

jkppr:
Here are some demo gifs that show how the feature looks in action.

1. A context link that triggers the redirect warning (in this case to VirusTotal)

![TS_ContextLinks_RedirectWarn](https://user-images.githubusercontent.com/99879757/205897124-00154a89-e58a-4cf0-a89e-ba3b4f3052d8.gif)

2. If a verification_regex is configured for a context link, it will only show the trigger when the value matches the regex. The gif shows two fields called hash, but only one shows the context link trigger.

[...]

PR #2440, opened 2022-12-01
Context lookup API endpoint part 1

jkppr:
**This PR is part one of working on issue #2439 to create a context lookup for event fields!** 

This PR includes:

- A configuration file `context_links.yaml` that is used to provide the context link configuration.
- Adjustments in existing configurations to load this file.
- An API endpoint resource `contextlinks.py` that loads and validates the config file.
- Adjusted routes to make the API endpoint available.


timesketch/api/v1/resources/contextlinks.py
```diff
@@ -0,0 +1,105 @@
...
+            else:
+                check_ok = False
+
+            # Verify that redirect_warning is defined and type bool
+            if not isinstance(item_dict.get("redirect_warning"), bool):
+                check_ok = False
+
+            if not check_ok:
+                logger.warning(
+                    "Failed to load context link: '{0:s}'! Please check the "
```
jaegeral, line 94: Couldn't we provide more info on what is missing instead?

timesketch/api/v1/resources/contextlinks.py
```diff
@@ -0,0 +1,105 @@
...
+        """Handles GET request to the resource.
+        Returns:
+            JSON object including version info
+        """
+        context_link_config = load_yaml_config("CONTEXT_LINKS_CONFIG_PATH")
+
+        response = {}
+
+        if not context_link_config:
+            logger.warning("The config file 'data/context_links.yaml' is empty!")
```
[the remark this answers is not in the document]
jkppr: Good point. However, after our talk about what to report in the logs and what not, I will opt for removing the warning. Since the default is an empty config file, this would create unnecessary log entries for a feature that the operator maybe just does not want to use. 
TL;DR: An empty config file is working as intended and hence does not need a warning.

timesketch/api/v1/resources/contextlinks.py
jkppr: Based on our discussion I refactored the input verification to be
1. better readable in code
2. bundled all failures in one message at the end to prevent spamming of the log file (1 message independent of how many entries are bad)
3. each error message now contains more granular info on which part of the config is bad

timesketch/api/v1/resources/contextlinks.py
```diff
@@ -0,0 +1,111 @@
...
+
+        response = {}
+
+        if not context_link_config:
+            return jsonify(response)
+
+        verification_results = []
+
+        # Check mandatory fields are correctly defined.
+        for item in context_link_config:
```
berggren, line 49: Why we are validating the YAML file every time the API endpoint executes? This implies that we don't trust the supplied, server side, YAML file to follow the needed format.

Instead, we could add a validation function to tsctl that the server admin can run to validate the yaml file before it is used in the system. Then it is up to the admin to check that it has a valid file. We could then use something like jsonschema to validate the fields to simplify this logic.

WDYT? Let's discuss this in person.

berggren (changes requested):

timesketch/api/v1/resources/contextlinks.py
jkppr: Thanks for the input. As discussed:

- Moved validation to tsctl for manual execution by an admin.
- Switched to jsonschema validation
- Removed validation from the API endpoint itself

timesketch/tsctl.py
```diff
@@ -21,6 +21,7 @@
 import click
 from flask.cli import FlaskGroup
 from sqlalchemy.exc import IntegrityError
+from jsonschema import validate, ValidationError
```
berggren, line 24: Do we need to add this to requirements.txt?

berggren (changes requested):
Looks good, one question we need to address before merge

timesketch/tsctl.py
jkppr: Added as discussed.

berggren (approved):

PR #2448, opened 2022-12-06
Context lookup part 3 - front-end UI

jkppr:
**This PR is part two of working on issue #2439 to create a context lookup for event fields.**

This PR includes:

- Adding a new `LinkRedirectWarning.vue` component that asks the user if they really want to continue to open a link to an external site.
- Refactored the `EventDetail.vue` to have a new second column that can host icon buttons.
- Added the indicators and triggers to the context links in the `EventDetail.vue`.
- Implemented the context links.

**Demo**
The first gif shows the redirect warning when trying to visit an external page (here Virustotal). If the warning is necessary needs to be configured by the admin in the config file.
TS_ContextLinks_RedirectWarn

The second gif shows a context link without an redirect warning, since the unfurl instance is hosted locally.
TS_ContextLinks_unfurl

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
@@ -37,7 +37,7 @@ limitations under the License.
                           v-on="on"
                           @click="formatXMLString = true"
                         >
-                          <v-tooltip top>
+                          <v-tooltip top close-delay=300 :open-on-click="false">
                             <template v-slot:activator="{ on }">
                               <v-icon
                                 v-on="on"
@@ -70,8 +70,7 @@ limitations under the License.
                           v-bind="attrs"
                           v-on="on"
                         >
-                          <!-- <v-icon x-small>mdi-magnify-expand</v-icon> -->
-                          <v-tooltip top>
+                          <v-tooltip top close-delay=300 :open-on-click="false">
                             <template v-slot:activator="{ on }">
                               <v-icon
                                 v-on="on"
--- a/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
@@ -24,17 +24,30 @@ limitations under the License.
                 <tr v-for="(value, key) in fullEventFiltered" :key="key">
                   <td>
                     {{ key }}
+                  </td>
+                  <td v-if="key.includes('xml') || checkContextLinkDisplay(key, value)" class="px-0">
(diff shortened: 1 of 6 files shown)


timesketch/frontend-ng/src/components/Explore/EventDetail.vue
```diff
@@ -43,8 +56,62 @@ limitations under the License.
...
+                      </template>
+                      <v-list dense>
+                        <v-list-item
+                          v-for="(item, index) in getContextLinkItems(key)"
+                          :key="index"
+                          style="cursor: pointer"
+                          @click.stop="contextLinkRedirect(key,item,value)"
+                        >
+                          <v-list-item-title v-if="getContextLinkRedirectState(key,item)" >{{ item }}</v-list-item-title>
+                          <v-list-item-title v-else >{{ item }}*</v-list-item-title>
```
berggren, line 94: Maybe replace the start with. "(internal)"?

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
```diff
@@ -24,17 +24,30 @@ limitations under the License.
                 <tr v-for="(value, key) in fullEventFiltered" :key="key">
                   <td>
                     {{ key }}
+                  </td>
+                  <td v-if="key.includes('xml') || checkContextLinkDisplay(key, value)" class="px-0">
```
berggren, line 28: Consider moving this into the menu instead of rendering its own icon. Not necessary in the PR, but maybe add a TODO?

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
jkppr: Since I leave the configuration for the redirect warning to each admin, the linked service can also be external and show now redirect warning (e.g. if they don't care about checking hashes in VT). In this situation it would show `Virustotal (internal)` which is not correct.
Hence the idea to go with just a marker for "no redirect warning", which can be improved.

For now I would like to keep the `*` for this PR and take the topic in my next discussion with UX.

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
jkppr: I thought about that earlier as well. However, I'm not sure how this will work with the expectations created by the icon type (link-extern). 
I opened #2449 to research and track this topic.

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
berggren: SG

berggren (approved):
LGTM

PR #2451, opened 2022-12-07
Scenarios conclusion

berggren:
This PR adds basic support for conclusions for DFIQ questions.

* Support add, edit and delete conclusion, per user
* Minor tweaks to the scenarios panel UI

Screen Shot 2022-12-07 at 13 20 03


timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+        """
+        sketch = Sketch.query.get_with_acl(sketch_id)
+        question = InvestigativeQuestion.query.get(question_id)
+
+        conclusion = InvestigativeQuestionConclusion.get_or_create(
+            user=current_user, investigativequestion=question
+        )
+
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sketch found with this ID")
```
jkppr, line 314: Same as above. If you do this check directly after you get the sketch you safe us the query for question and conclusion if the sketch does not exist/is not accessible.

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+
+        form = request.json
+        if not form:
+            form = request.data
+
+        conclusion_text = form.get("conclusionText")
+        if conclusion_text:
+            conclusion.conclusion = conclusion_text
+            db_session.add(conclusion)
+            db_session.commit()
```
jkppr, line 327: If `conclusion_text` is empty (e.g. because a user wants to remove their conclusion) this will do nothing and still return the old conclusion. Does this work as intended?

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+        return self.to_json(conclusions)
+
+    @login_required
+    def post(self, sketch_id, question_id):
+        """Handles POST request to the resource.
+
+        Adds or edits a conclusion.
+
+        Returns:
+            A JSON representation of the question.
```
jkppr, line 304: In line 239 you return a JSON representation of the conclusion not the question. Please fix the comment here.

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+            investigativequestion=question
+        ).all()
+
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sketch found with this ID")
+
+        return self.to_json(conclusions)
+
+    @login_required
+    def post(self, sketch_id, question_id):
```
jkppr, line 298: This post function is placed in the class `QuestionConclusionList`. But it actually just modifies / adds one conclusion of a specific question. Wouldn't it make more sense to place this function in the `QuestionConclusionResource` below? With the **List** part of the class I wouldn't expect it to modify one conclusion to be honest.

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+        conclusion = InvestigativeQuestionConclusion.query.get(conclusion_id)
+
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sketch found with this ID")
+
+        if not question:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No question found with this ID")
+
+        if not conclusion:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No conclusion found with this ID")
```
jkppr, line 352: Same as above. We would save some queries if we check directly after each get and abort before continue to query the next element.

jkppr (changes requested):
Some comments

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+    """Resource for investigative question conclusion."""
+
+    @login_required
+    def get(self, sketch_id, question_id):
+        """Handles GET request to the resource.
+
+        Returns:
+            A list of JSON representations of the conclusions.
+        """
+        sketch = Sketch.query.get_with_acl(sketch_id)
```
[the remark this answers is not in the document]
berggren: This is to make sure that the user has access to the sketch. The questions and conclusions are related to a Scenario, and the scenario =is tied to a sketch. So, if the user has access to the sketch, they should have access to the scenario (and in extension the questions and conclusions)

timesketch/api/v1/resources/scenarios.py
```diff
@@ -269,3 +270,96 @@ def post(self, sketch_id, scenario_id):
...
+        question = InvestigativeQuestion.query.get(question_id)
+
+        conclusions = InvestigativeQuestionConclusion.filter_by(
+            investigativequestion=question
+        ).all()
+
+        if not sketch:
+            abort(HTTP_STATUS_CODE_NOT_FOUND, "No sketch found with this ID")
```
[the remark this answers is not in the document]
berggren: Yes, good catch. I'll move it up

timesketch/api/v1/resources/scenarios.py
berggren: In REST and in our implementation, creating a new resource goes into the "list" part. For example, to create a new conclusion you would post to: `/sketches/1/questions/1/conclusions/` and to change an already existing one you PUT to `/sketches/1/questions/1/conclusions/123/`. I opted to not replicate the code, but I agree, you have a good point in this case. I'll add a PUT method on the QuestionConclusion for updates to make it more clear here.

timesketch/api/v1/resources/scenarios.py
berggren: copy paste error :) Fixed.

timesketch/api/v1/resources/scenarios.py
berggren: Good catch, I moved it up.

timesketch/api/v1/resources/scenarios.py
berggren: Good question. Do we want to remove the conclusion if the conclusionText is empty.. Maybe that is the expected result. WDYT?

timesketch/api/v1/resources/scenarios.py
berggren: fixed.

berggren (commented):
PTAL

timesketch/api/v1/resources/scenarios.py
jkppr: I see two options here:
1. We can prevent empty conclusions and always require at least a minimum of one symbol as input. (I at least don't see a use-case for an empty conclusion. But maybe I'm missing something?) This could be done in the frontend and therefore I don't see a need to change anything here.
2. We can try to interpret the most likely action the user intended and trigger the deletion of the conclusion. However, this would also be done best in the frontend I guess.

Will mark this comment as resolved and open a new one in the frontend files to go with one of the options.

timesketch/api/v1/resources/scenarios.py
jkppr: TIL, thanks!

timesketch/frontend-ng/src/components/Scenarios/QuestionConclusion.vue
```diff
@@ -0,0 +1,123 @@
...
+      this.showControls = false
+      this.conclusion.conclusion = ''
+      ApiClient.editQuestionConclusion(this.sketch.id, this.question.id, this.conclusion.id, this.conclusionText)
+        .then((response) => {
+          this.$store.dispatch('updateScenarios', this.sketch.id).then(() => {
+            this.loading = false
+          })
+        })
+        .catch((e) => {})
+    },
```
jkppr, line 100: In the context of [an earlier comment](https://github.com/google/timesketch/pull/2451#discussion_r1049616090) we can either check here if the `this.conclusionText` is empty and prevent empty text. Or we trigger the `deleteConclusion()` function and follow the interpretation of the users intention if the variable is empty?
WDYT?

jkppr (approved):
LGMT over all. Just one optional comment, if you want to address that topic.

timesketch/api/v1/resources/scenarios.py
berggren: I agree - lets enforce at least one char in the frontend 👍

timesketch/frontend-ng/src/components/Scenarios/QuestionConclusion.vue
berggren: Adding a check

PR #2456, opened 2022-12-12
[Sigma] try catch block if crypto.randomUUID is not supported

### LINKED ISSUE #2410: Bug: Chrome/FF-Uncaught TypeError: crypto.randomUUID is not a function
**Describe the bug**
After a clean installation of Timesketch on Fedora 36 server, I attempt to log in to the Timesketch UI using the account I set up in the installation. It successfully logs in; however, the UI does not come up. In the web developer tools console, I receive an error "Uncaught TypeError: crypto.randomUUID is not a function" in the SigmaRuleTemplates.js

**To Reproduce**
1. Install a clean version of Timesketch
2. Login 
3. Open developer tools and review console output

**Expected behavior**
The ability to interact with the web UI

 - OS: Fedora Server 36
 - FF 106/Chrome 107/Safari 16

jaegeral:

https://github.com/google/timesketch/issues/2410

This PR creates a static uuid if the function is not available.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

closes #2410

### DIFF
--- a/timesketch/frontend/src/utils/SigmaRuleTemplates.js
+++ b/timesketch/frontend/src/utils/SigmaRuleTemplates.js
@@ -15,8 +15,15 @@ limitations under the License.
 */
 
 // General first part of every Sigma rule:
+// try catch block if crypto.randomUUID is not supported
+try {
+    var rule_uuid = crypto.randomUUID()
+} catch (e) {
+    console.log('crypto.randomUUID() not supported, using a fixed value instead')
+    rule_uuid = '10a4fb8c-29d5-4eb6-905f-13d6a553d470'
+}
 const SkeletonFirst = `title: Foobar
-id: ${crypto.randomUUID()}
+id: ${rule_uuid}
 description: Detects suspicious FOOBAR
 references:
   - https://


jkppr (approved):
lgtm

jaegeral:
Can you provide an example output of the method you are suggesting?

Does it match the requirements https://github.com/SigmaHQ/sigma/wiki/Rule-Creation-Guide#id / https://en.wikipedia.org/wiki/Universally_unique_identifier ?

jnimmo:
Sounds like the only requirement for the ID is that it is unique, with a random UUID being one way of achieving that - a random number or string would also suffice.

If Timesketch doesn't utilise the UUID elsewhere (I'll go and have a look) then could also potentially just skip the ID field altogether as it is an optional field. But certainly the getRandomValue has been working in my testing.

jaegeral:
Yes that is true. The id field is optional and while there are no strict requirements. If you lookup the common practice used in the sigma GitHub repo, they all follow the uuidv4 format.

The UUID itself is referred in many places in timesketch (but again the format is not enforced and or checked, but I still want to keep it consistent.

With this pr if someone has issues he at least gets one UUID (that is static) and if the user tries to use the same UUID a second time, an error message will be shown as the UUID must be unique per rules installed on the timesketch server.

PR #2471, opened 2022-12-29
Add share dialog

### LINKED ISSUE #2479: Sharing dialog

berggren:
This PR ports over the sharing dialog to the new UI. It supports adding users and groups, as well as setting the general access mode for the sketch.

Screen Shot 2022-12-29 at 11 57 13

closes #2479


timesketch/api/v1/resources/user.py
```diff
@@ -186,7 +186,9 @@ def post(self, sketch_id):
                 sketch.revoke_permission(permission=permission, group=group)
 
         public = form.get("public")
-        if public == "true":
+        # TODO: Remove string check. Non-pythonic check is needed because the old UI
+        # returns a string of true or false and not a boolean.
+        if public is True or public == "true":
```
jaegeral, line 191: is there a chance that this would be `True` instead of `true`? If so we might want to check for `public.toLower() == "true"` WDYT?

timesketch/frontend-ng/src/components/ShareCard.vue
```diff
@@ -0,0 +1,237 @@
+<!--
+Copyright 2022 Google Inc. All rights reserved.
```
jaegeral, line 2: 2023 ;)

timesketch/frontend-ng/src/components/ShareCard.vue
```diff
@@ -0,0 +1,237 @@
...
+      let userToRemove = [userName]
+      let groupToRemove = [groupName]
+      ApiClient.revokeSketchAccess(this.sketch.id, userToRemove, groupToRemove)
+        .then((response) => {
+          this.$store.dispatch('updateSketch', this.sketch.id).then(() => {
+            this.isLoading = false
+          })
+        })
+        .catch((e) => {
+          this.errorSnackBar('Failed to remove access ' + e)
```
jaegeral, line 178: I suggest we also add the sketchid, user and or group to that error message so that it is clear to the user what was attempted

jaegeral (commented):
Some comments

timesketch/api/v1/resources/user.py
berggren: No, this is set from the backend and not by the user. It hasn't changed in years. I opt to keep this as is because we are deprecating the old stuff soon.

timesketch/frontend-ng/src/components/ShareCard.vue
berggren: Well, the PR is from before the new year :) So technically it was written in 2022.

timesketch/frontend-ng/src/components/ShareCard.vue
berggren: The sketchID is clear to the user already (they are in the sketch). And the user/group should be clear as well because the can only remove one at a time. The error message returned should be informative for the end user.

I opt to keep it like this.

jaegeral:
Lgtm. Need to see if I can click the approve button from mobile

jaegeral (commented):
Please fix the conflicts

jaegeral (approved):
LGTM

PR #2502, opened 2023-01-04
admin-cli analyzer-stats function

### LINKED ISSUE #2500: Analyzer stats in tsctl
To see which rules are performing how, a tsctl module could be useful to get a stats overview of Sigma analyzer runs and how long each one took and maybe filter on faulty ones.

jaegeral:
New module in tsctl to troubleshoot analyzer runs `tsctl analyzer-stats`.
* filter on analyzer name
* select a focus out of `[many_hits, long_runtime, recent]`
* search for a given text in the analyzer results. This comes handy when searching for a Sigma Rule Id.
* filter results on timeline id

Example:
```shell
tsctl analyzer-stats --help
Usage: tsctl analyzer-stats [OPTIONS] ANALYZER_NAME

  Prints sigma analyzer stats.

Options:
  --timeline_id TEXT         Timeline ID if the analyzer results should be
                             filtered by timeline.
  --focus TEXT               Focus on: [many_hits, long_runtime, recent]
  --result_text_search TEXT  Search in result text. E.g. for a specific
                             rule_id.
  --help                     Show this message and exit.
tsctl analyzer-stats sigma --focus many_hits --result_text_search 71a52
     runtime  hits                                                                          result  analysis_id                 created_at
36  0.083333  3657  3657 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           51 2023-01-03 21:33:14.475700
37  0.083333  3657  3657 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           52 2023-01-03 21:33:40.477309
39  0.083333  2344  2344 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           54 2023-01-03 21:33:40.594573
38  0.000000   145   145 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           55 2023-01-03 21:33:40.626931
47  0.100000   145   145 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           61 2023-01-04 17:09:04.078641
40  0.000000     0     0 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           53 2023-01-03 21:33:40.521002
41  0.000000     0     0 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           56 2023-01-03 21:33:40.658743
42  0.000000     0     0 events tagged for rule [Scheduler] (71a5257c-222f-4898-a117-694d6c63457c)           57 2023-01-03 21:33:40.696942
43  2.800000     0                                * Scheduler 71a5257c-222f-4898-a117-694d6c63457c           58 2023-01-04 17:09:03.751176
44  2.816667     0                                * Scheduler 71a5257c-222f-4898-a117-694d6c63457c           62 2023-01-04 17:09:04.112822
45  2.833333     0                                * Scheduler 71a5257c-222f-4898-a117-694d6c63457c           60 2023-01-04 17:09:04.046003
46  2.833333     0                                * Scheduler 71a5257c-222f-4898-a117-694d6c63457c           59 2023-01-04 17:09:04.014973
48  2.750000     0                                * Scheduler 71a5257c-222f-4898-a117-694d6c63457c           63 2023-01-04 17:09:04.148185
```

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #2500


docs/guides/admin/admin-cli.md
```diff
@@ -570,3 +570,45 @@ On instance['context_link']:
...
+```shell
+tsctl analyzer-stats --help
+Usage: tsctl analyzer-stats [OPTIONS] ANALYZER_NAME
+
+  Prints analyzer stats.
+
+Options:
+  --timeline_id TEXT         Timeline ID if the analyzer results should be
+                             filtered by timeline.
+  --focus TEXT               Focus on: [many_hits, long_runtime, recent]
```
berggren, line 595: "Focus" is not very intuitive naming IMO. What about `--scope` instead?

berggren (changes requested):

docs/guides/admin/admin-cli.md
jaegeral: good point, changed it

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+    "--result_text_search",
+    required=False,
+    help="Search in result text. E.g. for a specific rule_id.",
+)
+@click.option(
+    "--limit",
+    required=False,
+    help="Limit the number of results.",
+)
+def sigma_stats(analyzer_name, timeline_id, scope, result_text_search, limit):
```
jkppr, line 650: Should we keep the command and the function name consistent? The command name is `analyzer-stats` but the function called is `sigma_stats`.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+    """Prints sigma analyzer stats."""
```
jkppr, line 651: To my understanding this function shows stats for all analyzers. Please update the doc string accordingly.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
             print(f'=> OK: "{entry}"')
         except (ValidationError, SchemaError) as err:
             print(f'=> ERROR: "{entry}" >> {err}\n')
+
+
+# Analyzer stats cli command
+@cli.command(name="analyzer-stats")
+@click.argument("analyzer_name")
```
jkppr, line 629: Since you support an "all" keyword below, maybe this should be mentioned in the help for the command?

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+def sigma_stats(analyzer_name, timeline_id, scope, result_text_search, limit):
+    """Prints sigma analyzer stats."""
+
+    if timeline_id:
+        timeline = Timeline.query.get(timeline_id)
+        if not timeline:
+            print("No timeline found with this ID.")
+            return
+        analysis_history = Analysis.query.filter_by(
+            timeline=timeline, analyzer_name=analyzer_name
```
jkppr, line 659: If I run the command with the parameters `--timeline_id X all` I would expect it to return all analysis that was run on this timeline. But with this query it will try to find an analysis with the name "all" on the timeline, what is a different behaviour than expected with "all" in the check below.
I would recommend to add another check for all in here and adjust the query accordingly.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+        ).all()
+    elif analyzer_name == "all":
+        analysis_history = Analysis.query.filter_by().all()
+    else:
+        # analysis filter by analyzer_name is sigma
+        analysis_history = Analysis.query.filter_by(analyzer_name=analyzer_name).all()
```
jkppr, line 665: If the analyzer name does not exist, the script crashes. You might want to catch that or check if it exists in the first place.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+
+    df = pd.DataFrame()
+    for analysis in analysis_history:
+        if analysis.analyzer_name in [analyzer_name, "all"]:
```
jkppr, line 669: If I get your intention here, you want to limit the regex search below to the analyzers that are requested by the user. However, I don't think this works as intended: `analysis.analyzer_name` will contain the names of the analyzers collected above. So there will be no "all" and if the filter above worked, it will be all already filtered. Resulting in all queries using the "all" keyword for the analyzer_name are empty instead of listing all.

I would recommend to remove this check and improve the filters above when collecting the list of analysis.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+            # extract number of hits from result to a int so it could be sorted
+            try:
+                matches = re.search(r"\d+(?=\s+events)", analysis.result)
+            except TypeError:
+                matches = 0
+            numbers = 0
+            if matches:
+                numbers = int(matches.group())
```
jkppr, line 677: You could do the int conversion already in line 672 above and then have matches as resulting integer only. e.g.
```
try:
  matches = int(re.search(r"\d+(?=\s+events)", analysis.result).group())
except TypeError:
  matches = 0
```

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+            new_row = pd.DataFrame(
+                [
+                    {
+                        "runtime": analysis.updated_at - analysis.created_at,
+                        "hits": numbers,
+                        "result": analysis.result,
+                        "analysis_id": analysis.id,
+                        "created_at": analysis.created_at,
```
jkppr, line 685: Since there is the option to list all analyzers with no filter on the timeline, I would recommend to include a field containing the timeline ID here as well.

timesketch/tsctl.py
```diff
@@ -616,3 +622,91 @@ def validate_context_links_conf(path):
...
+                    }
+                ]
+            )
+            df = pd.concat([df, new_row], ignore_index=True)
+
+    # make the runtime column to only display in minutes and cut away days etc.
+    df["runtime"] = df["runtime"].dt.seconds / 60
```
jkppr, line 692: If the df is empty, e.g. because there was no analysis yet (or because the "all" keyword does not work) it will crash, since `df["runtime"]` does not exist. You might want to catch this or ensure further up that results exist.

jkppr (changes requested):
Cool feature! There are still some bugs that need to be addressed before we can merge it. Please see the comments below.

timesketch/tsctl.py
jaegeral: good point

timesketch/tsctl.py
jaegeral: true, I missed that aspect, fixing it

timesketch/tsctl.py
jaegeral: yep now it should be better, good catch

timesketch/tsctl.py
jaegeral: good point, added

timesketch/tsctl.py
jaegeral: good point, will return if `if df.empty:`

timesketch/tsctl.py
```diff
@@ -639,3 +645,99 @@ def validate_context_links_conf(path):
...
+        analysis_history = Analysis.query.filter_by().all()
+    else:
+        # analysis filter by analyzer_name
+        analysis_history = Analysis.query.filter_by(analyzer_name=analyzer_name).all()
+
+    df = pd.DataFrame()
+    for analysis in analysis_history:
+        # extract number of hits from result to a int so it could be sorted
+        try:
+            matches = int(re.search(r"\d+(?=\s+events)", analysis.result))
```
jkppr, line 701: Your approach to count "hits" is limited to only some analyzers. Mainly the sigma ones I guess. Since the result string is custom made by each analyzer author, it might declare "hits" differently (e.g. see the domains analyzers). 
I would recommend to count and display hits only for the sigma analyzer for now, since it is the most useful there.

timesketch/tsctl.py
```diff
@@ -639,3 +645,99 @@ def validate_context_links_conf(path):
...
+        except TypeError:
+            matches = 0
+        new_row = pd.DataFrame(
+            [
+                {
+                    "runtime": analysis.updated_at - analysis.created_at,
+                    "hits": matches,
+                    "timeline_id": analysis.timeline_id,
+                    "result": analysis.result,
```
jkppr, line 710: Since the result string can be arbitrarily long, I would recommend to place it to the end of the table/columns.

jkppr (changes requested):
Just two more nits.

timesketch/tsctl.py
jaegeral: done

timesketch/tsctl.py
jaegeral: adding a huge chunk of "if analyzer_name == sigma" seems to go in the wrong direction imho, then a dedicated sigma analyzer stats module might be better, I think this is fine for now, it will show 0 hits for analyzers that do not produce the certain regex based hits calculation, I added a todo in the code so it is clearer.

WDYT?

jkppr: As discussed: We will drop the hits if analyzer != sigma for now. This prevents an output where it says "hits=0" but in the results it says "250 hashes tagged".

jkppr (approved):
Looks good to me now :)

berggren (approved):

PR #2507, opened 2023-01-08
Return user defined aggregation name when using the API client

### LINKED ISSUE #2503: Inconsistencies in the Aggregation section of the Timesketch API
Some inconsistencies in the TS Python API when interacting with aggregation. This was something that I found when calling [list_aggregations](https://github.com/google/timesketch/blob/1cef69b041c488878f062ce9e515313aa05f96cd/api_client/python/timesketch_api_client/sketch.py#L545), and was confused as to why the `.name` attribute of the returning objects was not the one returned by the API (`"sketches/{self.id}/aggregation/"`). It seems the `name` and `description` of the Aggregation object are extracted from `resource_data`, which is populated by [from_saved](https://github.com/google/timesketch/blob/1cef69b041c488878f062ce9e515313aa05f96cd/api_client/python/timesketch_api_client/aggregation.py#L136).

So this goes like this:

1. list_aggregations, query to `sketches/{self.id}/aggregation`
2. At this point, I'd expect the funciton to return these results, but it goes on
3. `from_saved` is called for each aggregation (which I expected loaded pre-cached results from a database), calling another API endpoint, presumably to get fresh aggreagtor data. This is the data containing the right name and description for the aggregation.
4. `from_saved` calls `_run_aggregator` (which runs the aggregation once more?) and stores the result in the resource_data attribute of the `self` object. (maybe erroneously, instead of doing this with the results of the previous API call?)
5. When `list_aggregations` returns, the objects it returns are a list of results of _run_aggregator, which don't contain the expected data.

sydp:
* Updates the name attribute to reflect the user-defined name value.
* Adds a new attribute ```aggregator_name``` which holds the value previously held by ```name```.
* Adds unit tests for:
  * list_aggregations in the Sketch api client
  * Aggregation api client

New test data for aggregation is based on responses from the REST API using test data and aggregations.

Closes #2503 

+ What existing problem does this PR solve? #2503
+ What new feature is being introduced with this PR? Adds some unit tests for list_aggregations in the sketch python api

**Checks**

- [ ] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

closes #2503


api_client/python/timesketch_api_client/aggregation_test.py
```diff
@@ -0,0 +1,184 @@
+# Copyright 2020 Google Inc. All rights reserved.
```
berggren, line 1:
```suggestion
# Copyright 2023 Google Inc. All rights reserved.
```

berggren (approved):

PR #2511, opened 2023-01-11
SSH Brute Force Analysis

roshanmaskey:
This PR adds `SSHBruteForcePlugin` (`ssh_bruteforce.py`) that analysis SSH authentication events and provides the summary of brute force analysis report and automatically tags the relevant events. 

**SSHBruteForcePluing Result**

The figure below shows the result of the analyzer.

sshbruteforceplugin_result

**Labeled Events**

The figure below shows relevant SSH events labeled and stared.

sshbruteforceplugin_label

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


berggren:
Ack, I will review asap.

berggren:
Hey @roshanmaskey good stuff! I started looking at this today, and I have a couple of higher level thoughts.

- Maybe consider refactor to a proper plugin system for authentication analyzers (see the chain analyzers for example)
- The use of pyparsing - I'm wondering if we already have some of this data extracted via plaso? And if not, should we add it there?

I'll set up a sync for us together with @jkppr to progress this asap.

berggren:
Adding @jkppr to review from the analysis side

jkppr (commented):
As discussed offline: This PR will get some more work by @roshanmaskey with new ideas and approaches from the last month.
Main points are:
* Update the demo data to be compatible with the newest Timesketch/Plaso version
* If possible make use of the `syslog` plaso parser to extract the necessary fields. This would allow to skip the usage of pyparsing.
* If field  parsing is necessary, check out the feature extractor function to take care of this before running your analyzer.

roshanmaskey:
@jkppr ready for review.

timesketch/lib/analyzers/auth.py
jkppr, line 1: Let's discuss in our next 1:1 the pro & cons of building this as an analyzer plugin system.

timesketch/lib/analyzers/ssh_bruteforce_test.py
```diff
@@ -0,0 +1,124 @@
+# -*- coding: utf-8 -*-
```
jkppr, line 1: Not needed in python 3 anymore.

timesketch/lib/analyzers/ssh_bruteforce.py
```diff
@@ -0,0 +1,388 @@
+# -*- coding: utf-8 -*-
```
jkppr, line 1: Not needed in python 3 anymore.

jkppr (changes requested):
First round of comments.
I need a bit more time to think about the usage of pyparsing and if we could archive the same goal using the feature extractor. I will come back to this in round 2 then :)

PR #2521, opened 2023-01-19
Remove algorithm taken from variable

### LINKED ISSUE #2522: GOOGLE_OIDC_ALGORITHM leads to None variable
**Describe the bug**
Using the config variable `GOOGLE_OIDC_ALGORITHM=None` leads to no alogrithm value set. Besides that I could not find a reason why this variable is used in a function that is made for the `GOOGLE_IAP_*` variables.

**To Reproduce**
1. Start Timesketch on GCP with the variables `GOOGLE_IAP_*` set 
2. Open the Timesketch URL
3. Login on the IAP
4. See logs `timesketch.app/ERROR JWT validation error: The specified alg value is not allowed`

**Expected behavior**
1. Start Timesketch on GCP with the variables `GOOGLE_IAP_*` set 
2. Open the Timesketch URL
3. Login on the IAP
4. Succesfull login to Timesketch

Zawadidone:
Removes unnecessary code.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

Closes #2522


Zawadidone:
Could the failing of the job be related to this PR? Because I do not have the permission to retry the job.

jaegeral:
There seems to be a linter issue with black formatter. If you make a new commit to this pr, the jobs will re- run.

Zawadidone:
@jaegeral the checks have passed

jaegeral:
Good. Looking at the pr I am not sure why you want to suggest to remove the usage of the algorithm from the config and if none is present to use the one in the argument?

Zawadidone:
The old code used the variable `GOOGLE_OIDC_ALGORITHM ` but this variable has nothing to do with the IAP, from how I understand the code, the variable `GOOGLE_IAP_ALGORITHM` should be used. This variable is used as argument when calling the function `decode_jwt`.

berggren:
GOOGLE_OIDC_ALGORITHM and GOOGLE_IAP_ALGORITHM are two different authn methods. IAP will set it's algo in the calling code. IAP uses OIDC as it's authn method.

I'll opt to keep this as is for now.

berggren:
Looking at this one more time, there is an issue here and I see that we need to address it. It is a bit more involved than this PR though.

@Zawadidone thanks for looking into this! Let's continue the discussion on the bug #2522

PR #2523, opened 2023-01-23
Update features.yaml

itsmvd:
Bugfix for a regex that was capturing more than intended


tomchop:
To reduce ambiguity (and avoid future errors like this), could we consider using a named capture group?

https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups

berggren (approved):

PR #2529, opened 2023-01-31
Updates to the timesketch.org docs

jkppr:
Updating the following docs:
- https://timesketch.org/developers/getting-started/
  - Minor update to include the usage of `tsdev.sh` for easier dev env setup.
- https://timesketch.org/developers/analyzer-development/
  - Reworked the whole guide to get developers started using `l2tscaffolder` and up to a point where they can start adding their own analyzer logic.

### DIFF
--- a/docs/developers/analyzer-development.md
+++ b/docs/developers/analyzer-development.md
@@ -2,119 +2,342 @@
 hide:
   - footer
 ---
-# Write analyzers in Timesketch
+# Write analyzers for Timesketch
 
-## Multi Analyzer
+## Overview
 
-When you develop an analyzer that would benefit from creating smaller sub-jobs, you should use Multi Analyzer.
+This guide provides all steps to get you started with building a new Timesketch analyzer.
 
-For example The Sigma analyzer is such a Multi Analyzer. That means, the Sigma analyzer is calling `get_kwargs()` from [sigma_tagger.py](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/sigma_tagger.py). That will return a list of all Sigma rules installed on the instance. The Main celery job then spawns one celery job per Sigma rule that can run in parallel or serial depending on the celery config and sizing of the Timesketch instance.
+## Background
 
-If `get_kwargs()` is not implemented in the analyzer, [tasks.py](https://github.com/google/timesketch/blob/master/timesketch/lib/tasks.py) expects it is not a multi analyzer, thus creating only one celery job.
+Timesketch analyzers are programs that run when new data is indexed, e.g. when
+you upload a new plaso storage file or when adding an existing index to a
+sketch. You have access to a simple API that makes searching, commenting,
+tagging etc easy. Everything you can do in the UI you can do programmatically.
 
-## analyzer_run.py
+## Analyzers
 
-### Purpose
+An analyzer can be auto-run on every new timeline that is added to a sketch or
+manually triggered by the analyst. If you read somewhere about "sketch" and
+"index" analyzers, this info is deprecated. There is now only one type of
+analyzer.
 
-`analyzer_run.py` is a standalone python script made to bootstrap the
-development workflow to a minimum where only a file with events and
-a class file with your analyzer code is needed.
-You do not have to install Timesketch or any docker for that.
+An analyzer can search, correlate and enrich data in each given timeline. The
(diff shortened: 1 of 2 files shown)


docs/developers/analyzer-development.md
```diff
@@ -2,119 +2,342 @@
...
+
+### 1) Set Up your development environment
+
+Follow the instructions in
+[Timesketch development - Getting started](https://timesketch.org/developers/getting-started/)
+to get everything up and running.
+
+### 2) Generate the necessary analyzer templates
+
+Alright, you got your development environment up, great! The first thing we are
```
jaegeral, line 121: IMHO; it is grat to use "you" etc in the doc, but that sounds a bit to "Umgangssprachlich"

docs/developers/analyzer-development.md
```diff
@@ -2,119 +2,342 @@
...
+## Overview
 
-When you develop an analyzer that would benefit from creating smaller sub-jobs, you should use Multi Analyzer.
+This guide provides all steps to get you started with building a new Timesketch analyzer.
 
-For example The Sigma analyzer is such a Multi Analyzer. That means, the Sigma analyzer is calling `get_kwargs()` from [sigma_tagger.py](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/sigma_tagger.py). That will return a list of all Sigma rules installed on the instance. The Main celery job then spawns one celery job per Sigma rule that can run in parallel or serial depending on the celery config and sizing of the Timesketch instance.
+## Background
 
-If `get_kwargs()` is not implemented in the analyzer, [tasks.py](https://github.com/google/timesketch/blob/master/timesketch/lib/tasks.py) expects it is not a multi analyzer, thus creating only one celery job.
+Timesketch analyzers are programs that run when new data is indexed, e.g. when
```
jaegeral, line 13: Analyzers could also be run after the data is imported

docs/developers/analyzer-development.md
```diff
@@ -2,119 +2,342 @@
 hide:
   - footer
 ---
-# Write analyzers in Timesketch
+# Write analyzers for Timesketch
 
-## Multi Analyzer
+## Overview
 
-When you develop an analyzer that would benefit from creating smaller sub-jobs, you should use Multi Analyzer.
```
jaegeral, line 9: as far as I can tell your update would remove the whole info about multi vs non multi analyzer, is that by purpose?

docs/developers/getting-started.md
```diff
@@ -42,19 +42,23 @@ timesketch-dev         | User dev created/updated
...
+### Web server
+
+Now, start the `gunicon` server that will serve the Timsketch WSGI app.
+
+To make this task easier, we recommend using the `timesketch/contrib/tsdev.sh` script.
 
 In one shell:
 
 ```bash
-$ docker-compose exec timesketch gunicorn --reload -b 0.0.0.0:5000 --log-file - --timeout 120 timesketch.wsgi:application
```
jaegeral, line 57: how about we keep both? I know tsdev is maybe easier, but having the raw command here seems still useful to understand what happens.

docs/developers/getting-started.md
```diff
@@ -76,10 +80,12 @@ are not picked up by the Gunicorn servers but by **Celery workers**.
 If you're planning to work on those (or even just import timelines into your Timesketch instance), you'll need to launch
 a Celery worker, and re-launch it every time you bring changes to its code.
 
+You can use `timesketch/contrib/tsdev.sh` for this task as well.
+
 In a new shell, run the following:
 
 ```bash
-$ docker-compose exec timesketch celery -A timesketch.lib.tasks worker --loglevel info
```
jaegeral, line 82: same as above, I opt to keep both

jaegeral (changes requested):

docs/developers/analyzer-development.md
jkppr: Point taken :D

docs/developers/analyzer-development.md
jkppr: Good point. Will add the info here.

docs/developers/getting-started.md
jkppr: Yeah, why not :)

berggren (approved):

PR #2532, opened 2023-02-02
UI consistency and API resource naming

berggren:
Fixes to add consistency to UI elements and API resources.
- API resource naming for searchtemplates and sigma rules (frontend and api client updated)
- Dense list in drop down menus
- Reduced left panel width
- various smaller UI fixes 
- Bring back old Graph code that was accidentally removed in a previous PR. Without it we cannot build the legacy UI
- New UI builds


jaegeral (approved):
LGTM

PR #2535, opened 2023-02-06
Adjust generate timeline feature

### LINKED ISSUE #2356: Filter on field when generating timeline from index
When using the function `generate_timeline_from_es_index` I would like to add an argument based on a single key value pair. 
[https://github.com/google/timesketch/blob/40afb16415dc6ff67791472e56b13f0a0fef894c/api_client/python/timesketch_api_client/sketch.py#L1657](https://github.com/google/timesketch/blob/40afb16415dc6ff67791472e56b13f0a0fef894c/api_client/python/timesketch_api_client/sketch.py#L1657)
[https://timesketch.org/developers/api-upload-data/#import-data-already-ingested-into-opensearch](https://timesketch.org/developers/api-upload-data/#import-data-already-ingested-into-opensearch)

This values can be used in the Query DSL to filter documents based on the key value pair. For example `ID=1`.
[https://github.com/google/timesketch/blob/b7c7f428069a19564ceea7f31a2e19047848e855/timesketch/api/v1/resources/event.py#L1000](https://github.com/google/timesketch/blob/b7c7f428069a19564ceea7f31a2e19047848e855/timesketch/api/v1/resources/event.py#L1000)

This adds the following features to Timesketch when adding dat already ingested in to Opensearch:

* Prevents creation of 1500 indices limit in Opensearch
* This way multiply timelines can be genereated from a single index based on an unique ID suplied by the user. For example documents with the field `ID` is `1` or `2` are saved in in the same index. Using the key value pair `ID=1` and `ID=2` both timelines can be generated in a single index

Using a single index for a single timeline which is not compatible with 1500+ timelines.
[https://timesketch.org/guides/admin/scaling-and-limits/#opensearch-indices-limitation](https://timesketch.org/guides/admin/scaling-and-limits/#opensearch-indices-limitation)

Zawadidone:
The problem is that it is not possible to use a single index and generate two different timelines from it. Because the current implementation of the generate timeline feature is to add the `__ts_timeline_id` to every document in an index.

This leads to all documents having for example the `__ts_timeline_id=1` while there also documents that later will adjusted to `__ts_timeline_id=2`.

This PR adds the following features
* The field `__ts_timeline_filter_id` is used to filter timelines in a single index. 
  1. Index has 10 documents (5 with the field `__ts_timeline_filter_id=92D58BD6-332B-4C09-A989-8AE9623C3CCD` and 5 with the field `__ts_timeline_filter_id=AE706B1C-5014-4ED8-B11E-436005016985`)
  2. timeline 1 -> `generate_timeline_from_es_index([...], timeline_filter_id="92D58BD6-332B-4C09-A989-8AE9623C3CCD")`
  3. timeline 2 -> `generate_timeline_from_es_index([...], timeline_filter_id="AE706B1C-5014-4ED8-B11E-436005016985")`
  4. As result two timelines are added to Timesketch with both 5 events
* Because two timelines can be used by a single index the Explore page shows incorrect results. It counts the result of a query based on the amount of documents of an index and not the amount of results per timeline. https://github.com/google/timesketch/pull/2535/files#diff-d5019e2491cea14cac3f9dbeeaeff4a39f1dd13b6173944a1bba9c28a05d291cR283

Closes #2356 

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

Closes #2356


Zawadidone:
The previous PR's were created incorrectly

berggren:
Thanks for contributing. It is difficult to understand what the goal is for this PR. There are some substantial changes to critical functionality.

Can you explain what problem you are trying to solve?

berggren (changes requested):
Please update problem statement in the PR

Zawadidone:
Problem statement is adjusted

PR #2539, opened 2023-02-07
make tests into dedicated test cases

jaegeral:
Some tests in Sigma util test where bunched together and back in the day I agreed to split them into different test cases

closes#2329


jkppr (approved):
LGTM

PR #2542, opened 2023-02-08
[SigmaStudio]don't mutate the prop

jaegeral:
Small fix in Sigma Studio to not mutate a prob and adding a sample tag to the template, otherwise everyone using a template gets a error by default

closes #2542


jkppr (approved):
LGTM

jaegeral:
https://github.com/google/timesketch/issues/2541

PR #2546, opened 2023-02-09
Serve icons from local bundle

berggren:
In preparation for swithing to the new UI as default.
This PR removed the dependency on CDN delivery of mdi icons for the new UI. It bundles it in the app instead.

For devs: you need to do a `yarn add @mdi/font` to get them into your dev build


jaegeral:
> For devs: you need to do a `yarn add @mdi/font` to get them into your dev build

Should we add this somewhere in the docs as well?

jaegeral (approved):
LGTM, but I would recommend we add something in the dev docs, otherwise it would be missed by devs

PR #2551, opened 2023-02-10
fix: PR #2471 breaks import paths of some components

renzhexigua:
The relative import paths need to change as well, @berggren may forget it in PR #2471 

image


berggren:
Good catch! This was a mishap where the old Graph code was accidentally moved. This has been fixed since.

PR #2576, opened 2023-02-16
2550 2023 02 10 sigma parsing issue

### LINKED ISSUE #2550: Sigma parsing error with stars
There seem to be a problem in https://github.com/google/timesketch/blob/master/timesketch/lib/sigma_util.py#L124 when having a mixture of terms surrounded by "" and ** that the first word is always getting "" around them.

- [x] test cases
- [x] fix the issue

jaegeral:
Update test cases and the sanitise method.

Closes #2550


berggren:
What is the status of this?

jaegeral:
Need to fix the linters then it can be reviewed

jaegeral:
ok I will close this for the moment, maybe I will come back to it later....

Issue #1226, opened 2020-05-26
Keyboard shortcuts for timeline navigation


tomchop:
Pg Up / Pg Down to to to previsous / next page
Cmd / Ctrl + G to go to a specific date.

berggren:
Closing in favour of #214

berggren:
Potentially useful:
https://github.com/iFgR/vue-shortkey/blob/master/README.md

jaegeral:
Some more ideas for useful commands:

* CMD + * to star an event
* CMD + + or A to add a manual event
*

jaegeral:
This feature request is currently on hold since the initial vue component we wanted to use is not updated in a long time, eventually we will write our own wrapper, but that is not highest prio at the moment

PR #2577, opened 2023-02-17
Keyboard shortcuts

jaegeral:
Adds a Keyboard shortcut help page and related shortcuts to Timesketch
https://github.com/google/timesketch/issues/1226

Screenshot 2023-03-20 at 17 13 47

Note: there is no button or similar at the moment that would open the help menu, so a user must "know" it exists, this is for the moment to roll it out silently, but we could also consider just adding a button in the explore page.

Previous iterations:

Keyboart_shortcut_help

Keyboard Shortcut help

keyboard-shortcuts

shortcut


timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,22 @@ limitations under the License.
...
+    <!-- rounded vue element that goes over everything showing the shortcut help -->
+    <v-overlay v-if="showShortcutHelp" width="auto" v-model="dialog">
+      <template
+        >Control N: Next page<br />
+        Control P: Previous page<br />
+        Control T: Toggle Search History<br />
+        Control H: Toggle Shortcut help<br />
+      </template>
+    </v-overlay>
+    <!-- some invisible buttons that are used to trigger the shortcuts -->
```
jaegeral, line 272: this might need some more love and ideas on how to hide that

jaegeral:
Had a conversation with tomchop on this, as for now we mostly will do it in the Explore page (vs. general shortcuts) we will keep them in the main Explore component. If we find ourself using shortcuts in other components a lot, we can refactor kind of easy.

The downside of a component from the beginning is that we would have to pass variables around to understand from where it is called and then display context aware shortcuts.

jaegeral:
Screenshot 2023-03-03 at 13 25 32

jaegeral:
Removing @berggren for now from the reviewers to get @jkppr review in first and then go back to berggren

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -586,6 +612,77 @@ export default {
         this.meta.filter_labels.push(label)
       }
     },
+    toggleshowShortcutHelp: function () {
+      this.showShortcutHelp = !this.showShortcutHelp
+    },
+    nextpage(event) {
+      this.tableOptions.page++
```
jkppr, line 619: Since Johan moved the EventList into its own component, there is no `tableOptions` in this Explorer.vue component anymore. Hence this function throws an error.
I would recommend moving this feature to the EventList.vue component.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -586,6 +612,77 @@ export default {
         this.meta.filter_labels.push(label)
       }
     },
+    toggleshowShortcutHelp: function () {
```
jkppr, line 615: Since we moved the right side views into the canvas component the explorer component is loaded even when other components like stories, TI or analyzers are visible. To prevent the shortcuts from triggering on the other components I would recommend to wrap them in something like this:
`if (this.currentRouteName === 'Explore') `
Where `currentRouteName` is defined in the computed section:
```
currentRouteName() {
  return this.$route.name
}
```

jkppr (commented):
This is a great feature! Very useful for quick timeline navigation. 
Please see the comments below for bugs and ux feedback.

One general topic: Having the shortcut help dialog only displayed when someone already knows the shortcut is not optimal from an ux point of view. I think we need a `mdi-help-circle-outline` icon somewhere that would trigger this dialog. We could also include more basic information on how to search in timesketch. One option would be adding the (?) in the top right or inside the search bar or in the three dot menu in the top right. Lets discuss this offline.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -245,9 +245,9 @@ limitations under the License.
     </v-card>
 
     <!-- DFIQ context -->
-    <div class="mt-3 mx-3">
-      <div :class="[$vuetify.theme.dark ? 'dark-info-card' : 'light-info-card']" v-if="activeContext.question">
-        <v-toolbar dense flat color="transparent">
+    <div class="mt-3 mx-3" :class="[$vuetify.theme.dark ? 'dark-info-card' : 'light-info-card']">
```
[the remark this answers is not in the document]
jaegeral: nope, let me revert that.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,31 @@ limitations under the License.
       </div>
     </div>
 
+    <!-- write a v-dialog here that shows the shortcut help -->
+    <v-dialog v-model="showShortcutHelp" width="700" style="overflow: visible">
```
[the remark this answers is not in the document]
jaegeral: Reason for this was to mimic the style from https://github.com/google/timesketch/blob/master/timesketch/frontend-ng/src/components/UploadForm.vue#L18 
and
https://github.com/google/timesketch/blob/master/timesketch/frontend-ng/src/components/Explore/AddManualEvent.vue#L17
that when we do a overlay, it has a similar size.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,31 @@ limitations under the License.
...
+      <v-card width="700" style="overflow: visible" hover="true">
```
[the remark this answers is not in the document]
jaegeral: good point, will remove

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,31 @@ limitations under the License.
...
+        <v-card-title> Keyboard Shortcut help </v-card-title>
+
+        <v-card-text v-shortkey="['ctrl', 'n']" @shortkey="nextpage()"> Control N: Next page </v-card-text>
```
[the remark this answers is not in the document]
jaegeral: oh yes that looks way nicer

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,31 @@ limitations under the License.
...
+
+        <v-card-text v-shortkey="['ctrl', 'p']" @shortkey="previouspage()"> Control P: Previous page </v-card-text>
+
+        <v-card-text v-shortkey="['ctrl', 't']" @shortkey="toggleSearchHistory()">
```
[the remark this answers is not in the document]
jaegeral: moved it to Command + shift + button

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,31 @@ limitations under the License.
...
+        <v-card-actions>
+          <v-spacer></v-spacer>
+          <v-btn text color="primary" @click="toggleshowShortcutHelp">Close</v-btn>
+        </v-card-actions>
+      </v-card>
+    </v-dialog>
+
+    <!-- one button is needed to activate the Shortcuthelp -->
+
+    <button v-shortkey="['ctrl', 'h']" @shortkey="toggleshowShortcutHelp()"></button>
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/frontend-ng/src/views/Explore.vue
jaegeral: good point, done

timesketch/frontend-ng/src/views/Explore.vue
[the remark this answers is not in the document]
jkppr: When I check the dialogs for the "UploadForm" and "AddManualEvent" they have a v-card width of 700 in the code but are actually using a dialog with 1000px and 600px width (so not the same) which is actually just the space they need without resizing during the process.
The add manual event dialog size is set here for example: https://github.com/google/timesketch/blob/master/timesketch/frontend-ng/src/views/Explore.vue#L135

So, given that we have don't have one size dialog for all I would still recommend going with a dialog of the size that is needed only. Which is now probably `width="350"` after the changes.

timesketch/frontend-ng/src/views/Explore.vue
jkppr: I would recommend to do this check for all shortcuts and just trigger them on the page that is relevant. 
e.g. at the moment I still can trigger the shortcut help dialog on a story page, but there is actually no shortcut relevant for stories.
I also still can trigger next- & prev-page when looking at a story or graph. And it will paginate in the background even I'm not looking at the table. This is unexpected behaviour for me and I would recommend to make shortcuts only available when someone is viewing the element they are affecting.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,64 @@ limitations under the License.
...
+        </v-card-text>
+        <v-card-text>
+          <v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> Ctrl </v-btn>
+          +
+          <v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> Shift </v-btn>
+          +
+          <v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> h </v-btn>
+          = Toggle Shortcut help
+        </v-card-text>
+        <v-card-text v-shortkey="['ctrl', 'shift', 'm']" @shortkey="addEventBtn()">
```
jkppr, line 302: This function `addEventBtn()` is still missing. Either implement the feature or mark the shortcut as not active yet.

timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -260,6 +260,64 @@ limitations under the License.
...
+    <v-dialog v-model="showShortcutHelp" width="700" style="overflow: visible">
+      <v-card width="700" style="overflow: visible">
+        <v-card-title> Keyboard Shortcut help </v-card-title>
+
+        <v-card-text
+          ><v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> Ctrl </v-btn>
+          +
+          <v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> Shift </v-btn>
+          +
+          <v-btn style="text-transform: unset; cursor: unset" :ripple="false" x-small density="compact"> n </v-btn>
```
jkppr, line 273: Small optional nit: Since you are now using shift in the shortcut it would be fine to use uppercase letters again, since shift does make them uppercase.

jkppr (changes requested):
Please have a look at the added comments on the first review and the missing function in this one.

timesketch/frontend-ng/src/views/Explore.vue
jaegeral: should be better now, PTAL

timesketch/frontend-ng/src/views/Explore.vue
jaegeral: just noticed it as well, changed, now works

jkppr (approved):
LGTM from my side

jaegeral:
Re-adding @berggren  as @jkppr is fine with it, wanted to get your opinion in as well.

berggren:
@jaegeral We cannot merge this PR because the plugin you are adding (vue-shortkey) is not compatible with Vue3. The latest release of that plugin is from 2018, so it doesn't seem to be maintained either.

I opt we write our own component for global keyboard shortcuts. It's more or less a wrapper around normal key events from the DOM api so it shouldn't require much.

Let me know how you want to proceed.

timesketch/frontend-ng/package.json
```diff
@@ -29,6 +29,7 @@
     "vue-dragscroll": "^3.0.0",
     "vue-moment": "^4.0.0",
     "vue-router": "^3.0.2",
+    "vue-shortkey": "^3.1.7",
```
berggren, line 32: This plugin/library is not compatible with Vue3, and doesn't seem to be maintained (latest release is from 2018).

I opt to refactor this and write our own component for keyboard shortcuts instead.

berggren (changes requested):

PR #2586, opened 2023-02-28
Making timestamp leading value when importing a csv

### LINKED ISSUE #2579: Frontend-ng csv importer parses time wrong
**Describe the bug**
Importing a csv seems to miss a lot of accuracy of datetime.
E.g. this line:
```
2022-11-10T05-07-28+00:00,78Vh3h7,1668056848119860,Time Logged
```

Becomes: 

```
{
   "__ts_timeline_id": 17,
   "datetime": "2022-11-10T05:00:00+00:00",
   "message": "78Vh3h7",
   "timestamp": 1668056400000000,
   "timestamp_desc": "Time Logged"
}
```

Which clearly is wrong

**Expected behavior**
Datetime should be correct

jaegeral:
This PR changes the CSV import handling from using datetime as the leading value to using timestamp. 

It also calculates an offset between datetime and the provided timestamp to see if the values are too far apart.

The new code will first check if the row of the CSV file contains a timestamp. If it does, then it will use that timestamp as the leading value and calculate a datetime value.

The code will also calculate an offset between the datetime and the provided timestamp. This offset will be used to determine if the values are too far apart. If the offset is greater than a certain threshold, then the code will raise an error and create the datetime value based on the provided timestamp.

The PR also introduces a new csv with some test data that I hope we will increase over time. I opted to not make it mocked in the unit tests itself so that we can re-use the csv later in end2end tests or if someone needs test data with problems.

*Note:* As far as I can tell, in Importing JSON part of the util class, the timestamp is already the "leading" value - so this change will just get it to similar approach. But the JSOn import does not have a check yet on how far provided values are apart. 

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

closes #2579


timesketch/lib/utils_test.py
```diff
@@ -156,23 +176,54 @@ def test_invalid_CSV_file(self):
             # Call next to work around lazy generators.
             next(_validate_csv_fields(mandatory_fields, df_02))
 
+    def test_datetime_parsing_csv_file(self):
```
jaegeral, line 179: This is the only real new thing in the unit tests, rest is only black formatting

timesketch/lib/utils.py
```diff
@@ -332,6 +334,41 @@ def read_and_validate_csv(
                 _scrub_special_tags(row)
                 # Remove all NAN values from the pandas.Series.
                 row.dropna(inplace=True)
+                MAX_TIMESTAMP_DIFFERENCE = 100000  # 100 seconds
```
berggren, line 337: As discussed offline - Instead of rewriting the fualty datetime we should error out and let the user fiox the broken input.

berggren (changes requested):

timesketch/lib/utils.py
```diff
@@ -332,6 +334,32 @@ def read_and_validate_csv(
...
+                # check datetime plausibility
+                if "timestamp" in row and "datetime" in row:
+                    # if timestamp and datetime are both present,
+                    # calculate timestamp difference between the both
+                    # convert row datetime to timestamp
```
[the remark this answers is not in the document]
berggren: The comment doesn't add value IMO

berggren (approved):
LGTM, with one small nit.

berggren:
Take a look at the failing e2e test

berggren (changes requested):
failing e2e test. Please fix before merge

berggren (approved):

PR #2587, opened 2023-02-28
Port Graph to new UI

berggren:
This PR ports over the Graph functionality.

Screenshot 2023-02-28 at 17 24 09

Screenshot 2023-02-28 at 17 24 39


timesketch/frontend-ng/src/views/Graph.vue
```diff
@@ -0,0 +1,658 @@
...
+                ></v-radio>
+              </v-radio-group>
+
+              <h5>Transparency for unselected elements</h5>
+              <v-slider v-model="fadeOpacity" @change="changeOpacity" :max="100" :min="0" thumb-label>
+                <template v-slot:thumb-label="{ value }"> {{ value }}% </template>
+              </v-slider>
+            </v-card>
+          </v-menu>
+          <!-- Save graph dialog -->
```
jkppr, line 84: When a graph is saved it will be shown in the left panel only after a reload of the page. This feels unintuitive, since there is no feedback where the saved graph went when clicking the save button.
(if this will be fixed in a future version, please open an issue for tracking, thanks)

timesketch/frontend-ng/src/views/Graph.vue
```diff
@@ -0,0 +1,658 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<template>
+  <v-container fluid>
+    <v-card flat class="pa-3 pt-0 mt-n3" color="transparent">
+      <v-card class="d-flex align-start mb-1" outlined>
+        <v-text-field
```
jkppr, line 20: Note: When searching for nodes in the graph only lower case seems to work. Even the nodes are sometimes shown in full caps. E.g. when I have a node `SYSTEM` and search for `SYSTEM` in the search field, it will not find the node. But when searching for `system` it finds `SYSTEM` as expected.

jkppr (changes requested):
Overall it looks great! Please address the bug in the left panel `Graphs.vue` component.

timesketch/frontend-ng/src/views/Graph.vue
berggren: This is fixed. I moved this to the stgate (store) instead. The left panel will now update when new graph is saved.

berggren (commented):
PTAL

jkppr (approved):
LGTM

PR #2600, opened 2023-03-08
CLI: Add Tag to single event

jaegeral:
A fairly common uecase is one has a event ID and wants to codify to add a tag to it without using the API client. This PR adds this functionality.

E.G: 
```
timesketch timelines list
16 timeline test - First 10k (2)
17 new
18 timeline test - First 10k (3)
19 Manual events
root@5b49c41f6fef:/usr/local/src/timesketch/cli_client/python# timesketch timelines describe 19
Name: Manual events
Index: 49a318b0ba17867fd71b50903774a0c8
```

And then to add it:
```
timesketch timelines event-add-tags --event_id BNCewYYBe0OIiBUpeC37 --tags fooobarrrrrrr 19 --output-format json
{'events_processed_by_api': 1, 'number_of_events_passed_to_api': 1, 'number_of_events_with_added_tags': 1, 'tags_applied': 5, 'total_number_of_events_sent_by_client': 1}
```

It can be discussed if that function makes more sense under the timelines context of sketch context. The API method is in the sketch context, but without a timeline, it does not really works.

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/timelines.py
+++ b/cli_client/python/timesketch_cli_client/commands/timelines.py
@@ -54,13 +54,23 @@ def describe_timeline(ctx, timeline_id):
     click.echo(f"Index: {timeline.index_name}")
 
 
-@timelines_group.command("event-add-tag")
+@timelines_group.command("event-add-tags")
 @click.argument("timeline_id", type=int, required=False)
 @click.option("--event_id", required=True, help="ID of the event.")
-@click.option("--tag", required=True, help="Tag to add to the event.")
+@click.option(
+    "--tags",
+    required=True,
+    help="Comma seperated list of Tags to add to the event.",
+)
+@click.option(
+    "--output-format",
+    "output",
+    required=False,
+    help="Set output format (overrides global setting)",
+)
 @click.pass_context
-def event_add_tag(ctx, timeline_id, event_id, tag):
-    """Add a tag to an event."""
+def event_add_tags(ctx, timeline_id, event_id, tags, output):
+    """Add tags to an event."""
     sketch = ctx.obj.sketch
     timeline = sketch.get_timeline(timeline_id=timeline_id)
     if not timeline:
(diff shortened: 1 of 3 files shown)


cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -52,3 +52,27 @@ def describe_timeline(ctx, timeline_id):
...
+@click.option("--tag", required=True, help="Tag to add to the event.")
+@click.pass_context
+def event_add_tag(ctx, timeline_id, event_id, tag):
+    """Add a tag to an event."""
+    sketch = ctx.obj.sketch
+    timeline = sketch.get_timeline(timeline_id=timeline_id)
+    if not timeline:
+        click.echo("No such timeline")
+        return
+    events2 = [
```
tomchop, line 69: Shouldn't this be "event"?

tomchop (changes requested):

tomchop (approved):

cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -52,3 +52,43 @@ def describe_timeline(ctx, timeline_id):
...
+
+@timelines_group.command("event-add-tags")
+@click.argument("timeline_id", type=int, required=False)
+@click.option("--event_id", required=True, help="ID of the event.")
+@click.option(
+    "--tags",
+    required=True,
+    help="Comma seperated list of Tags to add to the event.",
+)
+@click.option(
```
berggren, line 65: drive by review, I know this has already been merged.

Whey is --output-format needed here? It has nothing to do with adding tags.

Please followup and remove if there isn't anything I'm missing here.

tomchop: This is to be able to tweak the output-format to JSON if the user wants to use the output of the sommand somewhere else.

cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -52,3 +52,43 @@ def describe_timeline(ctx, timeline_id):
         return
     click.echo(f"Name: {timeline.name}")
     click.echo(f"Index: {timeline.index_name}")
+
+
+@timelines_group.command("event-add-tags")
+@click.argument("timeline_id", type=int, required=False)
```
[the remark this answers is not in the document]
tomchop: maybe I'm misunderstanding how this works, but isn't the timeline ID required to identify the event correctly?

PR #2602, opened 2023-03-08
Fix SSL verify behavior in API client

### LINKED ISSUE #2601: API Client not working with SelfSigned SSL certificates
**Describe the bug**
When trying to connect to a Timesketch instance running with HTTPS and a SelfSigned certificate, the API client is not able to connect to the instance, even when setting the `verify` parameter in the config file to `True`.
This is due to the parameter being read from the config as `string`, but being used as `bool`. This leads to the API client always enforcing SSL verification.
I will add a PR in a minute, that will propse a fix for this.

**To Reproduce**
1. Run a Timesketch instance with a SelfSigned certificate
2. Run `ts_client = config.get_client()` and enter your settings interactively
3. Either set the `verify` parameter to `False` using the API or by directly editing `~/.timesketchrc`
4. Run `ts_client = config.get_client()`  again, forcing the client to reload its config from file
5. The client will still prompt `Unable to connect to the Timesketch server, are you connected to the network? Is the timesketch server running and accessible from your host? The error message is Timesketch server unreachable`

**Expected behavior**
When setting the `verify` parameter to `False`, a connection to a Timesketch instance running with SelfSigned certificates should be possible and SSL verification should not be enforced.

**Screenshots**
Not applicable here.

Not applicable here.

**Additional context**
Not applicable here.

FabFaeb:
This PR fixes a bug during the loading of the `.timesketchrc` config. The `verify` parameter was loaded as `string` but is used elsewhere (e.g. https://github.com/google/timesketch/blob/4ddf6605784f8fa70278babb8c7137e764da31c7/api_client/python/timesketch_api_client/client.py#L347) like a `boolean`. Therefore, the SSL verification could not be turned off.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

**Closing issues**

Closes #2601


berggren (approved):
LGTM thanks for fixing!

FabFaeb:
No worries - glad I could help. I just updated my branch with the latest changes from master.

PR #2603, opened 2023-03-08
Update sigmastudio snackbars

jkppr:
This small PR will update the sigmastudio to make use of the global snackbars instead of the old EventBus. With this fix, the snackbars will be green for success events and red for error events.


jaegeral (approved):
very cool

PR #2604, opened 2023-03-09
[CLI] events group to add comments and tags to an event

jaegeral:
This new cli group allows to annotate events in Timesketch.
It allows you to add comments to a specific event in a Timesketch sketch. 

* comments (comma seperated)
* tags (comma seperated)

Comments can be any text that you want to add to an event. They can be used to provide additional information about the event, to track changes to the event over time, or to communicate with other users about the event.

Comments are stored in the Timesketch database and can be accessed by anyone who has access to the sketch. They can also be exported to a file or shared with other users.

Adding comments to events is a powerful way to get more out of Timesketch. It can help you to keep track of your data, to communicate with other users, and to make better decisions about your data.

Example:

```
 timesketch events annotate --event-id A9BGwYYBe0OIiBUpWy3U --timeline-id 19 --output-format json --tags foobar2 --comments "adasd","comment2"
{'meta': {'comments': [{'comment': 'adasd', 'created_at': 'Fri, 10 Mar 2023 13:06:28 GMT', 'id': 3, 'updated_at': 'Fri, 10 Mar 2023 13:06:28 GMT', 'user': {'username': 'dev'}}, {'comment': 'comment2', 'created_at': 'Fri, 10 Mar 2023 13:06:28 GMT', 'id': 4, 'updated_at': 'Fri, 10 Mar 2023 13:06:28 GMT', 'user': {'username': 'dev'}}]}, 'objects': {'datetime': '2023-03-04T11:31:12', 'key': 'value', 'key2': 'value2', 'message': 'foobar', 'tag': ['foobar2', 'foobar'], 'timestamp': 1677929472000000, 'timestamp_desc': None}}
```

PR #2606, opened 2023-03-09
[CLI] add labels to one event

jaegeral:
Add labels to one event in the Timesketch CLI

```
timesketch timelines event-add-labels --event_id A9BGwYYBe0OIiBUpWy3U --labels "comment1 foobar2","asdasd2" 19 --output-format json
{'meta': {}, 'objects': [[{'created_at': '2023-03-09T15:24:40.519907', 'name': None, 'updated_at': '2023-03-09T15:24:40.543321', 'user': {'active': True, 'admin': True, 'groups': [], 'username': 'dev'}}]]}
{'meta': {}, 'objects': [[{'created_at': '2023-03-09T15:27:03.308873', 'name': None, 'updated_at': '2023-03-09T15:27:03.332113', 'user': {'active': True, 'admin': True, 'groups': [], 'username': 'dev'}}]]}
```

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/timelines.py
+++ b/cli_client/python/timesketch_cli_client/commands/timelines.py
@@ -92,3 +92,45 @@ def event_add_tags(ctx, timeline_id, event_id, tags, output):
         click.echo(return_value)
     else:
         click.echo(f"Tags {tags_list} added to event {event_id}")
+
+
+@timelines_group.command("event-add-labels")
+@click.argument("timeline_id", type=int, required=False)
+@click.option("--event_id", required=True, help="ID of the event.")
+@click.option(
+    "--labels",
+    required=True,
+    help="Comma seperated list of Labels to add to the event.",
+)
+@click.option(
+    "--output-format",
+    "output",
+    required=False,
+    help="Set output format (overrides global setting)",
+)
+@click.pass_context
+def event_add_labels(ctx, timeline_id, event_id, labels, output):
+    """Add labels to an event."""
+    sketch = ctx.obj.sketch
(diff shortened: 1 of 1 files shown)


cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -92,3 +92,45 @@ def event_add_tags(ctx, timeline_id, event_id, tags, output):
...
+)
+@click.pass_context
+def event_add_labels(ctx, timeline_id, event_id, labels, output):
+    """Add labels to an event."""
+    sketch = ctx.obj.sketch
+    timeline = sketch.get_timeline(timeline_id=timeline_id)
+    if not timeline:
+        click.echo("No such timeline")
+        return
+    events = [
```
berggren, line 119: The outout should be text by default. If you want to return a json string you could put that behind a `--json` flag?

cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -92,3 +92,45 @@ def event_add_tags(ctx, timeline_id, event_id, tags, output):
         click.echo(return_value)
     else:
         click.echo(f"Tags {tags_list} added to event {event_id}")
+
+
+@timelines_group.command("event-add-labels")
```
berggren, line 97: Same comment as oin the other PR about annotating an event. Consider refactor?
```
$timesketch timeline 19 event 1234 --label ...`
```

berggren (changes requested):

jaegeral:
This functionality will go in the events class that will be introduced in https://github.com/google/timesketch/pull/2604 and also potentially not be added at all, as we plan to sunset labels (in favour of just relying on tags)

PR #2607, opened 2023-03-09
[CLI] add a show_internal_columns and json output to the explore method…

jaegeral:
Add a `show_internal_columns` flag to the search method in Timesketch CLI and a new output format for the same method.

This enables analysts to output the raw timeline with OpenSearch Internals:

```
timesketch search --limit 20 --show_internal_columns --output-format json |jq .
{
  "__ts_timeline_id": 16,
  "comment": [],
  "datetime": 0,
  "message": "3QPcr6sadN",
  "timestamp": 0,
  "timestamp_desc": "Last Time Executed",
  "_id": "R6R9eYYBGeSzWlvssgk2",
  "_type": "_doc",
  "_index": "bfdc00e7d59f462a905de807cc1b35dc",
  "_source": null
}
```

This mostly caters when someone (e.g. developer) really wants to have ALL the attributes for an event, to debug things, to get the IDs for other commands etc.


jaegeral:
Adding @berggren  here as he might have opinions on the new verbose flag

cli_client/python/timesketch_cli_client/commands/search.py
berggren, line 1: --verbose is a bit ambiguous in this context. I opt that we are more direct.
I prefer `--show-internal-columns` as the flag instead. This is describing what it does, and is similar to the other flag `--show-headers`

berggren (changes requested):

berggren (approved):

PR #2613, opened 2023-03-15
Make tags reactive

### LINKED ISSUE #2614: Tags are not updating
When adding tags the list of tags is not updated until full refresh.

berggren:
This PR fixes a bug where tags did not update when added. This was due to the mutation relying on opensearch aggregation. The problem was that the aggregation was not returning the re-indexed document due to index update and stale index freshness.

* Plus a few minor UI consistency changes such as Save buttons.

closes #2614


timesketch/frontend-ng/src/components/LeftPanel/SigmaRules.vue
```diff
@@ -23,11 +23,9 @@ limitations under the License.
       @click="expanded = !expanded"
```
jkppr, line 23: **UX feedback:** If the element does not expand to show an "empty state" when there are no sigma rules (which makes sense), then it should not be rendered as clickable. (cursor: pointer).
I would recommend to change line 20 to `:style="!(sigmaRules && sigmaRules.length) ? '' : 'cursor: pointer'"`

timesketch/frontend-ng/src/components/LeftPanel/Stories.vue
```diff
@@ -24,7 +24,15 @@ limitations under the License.
       :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'"
```
jkppr, line 24: **UX feedback:** Same like with the Sigma Rules. If there are no stories and the element does not have any action when clicked, it should not be rendered as clickable. Recommend to change line 20 to `:style="!(meta.stories.length) ? '' : 'cursor: pointer'"`

jkppr (approved):
General LGTM, two smaller UX nits, please take a look.

timesketch/frontend-ng/src/components/LeftPanel/SigmaRules.vue
berggren: fixed

timesketch/frontend-ng/src/components/LeftPanel/Stories.vue
berggren: Good catch! Fixed now.

PR #2620, opened 2023-03-17
[CLI] Add Sigma list / describe to timesketch cli

jaegeral:
Adding a new command group to the Timesketch cli: `sigma`

This group moving forward will be used to interact with all things Sigma.

`list` and `describe` are used in a similar fashion as in search-template, where list will list all items and describe will further detail one element.

For the first iteration, it will list Sigma rules installed on a system:

```
timesketch sigma list --output-format text
                           rule_uuid                             title
8c10509b-9ba5-4387-bf6c-e347931b646f        SigmaRuleTemplateTitledddd
5266a592-b793-11ea-b3de-0242ac130004 Suspicious Installation of Zenmap
e5684ad6-5824-4680-9cc5-e8f0babd77bb                            Foobar
```

Because 
```
sigma_rule = api_client.get_sigmarule(rule_uuid)
```
Has no `to_json()` yet, the workaround is to use ```click.echo(sigma_rule.__dict__)``` instead to get all the attributes available in the object.

There is also:

```
sigma describe --rule-uuid 8c10509b-9ba5-4387-bf6c-e347931b646f --output-format text
Title: SigmaRuleTemplateTitledddd
Author: Timesketch
Status: experimental
Rule UUID: 8c10509b-9ba5-4387-bf6c-e347931b646f
Search: *foobar*
```

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] Documentation updated. --> will be done in a seperate PR


cli_client/python/timesketch_cli_client/commands/sigma.py
```diff
@@ -0,0 +1,86 @@
...
+    "--rule-uuid",
+    "rule_uuid",
+    required=True,
+    help="UUID of the sigma rule.",
+)
+@click.option(
+    "--output-format",
+    "output",
+    required=False,
+    help="Set output format (overrides global setting).",
```
jkppr, line 65: What output options are available? Looks to me just json or default? So why have a user parameter instead of a `--json` flag?

jkppr (changes requested):
Just some small nits. 
At least the list of available output formats needs be added to the help string from my point of view. (especially since both commands have different output formats available.

cli_client/python/timesketch_cli_client/commands/sigma.py
jaegeral: to keep the options (and help pages) consistent basically, every group / command has --output-format, even if there is only two options.

jkppr: makes sense.

cli_client/python/timesketch_cli_client/commands/sigma.py
```diff
@@ -0,0 +1,86 @@
...
+def sigma_group():
+    """Manage sigma rules."""
+
+
+@sigma_group.command("list")
+@click.option(
+    "--output-format",
+    "output",
+    required=False,
+    help="Set output format (overrides global setting).",
```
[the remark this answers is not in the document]
jaegeral: done

jkppr (approved):
LGTM from my side

cli_client/python/timesketch_cli_client/commands/sigma.py
```diff
@@ -0,0 +1,86 @@
...
+    except ValueError as e:
+        click.echo(e)
+        sys.exit(1)
+
+    if output == "json":
+        click.echo(sigma_rules.to_json(orient="records"))
+    elif output == "csv":
+        click.echo(sigma_rules.to_csv())
+    elif output == "text":
+        click.echo(sigma_rules.to_string(index=False, columns=["rule_uuid", "title"]))
```
berggren, line 49: add `header=show_headers` to enable the flag I mentioned above

berggren (changes requested):

cli_client/python/timesketch_cli_client/commands/sigma.py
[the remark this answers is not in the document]
jaegeral: adjusted it, now the user can select columns

jaegeral: done

cli_client/python/timesketch_cli_client/commands/sigma.py
```diff
@@ -0,0 +1,86 @@
...
+import click
+
+
+@click.group("sigma")
+def sigma_group():
+    """Manage sigma rules."""
+
+
+@sigma_group.command("list")
+@click.option(
```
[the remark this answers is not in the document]
jaegeral: good idea

berggren (approved):
LGTM

PR #2632, opened 2023-03-23
Sigma list speedup

jaegeral:
# Summary
This pull request optimizes the enrich sigma rule code and adjusts the UI frontend code for it.

## Changes
The following changes have been made:
The enrich sigma rule code has been optimized to improve performance.
The UI frontend code has been adjusted to reflect the changes to the enrich sigma rule code.


timesketch/lib/sigma_util.py
```diff
@@ -78,19 +78,28 @@ def get_sigma_config_file(config_file=None):
     return sigma_config
 
 
-def _enrich_sigma_rule_object(rule: SigmaRule):
-    """Helper function: Returns an enriched Sigma dict given a SigmaRule.
+def _enrich_sigma_rule_object(rule: SigmaRule, parse_yaml: bool = False):
```
berggren, line 81: This looks like the same function as in api/sigma.py? Why do we have two copies?

berggren (changes requested):
PTAL

berggren (approved):

timesketch/lib/sigma_util.py
jaegeral: https://github.com/google/timesketch/issues/2633

PR #2656, opened 2023-04-04
Empty state for Scenarios tab

berggren:
Handle empty state for when there are no scenarios added to the installation.
Screenshot 2023-04-04 at 13 40 35

### DIFF
--- a/timesketch/frontend-ng/src/views/Sketch.vue
+++ b/timesketch/frontend-ng/src/views/Sketch.vue
@@ -148,8 +148,11 @@ limitations under the License.
         <v-tab-item :transition="false">
           <ts-scenario v-for="scenario in activeScenarios" :key="scenario.id" :scenario="scenario"></ts-scenario>
           <v-row class="mt-0 px-2" flat>
-            <v-col cols="6">
-              <v-btn text color="primary" @click="scenarioDialog = true" style="cursor: pointer"
+            <v-col cols="12">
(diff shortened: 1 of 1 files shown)


timesketch/frontend-ng/src/views/Sketch.vue
```diff
@@ -148,8 +148,11 @@ limitations under the License.
         <v-tab-item :transition="false">
           <ts-scenario v-for="scenario in activeScenarios" :key="scenario.id" :scenario="scenario"></ts-scenario>
           <v-row class="mt-0 px-2" flat>
-            <v-col cols="6">
-              <v-btn text color="primary" @click="scenarioDialog = true" style="cursor: pointer"
+            <v-col cols="12">
+              <v-card v-if="!Object.keys(scenarioTemplates).length" flat class="pa-4"
+                >No scenarios available yet.</v-card
```
jkppr, line 153: So this message is displayed if there is no scenario template available. From an UX perspective it would make sense to tell the user what he needs to do to fix that. E.g. `No scenario templates available. Please run tsctl ... to import scenarios to use with the investigate feature.` 

I would also center the message, but that is personal preference ; )

jkppr (approved):
Overall LGTM. Just a small nit from UX perspective in the context of actionable error messages. But that can also be added later on.

timesketch/frontend-ng/src/views/Sketch.vue
berggren: Done

PR #2672, opened 2023-04-14
Scenarios model refactor

berggren:
This PR adds `approaches` to investigative questions. A questions can have multiple approaches and this align with the updated DFIQ spec.
* Move search templates and sigma rules to approcahes.
* Add dfiq_identifier field to Scenarios/Facet/Question/Approach


jkppr (approved):
Note: This PR will break the ability to add a Scenario in the Investigate tab since the new variable `dfiq_identifier` is not added to the API endpoint yet. `/usr/local/src/timesketch/timesketch/api/v1/resources/scenarios.py", line 130` 

Synced with Johan, and this error is expected and is fixed as part of another coming PR. Given that the feature is not yet ready and used, I will LGTM this PR in the scope of changing the DB model.

Issue #2676, opened 2023-04-14
UI: Analyzer Result searches on the wrong field


jkppr:
**Describe the bug**
When there are more than 10 different analyzers run, the Analyzer Result section in the left panel will display a search bar for better navigation. However, the search is based on the `analyzer_name` key and not on the `analyzer_display_name` value which is the value visible to the user.

**To Reproduce**
1. Run > 10 different analyzers, including the `Windows logon/logoff events` analyzer
2. When the search bar shows up, search for `Windows` 
3. This will not result in `Windows logon/logoff events` showing up, since the analyzer_name is `login`.
4. Search for `login` and `Windows logon/logoff events` will show up.

**Expected behavior**
1. The search should be based on the display_name that is visible to the user.
2. (optional) It would be great if the search would also work for timeline names.

jkppr:
@Annoraaq thanks for fixing the search so it does work on the display_name for now. Are you going to look into the optional part of searching for timeline names as well? Or shall we just close it out and keep it like this for now?

Annoraaq:
@jkppr I will look into the optional part, too.

PR #2679, opened 2023-04-14
Filter analyzer results by display_name

Annoraaq:
#2676 

Filter analyzer results on display name.

That fixes part 1 of #2676


jkppr (approved):
Thanks for fixing this!

Issue #2396, opened 2022-10-25
Frontend-ng "Untag an event" / Remove a Tag from an event


jaegeral:
If you tag an event e.g. with `suspicious` or `bad`or `good` you might have tagged the wrong event so you want to remove that tag, which is currently not possible.

This is a known issue and will be fixed in Q2, as this requires a new API endpoint, we did not want to introduce it as a new feature right before we make the big switch.

Expected solution:
Have a way to remove a tag from an event.

- [x] Add functionality to API https://github.com/google/timesketch/pull/2689
- [x] Add functionality to Python API client
- [x] Add to cli client
- [x] Add to Web UI

jaegeral:
This is blocked by https://github.com/google/timesketch/issues/2605

jkppr:
Frontend should be ready in my branch: https://github.com/jkppr/timesketch/tree/2396_ui_untag_events

I will review the backend PR in #2605 and then create a PR for the frontend when the backend is merged.

jkppr:
Draft PR: #2694 ready for review. Could be merged as soon as the backend is implemented and the PR adjusted.

jaegeral:
Done

PR #2689, opened 2023-04-18
add remove_tag to API

jaegeral:
While this PR: https://github.com/google/timesketch/pull/2663 has the back to back changes, this PR will only introduce the remove tag functionality to the API.

This is to make the change better to review.

https://github.com/google/timesketch/issues/2396


timesketch/lib/testlib.py
```diff
@@ -90,7 +90,7 @@ def __init__(self):
         self.indices = MockOpenSearchIndices()
 
     def search(
-        self, index, body, size=0, search_type=None
+        self, index, body, size=0, search_type=None, _source_include=None
```
jaegeral, line 93: needed to add this to support the tag events unit tests that where previously not there

timesketch/api/v1/resources/event.py
```diff
@@ -53,21 +53,24 @@
...
         tag_dict (dict): a dict that contains information to be returned
             by the API call to the user.
-        tags_to_add (list[str]): a list of strings of tags to add to each
-            event.
+        tags_to_modify (list[str]): a list of strings of tags to remove
+            or add on each event.
         datastore (opensearch.OpenSearchDataStore): the datastore object.
         flush_interval (int): the number of events to import before a bulk
             update is done with the datastore.
+        tag (bool): a boolean that decides if to tag or untag events.
```
jaegeral, line 70: I do not have a really good name for this variable, other option was to make an "action" as a string and pass "add" or "remove" to it to flow control the method, but I was not happy with it either.

timesketch/api/v1/resources_test.py
```diff
@@ -312,6 +312,120 @@ def test_invalid_index(self):
...
+        response = self.client.delete(
+            "/api/v1/sketches/1/event/tagging/",
+            json=json_value,
+            content_type="application/json",
+        )
+        self.assert200(response)
+        self.assertEqual(expected_response, response.json)
+
 
 class EventAddAttributeResourceTest(BaseTest):
```
jaegeral, line 430: from here on my hooks fixed some typos.

timesketch/api/v1/resources/event.py
jkppr: I would recommend to go with the string / flag approach, since it makes understanding the code much more easy.

jaegeral: done

jaegeral:
> Based on some offline discussion with @berggren :
> 
> The whole approach to tag / untag events (post & delete methods) look rather complicated in general, for such a "simple" task. The main question is, do we want to continue using the current approach with panda dataframes etc. or move to a simplified version? Other examples on handling event attribute manipulation are the handling of tags by the [analyzer framework](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/interface.py#L225) or the [annotation endpoint for comments](https://github.com/google/timesketch/blob/master/timesketch/api/v1/resources/event.py#L771).
> 
> Problem with changing the existing endpoint: Other clients might depend on the way this endpoint handles the tagging of events (e.g. providing the full event object instead of an event_id), etc.)
> 
> Idea/proposal: We could add a new API resource / endpoint in parallel to handle the tagging / untagging of events in a new and slimmer way. This could then be used by the tagging frontend I'm working on. This would allow backwards compatibility and buys some time to find all the parts that use the old resource to update towards the new one. And as a final step, we could remove the old resource when everything is happy with the new one.
> 
> What do you think?

So yes in theory it would be ideal to have a slimmer version that is easier to use.
However, the current way of handling adding tags was introduced (and kind of tested) over the course of months and years and just now came up as potentially complicated (I am not confident enough to say that it is to complicated, there might be good design reasons to have it the way it is).

That being said, introducing a v2 and or parallel way of tagging / untagging events while still maintaining the old, will certainly add complexity in the situation of trouble shooting. When someone added a tag and has problems, which method was used to add it?

The other aspect is time. As far as I understood it so far, the removal of tags is a feature needed by user sooner than later.   Sticking to the current concept of Tag handling in events and mimicking design decisions to get a `remove_tag´function for the API (and stripping out a bit of things like `verbose`) was doable in good amount of time. Creating a fork / v2 however we want to call it "simpler" version of the function will take time.

This was done as part of the Sigma refactoring where we moved from File to Database and it was not an easy task, for a function that was not used widely. Here we are talking about an API call that is used by analysts 100s of times each day.

So while I am all for refactoring and reviewing some parts of the API, I am not yet convinced that it would be worth for this API call.

Issue #2467, opened 2022-12-16
import jsonl error if have not 3 lines


lprat:
**Describe the bug**
If your file.jsonl must have 3 lines min else you receive a error

**To Reproduce**
sample:
```
{"message": "A message","timestamp": 123456789,"datetime": "2015-07-24T19:01:01+00:00","timestamp_desc": "Write time","extra_field_1": "foo"}
{"message": "Another message","timestamp": 123456790,"datetime": "2015-07-24T19:01:02+00:00","timestamp_desc": "Write time","extra_field_1": "bar"}
```

**Screenshots**
![image](https://user-images.githubusercontent.com/1132448/208158678-3e34ee03-371a-4976-bacd-8efbb9a8614e.png)

If you add one line more, you dont have error and you can upload jsonl.

**Additional context**
Timesketch docker version: "us-docker.pkg.dev/osdfir-registry/timesketch/timesketch   latest            915624b6d4e0   27 hours ago    737MB"

FazlOmar9:
Hey! If this is still relevant, I'd like to work on this. Could you please give me the details for this?

jasochen7:
Hello, I'm interested in working on this issue and am really passionate about the timesketch project. How can I get started in addressing this issue?

jaegeral:
Hey @FazlOmar9 and @jasochen7  I am not sure about your ask here, where do you struggle to get started?

jasochen7:
Just to clarify the error; are there any requirements that follow beyond the immediate description of the task? I.e. are there any reliability thresholds or other criteria that a solution must meet?

PR #2690, opened 2023-04-20
Import JSONL with less than three lines

jasochen7:
This is a fix for issue #2467


jaegeral:
Thx for the pull request, there are several linter errors, could you please take a look and fix those?

Thank you

berggren:
Closing this PR due to lack of response, and because this is not a fix for the problem and would need to be rewritten from scratch.

The issue is that the array of json lines includes the last empty string as well and it needs to be filtered out.

```
@@ -551,7 +551,7 @@ export default {
         if (e.target.readyState === FileReader.DONE) {
           /* 3a. Extract the headers from the CSV */
           let data = e.target.result
-          let rows = data.split('\n')
+          let rows = data.split('\n').filter((jsonlLine) => jsonlLine !== '')
           let i = Math.min(vueJS.staticNumberRows, rows.length)
           try {
             vueJS.headersString = JSON.parse(rows[0])
```

berggren:
fixed in #2793

PR #2691, opened 2023-04-20
Fix for the pagination problems

jkppr:
With the new UI there is an issue that the pagination of the event table does automatically resets every time someone switches a page.

This PR should fix this.

Additional new feature: The PR also adds a function to click on the "magnifying glass" in the search bar to trigger the search.


berggren (approved):
LGTM

PR #2699, opened 2023-04-21
filter analyzers by timeline

Annoraaq:
The analyzer search now also considers the name of the attached timelines. See #2676


jkppr (approved):
Thanks for fixing the analyzer filter.

PR #2701, opened 2023-04-24
Enable sketch archiving

### LINKED ISSUE #2695: New UI: handle archived sketches
In the new UI, when accessing a sketch that was archived it looks like a normal sketch, but all timelines are grey and not functional. 
We need some UI element to handle the information that the sketch is archived and needs un-archiving similar to the old UI.

berggren:
This PR adds archiving support to the new UI and also fixes rendering of sketches that are archived.
This PR also removes the non functional export feature for now. Reenable when we have fixed missing pieces (aggregations in stories)

fixes #2695


jkppr (approved):
lgtm

Issue #2659, opened 2023-04-04
UI feedback: Align elements in the left panel


jkppr:
**Describe the solution you'd like**
The left + symbols in the left panel for Stories and TI are a bit more left than the centered ones which align with the numbers.

**Screenshots**
![leftPanelpolish](https://user-images.githubusercontent.com/99879757/229834251-74d37ab2-1808-4e32-8ab9-cc5bc40f4c0b.png)

sosa-e:
Hello, I am new to open source and would like to work on this

PR #2704, opened 2023-04-24
Fix alignment for left panel menu

berggren:
Small polish. Count and icons was not aligned correctly. This PR fixes that.
(The many lines on AnalyzerResults.vue is due to Vetur formatting).

fixes: #2659


jkppr (approved):
lgtm

Issue #2705, opened 2023-04-25
Implement new analyzer output for TimeSketch


roshanmaskey:
Implement new analyzer output to be used by new analyzers.

PR #2706, opened 2023-04-25
AnalyzerOutput and supporting classes

roshanmaskey:
Addressing issue #2705 Implement new analyzer output for TimeSketch.

This PR contains AnalyzerOutput and it's supporting classes.


jkppr:
@roshanmaskey please check out my suggested changes in https://github.com/roshanmaskey/timesketch/pull/1

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+        self.dfiq_question_id = ""
+        self.dfiq_question_conclusion = ""
+        self.result_priority = "NOTE"
+        self.result_summary = ""
+        self.result_markdown = ""
+        self.references = []
+        self.attributes = {}
+        self.timesketch_instance = timesketch_instance
+        self.sketch_id = sketch_id
+        self.timeline_id = timeline_id
```
berggren, line 1230: Why do we need these?

I guess the json of an `AnalyzerOutput` will be added to an `Analysis` object in the database? If so then you can link back to sketch and timeline from there. 

timesketch_instance - Who would be the consumer of this information? This hints that the information would be consumed by something externally to TS?

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+        self.analyzer_identifier = analyzer_identifier
+        self.analyzer_name = analyzer_name
+        self.result_status = ""  # TODO: link to analyzer status/error?
+        self.dfiq_question_id = ""
+        self.dfiq_question_conclusion = ""
+        self.result_priority = "NOTE"
+        self.result_summary = ""
+        self.result_markdown = ""
+        self.references = []
+        self.attributes = {}
```
berggren, line 1227: The name `attributes` is too generic IMO. Can we rename this to something more telling, e.g. `result_attributes`?

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+            conclusion.
+        result_priority (str): [Required] Priority of the result based on the
+            analysis findings. Valid values are NOTE (default), LOW, MEDIUM, HIGH.
+        result_summary (str): [Required] A summary statement of the analyzer
+            finding. A result summary must exist even if there is no finding.
+        result_markdown (str): [Optional] A detailed information about the
+            analyzer finding in a markdown format.
+        references (List[str]): [Optional] A list of references about the
+            analyzer or the issue the analyzer attempts to address.
+        attributes (dict): [Optional] A list of object that holds the
```
berggren, line 1195: Is this a list or a dict? The docstring is ambiguous

berggren: And what type of 
```suggestion
        attributes (dict): [Optional] A list of objects that holds the
```

berggren: and what type of objects is this?

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+class AnalyzerOutputException(Exception):
+    """Analyzer output exception."""
+
+
+class AnalyzerOutput:
+    """A class to record timesketch analyzer output.
+
+    Attributes:
+        platform (str): [Required] Analyzer platfrom.
+        analyzer_identifier (str): [Required] Unique analyzer identifier.
```
berggren, line 1179: The analyzer_name is uniq as well. What is the identifier here?

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+        analyzer_name (str): [Required] Analyzer display name.
+        result_status (str): [Required] Analyzer result status.
+            Valid values are success, no-findings, error.
```
berggren, line 1182: `no-findings` is a bit ambiguous IMO. Is it because of an error or something else? Can we skip this status and only have `success` and `error`? Then to know if there are no findings you can check the result attributes.

timesketch/lib/analyzers/interface.py
```diff
@@ -1150,3 +1165,196 @@ def get_kwargs(cls):
...
+        dfiq_question_id (str): [Optional] Digital Forensics Investigative
+            Question (DFIQ) questions ID.
+        dfiq_question_conclusion (str): [Optional] DFQI questions answer/
+            conclusion.
```
berggren, line 1186: I opt to remove this for now. DFIQ is still in flux and we can add this later if needed.

An analyzer (analysis DB object) will be part of the DFIQ models, so we will have this link on that layer.

berggren (changes requested):

timesketch/lib/analyzers/interface.py
jkppr: The idea behind storing instance, sketch and timeline in the AnalyzerOutput dictionary is the option to export this easily for processing with other tools in the future.
As soon as those results are exported from Timesketch the information about instance, timeline and sketch become relevant again.

Instead of building the data during the export, I consider it as an easy solution to just include it in the output dict directly when the object is initialized.

timesketch/lib/analyzers/interface.py
jkppr: Yes, that should work as well.
```suggestion
        self.result_attributes = {}
```

timesketch/lib/analyzers/interface.py
jkppr: good catch, thanks. It is supposed to be a dictionary of key value pairs.
```suggestion
        result_attributes (dict): [Optional] A dict of key : value pairs that holds additional
```

timesketch/lib/analyzers/interface.py
jkppr: The analyzer_name is the display name and the analyzer_identifier is the unique analyzer name. The reason we did not go with "analyzer_name" as the unique identifier, is the potential for reuse of the output schema with other tools like turbinia analyzers.

timesketch/lib/analyzers/interface.py
jkppr: Yes, that should work as well.

timesketch/lib/analyzers/interface.py
jkppr: Even when the analyzer is part of the DFIQ models, this would only work TS internally. The reference would get lost when the results are exported to any other tool.

I'm fine with removing them for now, but we need to revisit how we want them added in the normed analyzer output later on again, when they become a thing.

roshanmaskey: Removed DFIQ attributes in analyzer output class.

berggren (approved):
LGTM

timesketch/lib/analyzers/interface.py
jkppr: I have added an object "platform_meta_data" to contain the platform specific fields.

jkppr (approved):
LGTM

PR #2707, opened 2023-04-25
Search with no timeline selected

### LINKED ISSUE #2682: [frontend] TS attempts a search with no timeline selected
**Describe the bug**
Non blocking bug: When a timeline is unselected in Timesketch, the current search is automatically run again to update the events shown. However, if no timeline is selected after unselecting the last (or only) one, timesketch will fall into an infinite loading loop an throw the following error:

```
POST http://127.0.0.1:5001/api/v1/sketches/3/explore/ 400 (BAD REQUEST)

Error: Request failed with status code 400
    at createError (createError.js:16:1)
    at settle (settle.js:17:1)
    at XMLHttpRequest.onloadend (xhr.js:54:1)
```

**To Reproduce**
1. Create a new sketch with at least one timeline.
2. Search for something on the timeline (e.g. `*`)
3. Unselect all timelines
4. Now Timesketch will throw the error above and go into the infinite searching loop.

**Expected behavior**
Timesketch should not attempt to search on no selected timeline. This edge case should be catched and handled.

berggren:
This PR adds a check to check that there are indices selected before doing a search. This removed the endless loading and error result.

fixes #2682

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventList.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventList.vue
@@ -650,9 +650,17 @@ export default {
       }
     },
     search: function (resetPagination = true, incognito = false, parent = false) {
+      // Exit early if there are no indices selected.
+      if (this.currentQueryFilter.indices && !this.currentQueryFilter.indices.length) {
+        this.eventList = emptyEventList()
+        return
+      }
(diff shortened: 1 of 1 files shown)


jkppr (approved):
lgtm

PR #2713, opened 2023-04-26
Saved View left panel fixes

berggren:
This PR changes how saved searches are handled in the left menu.
* No route is pushed when clicked. Using setView() instead.
* Added 3-dots menu with an option to link to the saved search

Screenshot 2023-04-26 at 09 52 07

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
@@ -30,16 +30,36 @@ limitations under the License.
     </div>
 
     <v-expand-transition>
-      <div v-show="expanded">
+      <div v-show="expanded" class="pb-2">
         <div
           v-for="savedSearch in meta.views"
           :key="savedSearch.name"
           @click="setView(savedSearch)"
           style="cursor: pointer; font-size: 0.9em"
         >
-          <v-row no-gutters class="pa-2 pl-5" :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'">{{
-            savedSearch.name
-          }}</v-row>
+          <v-row no-gutters class="py-1 pl-5 pr-3" :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'">
+            <v-col cols="11"
+              ><div class="mt-1">{{ savedSearch.name }}</div></v-col
+            >
+            <v-col cols="1">
+              <v-menu offset-y>
+                <template v-slot:activator="{ on, attrs }">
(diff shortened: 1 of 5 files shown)


timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
```diff
@@ -30,16 +30,36 @@ limitations under the License.
...
+              ><div class="mt-1">{{ savedSearch.name }}</div></v-col
+            >
+            <v-col cols="1">
+              <v-menu offset-y>
+                <template v-slot:activator="{ on, attrs }">
+                  <v-btn small icon v-bind="attrs" v-on="on">
+                    <v-icon small>mdi-dots-vertical</v-icon>
+                  </v-btn>
+                </template>
+                <v-card>
```
jkppr, line 51: What do you need the v-card component here for? This submenu should work without as well.

timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
```diff
@@ -30,16 +30,36 @@ limitations under the License.
...
+                  <v-list dense class="mx-auto">
+                    <v-list-item style="cursor: pointer" @click="copySavedSearchUrlToClipboard(savedSearch.id)">
+                      <v-list-item-title>
+                        <v-icon>mdi-link-variant</v-icon> Copy link to this search
```
jkppr, line 55: With other submenus in the Left panel (e.g. Sigma rule) we use small icons.

```suggestion
                        <v-icon small>mdi-link-variant</v-icon> Copy link to this search
```

Also, the sigma rules make use of the `v-list-item-icon` component. Which results in some space between icon and title. I'm fine with both, but we should decide on one general way for submenus.
Option A: Have icons in the `v-list-item-icon` components and separated from the `v-list-item-title` component.
Option B: Have icons within the `v-list-item-title` component.

jkppr (changes requested):
Generally looks great. Some small nits below.

timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
berggren: fixed

timesketch/frontend-ng/src/components/LeftPanel/SavedSearches.vue
berggren: Fixed, changed to be more consistent.

jkppr (approved):
lgtm

PR #2715, opened 2023-04-26
Context Search

berggren:
This PR adds context searches to the UI.

* Pivot from an event and search all timelines within a time window
* Context search result is displayed in a bottom sheet window (same as in graphs) 
* Option to "pop out" the context search and replace the current search window

Screenshot 2023-04-26 at 16 17 10

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventActionMenu.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventActionMenu.vue
@@ -14,23 +14,34 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <template>
-  <v-menu v-model="showMenu" offset-y transition="slide-y-transition">
-    <template v-slot:activator="{ on, attrs }">
-      <v-icon v-bind="attrs" v-on="on" class="ml-1">mdi-dots-vertical</v-icon>
-    </template>
-    <v-list dense class="mx-auto">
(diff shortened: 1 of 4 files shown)


jkppr (changes requested):
The approach with the bottom slider looks great! It is a very neat workflow 👍 
No blocking errors 👍 

Some UX nits:
* When using the "replace search" feature to set the time filter, it also overrides the settings on timeline visibility. After clicking the "replace search" button it activates all timelines again. As an analyst, I would expect it to keep the settings on active or disabled as before.
* A click outside of the slider does close the element. As an analyst I would expect it to stay there and let me interact with the part visible in the back while the slide is minimized or open.
* The context search slider table does not have the same fields selected as the main table. I would expect it to start with the same field selection as the main table.
* If the selection of visible fields is changed in the context search slider table and then "replace search" is triggered, it does not update the selected fields for the main table.
* Tags, Stars, Comments added in the context search slider table are not synced instantly to the main table. At least when the slider is closed I would expect it to update stars, tags and comments in the table. If using the "replace search" it does reload the table and therefore gets the new stars, tags and comments.

@berggren since there are no blocking errors, do you want to address those issues in this PR or merge it and get feature request issues assigned for the topics mentioned above?

jkppr (approved):
As discussed offline, lgtm with follow up feature request issues.

berggren:
Thanks for the thorough review, @jkppr ! I'll create FR for the feedback here.

PR #2716, opened 2023-04-27
[Analyzer] Sigma: NoneType error

jaegeral:
With the change of Sigma list to only parse the yaml when requested (due to performance issues in the WebUi) we did not check what data is needed for the analyzer. That was causing basically the Sigma analyzer to not work at all.

This now fixes the problem.

It was also found that in some instances, we still need `rule["id"]` and `rule["rule_id"]` while internally we try to use rule_uuid, it is worth to provide the `rule["id"]` as well when pulling a list.

Another long standing issue is fixed with this PR, if there are matches, it not only adds the rule_title to the events as an attribute but also the rule_id. In the past, we use to be able to lookup rules based on name (file). Now that it is switched to Database, we look up rules based on id.

This was the original error:

```
[2023-04-27 06:07:00,049] timesketch.tasks/INFO [sigma] result: Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/sigma_tagger.py", line 99, in run
    tagged_events_counter = self.run_sigma_rule(
  File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/sigma_tagger.py", line 55, in run_sigma_rule
    for event in events:
  File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/interface.py", line 996, in event_stream
    raise ValueError("Both query_string and query_dsl are missing")
ValueError: Both query_string and query_dsl are missing

During handling of the above exception, another exception occurred:

  File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/interface.py", line 1119, in run_wrapper
    result = self.run()
  File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/sigma_tagger.py", line 105, in run
    error_msg = "* {0:s} {1:s}".format(rule.get("title"), rule.get("id"))
TypeError: unsupported format string passed to NoneType.__format__
```


timesketch/lib/analyzers/sigma_tagger.py
```diff
@@ -119,7 +119,7 @@ def get_kwargs():
             sigma_rules All Sigma rules
         """
         sigma_rules = []
-        for rule in ts_sigma_lib.get_all_sigma_rules():
+        for rule in ts_sigma_lib.get_all_sigma_rules(parse_yaml=True):
```
jaegeral, line 122: this was the root cause

timesketch/lib/sigma_util.py
```diff
@@ -114,9 +115,12 @@ def enrich_sigma_rule_object(rule: SigmaRule, parse_yaml: bool = False):
     return parsed_rule
 
 
-def get_all_sigma_rules():
+def get_all_sigma_rules(parse_yaml: bool = False):
```
jaegeral, line 118: before it was not possible to make the function parse the rules, which in the analyzer way is needed. If nothing provided, the method will use the faster way. So one has to opt in to spent CPU

berggren (approved):

PR #2722, opened 2023-05-03
Fix active analyzer status check bug

### LINKED ISSUE #2673: New UI Analyzer on broken Timelines
**Describe the bug**
In the new UI it is possible to select broken Timelines (`status: fail`) in the timeline selection and also to execute analyzers on those timelines. This will result in the analyzers never finishing and hence the Analyzer result view on the left loading for infinity.

**To Reproduce**
1. Upload a broken timeline (e.g. a plaso file generated with an older plaso version)
2. Go to the Run Analyzer page
3. Select the broken timeline from the list (which should not be possible)
4. Execute any analyzer on this timeline and see the infinite loading loop.

**Expected behavior**
1. It should not be possible to select a timeline for analysis when the status is `fail`.
2. If an analyzer is run on a broken timeline anyway (e.g. via api or any other means) it should not fall into an infinite loading loop. We need to catch the edge case where the celery workers are already stopped/finished but the state of the analysis session is still in "started" and the queue is not updated anymore.

jkppr:
This PR updates the handling of the active analyzer status check.
+ Parallelize session status queries
+ Add dynamic interval per number of active sessions
+ Add timeout after 5 min
+ Fixed some logic bugs that ended in infinite loops

This should also fix #2673 
closes #2673

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
@@ -243,29 +243,33 @@ export default {
     async fetchActiveSessions() {
       const response = await ApiClient.getActiveAnalyzerSessions(this.sketch.id)
       let activeSessions = response.data.objects[0]['sessions']
-      if (activeSessions) {
+      if (activeSessions.length > 0) {
         activeSessions.forEach((sessionId) => {
           if (this.activeAnalyzerQueue.indexOf(sessionId) === -1) this.activeAnalyzerQueue.push(sessionId)
         })
+      } else {
+        this.activeAnalyzerQueue = []
       }
     },
     async fetchAnalyzerSessionData() {
       let activeAnalyzerSessionData = []
-      await this.activeAnalyzerQueue.forEach((sessionId) => {
-        ApiClient.getAnalyzerSession(this.sketch.id, sessionId).then((response) => {
+
+      const promises = this.activeAnalyzerQueue.map((sessionId) => {
+        return ApiClient.getAnalyzerSession(this.sketch.id, sessionId).then((response) => {
           let analyzerSession = response.data.objects[0]
           if (!analyzerSession) return
           activeAnalyzerSessionData.push(...analyzerSession['analyses'])
         })
       })
+      await Promise.all(promises)
       this.updateAnalyzerResultsData(activeAnalyzerSessionData)
     },
     triggeredAnalyzerRuns: function (data) {
-      console.log('triggeredAnalyzerRuns')
       this.activeAnalyzerTimerStart = Date.now()
       data.forEach((sessionId) => {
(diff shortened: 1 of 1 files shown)


timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -114,6 +128,10 @@ export default {
       analyzerResults: [],
       analyzerResultsData: {},
       activeAnalyzerQueue: [],
+      activeAnalyzerInterval: 15000,
+      activeAnalyzerTimeout: 300000,
```
berggren, line 132: I opt to add the unit here as well, e.g. milliseconds

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -274,18 +300,30 @@ export default {
...
+            if (this.activeAnalyzerTimerStart !== null && (Date.now() - this.activeAnalyzerTimerStart > this.activeAnalyzerTimeout)) {
               clearInterval(this.interval)
               this.interval = false
+              this.activeAnalyzerTimerStart = null
+              this.activeAnalyzerTimeoutTriggered = true
+              return
             }
+            // set dynamic interval
+            if (sessionQueue.length >= 10) this.activeAnalyzerInterval = 30000
+            if (sessionQueue.length >= 50) this.activeAnalyzerInterval = 60000
```
berggren, line 320: I know this is a stopgap solution before we write the new API endpoint. Please add a comment here mentioning that and a TODO to remove this later

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -114,6 +128,10 @@ export default {
       analyzerResults: [],
       analyzerResultsData: {},
       activeAnalyzerQueue: [],
+      activeAnalyzerInterval: 15000,
```
berggren, line 131:
```suggestion
      activeAnalyzerInterval: 15000, //milliseconds
```

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -274,18 +300,30 @@ export default {
...
-              this.fetchActiveSessions()
-            } else {
+            if (sessionQueue.length === 0 || this.activeAnalyzerQueue.length === 0) {
               // the queue is empty so stop the interval
+              clearInterval(this.interval)
+              this.activeAnalyzerTimerStart = null
+              this.interval = false
+              return
+            }
+            // check timeout
```
berggren, line 310: Please be a bit more descriptive. What timeout, why do we need to check it?

berggren (approved):
I left some comments and things I would like to see changed, but I LGTM and let you merge when you are ready.

PR #2724, opened 2023-05-05
Untag event API resource

jaegeral:
This is a new remove Tag APi endpoint.

Remove tags (max 500) from a list of events (max 500).

        Args:
            sketch_id: Integer primary key for a sketch database model
            in request form:
                events: list of events to remove tags from
                tags: list of tags to remove from events
                searchindex_id: the searchindex id or
                searchindex_name: the searchindex name

        Returns:
            HTTP_STATUS_CODE_OK if successful

        Raises:
            HTTP_STATUS_CODE_NOT_FOUND: if sketch or event does not exist
            HTTP_STATUS_CODE_BAD_REQUEST: if the request is malformed, e.g.
                events or tags are not lists or the number of events or tags
                exceeds the maximum allowed
            HTTP_STATUS_CODE_FORBIDDEN: if the user does not have write access
                to the sketch

### DIFF
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1301,8 +1301,6 @@ class EventTagResource(resources.ResourceMixin, Resource):
             HTTP_STATUS_CODE_NOT_FOUND: if sketch or event does not exist
         """
 
-        logger.error("TAG remove method called", exc_info=True)
-
         sketch = Sketch.query.get_with_acl(sketch_id)
         if not sketch:
             abort(HTTP_STATUS_CODE_NOT_FOUND, "No sketch found with this ID.")
@@ -1328,7 +1326,6 @@ class EventTagResource(resources.ResourceMixin, Resource):
                 "request".format(self.MAX_EVENTS_TO_TAG),
             )
 
-        # The API is agnostic and can take both here
         searchindex_id = form.get("searchindex_id")
         searchindex_name = form.get("searchindex_name")
 
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1285,20 +1285,26 @@ class EventTagResource(resources.ResourceMixin, Resource):
     @login_required
     def delete(self, sketch_id):
         """
-        Remove tags from a list of events (max 500).
+        Remove tags (max 500) from a list of events (max 500).
 
         Args:
             sketch_id: Integer primary key for a sketch database model
             in request form:
                 events: list of events to remove tags from
                 tags: list of tags to remove from events
+                searchindex_id: the searchindex id or
+                searchindex_name: the searchindex name
 
         Returns:
-            A JSON object with the event ID and the tag ID (instance of
-            flask.wrappers.Response)
+            HTTP_STATUS_CODE_OK if successful
 
         Raises:
             HTTP_STATUS_CODE_NOT_FOUND: if sketch or event does not exist
+            HTTP_STATUS_CODE_BAD_REQUEST: if the request is malformed, e.g.
+                events or tags are not lists or the number of events or tags
+                exceeds the maximum allowed
+            HTTP_STATUS_CODE_FORBIDDEN: if the user does not have write access
+                to the sketch
         """
 
         sketch = Sketch.query.get_with_acl(sketch_id)
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1386,7 +1386,13 @@ class EventTagResource(resources.ResourceMixin, Resource):
                     _event.get("_id"),
                     searchindex.index_name,
                 )
-                continue
+                abort(
+                    HTTP_STATUS_CODE_NOT_FOUND,
+                    "Unable to find event {0:s} in index {1:s} to untag".format(
+                        _event.get("_id"), searchindex.index_name
+                    ),
+                )
+
             existing_tags = result.get("_source").get("tag", [])
 
             new_tags = list(set(existing_tags) - set(tags))
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1350,8 +1350,7 @@ class EventTagResource(resources.ResourceMixin, Resource):
         for _event in events:
             # every event entry can have a dedicated searchindex_id or searchindex_name
             searchindex_id = _event.get("searchindex_id", None)
-            searchindex_name = _event.get("_searchindex_name", None)
-            searchindex_name = _event.get("_index")
+            searchindex_name = _event.get("_index", None)
 
             if not (searchindex_id or searchindex_name):
                 abort(
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1296,7 +1296,7 @@ class EventTagResource(resources.ResourceMixin, Resource):
                         (e.g. 56093e2566164c50bdf973643543571b)
                     searchindex_id: the searchindex id (e.g. 4) instead of
                         providing _index
-                tags: list of tags to remove from events
+                tags_to_remove: list of tags to remove from events
 
 
         Returns:
@@ -1336,11 +1336,11 @@ class EventTagResource(resources.ResourceMixin, Resource):
                 "request".format(self.MAX_EVENTS_TO_TAG),
             )
 
-        tags = form.get("tags", [])
-        if not isinstance(tags, list):
+        tags_to_remove = form.get("tags_to_remove", [])
+        if not isinstance(tags_to_remove, list):
             abort(HTTP_STATUS_CODE_BAD_REQUEST, "Tags need to be a list")
 
-        if len(tags) > self.MAX_TAGS_PER_REQUEST:
+        if len(tags_to_remove) > self.MAX_TAGS_PER_REQUEST:
             abort(
                 HTTP_STATUS_CODE_BAD_REQUEST,
                 "Cannot untag more than {0:d} tags in a single "
--- a/timesketch/api/v1/resources/event.py
+++ b/timesketch/api/v1/resources/event.py
@@ -1250,7 +1250,7 @@ class MarkEventsWithTimelineIdentifier(resources.ResourceMixin, Resource):
         return HTTP_STATUS_CODE_OK
 
 
-class EventTagResource(resources.ResourceMixin, Resource):
+class EventUnTagResource(resources.ResourceMixin, Resource):
     """Resource to add / remove a tag to an event."""
 
     # The maximum number of events to tag in a single request.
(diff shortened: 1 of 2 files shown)


timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,156 @@ def post(self, sketch_id):
...
+    """Resource to add / remove a tag to an event."""
+
+    # The maximum number of events to tag in a single request.
+    MAX_EVENTS_TO_TAG = 500
+
+    # The maximum number of tags in a single request.
+    MAX_TAGS_PER_REQUEST = 500
+
+    @login_required
+    def post(self, sketch_id, event_id):
```
jaegeral, line 1263: Happy to remove it again, but IMHO it would be nice for a user that this is not yet implemented and what to use in the meantime

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,156 @@ def post(self, sketch_id):
...
+
+    @login_required
+    def delete(self, sketch_id):
+        """
+        Remove tags (max 500) from a list of events (max 500).
+
+        Args:
+            sketch_id: Integer primary key for a sketch database model
+            in request form:
+                events: list of events to remove tags from
```
jkppr, line 1293: In the scope of minimizing data transfers, why does the developer need to submit the full event instead of just a list of event ids? In the rest of the method only the ` _event.get("_id")` is ever used anyway.

Benefit of the full event:
If the full event is provided, you don't need a separate parameter for searchindex_id/name, since `event._index` contains the name info already. This would also cover the use-case where tags from multiple events across multiple searchindex shall be removed. (see comment below)

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,156 @@ def post(self, sketch_id):
...
+                tags: list of tags to remove from events
+                searchindex_id: the searchindex id or
+                searchindex_name: the searchindex name
```
jkppr, line 1296: If you get the full event, just use `event.get("_index")` to get the name, instead of a separate parameter.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,156 @@ def post(self, sketch_id):
...
+                "Cannot untag more than {0:d} tags in a single "
+                "request".format(self.MAX_TAGS_PER_REQUEST),
+            )
+
+        datastore = self.datastore
+
+        for _event in events:
+            result = self.datastore.get_event(searchindex.index_name, _event.get("_id"))
+            if not result:
+                logger.warning(
```
jkppr, line 1381: This warning will not be surfaced to the end user in the frontend. If an event is provided that does not exist, I would recommend aborting instead of a warning.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,156 @@ def post(self, sketch_id):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Cannot untag more than {0:d} tags in a single "
+                "request".format(self.MAX_TAGS_PER_REQUEST),
+            )
+
+        datastore = self.datastore
+
+        for _event in events:
+            result = self.datastore.get_event(searchindex.index_name, _event.get("_id"))
```
jkppr, line 1379: If I get this part right, it will not find any event if the event is not part of the provided searchindex.

Problem: You allow for 500 events, but just one searchindex_id/name. So what happens if the event is not in this one searchindex?

Possible solution: Don't provide a static searchindex_id/name parameter and just dynamically get the searchindex for each event via something like this: 
```
searchindex = SearchIndex.query.filter_by(
                index_name = _event.get("_index")
            ).first()
```

jkppr (changes requested):
I like the slim solution and it works well.
I noticed some issues during the review, please take a look at the comments below.

timesketch/api/v1/resources/event.py
jaegeral: yep changed it

timesketch/api/v1/resources/event.py
jaegeral: ok my idea was to still continue with the provided data, but also happy to just err out here - will change it

timesketch/api/v1/resources/event.py
jaegeral: changed it

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,159 @@ def post(self, sketch_id):
...
+    def delete(self, sketch_id):
+        """
+        Remove tags (max 500) from a list of events (max 500).
+
+        Args:
+            sketch_id: Integer primary key for a sketch database model
+            in request form:
+                events: list of events to remove tags from with the following values:
+                    _id: the event id
+                    _index: the searchindex id or
```
jkppr, line 1295: iirc `_index` is the searchindex name and events don't carry the searchindex id information,

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,159 @@ def post(self, sketch_id):
...
+                "Cannot untag more than {0:d} tags in a single "
+                "request".format(self.MAX_TAGS_PER_REQUEST),
+            )
+
+        datastore = self.datastore
+
+        for _event in events:
+            # every event entry can have a dedicated searchindex_id or searchindex_name
+            searchindex_id = _event.get("_searchindex_id")
+            searchindex_name = _event.get("_searchindex_name")
```
jkppr, line 1353: I don't think `_event` has those keys, so this would fail. It contains `_index` which is the searchindex name.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,159 @@ def post(self, sketch_id):
...
+
+            if not (searchindex_id or searchindex_name):
+                abort(
+                    HTTP_STATUS_CODE_NOT_FOUND,
+                    "No search index information supplied.",
+                )
```
jkppr, line 1359: You could probably remove this check, since all events should have an `_index` field.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,159 @@ def post(self, sketch_id):
...
+
+            searchindex = None
+            # in both cases we are flexible, no matter what was supplied
+            if searchindex_name:
+                searchindex = SearchIndex.query.filter_by(
+                    index_name=searchindex_name
+                ).first()
+            elif searchindex_id:
+                searchindex = SearchIndex.query.get(searchindex_id)
```
jkppr, line 1368: I'm not sure if the event object carries the searchindex ID at all. So probably easier to just focus on the name and ignore the whole ID part.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,159 @@ def post(self, sketch_id):
...
+                abort(
+                    HTTP_STATUS_CODE_NOT_FOUND,
+                    "No search index information supplied.",
+                )
+
+            searchindex = None
+            # in both cases we are flexible, no matter what was supplied
+            if searchindex_name:
+                searchindex = SearchIndex.query.filter_by(
+                    index_name=searchindex_name
```
jkppr, line 1365: This will not work, since `searchindex_name = _event.get("_searchindex_name")` but the name is in `_event.get("_index")`

jkppr (changes requested):
some more nits below

timesketch/api/v1/resources/event.py
jaegeral: not if one was building it manually

jkppr (approved):
lgtm, thanks for building this!

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,167 @@ def post(self, sketch_id):
         )
 
         return HTTP_STATUS_CODE_OK
+
+
+class EventTagResource(resources.ResourceMixin, Resource):
```
berggren, line 1253: Can we rename this to EventUnTagResource. Two reasons:
1. We already have the old EventTaggingResource and having this is confusing
2. Move the logic from DELETE to POST. Dealing with delete requests using POST data is not recommended and goes against REST principles. I know that we are doing that in other places but I want to change that going forward.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,167 @@ def post(self, sketch_id):
...
+            HTTP_STATUS_CODE_NOT_FOUND: if sketch or event does not exist
+            HTTP_STATUS_CODE_NOT_IMPLEMENTED: if the method is not implemented
+        """
+        abort(
+            501,  # HTTP_STATUS_CODE_NOT_IMPLEMENTED
+            ("Method not yet implemented. Please use EventTaggingResource instead."),
+        )
+
+    @login_required
+    def delete(self, sketch_id):
```
berggren, line 1286: Make this a POST instead

timesketch/api/v1/resources/event.py
berggren: All events will have a _index field, regardless of where they were imported from. `_index` is a special opensearch field that is added by the system.

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,167 @@ def post(self, sketch_id):
...
+            abort(HTTP_STATUS_CODE_BAD_REQUEST, "Events need to be a list")
+
+        if len(events) > self.MAX_EVENTS_TO_TAG:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "Cannot untag more than {0:d} events in a single "
+                "request".format(self.MAX_EVENTS_TO_TAG),
+            )
+
+        tags = form.get("tags", [])
```
berggren, line 1339: rename to: `tags_to_remove`

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,167 @@ def post(self, sketch_id):
...
+                abort(
+                    HTTP_STATUS_CODE_NOT_FOUND,
+                    "Unable to find event {0:s} in index {1:s} to untag".format(
+                        _event.get("_id"), searchindex.index_name
+                    ),
+                )
+
+            existing_tags = result.get("_source").get("tag", [])
+
+            new_tags = list(set(existing_tags) - set(tags))
```
berggren, line 1400:
```suggestion
            new_tags = list(set(existing_tags) - set(tags_to_remove))
```

timesketch/api/v1/resources/event.py
```diff
@@ -1208,3 +1248,167 @@ def post(self, sketch_id):
...
+                continue
+
+            # write the new tags to the datastore
+            datastore.import_event(
+                index_name=searchindex.index_name,
+                event_id=_event.get("_id"),
+                event={"tag": new_tags},
+                flush_interval=datastore.DEFAULT_FLUSH_INTERVAL,
+            )
+            datastore.flush_queued_events()
```
berggren, line 1412: You want to do this outside of the loop. Otherwise you will flush the queue on every iteration and lose the benefit of the bulk change.

berggren (changes requested):

timesketch/api/v1/resources/event.py
jaegeral: ok cool, no aborting then, I will also not add a warning as it might clutter the logs quite a bit.

timesketch/api/v1/resources/event.py
jaegeral: yes true, but this is about potentially user supplied info, it is true that in most of our cases, this will be executed after one pulled an event. That does not rule out that someone is using the method just providing e.g. an event id but not an index.

timesketch/api/v1/resources/event.py
jaegeral: just to be clear here, that would also mean that if a user provided a event id that does not exist, nothing would happen

timesketch/api/v1/resources/event.py
jaegeral: done

timesketch/api/v1/resources/event.py
jaegeral: done

timesketch/api/v1/resources/event.py
jaegeral: done

timesketch/api/v1/resources/event.py
jaegeral: ok this method is gone now

berggren (approved):

PR #2726, opened 2023-05-06
Minor fixes/updates for AggregateDialog

### LINKED ISSUE #2685: UI bug: [Aggregation Dialog] Rare value data table overflows outside of card/dialog
Changing the # or rows in the rare values data table to greater than 10 can result in the table overflowing outside of its parent v-card and dialog.

### LINKED ISSUE #2686: UI feedback: [AggregationDialog] 500 Server error
When opening the aggregation dialog on an event, a 500 Server error will be returned when calling the aggregation API endpoint.

It appears that using the summary aggregator on certain fields, like text fields, will result in an OpenSearch timeout. 

Possible solutions:
* currently the summary aggregator tries to make multiple aggregation queries in a single call -> split these calls up and make them sequential
* increase the timeout

### LINKED ISSUE #2697: UI Bug: [AggregationDialog] Hour-Day graph displays data incorrectly
**Describe the bug**

Seems there is an off-by-one bug in the hour-day heatmap graph.  

**To Reproduce**

View any event that has >1 event count that falls on Sunday.

sydp:
This PR adds some minor fixes:

* fixes an off-by-one bug in the hour-day aggregation
* removes items per page table option for rare values.
* Separates the event time interval aggregation into a new aggregator and lazy loads the data based on the selected interval.  This is done to help overcome OpenSearch timeouts that are encountered with large data sets.

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**
* Closes #2697 
* Closes #2685 
* Closes #2686


timesketch/frontend-ng/src/components/Explore/AggregateDialog.vue
jkppr, line 1: When I try these changes in my dev env, my aggregations break with the following error in this component:
```
TypeError: Cannot read properties of undefined (reading 'buckets')
    at o.intervalChartOptions (AggregateDialog.vue:332:1)
    at t.get (vue.runtime.esm.js:3446:33)
    at t.evaluate (vue.runtime.esm.js:3547:27)
    at o.intervalChartOptions (vue.runtime.esm.js:5537:25)
    at o.ie (AggregateDialog.vue:1:7722)
    at Dn.t._render (vue.runtime.esm.js:2684:28)
    at o.r (vue.runtime.esm.js:3875:27)
    at t.get (vue.runtime.esm.js:3446:33)
    at t.run (vue.runtime.esm.js:3522:30)
    at Jr (vue.runtime.esm.js:4121:17)
```

jkppr (changes requested):
Generally looks good, I just noticed one breaking change. Please take a look below.

timesketch/frontend-ng/src/components/Explore/AggregateDialog.vue
jkppr, line 1: The DataLabels for the `Event distribution by` chart can get a bit crowded when selecting the year scope (see screenshot below). It also generates the following warning:
```
WARNING: DataLabels are enabled but there are too many to display. This may cause performance issue when rendering.
```

TS_agg_v2_labels

Can this be fixed by e.g. showing just a subset of data labels for orientation?

jkppr (changes requested):
Thanks for fixing the error. I noticed one more issue below.

timesketch/frontend-ng/src/components/Explore/AggregateDialog.vue
sydp: I set the max number of labels to 12 for this graph and on my test data/instance it doesn't overlap and I seem to no longer get that warning message.  how does it look on your end?

jkppr: I still get the warning for attributes that have data spanning many years, but this is not blocking from my point of view.
The solution with 12 labels looks great 👍

berggren (approved):

PR #2729, opened 2023-05-08
add untag_event and untag_events to API client

jaegeral:
This pull request adds the functionality to untag events in Timesketch. The new functions are:

`untag_event(self, event_id: str, index, tag: str):` - Untags the specified tag from the event with the specified ID.
`untag_events(self, events, tags_to_remove: list):` - Untags the specified tag from all of the events with the specified IDs.

# Motivation
The ability to untag events is a frequently requested feature. Currently, the only way to remove a tag from an event is to delete the event and recreate it without the tag. This is a time-consuming and error-prone process.

# Implementation
The new functions are implemented in the timesketch_api_client. They both use the `untag_events()` methods of the Timesketch API.

# Documentation
The new functions will be documented later.

# Demo

```
import timesketch_api_client

Create a client
client = timesketch_api_client.TimesketchAPIClient()
sketch = client.get_sketch(1)

Get the event ID
event_id = "k8P1MYcBkeTGnypeeKJL"

Untag the event
sketch.untag_event(event_id, "foobar")
```


jkppr (approved):
lgtm

Issue #2605, opened 2023-03-09
[API] events.py add functions for remove_label and remove_tag


jaegeral:
At the moment there is no API function (and thus no functionality in the API client, CLI client nor Web UI to remove a label and or Tag from an event.

This is becoming an increasing problem with the pre-populated labels in the new UI with good / bad / susp as there is no way to remove them once they are added to an event.

It is a known issue and movig forward we kind of want to have only one of the kind, so most likely it will be tags.

- [x] add `remove_tag` to events.py in api https://github.com/google/timesketch/pull/2724
- [x] add `remove_tag` to events.py in api-client https://github.com/google/timesketch/pull/2729
- [x] add `remove_tag` to events.py in JS-API-client https://github.com/google/timesketch/pull/2694
- [x] add `remove_tag` to events.py in Web UI  https://github.com/google/timesketch/pull/2694
- [x] add `remove_tag` to events.py in CLI client https://github.com/google/timesketch/pull/2732

PR #2732, opened 2023-05-08
Add remove_tag to CLI client

jaegeral:
This pull request adds a new option to the timesketch cli client to remove tags from an event. The new option, called `remove_tag`, takes a list of tags as its argument and removes those tags from the event. For example, the following command would remove the tags "suspicious" and "bad" from the event with the ID "12345".

```
timesketch events remove_tag --timeline-id 4 --event-id k8P1MYcBkeTGnypeeKJL --tag foobar333
200
```

https://github.com/google/timesketch/issues/2605


cli_client/python/timesketch_cli_client/commands/events.py
```diff
@@ -145,3 +145,57 @@ def add_event(ctx, message, date, attributes, timestamp_desc, output):
...
+
+    if tag:
+        # if tag is a string with commas, make it a list
+        if "," in tag:
+            tags = tag.split(",")
+            return_value = sketch.untag_events([event_id], timeline.index_name, tags)
+        else:
+            return_value = sketch.untag_event(event_id, timeline.index_name, tag)
+            click.echo(return_value)
+    else:  # just get the event
```
jkppr, line 193: Why not check if the event exists first, before going and try to untag something?

jkppr (approved):
generally looks good. Just one small question below.

cli_client/python/timesketch_cli_client/commands/events.py
jaegeral: Not sure I understand, this is to give the user the ability before removing any tags to call it without the tags argument, you will see the event, and then you can re-run it with tags to be removed.

It is independent if an event exists or not.

jaegeral: we just checked and we do the same in annotations.

PR #2733, opened 2023-05-08
New Cytoscape component (remove vue-cytoscape wrapper dependency)

berggren:
In preparation for migrating to Vue3.

This PR removes the dependency vue-cytoscape (which is a wrapper for CytoscapeJS for Vue2). The component is not maintained and there are no plans to make it compatible with Vue3.

- Remove dependence on vue-cytoscape
- Removes unused old dependencies
- Create "in-house" Cytoscape component
- Add rendering of graphs in Stories


timesketch/frontend-ng/src/components/Graph/Cytoscape.vue
```diff
@@ -0,0 +1,717 @@
...
+      edgeStyles: ['bezier', 'taxi'],
+      edgeStyle: 'bezier',
+      config: {
+        style: [
+          {
+            selector: 'node',
+            style: {
+              shape: 'roundrectangle',
+              width: 'label',
+              height: 'label',
```
jkppr, line 247: According to the warning on my console the style value of `label` is deprecated for width and height in cytoscape.
The recommended alternative I found is to measure the length of your test and then use this to determine the size.

jkppr (approved):
lgtm
Just one non blocking warning that you might want to take care of below.

timesketch/frontend-ng/src/components/Graph/Cytoscape.vue
berggren: Yeah, I opted to followup in another PRs for that. Tracking in: https://github.com/google/timesketch/issues/2743

Issue #2739, opened 2023-05-09
Pylint: Unable to import 'timesketch_api_client' (import-error)


jaegeral:
Multiple pull requests now had the following issue in the tests:

```
************* Module timesketch_cli_client.commands.search
cli_client/python/timesketch_cli_client/commands/search.py:22:0: E0401: Unable to import 'timesketch_api_client' (import-error)
Error: Process completed with exit code 2.
``` 

But nothing was changed in the API client.

jaegeral:
To mitigate at the moment the check is disabled via pylint, an example is here:

https://github.com/google/timesketch/actions/runs/4923351364/jobs/8795070221

jaegeral:
https://pipelines.actions.githubusercontent.com/serviceHosts/d395a850-1a03-4d7e-8ff4-331722f898b9/_apis/pipelines/1/runs/12073/signedlogcontent/10?urlExpires=2023-05-09T14%3A24%3A01.0196528Z&urlSigningMethod=HMACV1&urlSignature=CeoK%2Bab3CSVc85gbiU5luGAtTHAy0oG6RA44bUsSJxc%3D

jaegeral:
```
2023-05-09T15:18:46.2454835Z timesketch-api-client     20230208  /home/runner/work/timesketch/timesketch/api_client/python
2023-05-09T15:18:46.2456380Z timesketch-import-client  20210712  /home/runner/work/timesketch/timesketch/importer_client/python
```

jaegeral:
[10.txt](https://github.com/google/timesketch/files/11451750/10.txt)

jaegeral:
Ok the root cause seem to be: https://github.com/pylint-dev/pylint/issues/7701

And `SETUPTOOLS_ENABLE_FEATURES="legacy-editable"` does fix it.

jaegeral:
https://github.com/pylint-dev/pylint/issues/7306#issuecomment-1247862168

PR #2740, opened 2023-05-09
Github actions debugging

jaegeral:
Just for testing. Ignore

https://github.com/google/timesketch/issues/2739


jaegeral:
This is the actual fix for the problem: https://github.com/google/timesketch/pull/2748

Issue #2675, opened 2023-04-14
UI: Display an active analysis counter


jkppr:
At the moment we have only a loading circle displayed left to the analyzer results counter and a loading circle for each timeline that is still loading.
It would be great if:
* The loading circle on the "Analyzer Results" shows a counter of left active sessions.
* Each Analyzer also shows a loading circle if there is a matching active session (including a counter for the number of timelines processed with this analyzer at the moment)

Annoraaq:
There is still one part unresolved (the second bullet point)

PR #2749, opened 2023-05-12
Loading indicator active analyzers

Annoraaq:
Addressing the second part of #2675:

> Each Analyzer also shows a loading circle if there is a matching active session (including a counter for the number of timelines processed with this analyzer at the moment)

I added the currently active analyzers and timelines to the applications state so both the sidebar and the main content can access them in a clean way.


jkppr:
Thanks for adding this feature!

PR #2750, opened 2023-05-15
Save SearchTemplates via YAML only

berggren:
This PR disables the ability to save search templates from the API. This functionality is not yet supported in the UI and more work is needed for validating user supplied templates.

Long term solution will be tracked separately.


jaegeral (approved):
LGTM

PR #2763, opened 2023-05-23
Ensure that upload path is valid

berggren:
This PR fixes a bug where the upload path could be invalid based on the API request.
* Validate index name passed in from the request
* Make sure upload path is relative to the configured UPLOAD_PATH parameter
* Add tests


jkppr (approved):
lgtm

PR #2770, opened 2023-06-01
🐛 Handle None response from ip2geo

pemontto:
We found the GeoIP analyser would never complete with our timelines as we were getting `None` responses from ip2geo. E.g.
```python
Traceback (most recent call last): File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/interface.py", line 1134, in run_wrapper result = self.run() File "/usr/local/lib/python3.10/dist-packages/timesketch/lib/analyzers/geoip.py", line 301, in run continue TypeError: cannot unpack non-iterable NoneType object
```

This fix continues the loop early if we haven't got a result

#### Before 😢
image

#### After 🥳
image

### DIFF
--- a/timesketch/lib/analyzers/geoip.py
+++ b/timesketch/lib/analyzers/geoip.py
@@ -297,6 +297,9 @@ class BaseGeoIpAnalyzer(interface.BaseAnalyzer):
         for ip_address, ip_address_fields in ip_addresses.items():
             response = client.ip2geo(ip_address)
 
+            if not response:
+                continue
+
             try:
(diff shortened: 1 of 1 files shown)


jkppr:
Thanks @pemontto for the PR.

@sydp I'm assigning you as a reviewer since you're the expert for this analyzer. Can you please take a look at these changes and verify if we want to just continue when the response is None or handle the problem in a different way?

sydp (approved):
Lgtm

Issue #2554, opened 2023-02-13
UI feedback: keep color on deselected timelines


berggren:
**Describe the solution you'd like**
Keep color on deselected timelines, but lower transparency/saturation.

jaegeral:
I assume this ismedium effort and we could tackle it in Q2, but if not should also fine to move it to Q3

jkppr:
should be fixed with the new timeline chip design.

Issue #2774, opened 2023-06-02
Redesign of timeline chips in the new UI


jkppr:
User feedback has shown that the new design and functionality of the timeline chips in the new UI can be improved.

Goals of the redesign:
* Bring back the active/deactivated toggle with one click on the chip (see old UI).
* Move the color into a sphere on the left side to get a better and more accessible contrast for the timeline name.
* Add a three dot icon on the right side that triggers the sub menu
* Move the event counter into a small dedicated chip with good contrast (see mock). If not possible, use brackets.
* Keep the default size of timeline chips introduced with the new UI and truncate the timeline names accordingly.
  * Add a tool-tip on hover that shows the full timeline name if it is truncated.

Mock / Design idea for orientation:
![image](https://github.com/google/timesketch/assets/99879757/e61649da-9a1d-4df9-92ed-ba24de83191c)

jkppr:
See also this feature request: #2554 
It might be worth experimenting a bit with different ways to indicate and active and inactive timeline. The usage of icons for indication could be also an option.

jkppr:
Additional UX improvements that could be implemented as part of this issue (can also be a second PR):
1. For broken evidence, have the alert icon on the left side clickable and directly show the data source dialog with the error. 
2. For the submenu of broken timelines: 
  * Add some visual indicator that the error can be found in the data sources.
  * Remove the color selection for broken timelines.
3. From the timeline submenu we can probably remove the "timeline" from "Rename timeline" and "Delete timeline"?

jkppr:
I just noticed that the new design is not applied on the Analyzer page for the selection of timelines. 
re-opening this bug to adjust those timeline chips as well to the new design.

PR #2777, opened 2023-06-02
Restyle timeline chips

Annoraaq:
See #2774 

Redesigns the timeline chips.


timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
```diff
@@ -104,20 +104,42 @@ limitations under the License.
 
     <v-menu v-else offset-y :close-on-content-click="false" content-class="menu-with-gap" ref="timelineChipMenuRef">
       <template v-slot:activator="{ on }">
-        <v-chip v-on="on" :style="getTimelineStyle(timeline)" class="mr-2 mb-3">
-          <v-icon v-if="timelineStatus === 'fail'" left color="red"> mdi-alert-circle-outline </v-icon>
-          <span class="timeline-name-ellipsis">{{ timeline.name }}</span>
+        <v-chip  @click="toggleTimeline()" :style="getTimelineStyle(timeline)" class="mr-2 mb-3 pr-1 timeline-chip" >
+          <v-icon v-if="timelineFailed" left color="red"> mdi-alert-circle-outline </v-icon>
+          <v-icon v-if="!timelineFailed" left :color="timelineChipColor"> mdi-circle </v-icon>
```
jkppr, line 109: What do you think of making the timeline color spheres larger and match the border of the chip on the left side?

```suggestion
          <v-icon v-if="timelineFailed" left color="red" size="x-large"> mdi-alert-circle-outline </v-icon>
          <v-icon v-if="!timelineFailed" left :color="timelineChipColor" size="xx-large" class="ml-n3"> mdi-circle </v-icon>
```

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
```diff
@@ -104,20 +104,42 @@ limitations under the License.
...
+
+          <v-tooltip bottom>
+            <template v-slot:activator="{ on: onTooltip, attrs }">
+              <span class="timeline-name-ellipsis" :class="{ disabled: !isSelected && timelineStatus === 'ready'}"
+              v-bind="attrs"
+              v-on="onTooltip"
+              >{{ timeline.name }}</span>
+            </template>
+            <span>{{ timeline.name }}</span>
+          </v-tooltip>
```
jkppr, line 119: Is it possible to only show the tooltip when the text is actually truncated?

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
```diff
@@ -371,6 +393,18 @@ export default {
       if (this.timelineStatus === 'processing') return 'mdi-circle-slice-7'
       return 'mdi-alert-circle-outline'
     },
+    timelineFailed() {
+      return this.timelineStatus === 'fail';
+    },
+    timelineChipColor() {
+      if (this.timelineStatus === 'ready' && !this.isSelected) {
+        return 'grey'
```
jkppr, line 401: To fix #2554 as well, could you keep the timelineChipColor but instead of turning it grey if it is inactive lower transparency/saturation?

jkppr (changes requested):
Thanks for the swift turnaround on this feature! 

I have some smaller nits, please take a look below.

jkppr:
Some additional UX nits:
image

1. Is it possible to prevent the chips width from changing when active/inactive? 
2. It would be great if all chips have the same size per default (see screenshot above)
3. Currently the broken_evidence chip still shows a "pointer" and is clickable without that anything happens. Can you remove the pointer/click event for broken timelines and only allow a click on the three dot menu.
4. Broken timelines show a counter of events. Since the timeline is broken, this does not provide any benefit. What do you think about removing the counter for broken timelines at all?

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
Annoraaq: SG. Done.

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
Annoraaq: I added a delay of 300ms. It subjectively felt like a good value but we can still change it if you think it is not optimal.

I added a workaround for only showing the tooltip when the text is truncated. The problem is that the truncation is done by css. While there are some hacks that try to detect via JS when the css truncation happens, I added a simple but less precise solution: the tooltip is only shown if the timeline name has more than 30 characters. This does not always match with the css truncation, because characters have different widths.

LMK what you think.

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
Annoraaq: I changed it to keep the color. It has some transparency when disabled.

Annoraaq:
> Some additional UX nits: image
> 
> 1. Is it possible to prevent the chips width from changing when active/inactive?
> 2. It would be great if all chips have the same size per default (see screenshot above)
> 3. Currently the broken_evidence chip still shows a "pointer" and is clickable without that anything happens. Can you remove the pointer/click event for broken timelines and only allow a click on the three dot menu.
> 4. Broken timelines show a counter of events. Since the timeline is broken, this does not provide any benefit. What do you think about removing the counter for broken timelines at all?

I adressed all but point 3 (still on it).

timesketch/frontend-ng/src/components/Explore/TimelineChip.vue
jkppr: Thanks for the changes. 300ms feels very intuitive and the truncation workaround is fine with me as well.

jkppr (approved):
lgtm

PR #2782, opened 2023-06-06
DFIQ support

### LINKED ISSUE #2781: Support DFIQ
Support the upcoming DFIQ format to create Investigative Scenarios.

berggren:
This PR adds initial support for importing investigative scenarios using the DFIQ format.

To test this PR one needs to have DFIQ scenarios ready (not included). 

Screenshot 2023-06-02 at 12 59 58

closes #2781


timesketch/api/v1/resources/scenarios.py
```diff
@@ -92,84 +110,116 @@ def get(self, sketch_id):
...
 
-        This resource creates a new scenario for a sketch based on a template.
-        Templates are defined in SCENARIOS_PATH.
+        This resource creates a new scenario for a sketch based on a DFIQ template.
 
         Returns:
             A JSON representation of the scenario.
         """
         sketch = Sketch.query.get_with_acl(sketch_id)
+        dfiq = load_dfiq_from_config()
```
jkppr, line 119: Small (optional) nit: You could save on this operation in cases where the sketch does not exist or there is no access when you load the dfiq after the sketch error messages.

jkppr (approved):
generally lgtm

Just some minor optional nits below. Please take a look

timesketch/lib/dfiq.py
```diff
@@ -0,0 +1,340 @@
...
+
+        Returns:
+            A dict of DFIQ components.
+        """
+        if not yaml_data_path:
+            yaml_data_path = self.yaml_data_path
+        component_dict = {}
+        dfiq_files = os.listdir(
+            os.path.join(yaml_data_path, self.plural_map.get(dfiq_type))
+        )
```
jkppr, line 286: This line crashes the code if the `/timesketch/dfiq/` folder is empty. It needs to contain at least the required folders `approaches`,`facets`,`questions` and `scenarios`. Those folders can be empty.

I would recommend to either create those folders by default (even empty) or/and check for their existence before attempting to iterate the content of those folders. If they don't exist the code should return a usable error message instead of crashing and a 500.

berggren: Fixed now - until we merge the first scenario we are catching this. The message is shown to the user in the UI.

PR #2787, opened 2023-06-07
Fixing #2393 - analysis of broken timelines

### LINKED ISSUE #2393: Analyzer can be run on timelines that have not finished importing
It is possible to select Timelines to run analyzers while they are still importing. There should be:

a) no possibility to do that in the UI
b) a check in the analyzer code that checks for the status of a timeline before running it

jkppr:
This PR adds a check if the timeline status is not "ready" and will skipp such timelines. This will prevent any analysis to be run on timelines that are not in the ready state and therefore going stale / looping for ever.

closes #2393

### DIFF
--- a/timesketch/api/v1/resources/analysis.py
+++ b/timesketch/api/v1/resources/analysis.py
@@ -283,6 +283,14 @@ class AnalyzerRunResource(resources.ResourceMixin, Resource):
             timeline = Timeline.query.get(timeline_id)
             if not timeline:
                 continue
+            if not timeline.status[0].status == "ready":
+                logger.warning(
+                    'Timeline "{0:d}" is in status "{1:s}". Analyzer cannot be '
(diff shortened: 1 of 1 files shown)


timesketch/api/v1/resources/analysis.py
```diff
@@ -283,6 +283,14 @@ def post(self, sketch_id):
             timeline = Timeline.query.get(timeline_id)
             if not timeline:
                 continue
+            if not timeline.status[0].status == "ready":
+                logger.warning(
+                    'Timeline "{0:d}" is in status "{1:s}". Analyzer cannot be '
```
berggren, line 288: Do we need to log this state? I feel it will just clutter the logs.

berggren (changes requested):
see comment

berggren (approved):
As discussed offline - consider removing the logger. Otherwise LGTM, feel free to merge.

PR #2810, opened 2023-07-03
fixes #2809 UI bug

### LINKED ISSUE #2809: Analyzer result UI: result_summary does not exist
There was a small bug introduced in #2797 with the output for verbose analyzer output. When loading a new analyzer that supports the verbose output (e.g. SSH Bruteforce) the computed `verboseAnalyzerOutput` method can return undefined if the data is not ready yet. This will result in the following error:
```
[Vue warn]: Error in render: "TypeError: Cannot read properties of null (reading 'result_summary')"
```

jkppr:
This PR fixes the bug described in #2809 where the analyzer result list UI will throw an error if verbose analyzer results are loaded.
This fix will add checks for the usage of the computed variable and only show the results when they have finished loading.

**Closing issues**
closes #2809

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventTags.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventTags.vue
@@ -16,7 +16,6 @@ limitations under the License.
       <v-chip
         small
         class="mr-1"
-        :close="hover ? true : false"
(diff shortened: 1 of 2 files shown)


berggren (approved):
LGTM

Issue #2608, opened 2023-03-09
[CLI] Have a central option for json as output


jaegeral:
json should be a central output format that users can choose for all the calls and get all results as json

This involves adding the output format and adding the corresponding output methods in all calls where data is displayed to users.

- [x] Adding a central option for output format for JSON https://github.com/google/timesketch/pull/2819
- [x] Analyze https://github.com/google/timesketch/pull/2818
- [x] events
- [x] importer
- [x] search
- [x] sigma
- [x] sketch
- [x] timelines

jaegeral:
Ok this is now done, we now have a central way to define the output_format

PR #2819, opened 2023-07-06
[CLI] Add `output-format` as cli wide argument

jaegeral:
Adding a central flag for `output_format` like we have for `sketch` so that every command can be used.

There will be following PRs to remove the output format flag from individual modules and a PR to update the documentation.

This should not break any module but to keep it small I would like to keep this PR minimal.

So one would do:

```
timesketch --output-format text --sketch 1 analyze list
login
ntfs_timestomp
chain
tagger
ssh_sessionizer
``` 

instead of before:

```
timesketch --sketch 1 analyze list --output-format text
login
ntfs_timestomp
chain
tagger
ssh_sessionizer
```

The later will not work any more as the flags are tied to the module they are using them, here the root cli command (this is a click limitation) https://github.com/pallets/click/issues/567

Context: https://github.com/google/timesketch/issues/2608

### DIFF
--- a/cli_client/python/timesketch_cli_client/cli.py
+++ b/cli_client/python/timesketch_cli_client/cli.py
@@ -42,16 +42,26 @@ class TimesketchCli(object):
     Attributes:
         sketch_from_flag: Sketch ID if provided by flag
         config_assistant: Instance of ConfigAssistant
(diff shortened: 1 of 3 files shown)


jkppr (approved):
lgtm

PR #2821, opened 2023-07-07
[CLI] refactor to use central output-format

jaegeral:
This PR will refactor most (all?) modules in the CLI client to use the central output-format set in the context.
It will remove the need to pass and parse output format in every module making development of newer modules leaner.

It also updates the corresponding documentation for CLI client modules (where existent) and tests.

This is the main PR for https://github.com/google/timesketch/issues/2608


cli_client/python/timesketch_cli_client/commands/analyze.py
```diff
@@ -101,8 +95,7 @@ def list_analyzers(ctx, output):
             'json','text' or 'tabular'
     """
     sketch = ctx.obj.sketch
-    if not output:
-        output = ctx.obj.output_format
+    output = ctx.obj.output_format
```
jkppr, line 98: When I call `timesketch --sketch 1 analyze list` I get the following exception:
```
  File "/usr/local/bin/timesketch", line 33, in <module>
    sys.exit(load_entry_point('timesketch-cli-client', 'console_scripts', 'timesketch')())
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/src/timesketch/cli_client/python/timesketch_cli_client/commands/analyze.py", line 98, in list_analyzers
    output = ctx.obj.output_format
  File "/usr/local/src/timesketch/cli_client/python/timesketch_cli_client/cli.py", line 125, in output_format
    elif self.config_assistant.get_config("output"):
  File "/usr/local/lib/python3.10/dist-packages/timesketch_api_client/config.py", line 115, in get_config
    return self._config[name]
KeyError: 'output'
```

Looks to me like there is no default output defined?

If I run `timesketch --output-format json --sketch 1 analyze list` everything works. I would expect a default output format instead of an exception though.

cli_client/python/timesketch_cli_client/commands/events.py
```diff
@@ -108,18 +98,11 @@ def annotate(ctx, timeline_id, event_id, tag, comment, output):
...
-    help="Set output format (overrides global setting).",
-)
 @click.pass_context
-def add_event(ctx, message, date, attributes, timestamp_desc, output):
+def add_event(ctx, message, date, attributes, timestamp_desc):
     """Add an event to the sketch."""
     sketch = ctx.obj.sketch
-    if not output:
-        output = ctx.obj.output_format
+    output = ctx.obj.output_format
```
jkppr, line 105: When calling this function without any output definition I get a `KeyError: 'output'`. I would expect that the default output format is used if no format is defined on the cli.

cli_client/python/timesketch_cli_client/commands/events_test.py
jkppr, line 1: I would recommend to add a test that does not use any `--output-format` flag to catch problems with the default value as I ran into them.

cli_client/python/timesketch_cli_client/commands/search.py
```diff
@@ -160,9 +154,6 @@ def search_group(
     output_format = ctx.obj.output_format
```
jkppr, line 154: Another `KeyError: 'output'` since there is no default defined.

jkppr (changes requested):
I ran into multiple cases where the cli client crashed with `KeyError: 'output'` since it seems not to pick up on a default config for the output. 
Please fix this issue and then I'll do another round of review.

cli_client/python/timesketch_cli_client/commands/events.py
jaegeral: done

cli_client/python/timesketch_cli_client/commands/analyze.py
jaegeral: done

```
timesketch --sketch 1 analyze list
Using default output format: tabular
Name	Display Name	Is Multi
ssh_sessionizer	SSH sessions	False
sessionizer	Time based sessions	False
```

PR #2822, opened 2023-07-07
Support OpenSearch queries in DFIQ

berggren:
This PR adds support for plain opensearch queries from DFIQ. This complements the current support for search templates.
There is also some UI performance gains and refactoring.

```
        timesketch:
          - description: Remote Desktop Logons
            type: searchtemplate
            value: 08b0384b-92d8-46bb-b260-f87676afcf64
          - description: Remote Desktop Logons (query)
            type: opensearch-query
            value: 'data_type:"windows:evtx:record" AND event_identifier:4624 AND logon_type:RemoteInteractive'
```

Screenshot 2023-07-07 at 12 31 27


timesketch/frontend-ng/src/components/Scenarios/Question.vue
```diff
@@ -101,20 +124,22 @@ limitations under the License.
...
+    TsSearchChip,
     TsQuestionConclusion,
   },
   data: function () {
     return {
       expanded: false,
       fullDescription: false,
       conclusionText: '',
+      addConclusion: false,
+      opensearchQueries: [],
```
jaegeral, line 142: General comment for this PR, in other places (like https://github.com/google/timesketch/blob/master/timesketch/frontend-ng/src/components/LeftPanel/SearchTemplate.vue#L103) we are using `queryString`. So maybe we could use something similar here as well.

jaegeral (approved):
One small comment on naming variables, but overall looks good to me

timesketch/frontend-ng/src/components/Scenarios/SearchChip.vue
```diff
@@ -0,0 +1,56 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<template>
+  <v-chip x-small @click="search(queryString)">
+    {{ displayName }}
+  </v-chip>
```
jkppr, line 19: UX nit: Is there a way to keep the chip highlighted when it was clicked and the search query is in the search box? This would allow the user to understand easily where the current search came from.

timesketch/frontend-ng/src/components/Scenarios/ContextCard.vue
```diff
@@ -0,0 +1,47 @@
...
+  <v-card outlined class="mt-3 mx-3" v-if="activeContext.question">
+    <v-toolbar dense flat>
+      <strong>{{ activeContext.question.display_name }}</strong>
+      <v-spacer></v-spacer>
+      <v-btn small icon @click="$store.dispatch('clearActiveContext')" class="mr-1">
+        <v-icon>mdi-close</v-icon>
+      </v-btn>
+    </v-toolbar>
+    <v-divider></v-divider>
+    <div class="pa-4 markdown-body" v-html="toHtml(activeContext.question.description)" style="font-size: 0.9em"></div>
```
jkppr, line 26: Would it make more sense to show the approach text here instead of repeating the question?

jkppr (approved):
lgtm, just two small ux nits below.

timesketch/frontend-ng/src/components/Scenarios/Question.vue
berggren: I want to be explicit on what this is. We have queries coming from both searchtemplates and from "raw" opensearch queries.

timesketch/frontend-ng/src/components/Scenarios/SearchChip.vue
berggren: Yes, that would be great! I thought it wasn't possible but I figured it out. Added highlight color to the selected chip:

Screenshot 2023-07-07 at 15 33 25

timesketch/frontend-ng/src/components/Scenarios/ContextCard.vue
berggren: This is the description of the question that is in use. The idea is to evolve this "DFIQ card" to also show approaches etc. But that will be done later.

PR #2823, opened 2023-07-07
Add Windows Lateral Movement DFIQ scenario

berggren:
Add initial DFIQ scenario.


data/dfiq/questions/Q0001.yaml
```diff
@@ -0,0 +1,11 @@
+display_name: Who has used Remote Desktop (RDP) to login to the system?
+description: |
+  Remote Desktop Protocol (RDP) is a proprietary protocol developed by Microsoft which provides a user with a graphical
+  interface to connect to another computer over a network connection. The user employs RDP client software for this
+  purpose, while the other computer must run RDP server software.
```
jaegeral, line 5: as far as I can tell, this text is close to https://en.wikipedia.org/wiki/Remote_Desktop_Protocol should we give credit here?

jaegeral (approved):
Overall fine, only one question

data/dfiq/questions/Q0001.yaml
berggren: This text was not taken from there, but it is very similar. I'll add a link.

PR #2830, opened 2023-07-10
Update tsdev.sh to add creation of api client and cli client

jaegeral:
# Add building api client and cli client locally in tsdev to the shell script

When developing on the API client and or CLI client, both need to be rebuild on a regular base. This addition to `tsdev.sh`will help with that.
Just run `tsdev.sh build_api_cli`.


jkppr (approved):
lgtm

PR #2838, opened 2023-07-13
Always set active timelines

### LINKED ISSUE #2837: Can't search using _all when there is a deleted timeline in the sketch

berggren:
This PR fixes an issue where it was not possible to search if there are deleted timelines in the sketch.

fixes #2837


jkppr (approved):
lgtm

Issue #2558, opened 2023-02-14
UI feedback: Sort order (asc/desc) in new UI


berggren:
**Describe the solution you'd like**
Port over sort order (asc/desc) in new UI

jaegeral:
IMHO this is not blocking our Milestone https://github.com/google/timesketch/milestone/10, but having it in Q2 would be one of the things we should do

jkppr:
This was merged in #2843

PR #2843, opened 2023-07-14
Support event list sorting (asc/desc)

berggren:
This PR makes the Datetime column (server side) sortable (asc/desc).

Screenshot 2023-07-14 at 15 29 13


jkppr (approved):
Great solution for this long lasting issue with the new UI! lgtm

closes #2558

PR #2845, opened 2023-07-16
Rework comments

### LINKED ISSUE #2677: UI: Comments block a lot of space
In the new UI, the comments are displayed on the right side of the EventDetails. (see screenshot below)

Problems:
1. This occupies a lot of space that cannot be used otherwise (e.g. by displaying more data). 
2. A survey within the internal team has shown that only few events of a timeline are commented during an investigation. Therefore, it would be sufficient to show comments only if there is one.

TS_comments

Next steps:

- [x] Come up with a UX friendly mock that solves the two problems about comments above.
- [x] Implement the solution :)

jkppr:
This PR will hide the comments column in the event details per default and display a new "add comments" icon on the top instead. When clicked, the user can add a comment like it used to be. 

If the event already has a comment, it will be displayed by default.

 closes #2677

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/Comments.vue
+++ b/timesketch/frontend-ng/src/components/Explore/Comments.vue
@@ -45,9 +45,7 @@ limitations under the License.
 
             <v-card-actions v-if="comment.editable">
               <v-spacer></v-spacer>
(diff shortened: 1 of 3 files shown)


berggren (approved):
LGTM

PR #2850, opened 2023-07-19
Add TTY check for providing missing config values

### LINKED ISSUE #2849: Cannot provide missing config values if no TTY available
In scenarios where no TTY is available, TS will block on user input, but it will not be clear to the user why. This occurs when config values are missing. This problem exists for providing an auth token, and was fixed by checking for a tty first, and exiting with a useful error message if not. See: `timesketch/api_client/python/timesketch_api_client/client.py:260`. It is likely that this check just needs to be added to `timesketch/api_client/python/timesketch_api_client/config.py:configure_missing_parameters`.

ramo-j:
This change adds a check for a TTY before asking for missing config values. In scenarios where a TTY is not present (such as downstreams' dftimewolf using curses display) the UI will block indefinitely, but a user will not be able to tell why.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

Closes #2849


berggren (approved):
LGTM

api_client/python/timesketch_api_client/config.py
```diff
@@ -22,6 +22,7 @@
 import logging
 import os
 import requests
+import sys
```
berggren, line 25:
```suggestion
import sys
import requests
```

berggren (approved):

Issue #2799, opened 2023-06-20
Analyzer Result sync review and rework


jkppr:
This issue is meant to list all the problems that have been noticed with the analyzer result sync within the last weeks and track the process to fix those.

1. Review the handling of the active analyzer states between the `AnalyzerList.vue` and `AnalyzerResults.vue` component.
  * With PR #2749 we introduce handling of `activeAnalyzerTImelines` in the store. This is used to sync the "execute analyzers" buttons with the state of the analyzer. 
  * Review the Result components on the left panel and the analyzer list if the usage of the store can be used to replace the EventBus events that are currently used to reliably get a sync between active analyzers for all components.

2. Issue #2742 - even when the loading indicator on the Analyzer Results stops and disappears, it continues on the timeline level until the page is reloaded.

3. When the loading indicator disappears / is done for Analyzer Results it does not add the results to the list below automatically but needs to reload the page to see the results. This should be pushed automatically when the analyzer run finished.
  * The following GIF shows the problem with 2 & 3:
![TS_debug_gif](https://github.com/google/timesketch/assets/99879757/fdb94343-d246-4d0f-a998-a62a1203cea1)

4. Related to #2675 second part: We have an indicator on the general Analyzer Results level that shows how many analyzers are in the queue at the moment. We have an indicator on the timeline level that shows that a timeline is still being analyzed. But we miss this on the Analyzer level itself. I'm not sure if this might be too many circles but I would like to give it a try. Use-case is: An analyst can see easily which analyzers are still running and which are finished. See screenshot below for what I mean:
![image](https://github.com/google/timesketch/assets/99879757/bacafdc8-6357-42d5-b026-61f97d5d8639)

5. It would be great if we could display another icon for analyzers that have already been run for a given timeline instead of the normal play icon. The goal is to add a logic to the `AnalyzerList.vue` component that checks which analyzer did run on the selected timelines and replace the icon to a "run again" icon and maybe a pop-up asking the user if they really want to run it again. Since we can select multiple timelines and different combinations, it would also work to display the repeat icons as long as one selected timeline is already covered and allow for an "run only on timelines this was not run yet" option to the pop-up.

jkppr:
Closed via #2855 
Thanks for all the work!

PR #2855, opened 2023-07-21
Loading indicator active analyzers

Annoraaq:
Adresses all 5 points of #2799

We will keep the activeAnalyses and the finished analyses/analyzerResults in the store. That way we don't need the EventBus anymore.

As soon as there are new activeAnalyses the AnalyzerResults components starts polling the active analyses. If they change, the Analyzer results are loaded from the backend again and stored in the store.

In the AnalyzerList component, a different icon is shown if the analyzer has been run for ALL currently selected timelines.


timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResult.vue
```diff
@@ -36,8 +36,17 @@ limitations under the License.
...
-        <div class="ml-3">
-          <small>{{ Object.keys(analyzer.data.timelines).length }} </small>
+        <div v-if="isActive" class="ml-1">
+          <v-progress-circular
+            :size="20"
+            :width="1"
+            indeterminate
+          >
+            <small>{{ Object.keys(analyzer.data.timelines).length }}</small>
+          </v-progress-circular>
```
jkppr, line 46: The progress-circular seems not to be triggered here during my tests:

If I get this code right, this should address point 4 in #2799 , but in my tests I was not able to get the circle showing any activity.
image

A second problem I noticed and that is visible in the screenshot above: The progress circle is not always displayed for the timeline level either. Most times it works, but here I had still 2 feature extraction analyses running for the last 2 timelines.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -48,7 +48,7 @@ limitations under the License.
               small
               icon
               @click.stop=""
-              @click="triggeredAnalyzerRuns(activeAnalyzerQueue)"
+              @click="startPolling()"
             >
               <v-icon small>mdi-reload-alert</v-icon>
```
jkppr, line 53: I noticed a small bug when the AnalyzerResults on the left page run into the timeout and the `mdi-reload-alert` icon is displayed. Even when I trigger a new analyzer for any timeline, the AnalyzerResults progress does not restart but stays in the timeout.

In this screenshot I started the SSH analyzer after the timeout was triggered:
image

jkppr (commented):
First: Great solution with using the `store` for the analysis sessions. The whole implementation feels more solid, reactive and stable during my tests.

I ran into some smaller nits and bugs, please take a look below.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
Annoraaq: This should be resolved now, because the loading indicator on the right will disappear after x seconds. I could not artificially reproduce the issue with the timeline loading indicator on the left. Maybe you can check if you can still observe it.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -316,10 +316,25 @@ export default {
     },
     verboseAnalyzerOutput: function () {
       if (this.checkAnalyzerOutput) {
-        return JSON.parse(this.timeline.verdict)
+        // this can return null
+        const parsed = JSON.parse(this.timeline.verdict)
+      // normalize null to undefined
+        return parsed == null ? undefined : parsed;
       }
       return undefined
```
jkppr, line 324: During my tests I sometimes ran into an error, where this `verboseAnalyerOutput` function returned undefined even shortly after being assigned. Since it is not reproducible in a reliable way, my hypothesis is: There is a small chance for this function to return undefined before the analyzer output is parsed properly. 
I wonder if this could be solved with a small retry mechanism?

However, this is not blocking. If you have any idea to fix this, feel free to add it. If not, it is also good and can be fixed later on since it is not blocking.

Error message I got when triggering multiple analyzers for multiple timelines and having the results open on the left panel (including the detailed results) is below. I think this is triggered by the tooltip in line 81. `<span>Result Priority: {{ verboseAnalyzerOutput.result_priority }}</span>`

```
[Vue warn]: Error in render: "TypeError: Cannot read properties of undefined (reading 'result_priority')"

found in

---> <TsAnalyzerResultTimeline> at src/components/LeftPanel/AnalyzerResultTimeline.vue
       <TsAnalyzerResult> at src/components/LeftPanel/AnalyzerResult.vue
         <VData>
           <VDataIterator>
             <TsAnalyzerResults> at src/components/LeftPanel/AnalyzerResults.vue
               <VTabItem>
                 <VTabsItems>
                   <VNavigationDrawer>
                     <Sketch> at src/views/Sketch.vue
                       <VMain>
                         <VApp>
                           <App> at src/App.vue
                             <Root>


TypeError: Cannot read properties of undefined (reading 'result_priority')
    at Proxy.render (AnalyzerResultTimeline.vue:196:1)
    at Vue._render (vue.runtime.esm.js:2684:1)
    at VueComponent.updateComponent (vue.runtime.esm.js:3875:1)
    at Watcher.get (vue.runtime.esm.js:3446:1)
    at Watcher.run (vue.runtime.esm.js:3522:1)
    at flushSchedulerQueue (vue.runtime.esm.js:4121:1)
    at Array.eval (vue.runtime.esm.js:3143:1)
    at flushCallbacks (vue.runtime.esm.js:3065:1)
```

timesketch/frontend-ng/src/services/analyzerService.js
jkppr, line 1: What is the context for placing those two functions into a separate services js file? As far as I can see it, they are just used once by the `AnalyzerResults.vue`? 
Is there any benefit to this besides having them easily available for other components in the future?

jkppr (commented):
Awesome work over all. Just three small comments to be discussed. Afterwards we are ready to merge from my point of view. Thanks a lot for the efforts!

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
Annoraaq: Yes, it seems to be undefined in some cases.

We do not manually need to retry. If we use 'computed properties' like
```
resultPriority: function() {
    return this.verboseAnalyzerOutput && this.verboseAnalyzerOutput.result_priority
},
```

The change detection should update everything automatically as soon as  `this.verboseAnalyzerOutput` is defined.

I now fixed more places where we access `this.verboseAnalyzerOutput` in non safe ways.

jkppr (approved):

PR #2857, opened 2023-07-31
Don't exit with error if user chooses not to start timesketch

pemontto:
+ What existing problem does this PR solve? When running a non-interactive install the script shouldn't exit with an error if choosing not to start the compose project. E.g. `yes n | /tmp/deploy_timesketch.sh`
+ What new feature is being introduced with this PR? None

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.


berggren (approved):
LGTM, thanks!

PR #2866, opened 2023-08-10
Update troubleshooting.md

jaegeral:
Add a section on how to get the plaso version in Timesketch


jaegeral:
as this is only docs update, I will merge it right away

PR #2882, opened 2023-09-07
Export query result to CSV

### LINKED ISSUE #2879: UI feedback: port over the CSV export feature to the new UI
The old UI allowed for exporting CSV files of currently visible tables. It would be great to have something similar with the new UI as well.

berggren:
This PR adds back the functionality of exporting the current query results to a CSV file.

Screenshot 2023-09-07 at 14 10 37

closes #2879

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventList.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventList.vue
@@ -15,14 +15,10 @@ limitations under the License.
 -->
 <template>
   <div>
-    <v-dialog v-model="fullscreenLoading" width="700">
+    <v-dialog v-model="exportDialog" width="700">
       <v-card flat class="pa-5">
-        <v-toolbar flat>
-          <v-progress-circular indeterminate size="20" width="1"></v-progress-circular>
-          <span class="ml-5">Exporting {{ totalHits }} events</span>
-          <v-spacer></v-spacer>
-          <v-btn depressed @click="fullscreenLoading = false">cancel</v-btn>
-        </v-toolbar>
+        <v-progress-circular indeterminate size="20" width="1"></v-progress-circular>
+        <span class="ml-5">Exporting {{ totalHits }} events</span>
       </v-card>
     </v-dialog>
 
@@ -529,7 +525,7 @@ export default {
         suspicious: { color: 'orange', textColor: 'white', label: 'mdi-help-circle-outline' },
       },
       searchInProgress: false,
-      fullscreenLoading: false,
+      exportDialog: false,
       currentPage: 1,
       eventList: {
         meta: {},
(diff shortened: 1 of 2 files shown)


timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -15,6 +15,17 @@ limitations under the License.
 -->
 <template>
   <div>
+    <v-dialog v-model="fullscreenLoading" width="700">
+      <v-card flat class="pa-5">
+        <v-toolbar flat>
+          <v-progress-circular indeterminate size="20" width="1"></v-progress-circular>
+          <span class="ml-5">Exporting {{ totalHits }} events</span>
+          <v-spacer></v-spacer>
+          <v-btn depressed @click="fullscreenLoading = false">cancel</v-btn>
```
jkppr, line 24: Does "canceling" the export dialog stop the export process in any way?
I noticed that my timesketch instance gets very slow and/or freezes when I try (or accidentally) export >150k events. Even when I hit cancel right away.

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -145,6 +156,10 @@ limitations under the License.
                 </v-card>
               </v-dialog>
 
+              <v-btn icon @click="exportSearchResult()">
+                <v-icon>mdi-download</v-icon>
+              </v-btn>
```
jkppr, line 161:
```suggestion
              <v-tooltip top open-delay="500">
                <template v-slot:activator="{ on }">
                  <v-btn v-on="on" icon @click="exportSearchResult()">
                <v-icon>mdi-download</v-icon>
                  </v-btn>
                </template>
                <span>Download current view as csv</span>
              </v-tooltip>
```

UX nit: Add a tooltip to the icon.

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -514,6 +529,7 @@ export default {
         suspicious: { color: 'orange', textColor: 'white', label: 'mdi-help-circle-outline' },
       },
       searchInProgress: false,
+      fullscreenLoading: false,
```
jkppr, line 532: Why is it called `fullscreenLoading`? If I understand the code correct, it is used as v-model for the export dialog.
Wouldn't something like `exportDialog` make more sense?

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -839,10 +856,15 @@ export default {
           fileLink.href = fileURL
           fileLink.setAttribute('download', fileName)
```
jkppr, line 857: Comment on the `let fileName = 'export.zip'`, what do you think about adding timesketch in there? Like `timesketch_export.zip`? Best would also be to append a timestamp as well to prevent the pile of `export (1-100).zip` files in the Download folder.

This would allow for faster identification what those export archives in the download folder actually are.

jkppr (changes requested):
Great to have the export back :) 
Please see the comments below.

timesketch/frontend-ng/src/components/Explore/EventList.vue
berggren: It used to be a fullscreen loading element and not a dialoge. Let me rename it.

timesketch/frontend-ng/src/components/Explore/EventList.vue
berggren: Sure, this was ported over from the old code and we haven't changed that. I'll rename it.

timesketch/frontend-ng/src/components/Explore/EventList.vue
berggren: I removed the cancel function for now.

jkppr: Yeah, removing the cancel function does make sense to me as long as there will be side-effects when canceling.

timesketch/frontend-ng/src/components/Explore/EventList.vue
berggren: The new format is not timesketch_export_20230401T12.....csv

jkppr (approved):
lgtm

PR #2885, opened 2023-09-07
refactor verbose analyzer output

jkppr:
* Small refactor for the the AnalyzerOutput object
* Adding functions for more easy handling especially around adding entries to the meta attributes.
* Handling most meta entries transparent for the analyzer author.
* Updating the Domain analyzer to also make use of the verbose analyzer output!

### DIFF
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -253,15 +253,7 @@ class Event(object):
 
         # Add new tags to the analyzer output object
         if self._analyzer:
-            if "created_tags" in self._analyzer.output.platform_meta_data:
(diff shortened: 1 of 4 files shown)


timesketch/lib/analyzers/interface.py
```diff
@@ -1355,6 +1344,59 @@ def validate(self):
         except (ValidationError, SchemaError) as e:
             raise AnalyzerOutputException(f"json schema error: {e}") from e
 
+    def add_reference(self, reference):
```
berggren, line 1347: Add args to the docstring here, also applies to the other methods in this PR.

berggren (changes requested):
Looks good, but add args to docstrings

berggren (approved):
LGTM

PR #2886, opened 2023-09-08
Tag list refactor & bug fixes

### LINKED ISSUE #2795: UI feedback: Tag list in the omnibox
The tag list in the search omnibox is sorted by number of events per item. Except for Stars and Comments which are always displayed at the top.

With the introduction of the quick tags "bad", "suspicious" and "good" it would be great if they could be shown at the top of the list as well and get some small indicator to reflect their special role. (e.g. a colored icon)

TS_tag_omnibox

jkppr:
This PR refactors the tag list to use a single separate component, which makes maintenance more easy. It also introduces some UX improvements:

* Pin quick tags on top of the list, below the labels.
* Add quick tag icons from the quicktag config.
* Apply quick tags as filter chips.

image

closes #2795


berggren (approved):
LGTM

PR #2888, opened 2023-09-11
Event attribute include / exclude filters

### LINKED ISSUE #2573: UI feedback: Add a button in event row details to add filter on key / value
It would be great to have a button in Event Row details with the follwing options:

* Show events only that `have the selected KEY`, `do not have the selected key``

That could be useful to e.g. only get events that have the field hostname.

Technically it would just add an `AND _exists_:"*hostame*"` to the search query

**Screenshots**
TBD

jkppr:
This PR ports over the include & exclude feature for event attributes from the old UI. It also updates the formatting and handling of search queries when filter chips are applied.

image

I think this should close the issue below: 
closes #2573

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
@@ -59,7 +59,7 @@ limitations under the License.
                           <v-icon>mdi-filter-plus-outline</v-icon>
                         </v-btn>
                       </template>
-                      <span>'include' filter</span>
+                      <span>Filter for value</span>
                     </v-tooltip>
 
                     <!-- Exclude field:value as filter chip -->
--- a/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventDetail.vue
@@ -75,7 +75,7 @@ limitations under the License.
                           <v-icon>mdi-filter-minus-outline</v-icon>
                         </v-btn>
                       </template>
-                      <span>'exclude' filter</span>
+                      <span>Filter out value</span>
                     </v-tooltip>
 
                     <!-- Copy field name -->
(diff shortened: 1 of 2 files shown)


jkppr:
> Functionality looks good. Some UX ideas:
> 
> The + and - signs in the filter chips are a bit confusing. It feels like you shpuld click them. I opt to change this. Remove the icon for "include filters" and add a red text "Not" for exxlude filters. See how this is done in Opensearch dashboards:
>
> For the icons in the event detail view I opt to change them to magnify-minus-outline and magnify-plus-outline
> 
> WDYT?

Thanks for the input. I like the idea with the red NOT and will include it. 

About the `magnify-minus-outline` icons I'm a bit torn. They make sense from the "search" perspective. But the plus and minus are suuuper small and hard to distinguish if you have a good resolution screen (and don't zoom 200%), which is difficult from an accessibility point of view. I personally like the clean +/- icons. Looks modern and easy to distinguish to me. I could also imagine `mdi-filter-plus-outline` working, since the + and - are a bit larger. What do you think?

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
```diff
@@ -28,24 +28,76 @@ limitations under the License.
...
+                          @click.stop="applyFilterChip(key, value, 'must')"
+                          icon
+                          x-small
+                          class="mr-1"
+                          v-on="on"
+                        >
+                          <v-icon>mdi-filter-plus-outline</v-icon>
+                        </v-btn>
+                      </template>
+                      <span>'include' filter</span>
```
berggren, line 62:
```suggestion
                      <span>Filter for value</span>
```

timesketch/frontend-ng/src/components/Explore/EventDetail.vue
```diff
@@ -28,24 +28,76 @@ limitations under the License.
...
+                          @click.stop="applyFilterChip(key, value, 'must_not')"
+                          icon
+                          x-small
+                          class="mr-1"
+                          v-on="on"
+                        >
+                          <v-icon>mdi-filter-minus-outline</v-icon>
+                        </v-btn>
+                      </template>
+                      <span>'exclude' filter</span>
```
berggren, line 78:
```suggestion
                      <span>Filter out value</span>
```

berggren (approved):
Tooltip wording suggestion, otherwise LGTM

PR #2893, opened 2023-09-13
filter chip fixes

jkppr:
* Fix label formatting for stars and comments
* Apply tags from LeftPanel and Omnibox also as filter chips

### DIFF
--- a/timesketch/frontend-ng/src/views/Explore.vue
+++ b/timesketch/frontend-ng/src/views/Explore.vue
@@ -240,8 +240,9 @@ limitations under the License.
                 <v-icon v-if="chip.value === '__ts_comment'" left small>mdi-comment-multiple-outline</v-icon>
                 <v-icon v-if="getQuickTag(chip.value)" left small :color="getQuickTag(chip.value).color"
                   >{{ getQuickTag(chip.value).label }}</v-icon>
-                <span v-if="chip.operator === 'must_not' && chip.type === 'term'">
-                  <span style="color: red;">NOT </span>{{ (chip.field ? `${chip.field} : ${getTruncatedString(chip.value)}` : getTruncatedString(chip.value)) | formatLabelText }}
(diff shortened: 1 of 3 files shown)


timesketch/frontend-ng/src/views/Explore.vue
```diff
@@ -240,8 +240,9 @@ limitations under the License.
                 <v-icon v-if="chip.value === '__ts_comment'" left small>mdi-comment-multiple-outline</v-icon>
                 <v-icon v-if="getQuickTag(chip.value)" left small :color="getQuickTag(chip.value).color"
                   >{{ getQuickTag(chip.value).label }}</v-icon>
-                <span v-if="chip.operator === 'must_not' && chip.type === 'term'">
-                  <span style="color: red;">NOT </span>{{ (chip.field ? `${chip.field} : ${getTruncatedString(chip.value)}` : getTruncatedString(chip.value)) | formatLabelText }}
+                <span v-if="chip.operator === 'must_not'">
+                  <span style="color: red;">NOT </span>
+                  {{ (chip.field ? `${chip.field} : ${getTruncatedString(chip.value)}` : getTruncatedString(chip.value)) | formatLabelText }}
```
berggren, line 245: I missed this in the recent PR, but you don't have to do this in javascript. You can use CSS instead:

For example:
```
  white-space: nowrap;
  overflow: hidden;
  text-overflow: ellipsis;
  max-width: 150px;
  word-wrap: break-word;
```

This is what we do in other places, like timeline names etc.

berggren (changes requested):

timesketch/frontend-ng/src/views/Explore.vue
jkppr: Good point. Changed it

berggren (approved):
LGTM

PR #2894, opened 2023-09-15
Adding multi analyzer result support

### LINKED ISSUE #2747: UI Analyzer Results - muli support
Add support for multi analyzer results in the Analyzer UI.
* Example analyzers:
  * Feature Extractor
  * Sigma

image

jkppr:
This PR implements support for the multi analyzer results like the feature extractor in the new frontend.

* Porting multi analyzer results to the new UI.
* Updated analyzer result object formatting in the frontend to support multiple results.
* Fixed a small bug where the loading progress was not triggered.
* Removed some resolved comments.

**Screenshot**

MultiAnalyzerSupport

**Closing issues**

closes #2747

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
@@ -68,11 +68,7 @@ limitations under the License.
     </div>
 
     <v-expand-transition>
-      <div
-        v-if="!isMultiAnalyzer"
-        v-show="expanded"
-        :class="getHoverTheme"
-      >
+      <div v-if="!isMultiAnalyzer" v-show="expanded" :class="getHoverTheme">
         <v-simple-table v-if="checkAnalyzerOutput" dense class="ml-2 borderless">
           <tbody :class="getHoverTheme">
             <tr class="pr-3">
@@ -275,13 +271,10 @@ limitations under the License.
         <v-data-iterator
           :items="timeline.results"
           :items-per-page="10"
-          :hide-default-footer="(timeline.results.length < 10 ? true : false)"
+          :hide-default-footer="timeline.results.length < 10 ? true : false"
         >
           <template v-slot:default="props">
-            <div
-              v-for="(analyzer, index) in props.items"
-              :key="index"
-            >
+            <div v-for="(analyzer, index) in props.items" :key="index">
               <v-divider></v-divider>
               <v-row no-gutters class="pa-1 pl-5">
                 <span>{{ analyzer.verdict }}</span>
@@ -328,20 +321,20 @@ export default {
         // this can return null
         const parsed = JSON.parse(this.timelineFirstResult.verdict)
         // normalize null to undefined
-        return parsed == null ? undefined : parsed;
+        return parsed == null ? undefined : parsed
       }
       return undefined
     },
-    resultSummary: function() {
+    resultSummary: function () {
       return this.verboseAnalyzerOutput && this.verboseAnalyzerOutput.result_summary
     },
-    resultPriority: function() {
+    resultPriority: function () {
       return this.verboseAnalyzerOutput && this.verboseAnalyzerOutput.result_priority
     },
-    references: function() {
+    references: function () {
       return this.verboseAnalyzerOutput && this.verboseAnalyzerOutput.references
     },
-    resultStatus: function() {
+    resultStatus: function () {
       return this.verboseAnalyzerOutput && this.verboseAnalyzerOutput.result_status
     },
     getAnalyzerOutputMetaData: function () {
@@ -419,16 +412,27 @@ export default {
       }
     },
     timelineFirstResult: function () {
-      return (this.timeline && this.timeline.results.length > 0 && this.timeline.results[0]) ? this.timeline.results[0] : '... no results found'
+      return this.timeline && this.timeline.results.length > 0 && this.timeline.results[0]
+        ? this.timeline.results[0]
+        : '... no results found'
     },
     timelineCreated: function () {
       const firstEntry = this.timelineFirstResult
       if (!firstEntry) return '... invalid date'
-      const createdAt = (firstEntry.created_at && firstEntry.created_at.split('.').length) > 0 ? firstEntry.created_at.split('.')[0] : false
-      return createdAt ? createdAt : '... invalid date'
+      const createdAt =
+        (firstEntry.created_at && firstEntry.created_at.split('.').length) > 0
+          ? firstEntry.created_at.split('.')[0]
+          : '... invalid date'
+      return createdAt
     },
     getHoverTheme: function () {
-      return this.$vuetify.theme.dark ? (this.expanded ? 'dark-hover dark-bg' : 'dark-hover') : (this.expanded ? 'light-hover light-bg' : 'light-hover')
+      return this.$vuetify.theme.dark
+        ? this.expanded
+          ? 'dark-hover dark-bg'
+          : 'dark-hover'
+        : this.expanded
+        ? 'light-hover light-bg'
+        : 'light-hover'
     },
   },
   methods: {
@@ -441,7 +445,7 @@ export default {
     setSavedGraph(graphId) {
       EventBus.$emit('setSavedGraph', graphId)
     },
-    applySearch(searchQuery='', timelineId="_all") {
+    applySearch(searchQuery = '', timelineId = '_all') {
       let eventData = {}
       eventData.doSearch = true
       eventData.queryString = searchQuery
@@ -455,7 +459,7 @@ export default {
       }
       EventBus.$emit('setQueryAndFilter', eventData)
     },
-    applyFilterChip(term, termField='', termType='label', timelineId="_all") {
+    applyFilterChip(term, termField = '', termType = 'label', timelineId = '_all') {
       let eventData = {}
       eventData.doSearch = true
       eventData.queryString = '*'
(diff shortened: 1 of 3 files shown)


timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -161,7 +161,7 @@ limitations under the License.
                 <strong>Last run:</strong>
               </td>
               <td style="border: none">
-                <span> {{ timeline.created_at }} UTC </span>
+                <span> {{ timeline.results[0].created_at.split(".")[0] }} UTC </span>
```
Annoraaq, line 164: Given that this will crash if `results` or `created_at` are empty and that you also use this in another place, it would probably make sense to make it a computed property or at least a method.

Something like (untested):

```javascript
timelineCreated: function() {
  const firstResult = this.timeline && this.timeline.results.length > 0
    && this.timeline.results[0];
  if (!firstResult) return "invalid date";
  const createdAt = firstResult.created_at && firstResult.created_at.split(".").length > 0
    && firstResult.created_at.split(".")[0];
  
  return createdAt ? createdAt : "invalid date";
}
```

If we could use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) that would be even simpler.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -289,12 +289,79 @@ limitations under the License.
...
+              <td style="border: none">
+                <span> Multi analyzer</span>
+              </td>
+            </tr>
+            <tr>
+              <td style="border: none">
+                <strong>Last run:</strong>
+              </td>
+              <td style="border: none">
+                <span> {{ timeline.results[0].created_at.split(".")[0] }} UTC </span>
```
Annoraaq, line 328: See my comment above.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -264,7 +264,7 @@ limitations under the License.
               </td>
               <td style="border: none">
                 <span>
-                  {{ timeline.verdict }}
+                  {{ timeline.results[0].verdict }}
```
Annoraaq, line 267: This is similar to my other comment. You could also provide a computed property for the first result and reuse that. Then you could make sure that there actually is a first result. I am not sure how likely it is that there will be no results, though. But if there aren't the UI will crash.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -273,7 +273,7 @@ limitations under the License.
                 <strong>Last run:</strong>
               </td>
               <td style="border: none">
-                <span> {{ timeline.created_at }} UTC </span>
+                <span> {{ timeline.results[0].created_at.split(".")[0] }} UTC </span>
```
Annoraaq, line 276: See my comment above.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -289,12 +289,79 @@ limitations under the License.
...
       </div>
-      <div v-else v-show="expanded" class="ml-3 pb-1 mr-2">
-        <v-icon>mdi-alert-octagon-outline</v-icon>
-        <span class="ml-1">
-          Showing multi analyzer results is not supported in the new UI yet. Please visit the old UI to see these
-          results.
-        </span>
+      <div v-else v-show="expanded"
+        :class="
+          $vuetify.theme.dark
```
Annoraaq, line 294: That logic could go into a method or computed property, no? It would declutter the template a bit.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -289,12 +289,79 @@ limitations under the License.
...
+                ? expanded
+                  ? 'dark-hover dark-bg'
+                  : 'dark-hover'
+                : expanded
+                ? 'light-hover light-bg'
+                : 'light-hover'
+            "
+          >
+            <tr>
+              <td style="border: none">
```
Annoraaq, line 316: You could give the table a class "borderless" and then use 

```css
.borderless td {
  border: none;
}
```

Then it is easier to change, because you only have to change one part in the code.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -236,32 +238,90 @@ export default {
...
-            name: analysis.timeline.name,
-            color: analysis.timeline.color,
-            verdict: analysis.result,
-            created_at: analysis.created_at,
-            last_analysissession_id: analysis.analysissession_id,
-            analysis_status: analysis.status[0].status,
+          if (perAnalyzer[analysis.analyzer_name].analyzerInfo.is_multi) {
+            // this analyzer is a multi analyzer, so add all results for this analysis session
+            multiResults = this.getAnalyzerMultiResults(analyses, analysis.analyzer_name, analysis.analysissession_id, analysis.timeline.id)
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name] = {
```
Annoraaq, line 244: You could shorten this a bit. Then it is easier to see which parts are unique to multi analyzers:

```javascript
const analyzerInfo = {
  id: analysis.timeline.id,
  name: analysis.timeline.name,
  color: analysis.timeline.color,
  last_analysissession_id: analysis.analysissession_id,
};

if (perAnalyzer[analysis.analyzer_name].analyzerInfo.is_multi) {
  // this analyzer is a multi analyzer, so add all results for this analysis session
  multiResults = this.getAnalyzerMultiResults(analyses, analysis.analyzer_name, analysis.analysissession_id, analysis.timeline.id)
  perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name] = {
    ...analyzerInfo,
    analysis_status: multiResults.analysis_status,
    results: multiResults.results,
  }
} else {
  // this is not a multi analyzer, so add the results for this timeline
  perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name] = {
    ...analyzerInfo,
    analysis_status: analysis.status[0].status,
    results: [
      {
        verdict: analysis.result,
        created_at: analysis.created_at,
        analysis_status: analysis.status[0].status,
       },
     ]
  }
}
            

```

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -236,32 +238,90 @@ export default {
...
+          if (perAnalyzer[analysis.analyzer_name].analyzerInfo.is_multi) {
+            multiResults = this.getAnalyzerMultiResults(analyses, analysis.analyzer_name, analysis.analysissession_id, analysis.timeline.id)
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].last_analysissession_id =
+              analysis.analysissession_id
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].results[0].analysis_status = multiResults.analysis_status
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].results = multiResults.results
+          } else {
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].results[0].created_at = analysis.created_at
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].results[0].verdict = analysis.result
+            perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].last_analysissession_id =
```
Annoraaq, line 285: The `last_analysissession_id` seems to be the same in both branches of the `if-else`. So you could pull it out:

```javascript
perAnalyzer[analysis.analyzer_name].timelines[analysis.timeline.name].last_analysissession_id =
              analysis.analysissession_id
if (perAnalyzer[analysis.analyzer_name].analyzerInfo.is_multi) {
  // ...
} else {
  // ...
}
```

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -236,32 +238,90 @@ export default {
...
+            verdict: analysis.result,
+            created_at: analysis.created_at,
+            analysis_status: analysis.status[0].status,
+          })
+          status.add(analysis.status[0].status)
+        }
+      }
+      if (status.size === 1 && (status.has('DONE') || status.has('ERROR'))) {
+        return {
+          results: results,
```
Annoraaq, line 310: nit: use the short form
```javascript
 return {
  results,
  analysis_status: [...status][0],
}
```

The same for the other occurrences of `results: results`.

As an alternative you could also pull out the logic into a method that takes `status` and returns an `analysis_status` and use it like:

```javascript
return {
  results,
  analysis_status: this.analysisStatus(status),
}
```

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResultTimeline.vue
```diff
@@ -485,12 +405,35 @@ export default {
...
+        JSON.parse(this.timelineFirstResult.verdict)
         return true
       } catch (e) {
         return false
       }
     },
+    timelineFirstResult: function () {
+      return this.timeline && this.timeline.results.length > 0 && this.timeline.results[0]
+        ? this.timeline.results[0]
+        : '... no results found'
```
Annoraaq, line 417: If you return a string here, it will not work as expected.

in `timelineCreated` you check `if (!firstEntry)`, which will never be the case. To fix this, you could return undefined here. 

In any case all the places where you call `timelineFirstResult.verdict` will cause errors. So it might be a good idea to either check in those places via `this.timelineFirstResult && this.timelineFirstResult.verdict` or to create a computed property for this.

Annoraaq (approved):
Thanks!

PR #2910, opened 2023-09-27
Extract Windows event logs messages attributes

### LINKED ISSUE #2911: Windows Event Logs Message Fields Extraction
Windows event log contains details about the event in the message field (EventData XML attribute). Extracting the information in the EventData would enable analysts to query/filter event logs based on the attributes in EventData.

The figure below shows the high-level schema.

```xml
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
  </System>
  <EventData>
  </EventData>
</Event>
```

roshanmaskey:
Windows event log contains details about the event in the message field (EventData XML attribute). Extracting the information in the EventData would enable analysts to query/filter event logs based on the attributes in EventData.

The figure below shows the high-level schema.

```xml
<Event xmlns="http://schemas.microsoft.com/win/2004/08/events/event">
  <System>
  </System>
  <EventData>
  </EventData>
</Event>
```

Plaso extracts the fields and stores the values as a list in the field `strings`.

```
"strings": [ "S-1-5-18", "WIN-MDLVGLNGOM0$", "WORKGROUP", "0x00000000000003e7", "S-1-5-18", "SYSTEM", "NT AUTHORITY", "0x00000000000003e7", "5", "Advapi ", "Negotiate", "-", "{00000000-0000-0000-0000-000000000000}", "-", "-", "0", "0x000000000000026c", "C:\\Windows\\System32\\services.exe", "-", "-", "%%1833", "-", "-", "-", "%%1843", "0x0000000000000000", "%%1842" ]
```

This PR maps the extracted event log `strings` to the respective attributes using `winevt.yaml` (data/winevt.yaml`)

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

Put `closes #2911` in your comment to auto-close the issue that your PR fixes
(if such).


jaegeral:
fwiw, I think once this is in, it could mean we can make a lot of those field mappings better in https://github.com/google/timesketch/blob/master/data/sigma_config.yaml

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if not name:
+            raise ValueError("Feature name is empty")
+
+        if not config:
+            raise ValueError("Feature configuration value is empty")
+
+        source_name = config.get("source_name", None)
+        if not source_name:
+            raise ValueError("source_name is empyt")
```
jkppr, line 73: Would it make sense to check for all mandatory fields at this point and fail if any of them are not defined?

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+            raise TypeError(
+                f"Expecting source_name as a list in {name} and got {type(source_name)}"
+            )
+
+        source_name = source_name[0]
+
+        # Using -1 as a default value for event_identifier and event_version as
+        # 0 is considered as None.
+        event_identifier = int(config.get("event_identifier", -1))
+        event_version = int(config.get("event_version", -1))
```
jkppr, line 85: 1. Why do we need default values here? The `event_version` field is mandatory, so if it is not set the extraction should be skipped. The `event_identifier` is defined as "optional" but we should maybe revisit if this makes sense.

2. Using `-1` as a 0 adds an extra level of abstraction that is not necessary in my opinion. When using `int(config.get("event_version"))` and event_version is defined as 0, it should work as well. We just need to ensure to check for literal `int(0)` in the later checks to prevent interpretation as None.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if not source_name:
+            raise ValueError("source_name is empty")
```
jkppr, line 88: If `source_name` is not set, you are already raising an alert in line 73.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+
+        if event_identifier < 0:
+            raise ValueError(
+                f"event_identifier is {event_identifier}. Expecting the value >= 0"
+            )
```
jkppr, line 93: Per our current definition `event_identifier` is an optional Integer. It should therefore be enough to check if the key exists and if the value is an integer.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+            raise ValueError(f"event_version is {event_version}. Expecting value >= 0")
+
+        mappings = config.get("mapping", None)
+        if not mappings:
+            raise ValueError(f"mapping value for {name} is empty")
+
+        query = (
+            f"source_name: {source_name} AND event_identifier: {event_identifier}"
+            f" AND event_version: {event_version}"
+        )
```
jkppr, line 105: This query is missing the `provider_identifier` which is a required field per our design.

jkppr, line 105: How is this query handling multiple `provider_identifier` and `source_name`? I would recommend to have an option with an `OR` operator if multiple of those are defined.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+            attributes = {}
+
+            strings = event.source.get("strings", None)
+            if not strings:
+                logger.debug("No strings attribute in event ID: %s", event.event_id)
+                continue
+
+            for mapping in mappings:
+                attribute_name = mapping.get("name", None)
+                string_index = int(mapping.get("string_index", -1))
```
jkppr, line 122: 1. The mapping fields `name` and `string_index` are mandatory. If there is no definition this mapping should be skipped. 
2. Using `-1` here as the default value for `string_index` will result in an unexpected behavior in line 127. Since `strings[-1]` will return the last element of the string.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+                if attribute_aliases:
+                    for alias in attribute_aliases:
+                        attributes[alias] = attribute_value
+
+            event.add_attributes(attributes)
+            event.commit()
+
+            event_counter += 1
+
+        return f"{event_counter} events updated by {name}."
```
jkppr, line 144:
```suggestion
        return f"winevt feature extraction [{name}] extracted {event_counter} features."
```

For now let's keep the return value similar to the feature extraction one. Later we can update them to the new verbose output as well.

jkppr (changes requested):
Thanks for your work on this feature. Looking forward to the great possibilities we will get with the mapping.

Please see the comments below. If you have questions regarding the multi analyzer approach, let me know.

data/winevt.yaml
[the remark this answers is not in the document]
roshanmaskey: Agreed. Renamed `winevt.yaml` to `winevt_features.yaml`.

timesketch/lib/analyzers/feature.py
```diff
@@ -0,0 +1,58 @@
...
+
+        results = []
+
+        for feature_plugin in self._feature_plugins:
+            result = feature_plugin.run_plugin()
+            if not result:
+                logger.debug("No plugin result for %s", feature_plugin.NAME)
+            results.append(result)
+
+        return "\n".join(results)
```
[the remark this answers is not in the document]
roshanmaskey: Updated to the code to run multiple celery tasks - one celery task per feature.

timesketch/lib/analyzers/feature_plugins/winevt.py
```diff
@@ -0,0 +1,172 @@
...
+            # function `extract_features`.
+            feature_name = feature.get("feature", None)
+            feature_config = feature.get("feature_config", None)
+
+            result = self.extract_features(feature_name, feature_config)
+            if not result:
+                logger.debug("No result for plugin run for %s", feature_name)
+            results.append(result)
+
+        return "\n".join(results)
```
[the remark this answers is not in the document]
roshanmaskey: Updated the code to run multiple celery tasks.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: Removed default value for `event_identifier` and `event_version`. Both `event_identifier` and `event_version` are required and will throw a `TypeError` or `ValueError` depending on the situation.

Since, the feature extractions are registered as a new celery task, it will only affect the feature extraction that is missing the fields or have invalid value type.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: As per our offline discussion, this is handled in a new function `validate_feature_config`. It will only raise an exception if `source_name` and `provider_identifier` are not in the feature configuration file i.e. `winevt_features.yaml`.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: This is mandatory for the events that we are interested in. This is handled by using `int(config.get("event_identifier"))`.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: As per our offline discussion, this is handled in `validate_feature_config`.

roshanmaskey: As per offline discussion, handled in `winevt_features.py#L108`.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: Done. Checking for `name` and `string_index`  in `winevt_features.py#L139` and ``winevt_features.py#L145`.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: Done.

timesketch/lib/analyzers/feature_plugins/winevt.py
roshanmaskey: Done. Handled in `validate_feature_config`.

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+        self.validate_feature_config(name, config)
+
+        source_name = config.get("source_name", [""])[0]
+        provider_identifier = config.get("provider_identifier", [""])[0]
+
+        # If event_identifier and event_version does not exist or invalid value type,
+        # it will throw exception.
+        #
+        # Note: Each feature extraction is run as a new celery task so it only affects
+        # specific feature extraction.
```
jkppr, line 99: Moving this into the `validate_feature_config` function.

```suggestion
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+                extraction.
+
+        Returns:
+            str: Returns summary of the feature extraction.
+        """
+        return self.extract_features(name, config)
+
+    def validate_feature_config(self, name: str, config: dict) -> None:
+        """Validates the name and configuration.
+
```
jkppr, line 53: Moving the comment from line 95-99 here. A bit rephrased.

```suggestion
           If any of the required properties does not exist or is of the wrong
           type, a ValueError is raised and the feature definition is not extracted.
           This does not affect other feature extractions since each feature is
           executed in their own celery task.

```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+        #
+        # Note: Each feature extraction is run as a new celery task so it only affects
+        # specific feature extraction.
+        event_identifier = int(config.get("event_identifier"))
+        event_version = int(config.get("event_version"))
+
+        mappings = config.get("mapping")
+
+        if not mappings:
+            raise ValueError(f"mapping value for {name} is empty")
```
jkppr, line 106: This is not needed anymore since it is covered by the `validate_feature_config` function.

```suggestion
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+
+        # Building search query - Use source_name or provider_identifier and not both.
+        query = ""
+        if source_name:
+            query = f"source_name:{source_name}"
+        elif provider_identifier:
+            query = f"provider_identifier:{provider_identifier}"
```
jkppr, line 113: If both are provided, why not use both?

Also adding quotes to the values to prevent search errors!

```suggestion
        # Building search query
        query = ""
        if source_name and provider_identifier:
            query = f'source_name:"{source_name}" AND provider_identifier:"{provider_identifier}"'
        elif source_name:
            query = f'source_name:"{source_name}"'
        elif provider_identifier:
            query = f'provider_identifier:"{provider_identifier}"'
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+        for event in events:
+            attributes = {}
+            strings = event.source.get("strings", None)
+
+            if not strings or not isinstance(strings, list):
+                logger.debug(
+                    "Missing or invalid strings field in the event. Skipping the"
+                    " event %s.",
+                    event.event_id,
+                )
```
jkppr, line 134: Making the debug message tied to a specific feature definition.

```suggestion
                logger.debug(
                    "[%s] Missing or invalid strings field in the event. Skipping the"
                    " event %s.",
                    name,
                    event.event_id,
                )
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+
+                try:
+                    string_index = int(mapping.get("string_index"))
+                except (TypeError, ValueError) as e:
+                    logger.debug(
+                        "Skipping event %s. Missing string_index or invalid value. %s",
+                        event.event_id,
+                        str(e),
+                    )
+                    continue
```
jkppr, line 153: These checks should be done in the `validate_feature_config` function. If a feature definition is not following the defined schema, we should not map any of the values.

```suggestion
                string_index = int(mapping.get("string_index"))
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+                    attribute_value = strings[string_index]
+                except IndexError:
+                    logger.warning(
+                        "The index %d for field %s does not exist in strings. "
+                        "Skipping the event %s",
+                        string_index,
+                        attribute_name,
+                        event.event_id,
+                    )
+                    continue
```
jkppr, line 167: With my test data I ran into this warning for the feature extraction `security_4688_v2`. At least in my test winevt file I have multiple 4688 events that do not have index 9 to 14. 

If this happens we have 3 options in my opinion:

A) Raising a `ValueError` or `IndexError` exception and not mapping anything for this event.
B) Map only the fields that exist and ignore the ones with a missing index. 

My take: I would go with option B but surface the problem to the user by adding a comment to the event for each index that was not mapped. This allows for two things: 1. The analyst can check with the Timesketch operator if the feature definition needs adjusting. 2. The analyst is aware that for this specific event there might be some errors in the extraction.

My suggestion:
```suggestion
                    logger.warning(
                        "[%s] The index '%d' for field '%s' does not exist in strings. "
                        "Skipping the event '%s'",
                        name,
                        string_index,
                        attribute_name,
                        event.event_id,
                    )
                    event.add_comment(
                        f"Analyzer[{self.NAME}]: [{name}] The index '{string_index}' for field '{attribute_name}' does not exist in strings!"
                    )
                    continue
```

What do you think?

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+                if attribute_aliases:
+                    for alias in attribute_aliases:
+                        attributes[alias] = attribute_value
+
+            event.add_attributes(attributes)
+            event.commit()
+            event_counter += 1
+
+        logger.debug("%d features extracted using feature %s", event_counter, name)
+        return f"winevt feature extraction [{name}] extracted {event_counter} features."
```
jkppr, line 179:
```suggestion
        return (
            f"winevt feature extraction: [{name}] extracted {event_counter} features."
        )
```

timesketch/lib/analyzers/feature_plugins/winevt_features.py
```diff
@@ -0,0 +1,204 @@
...
+
+        Returns:
+            List[dict]: A list of dict containing features name and configuration.
+        """
+        # config_file winevt.yaml is located within the timesketch/data directory.
+        config_file = "winevt_features.yaml"
+
+        features_config = base_interface.get_yaml_config(config_file)
+        if not features_config:
+            logger.debug("No feature configuration data in %s", config_file)
```
jkppr, line 193: I think this can be a warning. If the configuration file does not exist, this should be surfaced in the logs to be found when someone comes looking for it without changing the logging level.

```suggestion
            logger.warning("No feature configuration data in %s", config_file)
```

timesketch/lib/analyzers/feature.py
```diff
@@ -0,0 +1,128 @@
...
+            str: A summary of sketch analyzer result.
+        """
+        try:
+            plugin_class = feature_manager.PluginManager.get_plugin(
+                self._plugin_name, self
+            )
+            if not plugin_class:
+                raise ValueError(
+                    f"Plugin class for plugin name {self._plugin_name} does not exist"
+                )
```
jkppr, line 97: If the `winevt_feature.yaml` config file does not exist this ValueError is raised with `self._plugin_name` being `None`. 
In the UI this is surfaced with the error message ` Error: Plugin class for plugin name None does not exist ` in the analyzer results. 

Can this error message be a bit more actionable with regards to what is missing and needs to be done? (e.g. check if your winevt_feature.yaml file exists.)

jkppr (changes requested):
Second round:
This works really well already! Main comments are around wording of errors and checking required dict entries.

What are your plans for moving the existing `feature_extraction.py` into the plugin system?

If you are fine with it, I would like to test both changes together, so either merge it in one PR or in two dependent PRs.

timesketch/lib/analyzers/feature_plugins/winevt_features.py
roshanmaskey: I like the idea of mapping the fields that exist and adding a comment.

timesketch/lib/analyzers/feature.py
roshanmaskey: Added code to check the value of `self._plugin_name` before processing it. Updated the error message to provide more information.

data/winevt_features.yaml
```diff
@@ -0,0 +1,560 @@
...
+      string_index: 0
+    - name: image_path
+      string_index: 1
+    - name: service_type
+      string_index: 2
+    - name: start_type
+      string_index: 3
+    - name: account_name
+      string_index: 4
+
```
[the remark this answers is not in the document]
jkppr: The updated config file is still missing in the latest state of this PR. Any reason against merging the suggested version above?

jkppr:
Notes from offline sync: 
* This PR will merge the new feature extraction plugins currently including the (old) regex extractor and the new winevt mapping extractor.
* The old feature extractor analyzer will be removed in a next separate PR.
* Documentation will be updated in a separate PR as well.

jkppr (approved):

PR #2913, opened 2023-09-27
Deprecate Sigma status CSV usage from code

jaegeral:
Removing some left overs from the old File based Sigma usage.

As the sigma rule status csv is not maintained, it is removed

If there is a new need / desire to track status of upstream Sigma rules and which work in Timesketch / which not, a new way needs to be established, a CSV does not play nice with the community.


berggren:
If the CSV file has been deprecated can you add a note in it that says so? Then we can mark it for deletion in a couple of months.

contrib/sigma_rule_status.csv
jkppr, line 1: Why would we want to move the csv into the contrib folder, now that it is considered deprecated? Shouldn't you just delete it?

jaegeral: well it is still maybe valuable to some to see which / why certain rules do not work (and it is also a tracker for myself which things need more love)

jkppr: I think I understand your use-case. These are my thoughts:

* Keeping the CSV for tracking only while the data itself can be changed quite quickly in the UI and database will result in an asynchronous status pretty fast.
* If it is needed for documentation purposes I would expect to find it in the documentation sections and not under contribution.
* Since the csv is not used in the code anymore, it is unclear to the viewer what they are looking at.

May I propose an alternative approach here:

* Add an export function into tsctl that will generate this csv based on up-to-date information from the database.
* This would allow you and every other Timesketch operator to have one source of truth on the state of sigma rules in their own Timesketch instance.
* If you want to have a general status tracker for the broader project, you can export it easily into the documentation from here. 

What do you think?

jaegeral:
> If the CSV file has been deprecated can you add a note in it that says so? Then we can mark it for deletion in a couple of months.

clarified in the PR description, the usage of the file has been deprecated, the content of the file itself might still be valuable for me / others, thus moving it to contrib outside of the data folder, as the file itself is no longer used in the code.

jkppr (approved):

PR #2919, opened 2023-09-30
Add Timeline, SearchIndex and Datasource creation to client api

### LINKED ISSUE #2918: Timeline creation in the api client
Our infrastructure uses plaso (psort.py) to load data into OpenSearch. At the moment, we've been using the `generate_timeline_from_es_index` function to load the data into timesketch. We'd rather use the `opensearch_ts` module so we don't need another pass over the data. However, for this, we need to have a timeline id to pass along. In general, it would be nice to create searchindexes and timelines via the api client.

Having looked through the code, my suggestion would be to split out the code that creates the timeline and searchindex in the `generate_timeline_from_es_index` function so these functions can also be called directly. That way, a timeline can be created via the api and passed along to psort.py to send data directly to the right timeline.

I've created a PR that fits my suggestion, but it's the first time working on this codebase so I am happy to amend anything, or throw it all away if a different approach is better! Happy to contribute either way.

Tijnoz:
This PR splits out the code to create the timeline, searchindex and datasource in the `generate_timeline_from_es_index` function so they can be called directly as well, which supports different usecases as described in #2918.

I'll reiterate: Happy to amend if this is not in line with the codebase, or go at this a different way!

closes #2918


Tijnoz:
@jaegeral I have amended the code to conform to the pylint spec (based on the error it gave before). Hope you can rerun the workflows for me.

jaegeral:
Ok I added typing to the PR to comply with our latest lint requirements

Tijnoz:
Thanks @jaegeral. If there's anything I can help with or add, let me know. Happy to bring this one over the line!

jaegeral:
> Thanks @jaegeral. If there's anything I can help with or add, let me know. Happy to bring this one over the line!

Thx, nah should be fine, I am just doing it for you, you waited way to long for that review. 
FWIW, the PR itself is fine, just want to make the tests happy and then we will get it in.

Again sorry that it took so long

jaegeral (approved):
Approved many thanks for the refactoring

PR #2929, opened 2023-10-09
Refactor base layout

berggren:
This PR cleans up our base layout, moving the nav drawer (left pane) under the main bar. This gives us a correct way to use the layout system. Earlier we were using the `app` parameter on the drawer and the top bar inside of the `v-main` element and this is discoraged by the framework and can introduce subtle bugs.

* Move top bar and left panel outside of the main component
* Each root component view (home and sketch) implements the top bar and left panel
* Removed the `Back to explore` button and added a section in the left panel at the top called `Search` that will always take you back to the main search interface.

Screenshot 2023-10-09 at 12 37 41

### DIFF
--- a/timesketch/frontend-ng/src/views/Sketch.vue
+++ b/timesketch/frontend-ng/src/views/Sketch.vue
@@ -247,17 +247,7 @@ limitations under the License.
       <v-divider></v-divider>
       <v-tabs-items v-model="leftPanelTab">
         <v-tab-item :transition="false">
-          <router-link
-            :to="{ name: 'Explore', params: { sketchId: sketchId } }"
-            tag="div"
-            class="pa-4"
-            style="cursor: pointer"
-            :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'"
-          >
-            <span> <v-icon left>mdi-magnify</v-icon> Search </span>
-          </router-link>
-          <v-divider></v-divider>
-
+          <ts-search></ts-search>
           <ts-saved-searches v-if="meta.views"></ts-saved-searches>
           <ts-data-types></ts-data-types>
           <ts-tags></ts-tags>
@@ -376,6 +366,7 @@ import TsSigmaRules from '../components/LeftPanel/SigmaRules'
 import TsIntelligence from '../components/LeftPanel/ThreatIntel'
 import TsGraphs from '../components/LeftPanel/Graphs'
 import TsStories from '../components/LeftPanel/Stories'
+import TsSearch from '../components/LeftPanel/Search'
 import TsUploadTimelineForm from '../components/UploadForm'
 import TsShareCard from '../components/ShareCard'
 import TsRenameSketch from '../components/RenameSketch'
@@ -397,6 +388,7 @@ export default {
     TsIntelligence,
     TsGraphs,
     TsStories,
+    TsSearch,
     TsAnalyzerResults,
     TsEventList,
   },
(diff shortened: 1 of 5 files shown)


timesketch/frontend-ng/src/views/Sketch.vue
```diff
@@ -355,11 +202,163 @@ limitations under the License.
...
+
+      <v-tabs v-model="leftPanelTab" grow>
+        <v-tab v-for="item in leftPanelTabItems" :key="item"> {{ item }} </v-tab>
+      </v-tabs>
+      <v-divider></v-divider>
+      <v-tabs-items v-model="leftPanelTab">
+        <v-tab-item :transition="false">
+          <router-link
+            :to="{ name: 'Explore', params: { sketchId: sketchId } }"
+            tag="div"
```
jkppr, line 252: According to the console error and [vue docs](https://router.vuejs.org/guide/migration/#Removal-of-event-and-tag-props-in-router-link-) the `tag` prop is deprecated and will be removed in Vue Router 4. Use the `v-slot` API instead.

timesketch/frontend-ng/src/views/Sketch.vue
```diff
@@ -355,11 +202,163 @@ limitations under the License.
...
+            class="pa-4"
+            style="cursor: pointer"
```
jkppr, line 254: UX nit: When the user is in the explore/search view they can click the "Search" menu entry again without anything happening nor the entry indicating that it is active.
As an alternative example: When going to the Analyzer or Sigma canvas elements, the `+` stayed grey marked as indicator that this is the currently selected menu item.

A possible option would be a grey highlight for the search entry in the left panel if the canvas is showing the explore view.

timesketch/frontend-ng/src/views/Sketch.vue
```diff
@@ -42,211 +42,58 @@ limitations under the License.
...
-            <ts-scenario v-for="scenario in hiddenScenarios" :key="scenario.id" :scenario="scenario"></ts-scenario>
-          </div>
-        </v-tab-item>
-      </v-tabs-items>
-    </v-navigation-drawer>
-
-    <!-- Top horizontal toolbar -->
-    <v-app-bar v-if="!loadingSketch" app hide-on-scroll clipped flat :color="$vuetify.theme.dark ? '#121212' : 'white'">
-      <v-btn icon v-show="!showLeftPanel && !loadingSketch" @click="toggleLeftPanel" class="ml-n1">
+      <v-btn icon @click="toggleLeftPanel">
```
jkppr, line 65: What do you think about making the menu icon change depending on the state of the left panel. This way it is more clear what it does from my point of view. E.g. by using `mdi-menu-close` if the left panel is visible.

jkppr (changes requested):
1 deprecation bug and 2 UX nits. Everything else looks good.

timesketch/frontend-ng/src/views/Sketch.vue
berggren: Good catch, done.

timesketch/frontend-ng/src/views/Sketch.vue
berggren: Ack, will consider this for the future. The current

berggren: Good point. I have implemented this now, with a highlight when the user is in any search context.

jkppr (approved):

PR #2930, opened 2023-10-09
Update yeti analyzer

tomchop:
Update the Yeti analyzer to use the v2 API that will be released soon.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added.
- [x] Documentation updated.

PR #2932, opened 2023-10-10
Adjusting regular expressions for features extraction

tomchop:
Some changes to the regexes in features.yaml, mostly relaxing them:

* Removed the header check for sshd/pid, given that there were two versions: `\[sshd, pid: \d+\]` and `\[sshd\] \[\d+\]`
* Removed the `.*` at the start of some regexes which must've made them much slower.


jkppr (changes requested):
Can you please decouple the changes to the features.yaml from the changes to the yeti analyzer? 
Have this PR just focused on the features changes and add the changes for the analyzer in your #2930 PR

tomchop:
I sent it too soon because I'm working out of the same branch, but it should get decoupled as soon as the other one is merged. :)

berggren (approved):

jkppr (approved):

PR #2937, opened 2023-10-11
ENable passing of "password" parameter programatically

tomchop:
Small fix to enable passing of the password parameter programatically, e.g.

```python
config_dict = {
  'host_uri': "http://timesketch-dev:5000",
  'username': "dev",
  'auth_mode': "userpass",
  'password': "dev",
}
ts_client = config.get_client(token_password='TOKEN_PASSWORD', config_dict=config_dict)
```

### DIFF
--- a/api_client/python/timesketch_api_client/config.py
+++ b/api_client/python/timesketch_api_client/config.py
@@ -61,6 +61,7 @@ class ConfigAssistant:
             "client_secret",
         ]
     )
+    USERPASS_AUTH_NEEDED = frozenset(['password'])
 
     EXTRA_FIELDS = frozenset(
         [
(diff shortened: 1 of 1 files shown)


tomchop:
Actually, the same can be achieved doing

```
ts_client = client.TimesketchApi("http://timesketch-dev:5000", "dev", "dev")
```

I don't know if it's worth keeping this PR.

jaegeral:
I opt not and keep the current things. I admit it is not really intuitive to use, that is why https://github.com/google/timesketch/pull/2938 should expose it in the doc how to do it.

tomchop:
Ack, closing it in that case.

PR #2959, opened 2023-10-25
Sanitise HTML from Unfurl

berggren:
* Small change to the layout of the unfurl graph.
* Sanitise html before render.

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
+++ b/timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
@@ -12,66 +12,57 @@ limitations under the License.
 -->
 <template>
   <v-card>
-    <v-card-title>
-      <v-img :src="getUnfurlLogo" max-height="32" contain class="mb-2"></v-img>
-    </v-card-title>
+    <v-toolbar flat color="transparent">
+      <v-img position="left" :src="getUnfurlLogo" max-height="24" contain class="ml-2"></v-img>
+    </v-toolbar>
+
     <v-card-subtitle class="pt-1">
-      <span><b>Input:</b> {{ url }}</span>
+      <div class="mb-2"><b>Input:</b> {{ url }}</div>
+      <div v-if="nodeContext">
+        <b>Selected node info: </b>
+        <span v-html="sanitizeHtml(nodeContext)"></span>
+      </div>
     </v-card-subtitle>
+
     <v-card-text>
       <div v-show="!unfurlReady">
         <v-progress-linear color="primary" indeterminate> </v-progress-linear>
       </div>
-      <v-divider></v-divider>
-      <v-container fluid v-show="unfurlReady">
+
+      <v-card v-show="unfurlReady" outlined class="mt-n5">
+        <v-toolbar dense flat color="transparent">
+          <v-spacer></v-spacer>
+          <v-tooltip top open-delay="500">
+            <template v-slot:activator="{ on }">
+              <v-btn icon v-on="on" @click="resizeCanvas()">
+                <v-icon>mdi-fit-to-page-outline</v-icon>
+              </v-btn>
+            </template>
+            <span>resize graph</span>
+          </v-tooltip>
+          <v-tooltip top open-delay="500">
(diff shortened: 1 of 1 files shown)


timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
```diff
@@ -12,66 +12,57 @@ limitations under the License.
...
-      <v-img :src="getUnfurlLogo" max-height="32" contain class="mb-2"></v-img>
-    </v-card-title>
+    <v-toolbar flat color="transparent">
+      <v-img position="left" :src="getUnfurlLogo" max-height="24" contain class="ml-2"></v-img>
+    </v-toolbar>
+
     <v-card-subtitle class="pt-1">
-      <span><b>Input:</b> {{ url }}</span>
+      <div class="mb-2"><b>Input:</b> {{ url }}</div>
+      <div v-if="nodeContext">
```
jkppr, line 21:
```suggestion
      <div>
```

I would highly recommend to keep the default call to action `Select a node in the graph below to get more information.` visible when the dialog opens the first time.

My user tests have shown that the call to action is needed for users to click the graph nodes and to get the best experience and information from unfurl.

timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
```diff
@@ -12,66 +12,57 @@ limitations under the License.
...
+
     <v-card-text>
       <div v-show="!unfurlReady">
         <v-progress-linear color="primary" indeterminate> </v-progress-linear>
       </div>
-      <v-divider></v-divider>
-      <v-container fluid v-show="unfurlReady">
+
+      <v-card v-show="unfurlReady" outlined class="mt-n5">
+        <v-toolbar dense flat color="transparent">
```
jkppr, line 33: 1. The transparent toolbar seems not to be doing anything since the graph canvas stops right below it. So this looks like a white box even with a transparent bg.
image

2. The toolbar in the graph canvas makes the graph not controllable in this white strip which is unintuitive, since I would expect to have control on the graph in the whole box.

jkppr (changes requested):
The graph controles inside the graph canvas look great, but they impact the space to operate the graph a lot due to the white blocking bar at the top. I would argue we move them out of the canvas to the top right or add a dotted border in the bottom of the toolbar.

timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
berggren: I see. This is the recomeded way to lay out the controls on the right. I'll see if I can make them float instead

timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
berggren: fair point, I'll add it back

timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
berggren: Moved them outside of the box for now.

timesketch/frontend-ng/src/components/Explore/UnfurlDialog.vue
```diff
@@ -12,66 +12,61 @@ limitations under the License.
...
       </div>
-      <v-divider></v-divider>
-      <v-container fluid v-show="unfurlReady">
-        <!-- Cytoscape container -->
-        <div style="font-size: medium" class="py-1 px-1">
-          <span style="font-weight: bold">Selected node context: </span>
-          <br />
-          <span style="font-style: italic" v-html="nodeContext"></span>
+
+      <v-toolbar dense flat color="transparent" class="mt-n8">
```
jkppr, line 33:
```suggestion
      <v-toolbar v-show="unfurlReady" dense flat color="transparent" class="mt-n8">
```

because: 
image

jkppr (approved):
One last UI issue to fix. Then lgtm

PR #2969, opened 2023-11-02
Removing old feature extractor analyzer

jkppr:
This PR removes the "old" feature extractor analyzer and replaces it with the new feature extraction plugin analyzer. The "old" regex based analyzer was moved into the new plugin one as part of #2910 .

* Removes the "old" feature extraction analyzer.
* Removes the old `feature.yaml` config to `regex_features.yaml` to better reflect its purpose.
  * The analyzer is backwards compatible with the old filename. The new one will be preferred though. 
  * Replaced `features.yaml` with `regex_features.yaml` in all set-up/deployment/config files.
  * Added `winevt_features.yaml` to all deployment scripts.
* Removes regex based extractions from the `regex_features.yaml`config that are now covered by the new `winevt_features.yaml` config.
* Takes care of dependency naming within the analyzers.
* Flips the run order for plugins: regex first, then the rest.
* Adds a small check in the winevt extraction plugin to add info comments only if they do not exist yet.


jkppr:
@roshanmaskey fyi

timesketch/lib/analyzers/feature_extraction_plugin.py
berggren, line 1: I would keep this named as `feature_extraction.py`. It's not a plugin, it's the main analyzer that runs all the plugins, right?

berggren (changes requested):

timesketch/lib/analyzers/feature_extraction_plugin.py
jkppr: Sure, works for me as well. I changed it.

berggren (approved):

PR #2979, opened 2023-11-14
SQLalchemy upgrade - step one

berggren:
Upgrading sqlalchemy. This PR prepare the current v1.4 codebase to be compatible with v2.0. Below are the tasks in the migration work that has been addressed with this PR.

### Prerequisites
1. Prerequisite Step one - A Working 1.3 Application ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#first-prerequisite-step-one-a-working-1-3-application)). 
2. Prerequisite Step two - A Working 1.4 Application ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#first-prerequisite-step-two-a-working-1-4-application)).

### Migrate to SA 2.0
1. Migration to 2.0 Step One - Python 3 only (Python 3.6 minimum) ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-one-python-3-only-python-3-6-minimum)).
3. Migration to 2.0 Step Two - Use RemovedIn20Warnings ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-two-turn-on-removedin20warnings))
4. Migration to 2.0 Step Three - Resolve all RemovedIn20Warnings ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-three-resolve-all-removedin20warnings))
   - query.get(id) has been deprecated and we use that in our AclBaseQuery. The 2.x way is to use the db session directly `db_session.get(User, 1)`.
   - Cascading backref has been depracated. This means that any assumption that any dependant model object will. be added to the DB session is not true any more. Refactor to make sure that all objects are explicitly added to the session before any db_session.commit(). This also means that our legacy way of setting default values for db models via __init__ constructor don't work anymore and could lead to subtle bugs. Refactor and remove any _init_ from all models and move defaults to the `default=` flag to the attribute instead.
6. Migration to 2.0 Step Four - Use the future flag on Engine ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-engine))
7. Migration to 2.0 Step Five - Use the future flag on Session ([ref](https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#migration-to-2-0-step-four-use-the-future-flag-on-session))

Part of #2979


jkppr (approved):
As discussed offline, lgtm

PR #2993, opened 2023-11-22
Mute noisy info logging in the feature extraction analyzer

jkppr:
This PR changes the logging for the feature extraction plugin manager from info to debug to prevent it from spamming the logs.


berggren (approved):
LGTM

PR #2994, opened 2023-11-22
Graph bug and layout fix

berggren:
This PR fixes some inconsistencies in the layout of the graph view to better align with the search page. It also fixes a bug where the graph settings menu was anchored to the `app` instead of the button.


jkppr (approved):
lgtm

PR #2552, opened 2023-02-10
Add retry logic to fetch_resource_data

tomchop:
Add some retry logic to `fetch_resource_data`

PR #3006, opened 2023-12-12
Fix error handling in the API client

jkppr:
This PR is an attempt to fix the error handling in the API client when requesting API resources.
* If the API returns any other status than `20X` (variations of OK), raise an error. 
* If the API client cannot json decode the API response, raise an error.
* The function `get_response_json()` now does not return empty dicts on error anymore, which was unexpected by users.
* Fixed the API client side re-try.
* The user now can handle the raised errors e.g. by sleep & retry.

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -368,23 +368,22 @@ class TimesketchApi:
             Dictionary with the response data.
 
         Raises:
-            RuntimeError: If response could not be JSON-decoded after
-                DEFAULT_RETRY_COUNT attempts.
+            RuntimeError: If response cannot be interpreted or is empty.
         """
         resource_url = "{0:s}/{1:s}".format(self.api_root, resource_uri)
         response = self.session.get(resource_url, params=params)
 
-        retry_count = 0
-        while True:
-            result = error.get_response_json(response, logger)
-            # Any dict with content is good enough for us to return.
-            if result:
-                return result
-            retry_count += 1
-            if retry_count >= self.DEFAULT_RETRY_COUNT:
-                raise RuntimeError(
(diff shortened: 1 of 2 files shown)


api_client/python/timesketch_api_client/client.py
```diff
@@ -368,23 +368,22 @@ def fetch_resource_data(self, resource_uri, params=None):
...
-            result = error.get_response_json(response, logger)
-            # Any dict with content is good enough for us to return.
-            if result:
-                return result
-            retry_count += 1
-            if retry_count >= self.DEFAULT_RETRY_COUNT:
-                raise RuntimeError(
-                    f"Unable to fetch JSON resource data. Response: {str(result)}"
-                )
+        result = error.get_response_json(response, logger, url=resource_url)
```
berggren, line 376: Why are you removing the retry here? It was added because sometimes the resource is not ready and a retry is needed. This will reintroduce issues that was fixed.

berggren (changes requested):

api_client/python/timesketch_api_client/client.py
berggren: I guess you want to delegate this to the client instead? I would argue that having a retry server side makes sense as well. With this we will need to change client code i several places in order to avoid subtle issues.

jkppr: > Why are you removing the retry here? It was added because sometimes the resource is not ready and a retry is needed. This will reintroduce issues that was fixed.

Yes, the idea was to delegate this to the client. However, I can see why we might want to have this also in the API client side covered.
Side note: The retry logic added here in #2552 does only re-try to parse the same response and does not do a new request to the server to my understanding.

I'll add a commit to fix both.

berggren (approved):

PR #3008, opened 2023-12-15
Collapsable left panel

berggren:
This PR adds a "mini" variant of the left panel when collapsed. Instead of hiding it completely it now shows the icons, and expands on click.

Other smaller changes in this PR:
* Small fixes for spacing consistency
* DFIQ card to follow UX in collapsing - i.e. move back to v-card where we have more control of the layout and easier to stay consitent. We will re-think the color etc when we have gone over a larger re-design of the DFIQ navigation and context card.

Screenshot 2023-12-15 at 17 46 33

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
@@ -14,7 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <template>
-  <div>
+  <div v-if="iconOnly" class="pa-4" style="cursor: pointer" @click="$emit('toggleDrawer')">
+    <v-icon left>mdi-auto-fix</v-icon>
+    <div style="height: 1px"></div>
+  </div>
+  <div v-else>
     <div
       class="pa-4"
       :style="!sortedAnalyzerResults.length ? '' : 'cursor: pointer'"
@@ -121,7 +125,9 @@ import ApiClient from '../../utils/RestApiClient'
 import TsAnalyzerResult from './AnalyzerResult.vue'
 
 export default {
-  props: [],
+  props: {
+    iconOnly: Boolean,
+  },
   components: {
     TsAnalyzerResult,
   },
--- a/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
@@ -14,7 +14,15 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <template>
-  <div v-if="iconOnly" class="pa-4" style="cursor: pointer" @click="$emit('toggleDrawer')">
+  <div
+    v-if="iconOnly"
+    class="pa-4"
+    style="cursor: pointer"
+    @click="
+      $emit('toggleDrawer')
+      expanded = true
+    "
+  >
     <v-icon left>mdi-auto-fix</v-icon>
     <div style="height: 1px"></div>
   </div>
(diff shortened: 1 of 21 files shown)


timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
```diff
@@ -14,7 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <template>
-  <div>
+  <div v-if="iconOnly" class="pa-4" style="cursor: pointer" @click="$emit('toggleDrawer')">
```
jkppr, line 17: (not just here but also for all other icons)
Instead of just toggling the left panel when clicked on the icon, what do you think about also opening the clicked point. Currently this is two clicks, which could be one.

Example: Left panel is collapsed, when I now click the Saved Searches icon, the left panel will open and I have to click the saved searches section again to get the list shown.

timesketch/frontend-ng/src/components/LeftPanel/TimelinesTable.vue
```diff
@@ -15,14 +15,19 @@ limitations under the License.
...
+  <div v-else>
     <div
       class="pa-4"
-      :style="'cursor: pointer'"
+      style="cursor: pointer"
       @click="expanded = !expanded"
       :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'"
     >
-      <span> <v-icon left>mdi-timeline-clock-outline</v-icon> Timelines </span>
+      <span> <v-icon left>mdi-clock-outline</v-icon> Timelines </span>
```
jkppr, line 30: Personally I prefer the `mdi-timeline-clock-outline` since it does not just represent the "time" factor but also the "line" or "table" factor. It also reflects the list view for timelines that we now have in the left panel.

What is your opinion on going with `mdi-timeline-clock-outline`?

jkppr (commented):
Looking great!
Just two small topics below.

timesketch/frontend-ng/src/components/LeftPanel/TimelinesTable.vue
berggren: I don't like because it doesn't align with the rest of the icons very well and I think it clutters the UI. It feels out of place. I prefer something simpler.

timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
berggren: I thought about this as well but it didn't behave like I wanted it to. I can try again and see if I can get it to work nicely 👍🏼

berggren: I was thinking of experimenting with it in a follow up PR, but I can give it a go in this one.

berggren: Done!

timesketch/frontend-ng/src/components/LeftPanel/Search.vue
```diff
@@ -14,7 +14,12 @@ See the License for the specific language governing permissions and
 limitations under the License.
 -->
 <template>
-  <div>
+  <div v-if="iconOnly" class="pa-4" style="cursor: pointer" @click="$emit('toggleDrawer')">
```
jkppr, line 17: When on another canvas (e.g. Analyzer or Graph) and clicking the search icon in the minimized version it just toggles the left panel but does not trigger the canvas to change.

jkppr (approved):
lgtm

One last optional nit below on triggering the Explore canvas.

timesketch/frontend-ng/src/components/LeftPanel/Search.vue
berggren: Good catch. Fixed now.

PR #3010, opened 2024-01-01
Fix: get and use access token for Yeti

tomchop:
Authentication via Yeti now needs fetching an access token. This change fixes the yeti analyzer to account for this.


berggren (approved):
LGTM

PR #3012, opened 2024-01-04
Improve analyzer result querying performance

jkppr:
The analyzer results are currently requested per timeline when a sketch is loaded. This results in some scaling issues when working with a sketch that has many timelines.

This PR improves the loading performance by adding a new API endpoint that returns only the latest analyzer session for each timeline and analyzer on a given sketch.

* Adding a `LatestAnalyzerSketchResource` API endpoint.
* Load analyzer results via the new API endpoint
* Fixing a small front-end bug


timesketch/api/v1/resources/analysis.py
```diff
@@ -167,6 +167,50 @@ def get(self, sketch_id, session_id):
...
+        sketch_analyzer_names = set()
+        for analyzer_session in sketch_analyzers:
+            sketch_analyzer_names.add(analyzer_session.name)
+
+        # Get the latest session for each analyzer for each sketch timeline
+        analyzer_sessions = []
+        for timeline in sketch.timelines:
+            for analyzer_name in sketch_analyzer_names:
+                analysis_history = (
+                    Analysis.query.filter_by(timeline=timeline, name=analyzer_name)
```
berggren, line 204: Did you find a way to make this into a more efficient query instead of loop in code?

berggren (changes requested):

timesketch/api/v1/resources/analysis.py
jkppr: Sadly not. Still looking into it.

PR #3022, opened 2024-01-18
Mark events with indicator's relevant_tags (Yeti)

tomchop:


timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -137,6 +137,10 @@ def mark_event(
             slug = re.sub(r"[^a-z0-9]", "-", n["name"].lower())
             slug = re.sub(r"-+", "-", slug)
             tags.append(slug)
+        for tag in indicator["relevant_tags"]:
```
berggren, line 140: I'm curious: Is it yeti that maintain this list of relevant tags? And what does it mean to be relevant? :)

tomchop: > what does it mean to be relevant?

Hitting heavy with deep questions... :D

Relevant tags in Yeti are tags that are "suggested" to be applied to observables that match said indicator. Otherwise yes, it's Yeti that maintains these tags for each indicator.

PR #3026, opened 2024-01-25
Adding form validation to prevent names > 255 char.

### LINKED ISSUE #3025: Timesketch form validation bug
Timesketch does not validate input fields for Sketch-, Timeline- or View-names. But the database only allows for max. 255 characters for a name. Currently a name > 255 characters will be accepted by front- & backend and result in a crash on the webserver as soon as it attempts to insert the name into the database.

Solution: Introduce form validation in the front- and backend.

jkppr:
This PR introduces some simple form validation for Sketch, Timeline and View names to be compliant with the database schema.

* Adding validation to the API endpoints with an actionable error message.
* Adding validation to the frontend with an actionable error message.

**Closing issues**

closes #3025


berggren (approved):
LGTM

I still think it is kind of silly we have this limitation :) I'll probably change it next time we update the schema.

PR #3028, opened 2024-01-28
Fix double escaping in sigma_util causing yaml.parser.ParserError

### LINKED ISSUE #3027: Error when importing sigma rule with single quote
**Describe the bug**
In certain cases, the import of Sigma rules fails, especially when they contain characters like `'`.

**To Reproduce**
1. Download triggering Sigma rule like  [proc_creation_lnx_gtfobin_vim.yml](https://github.com/SigmaHQ/sigma/blob/master/rules/linux/process_creation/proc_creation_lnx_gtfobin_vim.yml)
2. Run  `tsctl import-sigma-rules [location_of_rule]/proc_creation_lnx_gtfobin_vim.yml`
4. See error:

> ```
> yaml.parser.ParserError: while parsing a flow mapping
>   in "<unicode string>", line 1, column 1:
>     {'title': 'Vim GTFOBin Abuse - L ...
>     ^
> expected ',' or '}', but got '<scalar>'
>   in "<unicode string>", line 1, column 151:
>      ... 'Detects usage of "vim" and it\'s siblings as a GTFOBin to execu ...
**Expected behavior**
Successful import of rule

**Additional context**
I think the problem here lies in line 250 to 255 of `sigma_util.py`: https://github.com/google/timesketch/blob/57c585c0fae11accb1881c3a820a845b3af4c03b/timesketch/lib/sigma_util.py#L250-L255

in connection with [line 32 to 36 of `tools/sigma/parser/collection.py` in the `sigmatools` package](https://github.com/SigmaHQ/legacy-sigmatools/blob/1a0f514f19f5b9228e7b0a7f05acdf28c7436194/tools/sigma/parser/collection.py#L32-L36):

The function is called twice on the same input (once in `sigma_util.py`, once in `collection.py`). The "double-escaping" seems to lead to a ParserError.

lo-chr:
This PR fixes the issue https://github.com/google/timesketch/issues/3027, by simply passing the text of a sigma rule (`rule_text` instead of it's already escaped version `doc`) to `sigmatools`. `sigmatools` does the parsing and escaping of yaml by itself.

`yaml.safe_load_all` is still used for yaml-parsing (making sure, that the format is right) within this function, before calling the `sigmatools` library.

Like mentioned in the issue, the previous version led to ParserErrors, due to "double escaping" via `yaml.safe_load_all` in specific cases. 

The PR also fixes one `sigma_util.py` test, which was failing mistakenly.

**Checks**
- [x] All tests succeed.
- [x] Unit tests added.

**Closing issues**
closes #3027

### DIFF
--- a/timesketch/lib/sigma_util.py
+++ b/timesketch/lib/sigma_util.py
@@ -251,7 +251,7 @@ def parse_sigma_rule_by_text(rule_text, sigma_config=None):
 
         for doc in rule_yaml_data:
             parser = sigma_collection.SigmaCollectionParser(
-                str(doc), sigma_conf_obj, None
+                str(rule_text), sigma_conf_obj, None
             )
             parsed_sigma_rules = parser.generate(sigma_backend)
             rule_return.update(doc)
--- a/timesketch/lib/sigma_util_test.py
+++ b/timesketch/lib/sigma_util_test.py
@@ -383,7 +383,11 @@ level: medium
         self.assertIsNotNone(rule)
         self.assertEqual("5d2c62fe-3cbb-47c3-88e1-88ef73503a9f", rule.get("id"))
         self.assertIn(
-            'event_identifier:"10" AND (xml_string:"\\\\foobar.exe" AND xml_string:"10"',  # pylint: disable=line-too-long
+            'event_identifier:"10"',  # pylint: disable=line-too-long
+            rule.get("search_query"),
+        )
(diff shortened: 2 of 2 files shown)


jkppr (approved):
Thanks @lo-chr for raising an issue and also providing a PR to fix it. Much appreciated.

The PR looks good. We will merge it before the next release.

PR #3040, opened 2024-02-18
Add ApexChart based visualizations

sydp:
Adds ApexChart based visualizations to frontend-ng

**Checks**

- [x] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes
(if such).

[cast.webm](https://github.com/google/timesketch/assets/8305521/2923335d-b24f-4364-840b-bd7fd57f1e75)


berggren:
Hey @sydp This is a big PR :) It will take some time to review. I'll catch up with you offline to see if we can break it up a bit.

sydp:
Hi @berggren / @jkppr - as discussed, I simplified the layout and removed some of the components.  PTAL when you get a chance.

rocketeeer:
bump

timesketch/lib/aggregators/apex.py
```diff
@@ -0,0 +1,748 @@
...
+        if operator not in ("gte", "lte"):
+            raise ValueError(f"Unknown operator clause - {operator}.")
+
+        if clause not in self._VALID_QUERY_CLAUSES:
+            raise ValueError(f"Unknown boolean clause - {clause}.")
+
+        # try:
+        #     _ = datetime.fromisoformat(datetime_value)
+        # except ValueError as error:
+        # raise ValueError(f"{datetime_value} is not ISO formatted.") from error
```
jkppr, line 98: If we don't need it, let's remove it.

timesketch/frontend-ng/src/views/Visualization.vue
```diff
@@ -0,0 +1,42 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<template>
+  <v-container fluid>
+    <v-card flat class="mx-3">
```
jkppr, line 18:
```suggestion
    <v-card flat>
```

For the canvas container to align with the other elements like the DFIQ context box above.

jkppr (commented):
Alright, given the size of the PR I have to split the review work into multiple sessions. So this is part 1/X. I'll cover the files in the `Visualization` folder in part 2.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+        Visualizations 
+      </span>
+  
+      <v-btn
+        icon
+        text
+        class="float-right mt-n1 mr-n1"
+        :to="{ name: 'VisualizationNew' }"
+        @click.stop=""
+      >
```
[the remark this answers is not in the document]
sydp: makes sense :)

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+        >
+          <strong>
+            {{ visualizationCount }}
+          </strong>
+        </small>
+      </span>
+    </div>
+    <v-expand-transition>
+      <div v-show="expanded && savedVisualizations.length">
+        <v-subheader>Saved Visualizations</v-subheader>
```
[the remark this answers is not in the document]
sydp: Ah not just yet, I was exploring some other visualization under a different grouping (header) and this was left.  I have removed

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+          
+        >
+          <v-row 
+            no-gutters 
+            class="pa-2 pl-5" 
+            :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'"
+          >
+            <v-col
+              :class="$vuetify.theme.dark ? 'dark-font' : 'light-font'"
+              @click="navigateToSavedVisualization(savedVisualization.id)"
```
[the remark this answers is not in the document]
sydp: Previous aggregations will not be compatible with these changes so to resolve this, I propose to filter them out from the list and also update the aggregation on the analyzer side as a separate PR.  WDYT?

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+    <v-expand-transition>
+      <div v-show="expanded && savedVisualizations.length">
+        <v-subheader>Saved Visualizations</v-subheader>
+        <div
+          v-for="(savedVisualization, key) in savedVisualizations"
+          :key="key"
+          
+          style="cursor: pointer; font-size: 0.9em; text-decoration: none"
+          
+        >
```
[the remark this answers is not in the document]
sydp: Given the PR size, I propose to add pagination in a following PR.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+            }
+            let currentPath  = this.$route.fullPath
+            let deletedPath = this.$router.resolve(params).route.fullPath
+
+            if (currentPath === deletedPath) {
+              this.$router.push({ name: 'VisualizationNew', })
+            }
+          })
+          .catch((e) => {
+            console.error(e)
```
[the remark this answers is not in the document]
sydp: Agree, done.

timesketch/lib/aggregators/apex.py
sydp: Done.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
```diff
@@ -0,0 +1,248 @@
...
+            class="pa-2 pl-5" 
+            :class="$vuetify.theme.dark ? 'dark-hover' : 'light-hover'"
+          >
+            <v-col
+              :class="$vuetify.theme.dark ? 'dark-font' : 'light-font'"
+              @click="navigateToSavedVisualization(savedVisualization.id)"
+            >
+              <v-icon left small>
+                {{ getIcon(savedVisualization.chart_type) }}
+              </v-icon> {{ savedVisualization.name }}
```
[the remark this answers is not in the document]
sydp: I added this styling as suggested but not sure if it is working as intended.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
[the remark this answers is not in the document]
jkppr: Sounds good. Please filter them in the new UI and we will update the analyzers in a separate PR. 👍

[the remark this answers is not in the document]
sydp: I added a filter so legacy charts will not be displayed.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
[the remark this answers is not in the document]
sydp: Done.  lmk if this works for you.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
```diff
@@ -0,0 +1,186 @@
...
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+ -->
+ <template>
+  <v-card
+    flat
```
jkppr, line 18: I would recommend to go with `outlined` instead of flat. This would add a border around the visualization that helps understand what is part of the viz and what not. (e.g. tables look a bit lost without)

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
```diff
@@ -0,0 +1,186 @@
...
+        :height="height"
+        :width="width"
+        :xTitle="xTitle"
+        :showXLabels="showXLabels"
+        :yTitle="yTitle"
+        :showYLabels="showYLabels"
+        :showDataLabels="showDataLabels"
+      >
+      </TsHeatmapChart>
+      <TsLineChart
```
jkppr, line 91: The LineChart but also the other charts except the table have some overlay or cutout with their surrounding box. See screenshot below.
image

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
```diff
@@ -0,0 +1,186 @@
...
+        :height="height"
+        :width="width"
+        :xTitle="xTitle"
+        :showXLabels="showXLabels"
+        :yTitle="yTitle"
+        :showYLabels="showYLabels"
+        :showDataLabels="showDataLabels"
+      >
+      </TsLineChart>
+      <TsTableChart
```
jkppr, line 108: This might be something we do not need to "fix" or there might be no fix for it. But I noticed that when asking for the 10 most rare domains, I get 29 results (17 with count 1).

image

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
jkppr, line 1: The minimum settings we need to render an aggregation we need field, aggregation type and then some information depending on the type (e.g. max values and chart type). 
Would it be a lot of effort to trigger & auto-update the preview on the right side when the minimum information are provided or changed? (Instead of the user clicking the reload button)

timesketch/frontend-ng/src/components/Visualization/ChartConfig.vue
```diff
@@ -0,0 +1,236 @@
...
+          @selectedChartType="selectedChartType = $event"
+          required
+          :rules="[rules.required]"
+        >
+        </TsChartTypeSelect>
+      </v-col>
+    </v-row>
+    <v-row>
+      <v-col>
+        <v-text-field
```
jkppr, line 32: What do you think about removing the name field by default and only ask for a name when the user hits the Save button in a small dialog (similar to how we do saved searches)?

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
```diff
@@ -0,0 +1,418 @@
...
+                :yTitle="selectedYTitle"
+                :showYLabels="selectedShowYLabels"
+                :showDataLabels="selectedShowDataLabels"
+              ></TsChartCard>
+            </v-card-text>
+          </v-card>
+        </v-col>
+      </v-row>
+    </v-card-text>
+    <v-card-actions>
```
jkppr, line 91: What do you think about moving the actions into the config card? In the end the user wants to save this specific viz config they just created.

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
```diff
@@ -0,0 +1,418 @@
...
+      >
+        Clear
+      </v-btn>
+      <v-btn 
+        color="primary" 
+        @click="clear" 
+        :to="{ name: 'Explore' }"
+      >
+        Cancel
+      </v-btn>
```
jkppr, line 121: What is your use-case for the cancel button? Default behaviour with all the other views switching happens via the left-panel. The state of the canvas is stored on purpose (but can be deleted by clicking clear).

jkppr (commented):
(2/X) I continued with some of the chart files. Please take a look at the comments below.

timesketch/frontend-ng/src/components/LeftPanel/Visualizations.vue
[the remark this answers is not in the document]
sydp: ah my bad, i think accepting other commit suggestions then pulling them locally might have accidentally overwritten the local change.  thanks good catch.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: yeah this is WAI.  max document count for rare terms works differently to top terms in that it returns results for terms that have document counts up to the specified max value (10 in your example)

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
```diff
@@ -0,0 +1,418 @@
...
+                :maxItems="selectedMaxItems"
+                @updateMaxItems="selectedMaxItems = $event"
+                :interval="selectedInterval"
+                @updateInterval="selectedInterval = $event"
+                :intervalQuantity="selectedIntervalQuantity"
+                @updateIntervalQuantity="selectedIntervalQuantity = $event"
+                :splitByTimeline="selectedSplitByTimeline"
+                @updateSplitByTimeline="selectedSplitByTimeline = $event"
+              ></TsAggregationConfig>
+              <TsChartConfig
```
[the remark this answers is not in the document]
sydp: Done

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: Done.  Although this results in a "double" border when creating a new visualization.

timesketch/frontend-ng/src/views/Visualization.vue
```diff
@@ -0,0 +1,42 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<template>
+  <v-container fluid>
+    <v-card flat>
```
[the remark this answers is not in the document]
sydp: Done

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: Seems this will occur if the configured chart size is wider than the page dimensions.  

What is the desired behaviour you are expecting when this happens?

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
sydp: I am somewhat reluctant to auto update since this will trigger an aggregation on the server end which could be expensive/slow, depending on the selected field/aggregation type (and users may unintentionally select wrong values)  

Not sure what the effort, I don't imagine it would be "too much" (famous last words) but I'll hold off for now until you confirm this is what you would like to happen.

timesketch/frontend-ng/src/components/Visualization/ChartConfig.vue
sydp: At the create visualization stage it isn't a mandatory field to be able load data to preview and you cannot save without specifying a title so I'm not sure if breaking it out into a separate dialog makes a big difference.  Having said that, I have no strong opinions and can see if I can create a dialog if that's the direction you would like to take.

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
sydp: I'll message you offline on this.

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
sydp: The cancel follows the same behaviour as the Sigma editor where cancel switches the view back to the explore view.

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
```diff
@@ -0,0 +1,364 @@
...
+      </v-col>
+    </v-row>
+    <v-row v-if="!disableMaxItems">
+      <v-col>
+        <v-select
+          outlined
+          v-model="selectedMaxItems"
+          :items="[...Array(50).keys()].map((_, i) => i + 1)"
+          :label="labelMaxItems"
+        ></v-select>
```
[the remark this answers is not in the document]
sydp: Done

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
jkppr: Interesting. My page dimension has plenty of space left, but this still happens. Can this be some default chat size that does not match? 

I would go with fitting the chat into the borders. Clipping seems to be more of an unexpected behaviour. What do you think?

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
jkppr: Oh, okay. I need to check if the field explanation makes this clear for the user.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: Ok, you are probably right, I'm only working off a laptop screen so don't have much horizontal real estate. 

I think with the user-configurable dimension it might seem counter-intuitive to then shrink it since they would expect the dimension setting to match with what is displayed?

Also, I'm not 100% sure I can scale/shrink the chart once its dimensions are set (as its a 3P library) but I'll see what I can come up with.

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
sydp: > I guess this is due to changing the chart just being in the frontend and not re-running the aggregation on the backend?

Yes, that's correct - the Load button is only for retrieving the aggregated results so the user can preview them in the selected chart (and update in the front-end) before deciding to save/clear/cancel.

I'm fine with leaving as keeping the behaviour as is, but of course equally happy to tweak if desired.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
jkppr: This happens even with custom width settings for me:
image

sydp: Ah thanks for the screenshot, I think I know what's causing that.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
jkppr: I see. 
I think I have an idea on how to fix this and also fix the [overlapping border issue](https://github.com/google/timesketch/pull/3040#discussion_r1608588834) in one go. What do you think about the following:
1. Change the ChartCard.vue v-card back to flat.
2. Change the SavedVisualization.vue `<div>` back to a `<v-card class="mx-3" outlined>` with outlined border.
3. Change the VisualizationEditor.vue right side (TsChartCard col section) to `<v-card outlined v-show="chartSeries && selectedChartType">` so we don't have the empty v-card visible if no aggregation is created yet.

This should result in the border only showing when needed with sufficient spacing and no overlapping elements form the aggregation. 

Let me know what you think.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
jkppr: See also https://github.com/google/timesketch/pull/3040#discussion_r1609849771 before digging deeper into this.

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
```diff
@@ -0,0 +1,366 @@
...
+    <v-row>
+      <v-col>
+        <v-autocomplete
+          outlined
+          v-model="selectedAggregator"
+          :disabled="disableAggregator"
+          :items="aggregators"
+          label="Aggregate events by"
+          :rules="[rules.required]"
+        ></v-autocomplete>
```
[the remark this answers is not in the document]
sydp: by aggregation options, do you mean the aggregators themselves or the options for each aggregator?

sydp:
https://github.com/google/timesketch/pull/3040#discussion_r1609933790

I'm not able to comment to this in thread but this seems to be working as intended on my machine.  perhaps it could be stale data after changing that.  does clicking the "load button" produce the expected data?

timesketch/frontend-ng/src/components/Visualization/ChartConfig.vue
jkppr: Solved by the UI mock discussed offline.

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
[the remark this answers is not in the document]
jkppr: Sorry, I'm talking about the meaning of "Auto Time Interval", "Rare terms", "Time interval" and "Top K terms".

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
```diff
@@ -0,0 +1,366 @@
...
+    <v-row v-if="!disableMaxItems">
+      <v-col>
+        <v-text-field
+          :label="labelMaxItems"
+          outlined
+          v-model="selectedMaxItems"
+          single-line
+          type="number"
+          min=1
+        />
```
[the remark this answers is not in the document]
sydp: I think adding the `.number` helped resolve the console error.  As an added side-effect it may have fixed the issue here https://github.com/google/timesketch/pull/3040#discussion_r1609933790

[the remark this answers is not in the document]
sydp: Marking this as done.  Please re-open if its still causing issues

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: You are right it should show only up to 3 documents per domain.  Marking this is fixed as it appears it was related to the `maxItems` prop issue.

timesketch/frontend-ng/src/components/Visualization/ChartCard.vue
sydp: Done, thanks, works for me.

timesketch/frontend-ng/src/components/Visualization/ChartColumn.vue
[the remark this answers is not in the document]
sydp: It might be difficult to control when labels can be unexpectedly large.  Not an ideal "fix" but the user can simply turn off x-axis labels and hover over bar/data point to see the value.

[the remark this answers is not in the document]
sydp: For info, details on the behaviour of labels from ApexCharts:

https://apexcharts.com/docs/multiline-text-and-line-breaks-in-axes-labels/

timesketch/frontend-ng/src/components/Visualization/AggregationConfig.vue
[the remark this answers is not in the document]
sydp: I changed the names to closer reflect the actual opensearch aggregation being used and added a tooltip with a brief description of what is being aggregated.

jkppr (approved):
LGTM

Thanks for all the work on this feature!

PR #3041, opened 2024-02-21
DFIQ new UI and navigation

berggren:
This is the new DFIQ UI. 

## Selected question
Screenshot 2024-03-12 at 15 05 12

## Select or create new questions
Screenshot 2024-03-12 at 15 05 25


jkppr (approved):
Some minor nits, nothing blocking that can't be fixed in a follow-up PR.

PR #3049, opened 2024-03-15
Use subqueryload to make loading events with comments faster

tomchop:
When a sketch has a lot of commented events, performance starts to seriously degrade. In my dev environment, a sketch (450k total events) with 1 comment on 3k events would take 14 seconds to load.

The bottleneck was in the way comments are queried from the database via the `CommentMixin` class, which AFAIU dynamically creates a SQLAlchemy table definition. The Explore API goes through all events in a sketch to populate their comment dictionary. Using `subqueryload` to eagerly load comments in one go turned out to speed the queries up *a lot* (went from 14 seconds to 0.5 seconds for the same sketch).

### DIFF
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -321,7 +321,7 @@ class ExploreResource(resources.ResourceMixin, Resource):
 
         comments = {}
         if "comment" in return_fields:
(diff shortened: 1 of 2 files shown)


berggren (approved):
LGTM

Awesome stuff!

PR #3061, opened 2024-04-05
Support for obseravbles in Yeti analyzers

tomchop:
This PR introduces two changes in the yetiinidicators analyzer series:

* Support for observables (they are parsed into a wildcard query)
* Support for unrolling Entities of type Incident (i.e. register new Analyzer subclass)

### DIFF
--- a/timesketch/lib/analyzers/yetiindicators.py
+++ b/timesketch/lib/analyzers/yetiindicators.py
@@ -111,7 +111,9 @@ class YetiBaseAnalyzer(interface.BaseAnalyzer):
             )
 
         access_token = response.json()["access_token"]
-        self._yeti_session.headers.update({"authorization": f"Bearer {access_token}"})
+        self._yeti_session.headers.update(
+            {"authorization": f"Bearer {access_token}"}
+        )
 
     def _get_neighbors_request(self, params):
         """Simple wrapper around requests call to make testing easier."""
@@ -214,7 +216,9 @@ class YetiBaseAnalyzer(interface.BaseAnalyzer):
         event.add_tags(list(tags))
         event.commit()
 
-        msg = f'Indicator match: "{indicator["name"]}" (ID: {indicator["id"]})\n'
+        msg = (
+            f'Indicator match: "{indicator["name"]}" (ID: {indicator["id"]})\n'
+        )
         if neighbors:
             msg += f'Related entities: {[neighbor["name"] for neighbor in neighbors]}'
 
@@ -258,7 +262,9 @@ class YetiBaseAnalyzer(interface.BaseAnalyzer):
                 for ioc in self._intelligence_attribute["data"]
             }
         except ValueError:
-            print("Intelligence not set on sketch, will be created from scratch.")
+            print(
+                "Intelligence not set on sketch, will be created from scratch."
+            )
 
     def save_intelligence(self):
         self.sketch.add_sketch_attribute(
@@ -330,7 +336,9 @@ class YetiBaseAnalyzer(interface.BaseAnalyzer):
                 str(exception),
             )
             return None
-        return {"query": {"query_string": {"query": parsed_sigma["search_query"]}}}
+        return {
+            "query": {"query_string": {"query": parsed_sigma["search_query"]}}
+        }
 
     def run(self):
         """Entry point for the analyzer.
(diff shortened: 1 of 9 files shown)


timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -23,6 +24,21 @@
     "attack-pattern": "HIGH_VOLTAGE",
 }
 
+# Maps Yeti indicator locations to a Timesketch intelligence type
+INDICATOR_LOCATION_MAPPING = {"filesystem": "fs_path"}
+
+
+# Maps Yeti observable types to Timesketch observable types
+OBSERVABLE_INTEL_MAPPING = {
```
jkppr, line 32: This mapping looks like something that might change in the future. Would it make sense to have this mapping as part of the config instead of hardcoding it here?

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -241,33 +294,71 @@ def add_intelligence_entry(self, indicator, event, entity):
...
+            return intelligence_attribute, refs
+
         except ValueError:
-            print("Intelligence not set on sketch, will be created from scratch.")
+            logging.info(
+                "Intelligence not set on sketch, will be created from scratch."
+            )
+        return {"data": []}, set()
+
+    def save_intelligence(self) -> None:
```
jkppr, line 320: Should we add this to the general analyzer interface functions (like adding views or tags)? This would allow also other analyzers to make use of this function.

jkppr, line 320: Would it be beneficial for the analyzer to have an overview of added intelligence in the analyzer results? (similar to tags?)

timesketch/lib/analyzers/yetiindicators.py
jkppr, line 1: You have multiple info logging events in this analyzer. For whom are those log events? The analyzer log is not monitored/visible outside the TS admin role. Therefore I would recommend to log only actionable warnings and errors where the admin needs to fix something.

jkppr (commented):
Looks good. Some minor nits/questions, please take a look.

timesketch/lib/analyzers/yetiindicators.py
tomchop: It's more to be able to debug what is being searched for when expected tagging isn't happening. In that case, an admin could go dig in the logs, see what searches (if any) were made, and maybe better understand the root cause of the problem without having to debug the code.

timesketch/lib/analyzers/yetiindicators.py
tomchop: Yes, that would be a good idea! Maybe the number of new elements added would be enough.

tomchop: Yeah it could be a good idea, especially since the intelligence attribute is quite brittle (it's an array as a value of a dict...). I had quite a hard time making sure it was getting modified as I expected. Maybe revisiting the whole attribute would also be a good idea.

timesketch/lib/analyzers/yetiindicators.py
tomchop: I don't think so, at least not for now. These values are set in both codebases, and won't change from deployment to deployment, unless they fork Yeti and add new observable types that aren't in the main repo.

timesketch/lib/analyzers/yetiindicators.py
jkppr: Alright, let's add this as a separate PR and then refactor this analyzer again. Tracking in #3082

jkppr: This should be handled by the function requested in #3082

timesketch/lib/analyzers/yetiindicators.py
jkppr: So if it is for debug purposes I recommend to set the logging to the debug level. A default release deployment of Timesketch has their log level set to info. So these events would "spam" the log even if there is no issue that needs debugging. If you log them as debugging, the admin can just update the loglevel and then have a look in the log.

tomchop: Sounds good. Tweaked the info logging, and also switched the default from outbound to inbound (given that that's how most links are created in Yeti)

jkppr (approved):
lgtm

PR #3064, opened 2024-04-05
Add authorization token to use GCP IAP

Zawadidone:
Setting the header `Proxy-Authorization` does not work because it is
deleted: https://github.com/psf/requests/blob/2a438c27b5a5828c8ea0dc958112eecffca70b12/src/requests/sessions.py#L318-L319.


Zawadidone:
wrong repository

PR #3093, opened 2024-05-14
Fix analyzers fetching active sessions

### LINKED ISSUE #3092: API returns 404
**Describe the bug**
With release 20240508 Timesketch started to return a 404 error when calling the active analyzers API endpoint resulting in an error every few seconds.

The problem seems to be with the client requesting `GET /api/v1/sketches/[object%20Object]/analyzer/sessions/active/` which does not exist. Before PR #3058 this returned an empty 200, but with the new release a wrong path will return 404 as expected.

**To Reproduce**
1. Start an analyzer on your timeline
2. Open the dev console
3. Wait a moment and see the error pop up

**Expected behavior**
The API call should not fail and should actually call the right sketch ID and not `[object%20Object]`.

jkppr:
This PR fixes an issue that was discovered after releasing version 20240805 where an invalid API path that existed before is now returning a 404 error.

The problem was a unused parameter `this.store` in the `fetchActiveAnalyses` function. Removing the parameter from the function call fixes the URL path and even speeds up the analyzer UI 👍  

**Closing issues**

closes #3092

### DIFF
--- a/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
+++ b/timesketch/frontend-ng/src/components/LeftPanel/AnalyzerResults.vue
@@ -377,7 +377,7 @@ export default {
               return
             }
             const lastActiveCount = this.activeAnalyses.length
-            const activeAnalyses = await this.fetchActiveAnalyses(this.$store, this.sketch.id)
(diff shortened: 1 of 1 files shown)


Annoraaq (approved):
LGTM

PR #3106, opened 2024-06-06
MISP analyzer update

### LINKED ISSUE #3105: MISP analyzer update
# Analyzer update
Minor changes on MISP analyzer that increase significantly performances.

DavidCruciani:
This PR is for update MISP analyzer
fix #3105 .

This will increase significantly performances of the analyzer.


timesketch/lib/analyzers/contrib/misp_analyzer.py
jkppr, line 1: Here are some general design questions and possible optimization ideas to consider:

1. **Multi-Analyzer Parallelization**
* You're using the  `def get_kwargs():`  function, but you're not taking advantage of the parallelization that a [Multi-Analyzer](https://timesketch.org/developers/analyzer-development/#multi-analyzer) would allow.
* Currently, all queries are processed linearly by one celery task.
* Option A:
  * If you expect to have many more different queries in the future (e.g., for other attribute types), consider changing the analyzer into a real multi-analyzer.
  * Each query would then run in a separate Celery task.
  * Note that this means the MISP API would be hit by all tasks in parallel, so consider if this might be a potential issue.
  * The UI result would look similar to the feature_extraction or tagger analyzer.
  * Refer to the [tagger analyzer code](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/tagger.py) code to see how it uses the `def get_kwargs():` function.
  * Example UI output for option A:
image

* Option B:
  * If linear processing of queries is sufficient, remove the `def get_kwargs():`  function and move the setting of the `misp_url` and `misp_api_key`  variables directly into the init function.
  * As a non-multi-analyzer, it's also recommended to use the [newer output format](https://timesketch.org/developers/analyzer-development/#return-message), which has the UX benefit of directly linking created views and tags in the analyzer results.
  * Refer to the [domain analyzer](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/domain.py#L125) code for an example.
  * Example UI output for option B: 
  image

2. **Make attributes configurable**
To enhance the flexibility of the analyzer, consider enabling configuration of the `queries`, `attr`, and `timesketch_attr` variables used in the  function. This can be achieved by loading a YAML file located in the `timesketch/data/` directory. This approach would allow users to extend the analyzer based on their specific requirements without modifying the code. Refer to the [tagger analyzer code](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/tagger.py) for implementation details. 

Let me know if you have questions on those recommendations and if you want to tackle this as part of this PR or maybe in a separate one (if at all).

jkppr (commented):
Hi @DavidCruciani, thanks for those improvements!

timesketch/lib/analyzers/contrib/misp_analyzer.py
DavidCruciani: Hi, @jkppr, thanks a lot for your response and recommendations.
I think that could be great to use parallelization as a lot of data can be treated. Also, I'd like to do it in this PR as it's for optimization.

jkppr: Great! I'm happy to review the PR when you have added the parallelization. Let me know if you have any questions regarding the implementation. Feel free to hit me up on the OSDFIR slack channel as well.

timesketch/lib/analyzers/contrib/misp_analyzer.py
```diff
@@ -28,29 +28,29 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
...
-        self.misp_url = kwargs.get("misp_url")
-        self.misp_api_key = kwargs.get("misp_api_key")
+        self.misp_url = current_app.config.get("MISP_URL")
+        self.misp_api_key = current_app.config.get("MISP_API_KEY")
         self.total_event_counter = 0
-        self.request_set = set()
         self.result_dict = dict()
+        self._query_string = kwargs.get("query")[0]
+        self._attr = kwargs.get("query")[1]
+        self._timesketch_attr = kwargs.get("query")[2]
```
jkppr, line 37: Optional suggestion to make it easier to read the code: 

```suggestion
        self._query_string = kwargs.get("query_string")
        self._attr = kwargs.get("attr")
        self._timesketch_attr = kwargs.get("timesketch_attr")
```

timesketch/lib/analyzers/contrib/misp_analyzer.py
```diff
@@ -28,29 +28,29 @@ def __init__(self, index_name, sketch_id, timeline_id=None, **kwargs):
...
-            return []
-
-        matcher_kwargs = [{"misp_url": misp_url, "misp_api_key": misp_api_key}]
-        return matcher_kwargs
+        to_query = [
+            {"query": ["md5_hash:*", "md5", "md5_hash"]},
+            {"query": ["sha1_hash:*", "sha1", "sha1_hash"]},
+            {"query": ["sha256_hash:*", "sha256", "sha256_hash"]},
+            {"query": ["filename:*", "filename", "filename"]},
+        ]
```
jkppr, line 52: Optional suggestion to make it easier to read the code:

```suggestion
        to_query = [
            {"query_string": "md5_hash:*",  "attr": "md5", "timesketch_attr": "md5_hash"},
            {"query_string": "sha1_hash:*",  "attr": "sha1", "timesketch_attr": "sha1_hash"},
            {"query_string": "sha256_hash:*", "attr": "sha256", "timesketch_attr": "sha256_hash"},
            {"query_string": "filename:*", "attr": "filename", "timesketch_attr": "filename"},
        ]
```

jkppr (commented):
Two optional recommendations for a better code readability.

timesketch/lib/analyzers/contrib/misp_analyzer_test.py
```diff
@@ -43,7 +49,7 @@ def setUp(self):
     @mock.patch("requests.post")
     def test_attr_match(self, mock_requests_post):
         """Test match"""
-        analyzer = misp_analyzer.MispAnalyzer("test_index", 1)
+        analyzer = misp_analyzer.MispAnalyzer("test_index", 1, None, **QUERY_MISP)
         analyzer.misp_url = "https://test.com/"
         analyzer.misp_api_key = "test"
```
jkppr, line 54: You have defined the data already in the set-up function above. No need for those two lines.

timesketch/lib/analyzers/contrib/misp_analyzer_test.py
```diff
@@ -70,7 +76,7 @@ def test_attr_match(self, mock_requests_post):
     @mock.patch("requests.post")
     def test_attr_nomatch(self, mock_requests_post):
         """Test no match"""
-        analyzer = misp_analyzer.MispAnalyzer("test_index", 1)
+        analyzer = misp_analyzer.MispAnalyzer("test_index", 1, None, **QUERY_MISP)
         analyzer.misp_url = "https://test.com/"
         analyzer.misp_api_key = "test"
```
jkppr, line 81: You have defined the data already in the set-up function above. No need for those two lines.

timesketch/lib/analyzers/contrib/misp_analyzer_test.py
jkppr, line 1: The hash in line 12 is not used, is it? If not, do we need it?

timesketch/lib/analyzers/contrib/misp_analyzer_test.py
```diff
@@ -86,11 +92,18 @@ def test_attr_nomatch(self, mock_requests_post):
...
             headers={"Authorization": "test"},
             verify=False,
         )
+
+    @mock.patch("timesketch.lib.analyzers.interface.OpenSearchDataStore", MockDataStore)
+    def test_get_kwargs(self):
+        analyzer_init = misp_analyzer.MispAnalyzer("test_index", 1)
+        queries = analyzer_init.get_kwargs()
+        self.assertIsNotNone(queries)
+        self.assertGreaterEqual(len(queries), 4)
```
jkppr, line 109: Some minor changes/additions I would recommend:
* Checking for len >= 1 should be sufficient. Then you don't need to adjust the test in case you make the queries configurable or add hardcoded queries in the future.
* Also adding a check if all loaded kwargs have the correct keys would be a good idea.

jkppr (commented):
Thanks for updating the tests. I have some minor nits below. Afterwards, we should be good to merge the changes :)

timesketch/lib/analyzers/contrib/misp_analyzer_test.py
jkppr: Maybe something like this:
```
required_keys = ["query_string", "attr", "timesketch_attr"]
for query in misp_queries:
    self.assertTrue(all(key in query for key in required_keys))
```

jkppr (approved):
lgtm

PR #3108, opened 2024-06-06
fix #3107 - Add only tags created by an analyzer to the output

### LINKED ISSUE #3107: Analyzer output lists always all tags
**Describe the bug**
With the new verbose analyzer output, the analyzer results show and link created tags for an analyzer session. However, there is a bug that adds all tags to the list, including the ones manually added or added by different analyzers.

**To Reproduce**
1. Upload a browser history plaso file
2. Add some manual tags to different events
3. Run the "Domain" analyzer
4. See the Tags list in the Analyzer results containing also a list of manually added tags.

**Expected behavior**
The analyzer results should only list tags that have been added by this specific analyzer.

**Screenshots**
image

jkppr:
This PR fixes #3107 by adding only tags created by the analyzer itself to the analyzer output metadata field.

Before:
image

After:
image

**Closing issues**

closes #3107

### DIFF
--- a/timesketch/lib/analyzers/interface.py
+++ b/timesketch/lib/analyzers/interface.py
@@ -257,7 +257,7 @@ class Event(object):
 
         # Add new tags to the analyzer output object
         if self._analyzer:
-            self._analyzer.output.add_created_tags(new_tags)
+            self._analyzer.output.add_created_tags(tags)
 
     def add_emojis(self, emojis):
         """Add emojis to the Event.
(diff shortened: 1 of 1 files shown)


timesketch/lib/analyzers/interface.py
```diff
@@ -1432,7 +1432,12 @@ def add_created_tags(self, tags):
         Args:
             tags: The tags to add to the list of created_tags.
         """
-        self.add_meta_item("created_tags", tags)
+        existing_tags = self.platform_meta_data["created_tags"]
+        if existing_tags:
+            analyzer_tags = list(set().union(existing_tags, tags))
```
tomchop, line 1437: `Set(existing) | set(tags)` would probably be a bit nicer

tomchop (approved):
One small syntax comment

PR #3113, opened 2024-06-20
Merge multiple intelligence attributes if present

tomchop:
This is a workaround to a strange race condition (https://github.com/google/timesketch/issues/3114) encountered when an analyzer runs on multiple timelines and decides to overwrite a sketch attribute, a situation which the yetiindicators.py analyzer seems to run into fairly often.


berggren (approved):
LGTM

PR #3139, opened 2024-08-07
Fix table row height in Firefox

Annoraaq:
There is still a difference in the height of the background color, but at least it is readable again.

Chrome: 
image

Firefox:
image


jkppr (approved):
thanks for the fix!

PR #3140, opened 2024-08-15
Add timeline selection to visualization editor

sydp:
This PR adds a timeline selection component to the visualization editor to allow the user to select specific timelines when creating aggregations/charts.

- [x] All tests succeed.
- [x] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

### DIFF
--- a/timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
+++ b/timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
@@ -39,7 +39,7 @@ limitations under the License.
     <v-row class="mt-3">
       <v-col >
         <v-container class="ma-0">
-          <ts-timeline-search @selectedTimelines="selectedTimelineIDs = $event"></ts-timeline-search>
+          <ts-timeline-search componentName="visualization" @selectedTimelines="selectedTimelineIDs = $event"></ts-timeline-search>
         </v-container>
       </v-col>
     </v-row>
--- a/timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
+++ b/timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
@@ -46,6 +46,7 @@ limitations under the License.
     <v-row class="mt-3">
       <v-col>
         <TsAggregationConfig
+          v-if="selectedTimelineIDs.length > 0"
           :field="selectedField"
           @updateField="selectedField = $event"
           :aggregator="selectedAggregator"
@@ -62,6 +63,7 @@ limitations under the License.
           @updateSplitByTimeline="selectedSplitByTimeline = $event"
         ></TsAggregationConfig>
         <TsChartConfig
+          v-if="selectedTimelineIDs.length > 0"
           :aggregatorType="selectedAggregator"
           :chartType="selectedChartType"
           @updateChartType="selectedChartType = $event"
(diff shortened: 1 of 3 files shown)


timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
```diff
@@ -36,9 +36,20 @@ limitations under the License.
       </v-btn>
     </v-toolbar>
     <v-divider class="mx-3"></v-divider>
+    <v-row class="mt-3">
+      <v-col >
+        <v-container class="ma-0">
+          <ts-timeline-search 
+            componentName="visualization" 
+            :analyzerTimelineId="selectedTImelineIDs"
```
jkppr, line 44: Loading a sketch with this PR throws the following error on the console:

```
[Vue warn]: Property or method "selectedTImelineIDs" is not defined on the instance but referenced during render. Make sure that this property is reactive, either in the data option, or for class-based components, by initializing the property. See: https://v2.vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties.

found in

---> <TsVisualizationEditor> at src/components/Visualization/VisualizationEditor.vue
       <Visualization> at src/views/Visualization.vue
         <Canvas> at src/views/Canvas.vue
           <VMain>
             <Sketch> at src/views/Sketch.vue
               <VApp>
                 <App> at src/App.vue
                   <Root>
```
Deleting the line should fix it, it was used to pre-select timelines based from the Timeline submenu. We can add that later for visualizations.

```suggestion
```

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
jkppr, line 1: Small UX nit: Can you add a filter to the field selection so that it only shows fields available in at least one of the selected Timelines?
Otherwise something like this can happen:
image

jkppr (changes requested):
Overall looking good, thanks for adding those filters.

Please check the comments below.

Additionally, can you add an indicator on what timelines a saved visualization was performed? E.g. looking at this screenshot, it is not clear that the data only comes from the BitsClient timeline if I would not have written it into the title.

image

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
sydp: The field names are currently pulled from the vue state where the field names are cached.  Maybe I am thinking about this wrong but to be able to do this I think there might be two options:
* update the state to store field mappings per index
* have a REST endpoint that can be queried to retrieve field name mappings for given indexes/timelines.

sydp:
> Overall looking good, thanks for adding those filters.
> 
> Please check the comments below.
> 
> Additionally, can you add an indicator on what timelines a saved visualization was performed? E.g. looking at this screenshot, it is not clear that the data only comes from the BitsClient timeline if I would not have written it into the title.
> 
> image


If i understand correctly, we want a separate visual element that displays the underlying timeline being used?   Does this have to be in the chart itself or somewhere else in the UI?  Perhaps we need to design/mock this further before implementing?

In saying that, imho I think this is best left to the analyst (or the analyzer) creating the chart to set the title value to appropriately reflect what is being displayed.   Perhaps an auto-generated (template or GenAI suggested) title based on the field, aggregation and timelines selected could achieve this?

timesketch/frontend-ng/src/components/Visualization/VisualizationEditor.vue
jkppr: Good point.
I think we should explore an option to have this information in the store, since it might be use-full for even more use-cases.
But lets do this in a separate PR, since this might be more involved.

jkppr (approved):
As discussed, we will tackle the open UX topics in a follow-up PR. For now it is more important to get the Timeline filter in for the next release.

Issue #3096, opened 2024-05-19
Run selected/analyzers from "timesketch importer"


hasamba:
im using a script that creates a plaso file from kape output and upload the output file to timesketch,
it would be very helpful if there will be an argument for timesketch_importer that will auto run selected or all analyzers/tagger after uploads and indexing finishes.

thanks

jkppr:
Hi @hasamba
We are using dftimewolf for this use-case and its `TimesketchExporter` module supports triggering Analyzers on the uploaded timelines. For example with the [upload_ts](https://github.com/log2timeline/dftimewolf/blob/main/data/recipes/upload_ts.json) recipe.

Adding this feature to the timesketch_importer is not on the road map for now, but something that sounds like a great opportunity for a community contribution. I'm happy to review the PR if anyone wants to take a stab.

wiredinhp:
@hasamba @jkppr Hi there ! I am a new contributor to this repository and would love to contribute by solving this issue. Could you please assign this issue to me ?

jkppr:
Hi @wiredinhp thanks for offering to implement this feature request. I have assigned you the issue.
* Please find the getting started guide here: https://timesketch.org/developers/getting-started/
* Feel free to submit a draft PR early to gather some review feedback.

jkppr:
@wiredinhp How is it going with this issue? Are you still working on it?

wiredinhp:
@jkppr Hi ! Actually I am working on this issue and I would try to finish it in 3-4 days. My apologies I got paused in between due to some reasons and not informed. Am still working on it and if have some issue I would contact you here 👍 . Thanks for your support 👍

abroglesc:
@wiredinhp Do you happen to have an update on this? We are facing this same issue and deciding whether we should invoke the analyzers via API call or wait for this fix now.

YiChiCanCode:
@jkppr Could you assign me to resolve this issue please? Thanks!

jkppr:
Reassigning this issue due to inactivity.

PR #3143, opened 2024-08-19
Enable passing on auto-run analyzers parameter when using importer library

YiChiCanCode:
**Problem:**
According to [issue 3096](https://github.com/google/timesketch/issues/3096), analyzers cannot be selectively automatically run right after timelines are being uploaded.

**New Feature:**
Allow users to pass on analyzers and timeline ids to automatically run analyzers on the specified timelines after uploading the timelines with the importer library.

**Changes Overview:**

- Add a new function run_analyzers(self, analyzer_names=None), which allows users to run the required analyzers on all the uploaded timelines.
- add the corresponding flags for timesketch importer tools

**For user purpose**
To use the updated importer library, one could use the following code
```
analyzer_names=['sigma', 'feature_extraction'] # or specify the analyzer names you want to use
with importer.ImportStreamer() as stream:
    stream.set_sketch(<sketch_object>)
    stream.set_timeline_name(<timeline_name>)
    stream.add_file(<input_file_object>)
    stream._trigger_analyzers(analyzer_names=analyzer_names) # run analyzers on uploaded timelines
```

**Checks:**

- [x] All tests succeed.
- [x] Tested code on local development, worked as intended
- [x] Tested the CLI script


importer_client/python/timesketch_import_client/importer.py
```diff
@@ -708,21 +708,24 @@ def celery_task_id(self):
...
-        """Close the streamer."""
+    def close(self, analyzer_names=None):
+        """Close the streamer, and run the analyzers based on the analyzer_names"""
         try:
             self._ready()
         except ValueError:
             return
 
+        if analyzer_names is None:
+            analyzer_names = []
```
jkppr, line 719: Instead of this check here you could also set the default for the parameter directly to `[]`.

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -708,21 +708,24 @@ def celery_task_id(self):
...
         if self._data_lines:
             self.flush(end_stream=True)
 
         # Trigger auto analyzer pipeline to kick in.
         pipe_resource = "{0:s}/sketches/{1:d}/analyzer/".format(
             self._sketch.api.api_root, self._sketch.id
         )
-        data = {"index_name": self._index}
+        data = {"index_name": self._index, "analyzer_names": analyzer_names}
         _ = self._sketch.api.session.post(pipe_resource, json=data)
```
jkppr, line 729: You only need to execute this part if the user requested any analyzers to run. I would recommend to do a check for provided analyzer_names.

importer_client/python/timesketch_import_client/importer_test.py
jkppr, line 1: Let's ignore the tests for now and focus on getting the functionality to work.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -143,6 +143,12 @@ def upload_file(
         timeline = streamer.timeline
         task_id = streamer.celery_task_id
 
+        analyzers_name = config_dict.get("analyzers_name")
```
jkppr, line 146: This is always `None`. The config_dict setting below says the variable name is `analyzer_names` not `analyzers_name`.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -462,6 +468,19 @@ def main(args=None):
...
+    config_group.add_argument(
+        "--analyzer_names",
+        "--analyzer-names",
+        type=str,
+        action="store",
+        dest="analyzer_names",
+        default=[],
+        help=(
+            "Set of analyzers that we will automatically run right after the"
+            "timelines are uploaded. The input needs to be the analyzers names."
```
jkppr, line 480: Please explain what is expected as input. A List, a Set, a comma separated string?

jkppr (changes requested):
Please take a look at the comments below.

importer_client/python/timesketch_import_client/importer.py
chuangsc-kid: Thanks! Originally that was this implementation (set the default for the parameter directly to `[]`), but there is a pylint warning saying that defaulting parameter to `[]` is dangerous.

YiChiCanCode: Thanks! Originally that was this implementation (set the default for the parameter directly to []), but there is a pylint warning saying that defaulting parameter to [] is dangerous.

importer_client/python/timesketch_import_client/importer_test.py
YiChiCanCode: Sg!

importer_client/python/tools/timesketch_importer.py
```diff
@@ -462,6 +482,34 @@ def main(args=None):
...
+        type=comma_separated_int_list,
+        action="store",
+        dest="timeline_ids",
+        default=[],
+        help=(
+            "Set of timelines that we will automatically analyze right after the "
+            "timelines are uploaded. The input needs to be the analyzers IDs."
+            "Provided as a comma-separated integer list."
+        ),
+    )
```
jkppr, line 511: Do we need the user to provide this information? I don't think the user knows the timeline ID prior to uploading in a typical workflow. 
The upload function returns a timeline object that contains the ID of the just uploaded timeline, so this could be used instead?
Let me know if I'm missing the use-case here.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -462,6 +482,34 @@ def main(args=None):
...
+        type=comma_separated_list,
+        action="store",
+        dest="analyzer_names",
+        default=[],
+        help=(
+            "Set of analyzers that we will automatically run right after the "
+            "timelines are uploaded. The input needs to be the analyzers names."
+            "Provided as a comma-separated string."
+        ),
+    )
```
jkppr, line 497: Instead of using a comma separated string (potentially error prone and requires extra parsing) I would recommend going with the [argparse nargs](https://docs.python.org/3/library/argparse.html#nargs) functionality to handle multiple analyzers.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -143,10 +143,30 @@ def upload_file(
         timeline = streamer.timeline
```
jkppr, line 143: To make it easier for yourself, you could use this `timeline` object and use its `run_analyzers()` function instead of directly speaking to the API.
See the function of the timeline object in the API client here: https://github.com/google/timesketch/blob/master/api_client/python/timesketch_api_client/timeline.py#L225C9-L225C22

 What do you think?

importer_client/python/tools/timesketch_importer.py
```diff
@@ -143,10 +143,30 @@ def upload_file(
...
         task_id = streamer.celery_task_id
 
+        analyzer_names = config_dict.get("analyzer_names")
+        timeline_ids = config_dict.get("timeline_ids")
```
jkppr, line 147: `analyzer_names` & `timeline_ids` are None here, since they are not set in the `config_dict` that is passed to this function.

importer_client/python/tools/timesketch_importer.py
jkppr: This `timeline` object also returns the ID of the timeline that was just uploaded! (`timeline.id`)

importer_client/python/tools/timesketch_importer.py
```diff
@@ -143,10 +143,30 @@ def upload_file(
...
+        if analyzer_names and timeline_ids:
+            streamer.close(analyzer_names=analyzer_names, timeline_ids=timeline_ids)
+        else:
+            streamer.close()
```
jkppr, line 151: Instead of using the streamer.close function I would recommend to break the Analyzer execution into a separate function. This allows for a workflow that separates upload and analyzer execution.
This extra function can then be called at the end of the main function when the timeline is confirmed uploaded and if analyzers are provided as arguments.

jkppr (changes requested):
Please have a look at the comment below.

jkppr (approved):

PR #3158, opened 2024-08-22
[cli] timeline delete

jaegeral:
Introduce a new method in timesketch cli to delete timelines.

At the moment they are only marked as deleted. But moving forward that will change.

(It simply mimics what is already possible via API and UI).

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/timelines.py
+++ b/cli_client/python/timesketch_cli_client/commands/timelines.py
@@ -110,3 +110,29 @@ def rename_timeline(ctx, timeline_id, new_name):
         click.echo(f'Unsupported output format: "{output}" - using "text" instead')
     else:
         click.echo(f"New name: {timeline.name}")
+
+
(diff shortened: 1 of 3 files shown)


jkppr (approved):
lgtm

cli_client/python/timesketch_cli_client/commands/timelines.py
```diff
@@ -110,3 +110,29 @@ def rename_timeline(ctx, timeline_id, new_name):
...
+    if not timeline:
+        click.echo("No such timeline")
+        return
+    timeline.lazyload_data()
+    if click.confirm(
+        f"Confirm to mark the timeline deleted:{timeline_id} {timeline.name}?"
+    ):
+        timeline.delete()
+
+    click.echo("Deleted")
```
pyllyukko, line 137: Now it prints "Deleted" even if the user chooses not to, which is confusing.

Issue #2432, opened 2022-11-18
drop-db issue


michael2425:
Unable to drop database in timesketch-web instance using "drop-db" command. Error shown as below:

```
Do you really want to drop all the database tables? [y/N]: y
Are you REALLLY sure you want to DROP ALL the database tables? [y/N]: y
All tables dropped. Database is now empty.
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/sqlalchemy/engine/base.py", line 1245, in _execute_context
    self.dialect.do_execute(
  File "/usr/local/lib/python3.10/dist-packages/sqlalchemy/engine/default.py", line 581, in do_execute
    cursor.execute(statement, parameters)
psycopg2.errors.DependentObjectsStillExist: cannot drop table investigativequestion because other objects depend on it
DETAIL:  constraint investigativequestion_searchtempl_investigativequestion_id_fkey on table investigativequestion_searchtemplate depends on table investigativequestion
constraint investigativequestion_sigmarule_investigativequestion_id_fkey on table investigativequestion_sigmarule depends on table investigativequestion
HINT:  Use DROP ... CASCADE to drop the dependent objects too.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/bin/tsctl", line 8, in <module>
    sys.exit(cli())
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/flask/cli.py", line 601, in main
    return super().main(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/lib/python3.10/dist-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/click/decorators.py", line 26, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/flask/cli.py", line 445, in decorator
[...]

PR #3164, opened 2024-08-22
move the user message behind the actual drop command.

jaegeral:
By working on https://github.com/google/timesketch/issues/2432 I discovered that we give the user a successful message, even if the command was not actually executed. So even if something failed, the user would have gotten the message all fine.

That is why I moved the user output after the actual command, so it can fail.


jkppr (approved):
lgtm

PR #3174, opened 2024-09-03
[CLI] export archive and unarchive a sketch

jaegeral:
Adding CLI capabilities to `archive` `unarchive` and `export` a sketch.

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -13,11 +13,13 @@
 # limitations under the License.
 """Commands for sketches."""
 
+import time
 import json
 import click
 import pandas as pd
 
 from timesketch_cli_client.commands import attribute as attribute_command
+from timesketch_api_client import search
 
 
 @click.group("sketch")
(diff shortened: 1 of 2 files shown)


cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -93,3 +95,67 @@ def create_sketch(ctx, name, description):
...
+    click.echo("Executing export . . . ")
+    click.echo("Depending on the sketch size, this can take a while")
+    # start counting the time the export took
+    start_time = time.time()
+    try:
+        search_obj = search.Search(sketch=sketch)
+
+        click.echo(f"Number of events in that sketch: {search_obj.expected_size}")
+
+        search_obj.to_file(filename)
```
jkppr, line 119: You are only catching `ValueError` but what about e.g. `IOError` if there are issues with writing the file?

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -93,3 +95,67 @@ def create_sketch(ctx, name, description):
...
+        # Using the sketch.export function might also work
+        # TODO: https://github.com/google/timesketch/issues/2344
+        # sketch.export("./test.zip")
```
jkppr, line 122: Does it work or not? Maybe we should be a bit more clear on the message here :)

jkppr (approved):
lgtm - some minor nits below

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: I did some tests and with more than 10 mil events on a local system it timed out, but this is not related to the CLI client

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -93,3 +95,67 @@ def create_sketch(ctx, name, description):
...
+        end_time = time.time()
+        click.echo(f"Export took {end_time - start_time} seconds")
+        click.echo("Finish")
+    except ValueError as e:
+        click.echo(f"Error: {e} {e.message}")
```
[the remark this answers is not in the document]
jaegeral: true

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: but I think I will keep it :-)

PR #3183, opened 2024-09-20
Smaller refactoring, adding readmes to folders

jaegeral:
As agreed in a ts dev meeting, first round of small refactorings and adding README files to folders


jkppr (approved):
lgtm

PR #3186, opened 2024-09-20
[Tech dept] update contrib readme, update utils readme and move tsdev from contri…

jaegeral:
* update the readme file in contrib
* update the readme file in util
* moving the tsdev.sh script to util (because ts-dev team uses and maintains it)


jkppr (approved):
lgtm

PR #3187, opened 2024-09-20
Remove analyzer_run.py

jaegeral:
Originally, I wanted to move the `test_tools` folder to tests folder. But after discussion with Johan and Janosch, we decided that analyzer_run.py is currently not used, therefor we will remove it. If we see it is used anywhere, we could bring it back.

While moving it, codespell corrected a minor thing.

I hope the tests (unit / end2end still work)


jkppr (approved):
lgtm

PR #3188, opened 2024-09-26
Frontend v3 Scaffold

berggren:
This PR adds scaffolded Vue3 frontend.

Screenshot 2024-10-16 at 16 55 24


timesketch/frontend-v3/src/layouts/default.vue
Annoraaq, line 1: I had to rename this file from `default.vue` to `Default.vue` in order to avoid a build error.

package.json
Annoraaq, line 1: Isn't this supposed to go in `timesketch/frontend-v3/package.json`?

I got a build error that axios could not be found until I added it to the frontend-v3 package.json.

Annoraaq (approved):
Very nice!

package.json
berggren: Good catch, fixed.

timesketch/frontend-v3/src/layouts/default.vue
berggren: Good catch, fixed now.

PR #3192, opened 2024-10-01
Prevent opensearch from aggregating across all indices.

### LINKED ISSUE #3191: Sketch label aggregation scaling issue
**Describe the bug**
For a Timesketch deployment with many sketches there seems to be an issue with a `datastore.get_filter_label()` function crashing with an `urllib3.exceptions.ProtocolError: ('Connection aborted.', HTTPException('got more than 100 headers'))` error.

This happens because the Sketch API endpoint tries to get filter labels even for sketches that have no index (e.g. empty sketches). However, opensearchpy handles an empty value for index as "search all indices". 
This is not expected but is also not a big issue. Until the amount of tasks created to run the aggregation across all indices exceeds 100 and therefore running into the max number of headers urllib3 can handle.

Affected code: https://github.com/google/timesketch/blob/master/timesketch/api/v1/resources/sketch.py#L472

To fix this, we should not request the filter labels for sketches without and index.

jkppr:
This PR addresses an issue where Timesketch attempts to execute an OpenSearch query with an empty index list when fetching labels for sketches without active timelines. This defaults to querying `_all` indices, which can lead to performance issues and the "got more than 100 headers" error due to excessive task management headers and/or large responses in the aggregation.

The following changes were made:

- In the `SketchResource.get` function (`timesketch/api/v1/resources/sketch.py`), a conditional check was added to ensure `get_filter_labels` is called only if `sketch_indices` is not empty.  If the list is empty (no indices to query), an empty list is directly assigned to `filter_labels` in the metadata.
- In the `get_filter_labels` function (`timesketch/lib/datastores/opensearch.py`) a check was added at the beginning of the function to immediately return an empty list if the provided `indices` parameter is empty.  This prevents the aggregation query from being constructed and executed unnecessarily.  Explanatory comments were added to clarify the purpose of these changes.

These modifications prevent querying `_all` indices with empty `sketch_indices`, improving performance and preventing header limit errors when fetching sketch details for empty sketches or sketches without active timelines.

**Closing issues**

closes #3191

### DIFF
--- a/timesketch/api/v1/resources/sketch.py
+++ b/timesketch/api/v1/resources/sketch.py
@@ -469,7 +469,11 @@ class SketchResource(resources.ResourceMixin, Resource):
             stats_per_timeline=stats_per_timeline,
             last_activity=utils.get_sketch_last_activity(sketch),
             sketch_labels=[label.label for label in sketch.labels],
-            filter_labels=self.datastore.get_filter_labels(sketch.id, sketch_indices),
+            filter_labels=(
+                self.datastore.get_filter_labels(sketch.id, sketch_indices)
+                if sketch_indices
+                else []
+            ),
         )
         return self.to_json(sketch, meta=meta)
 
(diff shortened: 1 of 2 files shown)


berggren (approved):
LGTM

PR #3214, opened 2024-10-21
Fix: Resolve pytype --strict-none-binding issue in the api client

jkppr:
This PR addresses pytype errors raised by the `--strict-none-binding` flag in the api client `Search` class. The issue stemmed from `self._raw_response` being initialized to `None` and then populated later by `_execute_query`, leading to potential unsoundness when accessing it directly.

The fix modifies `_execute_query` to always return a value: a dictionary with the search results if the results are not saved to a file and `count=False`, the total count if `count=True`, or `None` if the results are saved to a file.  The `to_dict` method now calls `_execute_query` if  `self._raw_response` is `None` and handles the case where `_execute_query` returns `None` (indicating a previous `to_file` call) by raising a ValueError. This approach avoids the less robust solution of using assertions and ensures that `to_pandas` and other methods always have a valid dictionary to work with.

A similar issue was fixed in `story.py` to handle cases where a view block did not have a populated `view` attribute, preventing a potential error.

These changes improve type safety and robustness in the `Search` and `Story` classes.


api_client/python/timesketch_api_client/search.py
```diff
@@ -1074,8 +1082,10 @@ def scrolling_enable(self):
 
     def to_dict(self):
         """Returns a dict with the respone of the query."""
-        if not self._raw_response:
+        if self._raw_response is None:
             self._execute_query()
+            if not self._raw_response:
+                raise ValueError("No results to return.")
```
tomchop, line 1088: If _raw_response contains an empty list (no results) will this still evaluate to True? Why not re-use the `is None` clause instead? Same comment for L 1113

tomchop (changes requested):

PR #3215, opened 2024-10-22
Upgrade unfurl and aiplatform dependencies

jkppr:
Upgrade `dfir-unfurl` to `20240627` and add the `google-cloud-aiplatform` lib in version `1.70.0`.


berggren (approved):

jkppr:
Looks like `dfir-unfurl` is not compatible with python 3.8 and requires at least >=3.9 .

PR #3237, opened 2024-11-29
add context menu and sketch creation to homepage

Annoraaq:
image (4)


timesketch/frontend-v3/src/layouts/AppBar.vue
```diff
@@ -29,17 +29,62 @@ limitations under the License.
...
+              prepend-icon="mdi-brightness-6"
+              v-on:click="toggleTheme"
+            >
+              <v-list-item-title>Toggle theme</v-list-item-title>
+            </v-list-item>
+            <v-list-item
+              href="/legacy/"
+              prepend-icon="mdi-view-dashboard-outline"
+            >
+              <v-list-item-title>Use the old UI</v-list-item-title>
```
berggren, line 53: I wonder if we should skip this for the new UI? @jkppr WDYT?

timesketch/frontend-v3/src/router/index.js
```diff
@@ -12,6 +12,7 @@ import Default from "@/layouts/Default";
 
 // Import App views
 import Home from "@/views/Home.vue";
+import Overview from "@/views/Overview.vue";
```
berggren, line 15: What's the idea with the "Overview" page? :) Is this part of the "Sketch" view?

berggren (changes requested):
Looks good, a couple of comments. Also, take a look at the conflicts and resolve / merge master branch.

timesketch/frontend-v3/src/layouts/AppBar.vue
jkppr: Yes, I think with v3 we can finally deprecate the old legacy front-end.

Do we need a link to the frontend-ng folder though until we have everything ported?

timesketch/frontend-v3/src/router/index.js
Annoraaq: I think I accidentally took that over from frontend-ng version:

```
 {
        path: '',
        name: 'Overview',
        redirect: { name: 'Explore' },
      },
```

I guess we don't need the Overview component and replace it with Canvas. I'll remove it from this CL as it is irrelevant for it anyway.

timesketch/frontend-v3/src/layouts/AppBar.vue
Annoraaq: The route does not point anywhere yet, so we can still decide later what we mean by "old ui". I'd say we definitely don't need v1, but maybe ng is a good idea for the time of migration.

berggren: Sounds good

berggren (approved):

PR #3250, opened 2024-12-20
Resolve unsoundness caught by pytype --strict-none-binding.

hnbdgr379:
Resolve unsoundness caught by pytype --strict-none-binding.

In short, pytype was previously more permissive when variables were initialized to None but may have been re-assigned later. This change improves pytype's ability to catch unsoundness in the affected files.

**Checks**

- [x] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


jkppr (approved):
lgtm, thanks

PR #3254, opened 2024-12-23
Add AIStudio as a supported LLM library

itsmvd:
This will make it easier for users to enable AI enabled features in Timesketch (just pull an API key from https://aistudio.google.com/), instead of having to export a GCP service account key.

This library also supports a `response_schema`, instructing/enforcing a response from a LLM model in a provided structure.

Example:
```
schema = {
    "type": "object",
    "properties": {
        "_id": {"type": "string"},
        "message": {"type": "string"}
    }
}
response = llm.generate("Give me the most suspicious event.", response_schema=schema)
print(response)
```


jkppr (approved):
LGTM

PR #3262, opened 2025-01-15
[CLI] Sketch label management

jaegeral:
This PR adds three new commands to the Timesketch CLI for managing labels on sketches:

- **add_label**: Adds a label to a sketch.
- **remove_label**: Removes a label from a sketch.
- **list_labels**: Lists all labels associated with a sketch.

These commands provide a convenient way to manage sketch labels directly from the command line, improving workflow efficiency for users who prefer CLI interactions.

Documentation for it is also updated

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -158,3 +158,40 @@ def unarchive_sketch(ctx):
     if sketch.is_archived():
         sketch.unarchive()
         click.echo("Sketch unarchived")
+
+
+@sketch_group.command("delete", help="Delete a sketch")
+@click.pass_context
+def delete_sketch(ctx):
+    """Delete a sketch."""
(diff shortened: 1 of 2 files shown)


cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -158,3 +158,40 @@ def unarchive_sketch(ctx):
...
         click.echo("Sketch unarchived")
+
+
+@sketch_group.command("delete", help="Delete a sketch")
+@click.pass_context
+def delete_sketch(ctx):
+    """Delete a sketch."""
+    sketch = ctx.obj.sketch
+    sketch.delete()
+    click.echo("Sketch deleted")
```
jkppr, line 169: I don't think this should be part of this PR?

```suggestion
```

jkppr (approved):
Just one minor comment below. Otherwise LGTM

PR #3264, opened 2025-01-17
Enhance tsctl with User Status and Group Membership Information

jaegeral:
This pull request improves the tsctl command-line tool by adding two new features:

- **User Status in list-users**: The tsctl list-users command now displays the status of each user (e.g., active, inactive, locked). This allows administrators to quickly check the status of users in the system.

- **Group Membership** in list-groups: The tsctl list-groups command now lists the members of each group. This makes it easier to understand the composition of groups and manage user access controls.

These enhancements provide administrators with better visibility into user and group information, making it easier to manage Timesketch users and permissions.

Examples:

```
tsctl list-users --status
dev (active: True)
admin (active: True)
foobar2 (active: True)
foobar (active: False)
```

```
tsctl list-groups --showmembership
foobar:
foobar-group: foobar2, foobar
```

### DIFF
--- a/timesketch/tsctl.py
+++ b/timesketch/tsctl.py
@@ -49,14 +49,18 @@ def cli():
 
 
 @cli.command(name="list-users")
-def list_users():
+@click.option("--status", "-s", is_flag=True, help="Show status of the users.")
+def list_users(status):
     """List all users."""
     for user in User.query.all():
         if user.admin:
             extra = " (admin)"
         else:
             extra = ""
-        print(f"{user.username}{extra}")
+        if status:
+            print(f"{user.username}{extra} (active: {user.active})")
+        else:
+            print(f"{user.username}{extra}")
 
 
 @cli.command(name="create-user")
--- a/timesketch/tsctl.py
+++ b/timesketch/tsctl.py
@@ -196,10 +196,14 @@ def list_sketches():
 
 
 @cli.command(name="list-groups")
-def list_groups():
+@click.option("--showmembership", is_flag=True, help="Show members of that group.")
+def list_groups(showmembership):
     """List all groups."""
     for group in Group.query.all():
         print(group.name)
+        if showmembership:
+            for user in group.users:
+                print(f"\t{user.username}")
 
 
 @cli.command(name="create-group")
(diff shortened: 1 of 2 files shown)


timesketch/tsctl.py
```diff
@@ -192,10 +196,14 @@ def list_sketches():
...
 @cli.command(name="list-groups")
-def list_groups():
+@click.option("--showmembership", is_flag=True, help="Show members of that group.")
+def list_groups(showmembership):
     """List all groups."""
     for group in Group.query.all():
         print(group.name)
+        if showmembership:
+            for user in group.users:
+                print(f"\t{user.username}")
```
jkppr, line 206: I think we can skip this flag since we have a `list-group-members` command already.

timesketch/tsctl.py
```diff
@@ -49,14 +49,18 @@ def cli():
 
 
 @cli.command(name="list-users")
-def list_users():
+@click.option("--status", "-s", is_flag=True, help="Show status of the users.")
```
jkppr, line 52: Can we also get an option to show group memberships of for each user? Some output like this: `user:group1,group2,group3` maybe?

timesketch/tsctl.py
```diff
@@ -49,14 +49,18 @@ def cli():
...
+def list_users(status):
     """List all users."""
     for user in User.query.all():
         if user.admin:
             extra = " (admin)"
         else:
             extra = ""
```
jkppr, line 59: It is not related to the PR, but what do you think about moving the admin marker inside the status option?

jkppr (commented):
Looks good overall, please see comments below.

timesketch/tsctl.py
jkppr: Fair point. But I think not to change something because something external might break would hinder any improvement in general so should not be a factor.
However, for this specific suggestion, there is no technical need to change it, so let's keep it as is.

timesketch/tsctl.py
jaegeral: done

jkppr (approved):

PR #3267, opened 2025-01-23
Dynamically update Star/Comment label counts in the left panel

### LINKED ISSUE #2902: "Star" label is not updated in real-time
When staring events on a timeline it takes a reload to update the "Tags" and "omnibar" to display the number of stared events and if there was no starred event before, even to show the `All starred events ()` entry.

This entry should appear in real time as soon as the first event is marked with a star without the need to reload the component.

jkppr:
This PR addresses the issue where the star and comment label counts in the left panel did not update in the UI after starring or commenting on events, requiring a manual page refresh.

The proposed solution updates the counts dynamically whenever an event's star or comment status changes, providing instant feedback to the user. The update is achieved by recalculating and updating the `meta.filter_labels` array in the Vuex store after each star/comment action, ensuring the UI reflects the accurate label counts without the need for watchers or page reloads.

Changes:

- Added logic to `EventList.vue` to update `meta.filter_labels` in the store after `toggleStar` and `toggleMultipleStars` actions.
- Added a new mutation `SET_EVENT_LABELS` and action `updateEventLabels` to the store to handle label count updates.
- Modified `Comments.vue` to trigger `updateEventLabels` after adding or deleting a comment.
- Updated `TagsList.vue` to use a `v-data-iterator` for pagination and searching of tags and labels. This refactors the existing tag and label lists.

This improvement enhances the user experience by providing real-time feedback on label changes and a more efficient way to manage and filter events based on their annotations.

closes #2902

### DIFF
--- a/timesketch/frontend-ng/src/store.js
+++ b/timesketch/frontend-ng/src/store.js
@@ -242,18 +242,18 @@ export default new Vuex.Store({
         })
         .catch((e) => {})
     },
-    updateEventLabels(context, payload) {
-      if (!payload.labels || !payload.label || !payload.num)  {
+    updateEventLabels(context, {label: inputLabel, num}) {
+      if (!inputLabel || !num)  {
         return
       }
-      let labels = payload.labels
-      let starLabelIndex = labels.findIndex(label => label.label === payload.label);
-      if (starLabelIndex > -1) {
-        labels[starLabelIndex].count = labels[starLabelIndex].count + payload.num
+      let allLabels = context.state.meta.filter_labels
+      let label = allLabels.find(label => label.label === inputLabel);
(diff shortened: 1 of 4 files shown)


timesketch/frontend-ng/src/store.js
```diff
@@ -239,6 +242,19 @@ export default new Vuex.Store({
         })
         .catch((e) => {})
     },
+    updateEventLabels(context, payload) {
```
Annoraaq, line 245: You can also get the labels from the context with `context.state.meta.filter_labels`. Then you don't need to pass them here. I saw that you are always passing the labels from the state so far.

Annoraaq (approved):

PR #3272, opened 2025-01-31
Fix: Handle "query_shard_exception" in OpenSearch error handling

jaegeral:
This pull request enhances the error handling for OpenSearch queries to include specific handling for the "query_shard_exception" error.  Previously, only the "too_many_nested_clauses" error was handled explicitly.  This change ensures that "query_shard_exception" is also caught, displayed to the user and logged appropriately, allowing for more informative error reporting and potential future recovery logic.

Specifically:

*   Modified the error handling logic in `EventList` to include a check for the "query_shard_exception" message.

This change improves the robustness of the OpenSearch integration by providing more granular error handling.  It does not introduce any new functionality or change existing behavior other than improved error reporting.

It does not fix the underlying issue.


jkppr (approved):
lgtm thanks

PR #3275, opened 2025-01-31
V3 sketch page top bar

Annoraaq:
The top bar now allows renaming and opening the menu.

Archiving works, but it throws an error which I still have to look into, which has the effect that you end up with a blank page and have to manually refresh. The navigation item in the side bar is a dummy and does not yet redirect.

Note: this is still marked as draft as it contains the changes of a previous PR.

image (7)


berggren (approved):
LGTM

PR #3278, opened 2025-02-03
Refactor LLM manager so that users can configure an LLM provider per feature

itsmvd:
Refactors the LLM manager to allow users to configure different LLM providers for specific features. This enables more granular control over which LLM is used for each LLM powered feature within Timesketch, defaulting to a global LLM provider.


berggren (approved):
LGTM

berggren (approved):

PR #3290, opened 2025-02-10
Add User Settings Menu to Home Page

jaegeral:
This pull request introduces a user settings menu accessible from the home page, allowing users to manage their Timesketch sketches and preferences.

* Implemented a dropdown/modal menu item that appears when the icon is clicked.

### DIFF
--- a/timesketch/frontend-ng/src/views/Home.vue
+++ b/timesketch/frontend-ng/src/views/Home.vue
@@ -56,6 +56,15 @@ limitations under the License.
                 </v-list-item-content>
               </v-list-item>
 
+              <v-list-item @click="showSettingsDialog = true">
+                  <v-list-item-icon>
+                    <v-icon>mdi-cog-outline</v-icon>
+                  </v-list-item-icon>
+                  <v-list-item-content>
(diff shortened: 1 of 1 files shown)


jkppr:
Blocked by https://github.com/google/timesketch/pull/3292

timesketch/frontend-ng/src/views/Home.vue
```diff
@@ -146,6 +166,12 @@ export default {
...
       return this.$store.state.currentUser
     },
+    userSettings() {
+      return this.$store.state.settings
+    },
+    systemSettings() {
+      return this.$store.state.systemSettings
+    },
   },
   methods: {
```
jkppr, line 176: When the component is mounted it needs to update its settings to have them synced across views. Otherwise the component can't access the system settings and will display user settings only in the default setting.

```suggestion
  },
  mounted() {
    this.$store.dispatch('updateSystemSettings')
    this.$store.dispatch('updateUserSettings')
  },
  methods: {
```

jkppr (approved):
Overall LGTM. Just one important change below.

PR #3291, opened 2025-02-10
TagList bug fix

jkppr:
This PR fixes a small bug introduced in #3267 where in rare situations a tag or label element can be `undefined` leading to the [localCompare in the allTagsAndLabels() function](https://github.com/google/timesketch/blob/master/timesketch/frontend-ng/src/components/LeftPanel/TagsList.vue#L115) to fail, rendering the TagList on the left panel broken.

This is fixed by filtering the data prior to sending it to the compare function.


Annoraaq (approved):

jkppr:
Blocked by https://github.com/google/timesketch/pull/3292

PR #3295, opened 2025-02-12
UI build 2025 01 12

jaegeral:
New UI Build


timesketch/frontend-ng/dist/empty-state.png
jkppr, line 1: I don't know why `yarn build` wants to delete this file, but if we delete it our empty state would be broken. 
So I would recommend to not delete this file and open an issue to fix this for future ui builds.

image

jkppr (approved):
General LGTM, just bring back the image please.

PR #3296, opened 2025-02-12
Bugfix in llm_summarize and introduce  initial tests

itsmvd:


jkppr (approved):
Thanks for fixing the "no-results" bug and adding tests. 

Approving now, but please fix the failing unit-test before merging.

PR #3304, opened 2025-02-24
Filtered back-ticks and other trailing characters from the resulting query

dianakramer:
- Consed an internal bug
- Filtered back-ticks and other trailing characters from the resulting query
- Added testing of the result from AI

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


timesketch/api/v1/resources_test.py
```diff
@@ -1408,6 +1408,50 @@ def test_nl2q_llm_error(self, mock_aggregator, mock_create_provider):
...
+        self.login()
+        data = dict(question="Question for LLM?")
+        mock_AggregationResult = mock.MagicMock()
+        mock_AggregationResult.values = [
+            {"data_type": "test:data_type:1"},
+            {"data_type": "test:data_type:2"},
+        ]
+        mock_aggregator.return_value = (mock_AggregationResult, {})
+        mock_llm = mock.Mock()
+        mock_llm.generate.return_value = " \t`LLM generated query`\n "
```
jkppr, line 1426: Can we test multiple variations the LLM likes to add backticks?
e.g. also check `````" \t```LLM generated query```\n "````` or uneven backticks like ```" \t`LLM generated query``\n "```

timesketch/api/v1/resources/nl2q.py
```diff
@@ -222,5 +222,5 @@ def post(self, sketch_id):
             )
             return jsonify(result_schema)
 
-        result_schema["query_string"] = prediction.strip("```")
+        result_schema["query_string"] = prediction.strip("`\n\r\t ")
```
jkppr, line 225: How does this solution compare to e.g. using ```prediction.replace("`", "")```?

jkppr (commented):
Please have a look at the comments below.

timesketch/api/v1/resources/nl2q.py
dianakramer: It would assure it does not remove the "`" if it's legitimate part of the query. Strip is used to remove characters from the beginning (leading) and end (trailing) of a string.

jkppr (approved):
lgtm

PR #3308, opened 2025-02-26
Add LLM features manager and interface

itsmvd:


timesketch/lib/llms/features/interface.py
```diff
@@ -0,0 +1,53 @@
...
+            sketch_id: The ID of the sketch.
+            kwargs: Feature-specific keyword arguments for prompt generation.
+
+        Returns:
+            The generated prompt string.
+        """
+        raise NotImplementedError()
+
+    @abstractmethod
+    def process_response(self, llm_response: str, **kwargs: Any) -> dict[str, Any]:
```
berggren, line 41: It's not clear to me what processing means in this context. What is expected from the processing? Perhaps update the doc string with some use cases and/or examples.

timesketch/lib/llms/features/interface.py
```diff
@@ -0,0 +1,53 @@
...
+from typing import Any, Optional
+from abc import ABC, abstractmethod
+from timesketch.models.sketch import Sketch
+
+
+class LLMFeatureInterface(ABC):
+    """Interface for LLM features."""
+
+    NAME: str = "llm_feature_interface"  # Must be overridden in subclasses
+    RESPONSE_SCHEMA: Optional[dict[str, Any]] = None
```
berggren, line 25: Perhaps add an explanation in the doc string on what the RESPONSE_SCHEMA can be used for. Audience would be developers that implements the interface.

berggren (changes requested):
Looks good - just a couple of comments around describing a few things for developers implementing the interface.

timesketch/lib/llms/features/__init__.py
jkppr, line 1: I assume you are going to load new feature modules in this init file. 

You could consider adding an automated module loading function like we do it with the dfiq_analyzers here. https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/dfiq_plugins/__init__.py#L5
https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/dfiq_plugins/manager.py#L31

This would allow people to add feature modules without restarting the whole instance every time.

jkppr (commented):
Looks good to me too. I added one suggestion below. Once the topics from jbn are addressed, we can stamp it.

timesketch/lib/llms/features/interface.py
itsmvd: Done, good point

timesketch/lib/llms/features/__init__.py
itsmvd: Implemented + added test for the function. Good suggestion, will add it to the LLM provider manager later as well.

jkppr (approved):
lgtm

PR #3310, opened 2025-02-27
Introduce LLMResource API method, tests, and add it as a method for the frontend

itsmvd:
The LLMResource will be the gateway to LLM features and will track common LLM metrics for all LLM features.

This will enable us to, in a next PR, remove the `nl2q` and `getEventSummary` methods in the frontend, and instead call a single `llmRequest` method on the `RestApiClient.js`.
Which will then enable us to move the LLM features to the `features` folder, and out of the `resources` folder.


timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+        """Handles POST requests to the resource."""
+        start_time = time.time()
+        sketch = self._validate_sketch(sketch_id)
+        form = self._validate_request_data()
+        feature = self._get_feature(form.get("feature"))
+
+        self._increment_request_metric(sketch_id, feature.NAME)
+
+        timeline_ids = self._validate_indices(sketch, form.get("filter", {}))
+        prompt = self._generate_prompt(feature, sketch, form, timeline_ids)
```
jkppr, line 72: If I understand this correct, the llm feature lib is called twice here. Once to generate the prompt and once to process the response.
Is there a particular reason to run this sequentially from the API endpoint?

Would something like the [analyzer run()](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/account_finder.py#L18) function be a solution to consider? This way each feature would have more control about when to create the prompt and what to do with the results and the API endpoint here just gets the stuff it is supposed to return. This would also make it easier later on to move features into celery tasks I think.

The `_execute_llm_call` function could therefore move into the feature interface or util.

What do you think?

timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+        ),
+        "llm_duration_seconds": prometheus_client.Summary(
+            "llm_duration_seconds",
+            "Time taken to process an LLM request (in seconds)",
+            ["sketch_id", "feature"],
+            namespace=METRICS_NAMESPACE,
+        ),
+    }
+
+    _LLM_TIMEOUT_WAIT_SECONDS = 30
```
jkppr, line 59: I think this should be configurable via the `timesketch.conf` to allow for users with slow AI endpoints (e.g. local ollama) to still experiment with the feature by adjusting this setting.

timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+            if "error" in response:
+                self.METRICS["llm_errors_total"].labels(
+                    sketch_id=str(sketch_id),
+                    feature=feature.NAME,
+                    error_type="llm_api_error",
+                ).inc()
+                abort(
+                    definitions.HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR,
+                    "Error during LLM processing.",
+                )
```
jkppr, line 184: Is there any actionable information we could give the user with this error message?

timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+                logger.warning(
+                    "LLM call timed out after %d seconds.",
+                    self._LLM_TIMEOUT_WAIT_SECONDS,
+                )
+                process.terminate()
+                process.join()
+                self.METRICS["llm_errors_total"].labels(
+                    sketch_id=str(sketch_id), feature=feature.NAME, error_type="timeout"
+                ).inc()
+                abort(definitions.HTTP_STATUS_CODE_BAD_REQUEST, "LLM call timed out.")
```
jkppr, line 172: What is the expected action from the user here? Try again? Contact admin if repeatedly happening?

timesketch/api/v1/resources/llm.py
jkppr, line 1: Small nit: Could you add docstrings to the class and methods to explain their parameters and return values?

timesketch/api/v1/resources_test.py
```diff
@@ -1692,3 +1691,133 @@ def test_llm_summarize_with_events(self, mock_create_provider):
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(response.get_data(as_text=True))
         self.assertEqual(response_data.get("summary"), "Mock summary from LLM")
+
+
+@mock.patch("timesketch.api.v1.resources.OpenSearchDataStore", MockDataStore)
+class LLMResourceTest(BaseTest):
```
jkppr, line 1697: You are mainly testing the "happy path" and input errors. Consider also testing error edge cases like errors raised by the feature prompt generation or during the llm execution.

jkppr (changes requested):
Looks good overall. Some comments below to be addressed. Thanks.

timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+
+    _LLM_TIMEOUT_WAIT_SECONDS = 30
+
+    @login_required
+    def post(self, sketch_id: int):
+        """Handles POST requests to the resource."""
+        start_time = time.time()
+        sketch = self._validate_sketch(sketch_id)
+        form = self._validate_request_data()
+        feature = self._get_feature(form.get("feature"))
```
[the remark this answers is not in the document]
itsmvd: `_get_feature` already deals with this, so if `feature` does not exist on the form, None gets passed into `_get_feature` which handles this

timesketch/api/v1/resources/llm.py
itsmvd: Agree this is a good suggestion, but lets address this in a separate PR (I'll make a note), as it requires making changes to timesketch.conf.

timesketch/api/v1/resources/llm.py
itsmvd: Provided details of the error in the error message

timesketch/api/v1/resources/llm.py
itsmvd: Good point, could be more intuitive. I've added this for now:
```
                abort(
                    definitions.HTTP_STATUS_CODE_BAD_REQUEST,
                    "LLM call timed out. Please try again. "
                    "If this issue persists, contact your administrator.",
                )
```

timesketch/api/v1/resources/llm.py
itsmvd: Done

timesketch/api/v1/resources_test.py
itsmvd: Good point. Added two additional tests, `test_post_prompt_generation_error` and `test_post_llm_execution_timeout`.

timesketch/api/v1/resources/llm.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def _validate_request_data(self) -> dict:
+        """Validates the presence of request JSON data."""
+        form = request.json
+        if not form:
+            abort(
+                definitions.HTTP_STATUS_CODE_BAD_REQUEST,
+                "The POST request requires data",
+            )
+        return form
```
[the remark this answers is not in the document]
itsmvd: I wasn't aware of this library, I've made a personal note to look into it to assess what the best way is to deal with form submissions through the API endpoint, but did not find direct added value to add it now (given the thorough backend validation we do and that documentation on what to provide to the API calls is clear)

jkppr (approved):

PR #3316, opened 2025-03-05
adding unit tests for archiving a sketch

jaegeral:
Introduce more unittests to cover archived sketch handling for:

* archive a sketch
* attempt to delete an archived sketch

Attempting to delete an archived sketch currently results in a 200, it has to be decided if that is a behaviour we want to keep or if it needs to be unarchived first (from an OS perspective I think it has to be).


timesketch/api/v1/resources_test.py
```diff
@@ -159,6 +159,49 @@ def test_append_label_to_sketch(self):
...
+
+        # Pull the sketch again to get the status
+        response = self.client.get(f"/api/v1/sketches/{created_id}/")
+        self.assertEqual(
+            response.json["objects"][0]["name"],
+            "test_archive_sketch",
+        )
+        self.assert200(response)
+        self.assertIn("archived", response.json["objects"][0]["status"][0]["status"])
+
```
jkppr, line 204: You could add one more step in here to include testing the `unarchive` function as well. What do you think?

timesketch/api/v1/resources_test.py
```diff
@@ -205,6 +248,53 @@ def test_attempt_to_delete_protected_sketch(self):
...
+        data = dict(
+            name="test_delete_archive_sketch",
+            description="test_delete_archive_sketch",
+        )
+        response = self.client.post(
+            "/api/v1/sketches/",
+            data=json.dumps(data, ensure_ascii=False),
+            content_type="application/json",
+        )
+        created_id = response.json["objects"][0]["id"]
```
jkppr, line 265: We should consider outsourcing the creation of test sketches into a helper function. Just as a general idea for the future, not blocking on your PR.

jkppr (approved):
General looking good. Two ideas below.

timesketch/api/v1/resources_test.py
jaegeral: yeah I was considering this, I want to check further if I want to have several things tested in the same case, usually I try to keep it separated.

PR #3318, opened 2025-03-05
Update nginx.conf

jkppr:
This PR adds a filter to the nginx.conf to filter some potential unauthenticated ReDoS attempts on the `/login` endpoint. This solution is to prevent low hanging attempts and should not be your only option to ensure the availability of your Timesketch service.
More details on the issue: https://blog.yossarian.net/2022/12/28/ReDoS-vulnerabilities-and-misaligned-incentives


itsmvd (approved):
lgtm

PR #3323, opened 2025-03-07
Fix: Removal Logic Bug in Annotation Mixins

jaegeral:
This pull request addresses a bug in the `StatusMixin`, `CommentMixin`, and `LabelMixin` classes that caused incorrect removal of elements (statuses, comments, and labels).

**Problem:**

The original code iterated through lists (`self.status`, `self.comments`, `self.labels`) and removed elements directly within the loop. This is a known cause of errors in Python, as it modifies the list during iteration, potentially leading to skipped elements.

**Solution:**

The fix involves the following changes:

* **Refactored Removal Logic:** The removal logic in `StatusMixin.set_status`, `CommentMixin.remove_comment`, and `LabelMixin.remove_label` has been refactored to avoid modifying the list during iteration.
* **Intermediate List for Removal:** The code now first identifies the elements to be removed and stores them in a separate list. After the iteration is complete, these elements are removed from the original list.

**Benefits:**

* **Corrected Removal Behavior:** This change ensures that elements are removed correctly and consistently, without skipping any elements.
* **Increased Reliability:** The fix prevents potential errors and improves the overall reliability of the annotation mixins.

It also adds a mechanism in tsctl to potentially see multiple status.

(and fixes some typos)


timesketch/models/annotations.py
```diff
@@ -339,12 +360,21 @@ def set_status(self, status):
...
             The status as a string
         """
         if not self.status:
             self.status.append(self.Status(user=None, status="new"))
-        return self.status[0]
+        if len(self.status) > 1:
+            # TODO: at some point replace the waning with an exception:
+            # raise RuntimeError("More than one status available")
+            logging.warning("More than one status available")
+        return self.status[0]  # always return the last element in the array
```
jkppr, line 377: Some minor changes to the wording and adding some data to make the warning and exception more debuggable for admins.

```suggestion
            self_id = self.id if hasattr(self, "id") else None
            # TODO: Change from warning to raising an exception once we ensured
            # it won't affect the deployment.
            # raise RuntimeError(
            # "More than one status available for object [%s] with ID: [%s]",
            #     str(type(self).__name__),
            #     str(self_id)
            # )
            logging.warning(
                "More than one status available for object [%s] with ID: [%s]",
                str(type(self).__name__),
                str(self_id)
            )
        return self.status[0]
```

jkppr (approved):
Thanks, I have added some minor changes.

PR #3329, opened 2025-03-10
Update pylint & astroid

jkppr:
This PR updates pylint to version `2.17.7` and astroid to `2.15.8`. This change should fix the container build problems we see for the `timesketch-dev` container.

It also updates the `.pylintrc` file to reflect the newer version.

I also have disabled 2 pylintrc workflows for `ubutnu-20.04`. In theory there should be no different outcome and we don't use that version in our builds anymore.


jaegeral (approved):
LGTM; great work, love typing

PR #3330, opened 2025-03-11
Improve snackbar.js: add support for custom timeouts & small refactor

itsmvd:
We also now set all objects at creation time, we shouldn't use `let` here because we are not reassigning any variables after the init.

When setting a snackbar from the frontend users can now set a custom timeout on the snackbar like:
`this.warningSnackBar('This feature is currently limited to 1000 starred events. This limit will be increased soon.', 10000);`


timesketch/frontend-ng/src/mixins/snackBar.js
```diff
@@ -1,55 +1,52 @@
...
+                timeout: timeout
+            }
             this.$store.dispatch('setSnackBar', snackbar)
         },
-        warningSnackBar(message) {
-          let snackbar = defaultSnackBar
-          snackbar.message = message
-          snackbar.color = "warning"
-          this.$store.dispatch('setSnackBar', snackbar)
+        infoSnackBar(message, timeout = 2000) {
```
jkppr, line 43: Why should the infoSnackBar use a different default timeout than all others?

itsmvd: Keeping as is, callers can overwrite the default per this PR

jkppr (approved):

PR #3332, opened 2025-03-11
New LLM feature: Starred events to forensic report (stored as a Story)

itsmvd:
This PR adds a new LLM-powered feature to automatically generate forensic reports (stored as a Story) from starred events.

- Adds a "Generate forensic report" button that appears when starred events filter is active
- Processes starred events through an LLM provider to create a forensic analysis report
- Stores the generated report as a Story
- Implements metrics for tracking events processed and stories created
- Limits initial implementation to 1000 starred events per report (will be expanded once we add max_token_count tracking for LLM providers)

**New button in EventList.vue** _(only visible when viewing starred events)_
image

**Example LLM generated report from starred events**
image


itsmvd:
Two notes:
* Will add tests after first review is done
* `_run_timesketch_query` will be abstracted to a helper function in a next PR, as we refactor it to use streaming instead

jaegeral:
comment from the sideline, I believe keeping the limit to 1000 events is plenty, even if we increase the possible number to use for llm in general, having somewhat of a limit here could be just fine.

data/timesketch.conf
```diff
@@ -380,11 +380,17 @@ LLM_PROVIDER_CONFIGS = {
         },
     },
     'llm_summarize': {
-        'aistudio': {
-            'model': 'gemini-2.0-flash-exp',
+        'vertexai': {
+            'model': 'gemini-2.0-flash-001',
             'project_id': '',
```
jkppr, line 385: Is changing this default intended?

docker/dev/build/docker-entrypoint.sh
```diff
@@ -25,6 +25,7 @@ if [ "$1" = 'timesketch' ]; then
   ln -s /usr/local/src/timesketch/data/plaso_formatters.yaml /etc/timesketch/plaso_formatters.yaml
   ln -s /usr/local/src/timesketch/data/nl2q /etc/timesketch/
   ln -s /usr/local/src/timesketch/data/llm_summarize /etc/timesketch/
+  ln -s /usr/local/src/timesketch/data/llm_forensic_report /etc/timesketch/
```
jkppr, line 28: I think we should start looking into a prompt folder...

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -986,6 +1002,35 @@ export default {
           this.isSummaryLoading = false
         })
     },
+    generateForensicReport() {
```
jkppr, line 1005: Do you really want to call it a `Forensic Report`? The term could be misunderstood, so maybe something more specific to what this function is doing: `keyEventReport`?

timesketch/frontend-ng/src/mixins/snackBar.js
jkppr, line 1: Aren't you changing this file in https://github.com/google/timesketch/pull/3330 already?

timesketch/lib/llms/actions.py
jkppr, line 1: In the general structure of Timesketch code this would be something living in the `interface.py`. See the analyzer interface for example: https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/interface.py#L639

jkppr (changes requested):

data/timesketch.conf
itsmvd: Yes, as the flash model has gone to GA since and is no longer experimental

docker/dev/build/docker-entrypoint.sh
itsmvd: Correct, I've filed an internal bug for this a couple of weeks back

timesketch/frontend-ng/src/mixins/snackBar.js
itsmvd: Yep, I created that PR after this one. So I'll merge the changes from that PR into this one, once that one is merged.

timesketch/lib/llms/actions.py
itsmvd: As discussed I don't think a utility/helper function should be inside an interface class of which the purpose is to define an abstract base class. So this is better off in a helper/util file, and the best place for this imo in this case is to create a `utils.py` file under the `lib/stories.py` folder. Changed this in the comment, lmk what you think.

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+            logger.error("Prompt template is missing the <EVENTS_JSON> placeholder")
+            raise ValueError(
+                "LLM forensic report prompt template is missing the "
+                "required <EVENTS_JSON> placeholder."
+            )
+
+        prompt_text = prompt_template.replace("<EVENTS_JSON>", json.dumps(events_dict))
+        return prompt_text
+
+    def _run_timesketch_query(
```
jkppr, line 111: As discussed 1:1 have a look if you can replace this with the [OpenSearchDataStore.search()](https://github.com/google/timesketch/blob/master/timesketch/lib/datastores/opensearch.py#L581) function instead.

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+
+        if id_list:
+            id_query = " OR ".join([f'_id:"{event_id}"' for event_id in id_list])
+            query_string = id_query
+
+        all_indices = list({t.searchindex.index_name for t in sketch.timelines})
+        indices_from_filter = query_filter.get("indices", all_indices)
+
+        if "_all" in indices_from_filter:
+            indices_from_filter = all_indices
```
jkppr, line 147: This is actually a logic error (we just fixed it in PR #3359 ). So this should be:

```suggestion
        all_timeline_ids = [t.id for t in sketch.timelines]
        indices = query_filter.get("indices", all_timeline_ids)

        if "_all" in indices_from_filter:
            indices_from_filter = all_timeline_ids
```

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+
+        return export.query_results_to_dataframe(result, sketch)
+
+    def generate_prompt(self, sketch: Sketch, **kwargs: Any) -> str:
+        """Generates the forensic report prompt based on events from a query.
+        Args:
+            sketch: The Sketch object containing events to analyze.
+            **kwargs: Additional arguments including:
+                - form: Form data containing query and filter information.
+                - datastore: OpenSearch instance for querying.
```
jkppr, line 173: On the backend, you actually don't need the datastore object to query. You only need the sketch_id and can then use the wrapper I mentioned above.

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+            str: Generated prompt text with events to analyze.
+        Raises:
+            ValueError: If required parameters are missing or if no events are found.
+        """
+        form = kwargs.get("form")
+        datastore = kwargs.get("datastore")
+        timeline_ids = kwargs.get("timeline_ids")
+
+        if not form:
+            raise ValueError("Missing 'form' data in kwargs")
```
jkppr, line 185: If form is a required argument, why is it in **kwargs and not in the normal args list?

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+        events_df = self._run_timesketch_query(
+            sketch,
+            query_string,
+            query_filter,
+            datastore=datastore,
+            timeline_ids=timeline_ids,
+        )
+
+        if events_df is None or events_df.empty:
+            return "No events to analyze for forensic report."
```
jkppr, line 199: So this will be the response handled by the llm API endpoint. iirc it will then still call the "process results" function? Something you want to address here?

timesketch/lib/llms/features/llm_forensic_report.py
```diff
@@ -0,0 +1,304 @@
...
+
+        events_df = self._run_timesketch_query(
+            sketch,
+            query_string,
+            query_filter,
+            datastore=datastore,
+            timeline_ids=timeline_ids,
+        )
+
+        total_events_count = len(events_df)
```
jkppr, line 276: Unlikely since the icon only shows when there are stared events, but maybe catch if there are no results.

jkppr (changes requested):
round 2

timesketch/lib/llms/features/llm_forensic_report.py
itsmvd: Which wrapper "mentioned above" are you referring to? I can't seem to find it

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -986,6 +1002,35 @@ export default {
...
+      if (this.totalHits > 1000) {
+        this.warningSnackBar('This feature is currently limited to a 1000 starred events, try setting a timerange filter. ' +
+        'This limit will be increased soon.', 10000);
```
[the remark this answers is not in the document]
itsmvd: Will remove this last part of the message. Making it configurable for an admin I'm not a fan of at the moment, because if they increase the limit to a size that won't fit inside the chosen LLM provider's context window, the feature will break. So better if we can control this setting for now until we can include such granular settings.

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -631,6 +643,11 @@ export default {
     }
   },
   computed: {
+    isStarredEventsFilterActive() {
+      return this.filterChips.some(chip => 
+        chip.type === 'label' && chip.value === '__ts_star'
+    )
+    },
```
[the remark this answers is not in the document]
itsmvd: Improved readability, perhaps this is a personal taste but having a function like `isStarredEventsFilterActive` I would always prefer of a long in-line if-statement.

timesketch/lib/llms/features/llm_forensic_report.py
jkppr: #3379 was merged.

jaegeral:
@itsmvd looks like there are merge conflicts now, any chance you could address those?

itsmvd:
Filed https://github.com/google/timesketch/issues/3705 to track using comments/tags for the report as well as that would be cool to have

timesketch/lib/llms/features/llm_forensic_report.py
itsmvd: WIll introduce a `SkipLLMError` type of custom exception in a next PR.

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -255,6 +255,17 @@ limitations under the License.
                   <v-icon title="Download current view as CSV">mdi-download</v-icon>
                 </v-btn>
 
+                <v-btn 
+                  icon 
+                  @click="generateForensicReport()" 
+                  class="ml-2" 
+                  :loading="isGeneratingReport"
+                  v-if="isStarredEventsFilterActive">
+                    <div class="ts-llm-icon-wrapper" v-if="!isGeneratingReport">
```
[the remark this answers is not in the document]
itsmvd: Let's decide on an icon together

itsmvd:
> @itsmvd looks like there are merge conflicts now, any chance you could address those?

Done

timesketch/lib/llms/features/llm_forensic_report.py
itsmvd: Fair point, now addressed

itsmvd:
Last thing to do (I think): settle upon an icon to use for this feature. Edit: Done

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,289 @@
...
+import logging
+import time
+from typing import Any, Optional
+
+import pandas as pd
+import prometheus_client
+from flask import current_app
+
+from timesketch.lib import utils
+from timesketch.api.v1 import export
```
jkppr: Importing functions from the API on the backend feels a bit brittle to me. What are you trying to do here that can't be done with other means?

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,289 @@
...
+            timeline_ids: List of timeline IDs to query.
+        Returns:
+            pd.DataFrame: DataFrame containing query results.
+        Raises:
+            ValueError: If no valid indices are found.
+        """
+        datastore = OpenSearchDataStore(
+            host=current_app.config.get("OPENSEARCH_HOSTS")[0]["host"],
+            port=current_app.config.get("OPENSEARCH_HOSTS")[0]["port"],
+        )
```
jkppr: You don't need the host & port anymore. This is now done in the OpenSearchDataStore by default. See https://github.com/google/timesketch/blob/master/timesketch/lib/datastores/opensearch.py#L199

So calling `datastore = OpenSearchDataStore()` is fully sufficient for you.

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,289 @@
...
+        all_timeline_ids = [t.id for t in sketch.timelines]
+        indices = query_filter.get("indices", all_timeline_ids)
+        if "_all" in indices:
+            indices = all_timeline_ids
+        indices, timeline_ids = utils.get_validated_indices(indices, sketch)
+        if not indices:
+            raise ValueError(
+                "No valid search indices were found to perform the search on."
+            )
+        result = datastore.search(
```
jkppr, line 157: On the frontend you're implementing an upper limit check for `STARRED_EVENTS_REPORT_LIMIT`.

This size limit is not enforced on the backend though. So when this llm feature is called from anywhere else (e.g. by an analyzer) or via api client, it can attempt to summarize way more events that intended.

timesketch/lib/stories/utils.py
```diff
@@ -0,0 +1,57 @@
+# Copyright 2025 Google Inc. All rights reserved.
```
jkppr:
```suggestion
# Copyright 2026 Google Inc. All rights reserved.
```

timesketch/lib/stories/utils.py
```diff
@@ -0,0 +1,57 @@
...
+            with timestamp will be used.
+    Returns:
+        The ID of the newly created story.
+    Raises:
+        ValueError: If there's an error creating the story.
+    """
+    if title is None:
+        title = f"Investigation Report - {time.strftime('%Y-%m-%d %H:%M')}"
+    try:
+        story = Story(title=title, sketch=sketch, user=sketch.user)
```
jkppr: This will create the story always attributed to the sketch owner. But a story can also be created by other users on the sketch. So I would recommend using `current_user` form `flask_login` instead to get the caller of the method and attribute the story correctly.

timesketch/frontend-ng/src/assets/main.scss
jkppr, line 1: Is this needed? We won't use gradient anyway sorry.

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,289 @@
...
+        if "<EVENTS_JSON>" not in prompt_template:
+            logger.error("Prompt template is missing the <EVENTS_JSON> placeholder")
+            raise ValueError(
+                "LLM starred events report prompt template is missing the "
+                "required <EVENTS_JSON> placeholder."
+            )
+        prompt_text = prompt_template.replace("<EVENTS_JSON>", json.dumps(events_dict))
+        return prompt_text
+
+    def _run_timesketch_query(
```
jkppr, line 119: You can probably make your tests a bit easier by adding `datastore: Optional[OpenSearchDataStore] = None` to the accepted arguments here. And in the test you can just pass `MockDataStore` instead of loading `timesketch.lib.llms.features.llm_starred_events_report.OpenSearchDataStore`.

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -536,6 +544,8 @@ const defaultQueryFilter = () => {
   }
 }
 
+const STARRED_EVENTS_REPORT_LIMIT = 1000
```
jkppr, line 551: Should this be passed to the backend too to ensure the same limit is applied?

timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -241,6 +241,14 @@ limitations under the License.
                   <v-icon title="Download current view as CSV">mdi-download</v-icon>
                 </v-btn>
 
+                <v-btn 
+                  icon 
+                  @click="generateStarredEventsReport()" 
+                  :loading="isGeneratingReport"
+                  v-if="isStarredEventsFilterActive">
+                    <v-icon v-if="!isGeneratingReport" title="Generate report from starred events">mdi-file-star-four-points</v-icon>
+                </v-btn>
```
jkppr, line 250: Please make the button dependent on the settings. I'm fine with not adding a toggle to the user-settings dialog since this is a manual action. But you still need to handle the situation where someone has no LLM provider configured and hence the button won't work. 

TL;DR: If no provider is configured = don't show the button or disable it with a tooltip explaining why.

Otherwise users will get an error when clicking the button:
```
Error generating report: Error initializing LLM provider: Ollama provider requires a 'server_url' in its configuration.
```

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,289 @@
...
+    ),
+}
+
+
+class LLMStarredEventsReportFeature(LLMFeatureInterface):
+    """LLM Starred Events Report feature."""
+
+    NAME = "llm_starred_events_report"
+    PROMPT_CONFIG_KEY = "PROMPT_LLM_STARRED_EVENTS_REPORT"
+    RESPONSE_SCHEMA = {
```
jkppr: At least during my testing I had multiple executions where the returned data was not valid json failing the schema test resulting in an error 500.

Given that our stories do support markdown natively and LLMs are better at generating markdown than json, would it be an option to drop the shema and just work with the markdown output and send this to the story?

itsmvd: Yes this is a very good point. Outdated design pattern based on my testing a long time ago

timesketch/frontend-ng/src/components/Explore/EventList.vue
itsmvd: Yes, something I thought I had done for sure but apparently forgot, thanks for catching!

timesketch/frontend-ng/src/components/Explore/EventList.vue
itsmvd: Yep good catch

timesketch/lib/llms/features/llm_starred_events_report.py
itsmvd: Good idea

timesketch/lib/stories/utils.py
itsmvd: Had not thought about that one, makes sense

timesketch/lib/llms/features/llm_starred_events_report.py
itsmvd: Good to know!

timesketch/lib/llms/features/llm_starred_events_report.py
itsmvd: Agreed, was just the easiest way to quickly poc this and then didn't reconsider after so many months to update. Lesson to myself to not let open PRs open for that long ;-) The llm_summarize feature used the same thing.

In any case, I think it's clear what the purpose is: to normalize results from OpenSearch to a dataframe. I've opted to move the function that does this to `lib/utils.py` and changed the call sites in this LLM feature as well as in export.py and some other places to that function.

timesketch/frontend-ng/src/assets/main.scss
itsmvd: Yes it's still used by the LLM summarization feature, where I still think it looks good and is in line with common LLM design patterns

itsmvd:
> Thanks for updating the PR. Here is the next round of comments.

Thanks!

data/llm_starred_events_report/prompt.txt
jkppr, line 1: Can we add the following things to the prompt please:
1. The title of the report should be "Starred Events Report" (not "Forensics Report" since this can imply stuff we can't guarantee).
2. Add a context info how many events have been used to generate this report. (e.g. ``Add `[*This report was generated based on <LEN_EVENTS_JSON> stared events.*]` in the top of the report.``

timesketch/lib/llms/features/llm_starred_events_report.py
```diff
@@ -0,0 +1,329 @@
...
+        if datastore is None:
+            datastore = OpenSearchDataStore()
+
+        all_timeline_ids = [t.id for t in sketch.timelines]
+        indices = query_filter.get("indices", all_timeline_ids)
+        if "_all" in indices:
+            indices = all_timeline_ids
+        indices, timeline_ids = utils.get_validated_indices(indices, sketch)
+
+        event_count = datastore.search(
```
jkppr, line 206: This call throws an error for me: 
```
  File "/usr/local/src/timesketch/timesketch/lib/datastores/opensearch.py", line 873, in search
    count_result = self.client.count(
                   ^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/client/utils.py", line 176, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/client/__init__.py", line 707, in count
    return self.transport.perform_request(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/transport.py", line 457, in perform_request
    raise e
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/transport.py", line 418, in perform_request
    status, headers_response, data = connection.perform_request(
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/connection/http_urllib3.py", line 308, in perform_request
    self._raise_error(
  File "/opt/venv/lib/python3.12/site-packages/opensearchpy/connection/base.py", line 315, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
opensearchpy.exceptions.RequestError: RequestError(400, 'parsing_exception', 'request does not support [size]')
```

PR #3336, opened 2025-03-13
Update api_client code for new pylint version

jkppr:
Update the API client code to be compatible with the new pylint version.


jaegeral (approved):
LGTM

PR #3343, opened 2025-03-14
Update documentation.yml

jaegeral:
Foobar

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -13,5 +13,19 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: 3.8
-      - run: pip install mkdocs-material mkdocs-redirects
-      - run: mkdocs gh-deploy --force
+      - name: Install dependencies
+        run: pip install mkdocs-material mkdocs-redirects
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./site #the default output directory of mkdocs.
+          publish_branch: gh-pages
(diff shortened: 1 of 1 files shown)


.github/workflows/documentation.yml
```diff
@@ -13,5 +13,19 @@ jobs:
...
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./site #the default output directory of mkdocs.
+          publish_branch: gh-pages
+          deploy_key: ${{ secrets.DEPLOY_KEY }}
+          external_repository: google/timesketch #add this line.
+          user_name: github-actions[bot]
+          user_email: 41898282+github-actions[bot]@users.noreply.github.com
+      - name: Build site
+        run: mkdocs build
```
jkppr, line 29: iirc you need to build the site first before you can deploy them. So this part should go before the "Deploy to GitHub pages" step.

.github/workflows/documentation.yml
```diff
@@ -13,5 +13,19 @@ jobs:
...
         with:
           python-version: 3.8
-      - run: pip install mkdocs-material mkdocs-redirects
-      - run: mkdocs gh-deploy --force
+      - name: Install dependencies
+        run: pip install mkdocs-material mkdocs-redirects
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
jkppr, line 21: Do we still need the `github_token` if we already have a `deploy_key`? 

```suggestion
```

jkppr (approved):
We could also limit the workflow to only trigger if something changed in the `docs` path by adding:
```
on:
  push:
    branches:
      - master
    paths:
      - 'docs/**'
```

PR #3345, opened 2025-03-14
Update Frontend-NG Build and Deployment Workflow

jaegeral:
## Update Frontend-NG Build and Deployment Workflow

This pull request updates the GitHub Actions workflow responsible for building and deploying the `timesketch/frontend-ng` application. The changes ensure that the `dist` folder is automatically updated on the `main` branch whenever changes are pushed to the frontend code (excluding changes directly within the `dist` folder).

**Changes:**

*   **Workflow Trigger:**
    *   The workflow is triggered on pushes to the `main` branch.
    *   The workflow only runs if changes are detected within the `timesketch/frontend-ng/` directory.
    *   Changes within the `timesketch/frontend-ng/dist/` directory are explicitly excluded from triggering the workflow to prevent infinite loops.
*   **Build Process:**
    *   The workflow checks out the code.
    *   Sets up Node.js version 18 (can be adjusted if needed).
    *   Installs frontend dependencies using `yarn install`.
    *   Builds the frontend application using `yarn build`.
*   **Deployment:**
    *   Instead of deploying to a separate `dist` branch, the workflow now directly commits and pushes the contents of the `timesketch/frontend-ng/dist/` folder to the `main` branch.
    *   A commit message `chore: Update frontend-ng dist [skip ci]` is used. The `[skip ci]` tag prevents the workflow from re-triggering itself.
    *   Handles the case where there are no changes in the `dist` folder after the build, preventing the workflow from failing.
    * The correct branch is checked out before the build.

**Reasoning:**

*   This change streamlines the frontend deployment process by directly updating the `main` branch with the built code.
*   The exclusion of the `dist` folder from the trigger prevents infinite workflow loops.
*   The `[skip ci]` tag in the commit message further prevents unnecessary workflow runs.
* The workflow now handles the case where there are no changes to commit.

### DIFF
--- a/.github/workflows/build_frontend-ng.yml
+++ b/.github/workflows/build_frontend-ng.yml
@@ -29,11 +29,11 @@ jobs:
         working-directory: timesketch/frontend-ng
         run: yarn build
 
-      - name: Deploy to dist branch
-        uses: JamesIves/github-pages-deploy-action@v4
-        with:
-          branch: dist # The branch the dist folder will be pushed to
-          folder: timesketch/frontend-ng/dist # The folder to deploy
-          clean: true # Remove previous files in the dist branch
-          token: ${{ secrets.GITHUB_TOKEN }} # Use the default GITHUB_TOKEN
-          #target-folder: frontend-ng # Optional: If you want to deploy to a subfolder in the dist branch
+      - name: Commit and push dist folder
+        working-directory: timesketch/frontend-ng
+        run: |
+          git config --local user.email "github-actions[bot]@users.noreply.github.com"
+          git config --local user.name "github-actions[bot]"
+          git add dist
+          git commit -m "chore: Update frontend-ng dist [skip ci]" || echo "No changes to commit"
+          git push origin HEAD:${{ github.ref_name }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build_frontend-ng.yml
```diff
@@ -0,0 +1,39 @@
+name: Build and Deploy Frontend-NG
+
+on:
+  push:
+    branches:
+      - main # Or your main development branch
```
jkppr, line 6: We have not renamed the default branch yet.

```suggestion
      - master
```

.github/workflows/build_frontend-ng.yml
```diff
@@ -0,0 +1,39 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '18' # Or your preferred Node.js version
```
jkppr, line 22: I think we are running on Node 20 by now.

```suggestion
          node-version: '20'
```

jkppr (approved):
Some small fixes below.

.github/workflows/build_frontend-ng.yml
jaegeral: ack

.github/workflows/build_frontend-ng.yml
jaegeral: ack

PR #3351, opened 2025-03-19
Fix: Handle Search Results for Events in Removed Timelines + E2E Test

jaegeral:
## Description

This pull request addresses an issue where search results could include events from timelines that have been removed from a sketch. This could lead to unexpected behavior and inaccurate results for users. In the UI that never happened because it is filtered in the frontend code (but the events where sent to the user)

Note: The events are still PRESENT in the OS cluster. This will be part of a larger rewrite effort.

**Problem:**

Previously, when a timeline was deleted from a sketch, the underlying search index was not always immediately removed or excluded from future searches. This meant that events from the deleted timeline could still appear in search results, even though the timeline was no longer visible in the sketch.

**Solution:**

This PR implements the following changes to resolve this issue:

1.  **Enhanced Index Validation:**
    *   The `get_validated_indices` function in `timesketch/lib/utils.py` has been updated to more accurately identify and exclude indices associated with deleted timelines.
    *   It now checks the status of timelines within a sketch and only includes indices from timelines that are in a "ready" state.
    *   It also now checks if a timeline is deleted and removes it from the list of valid timelines.
    *   This ensures that only active timelines are considered when building the list of valid indices.

2.  **Search Index Filtering:**
    *   The `ExploreResource` in `timesketch/api/v1/resources/explore.py` now utilizes the updated `get_validated_indices` function to filter out invalid indices before performing a search.
    *   This guarantees that searches are only executed against indices that are associated with active timelines in the sketch.

3.  **End-to-End Test:**
    *   A new end-to-end test, `test_delete_timeline`, has been added to `end_to_end_tests/client_test.py` to specifically verify this fix.
    *   The test performs the following steps:
        *   Creates a new sketch.
        *   Imports two timelines into the sketch.
        *   Verifies the correct number of events are present.
        *   Deletes one of the timelines.
        *   Verifies that the number of events in the sketch is updated correctly after deleting the timeline.
        *   Verifies that the number of timelines in the sketch is updated correctly after deleting the timeline.
        *   Verifies that the deleted timeline is no longer included in search results.

**Benefits:**

*   **Accurate Search Results:** Users will now receive more accurate search results that only include events from active timelines within a sketch.
*   **Consistent Behavior:** The behavior of the search functionality will be more consistent and predictable, as deleted timelines will no longer affect search results.
*   **Improved Data Integrity:** The system will maintain better data integrity by ensuring that searches only operate on valid and relevant data.
* **Test Coverage:** The new end-to-end test provides confidence that the fix works as intended and will help prevent regressions in the future.

**Testing:**

*   The new end-to-end test (`test_delete_timeline`) has been added and passes successfully.
*   Manual testing has been performed to verify that events from deleted timelines are no longer included in search results.
*   All existing tests pass.

**Checklist:**

-   [x] I have performed a self-review of my own code.
-   [x] I have added a test that proves my fix is effective or that my feature works.
-   [x] New and existing unit tests pass locally with my changes.
-   [x] I have added necessary documentation (if appropriate).

### DIFF
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -156,6 +156,8 @@ class ExploreResource(resources.ResourceMixin, Resource):
 
         all_indices = list({t.searchindex.index_name for t in sketch.timelines})
         indices = query_filter.get("indices", all_indices)
+        all_timeline_ids = [t.id for t in sketch.timelines]
+        timeline_ids = query_filter.get("timelines", all_timeline_ids)
 
         # If _all in indices then execute the query on all indices
         if "_all" in indices:
@@ -163,7 +165,12 @@ class ExploreResource(resources.ResourceMixin, Resource):
 
         # Make sure that the indices in the filter are part of the sketch.
         # This will also remove any deleted timeline from the search result.
+        # comment: jaegeral: since we write all things now to only one
+        # searchindex, this will be pointless
         indices, timeline_ids = get_validated_indices(indices, sketch)
+        # the timeline_ids seem to be [] all the time, make a check if it its empty
+        if not timeline_ids:
+            timeline_ids = all_timeline_ids
 
         # Remove indices that don't exist from search.
         indices = utils.validate_indices(indices, self.datastore)
@@ -187,6 +194,8 @@ class ExploreResource(resources.ResourceMixin, Resource):
             )
 
         # Aggregate hit count per index.
+        # TODO: the value for indices might be misleading since we only have one
+        # one index per sketch since a while.
         index_stats_agg = {
             "indices": {
                 "terms": {
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -929,9 +929,7 @@ class Sketch(resource.BaseResource):
         return template_list
 
     def list_timelines(self):
-        """List all timelines for this sketch.
-
-        Retrieves a list of all timelines associated with the current sketch.
+        """Retrieves a list of all timelines associated with the current sketch.
         This method fetches the sketch data and parses the 'timelines'
         information to create a list of Timeline objects.
 
@@ -953,24 +951,13 @@ class Sketch(resource.BaseResource):
             raise RuntimeError("Unable to list timelines on an archived sketch.")
 
         timelines = []
+
         data = self.lazyload_data()
         objects = data.get("objects")
-
-        if not isinstance(objects, list):
-            return timelines
-
         if not objects:
             return timelines
 
-        first_object = objects[0]
-        if not isinstance(first_object, dict):
-            return timelines
-
-        timeline_list = first_object.get("timelines")
-        if not isinstance(timeline_list, list):
-            return timelines
-
-        for timeline_dict in timeline_list:
+        for timeline_dict in objects[0].get("timelines", []):
             timeline_obj = timeline.Timeline(
                 timeline_id=timeline_dict["id"],
                 sketch_id=self.id,
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -931,20 +931,46 @@ class Sketch(resource.BaseResource):
     def list_timelines(self):
         """List all timelines for this sketch.
 
+        Retrieves a list of all timelines associated with the current sketch.
+        This method fetches the sketch data and parses the 'timelines'
+        information to create a list of Timeline objects.
+
         Returns:
-            List of timelines (instances of Timeline objects)
+            list[Timeline]: A list of Timeline objects representing the
+                timelines in the sketch. Returns an empty list if:
+                - The sketch is archived.
+                - The sketch has no timelines.
+                - The sketch data is malformed or missing the expected
+                  'timelines' structure.
+                - The 'objects' key is not a list.
+                - The first element of the 'objects' list is not a dict.
+                - The 'timelines' key is not a list.
+
+        Raises:
+            RuntimeError: If the sketch is archived.
         """
         if self.is_archived():
             raise RuntimeError("Unable to list timelines on an archived sketch.")
 
         timelines = []
-
         data = self.lazyload_data()
         objects = data.get("objects")
+
+        if not isinstance(objects, list):
+            return timelines
+
         if not objects:
             return timelines
 
-        for timeline_dict in objects[0].get("timelines", []):
+        first_object = objects[0]
+        if not isinstance(first_object, dict):
+            return timelines
+
+        timeline_list = first_object.get("timelines")
+        if not isinstance(timeline_list, list):
+            return timelines
+
+        for timeline_dict in timeline_list:
             timeline_obj = timeline.Timeline(
                 timeline_id=timeline_dict["id"],
                 sketch_id=self.id,
--- a/end_to_end_tests/client_test.py
+++ b/end_to_end_tests/client_test.py
@@ -450,9 +450,6 @@ level: high
 
         timeline = sketch.list_timelines()[0]
 
-        # Wait for events to be indexed
-        time.sleep(5)
-
         # Check that there are 6 events in total
         search_client = search.Search(sketch)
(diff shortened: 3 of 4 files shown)


timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,26 @@ def post(self, sketch_id: int):
 
         all_indices = list({t.searchindex.index_name for t in sketch.timelines})
         indices = query_filter.get("indices", all_indices)
+        all_timeline_ids = [t.id for t in sketch.timelines]
```
jkppr, line 159: Would `sketch.timelines` return also timelines that have been marked deleted?

timesketch/api/v1/resources/explore.py
jkppr, line 1: Can you explain how this PR does remove timelines that are marked deleted from the search results?

timesketch/api/v1/resources/explore.py
jaegeral: no, they are not part of the timelines lookup, there is no "marked as deleted" thing, they are either there or not.

timesketch/api/v1/resources/explore.py
jaegeral: it is not, that was a misstake in the PR description. It only "adds current timelines" to the search filter.

timesketch/api/v1/resources/explore.py
```diff
@@ -187,6 +198,8 @@ def post(self, sketch_id: int):
             )
 
         # Aggregate hit count per index.
+        # TODO: the value for indices might be misleading since we only have one
+        # one index per sketch since a while.
```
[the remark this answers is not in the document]
jaegeral: hm true, so this can have multiple values, then I can remove the todo

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,26 @@ def post(self, sketch_id: int):
...
+        timeline_ids = query_filter.get("timelines", all_timeline_ids)
 
         # If _all in indices then execute the query on all indices
         if "_all" in indices:
             indices = all_indices
 
         # Make sure that the indices in the filter are part of the sketch.
         # This will also remove any deleted timeline from the search result.
+        # comment: jaegeral: since we write all things now to only one
+        # searchindex, this will be pointless
```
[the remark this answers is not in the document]
jaegeral: removed the comment

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,23 @@ def post(self, sketch_id: int):
 
         all_indices = list({t.searchindex.index_name for t in sketch.timelines})
         indices = query_filter.get("indices", all_indices)
+        all_timeline_ids = [t.id for t in sketch.timelines]
+        timeline_ids = query_filter.get("timelines", all_timeline_ids)
 
         # If _all in indices then execute the query on all indices
         if "_all" in indices:
             indices = all_indices
```
jkppr, line 164: I think the error you try to solve is actually here. When someone sends "_all" (most likely when using the api client) instead of the list of timelines (like our frontend does), it replaces the list of timeline ids with the list of searchindices which then results in an empty `timeline_ids` after the `get_validated_indices(indices, sketch)` call below.

See the following debug outputs:
* Request via frontend on a sketch with 2 timelines but only 1 selected
```
[2025-03-20 17:48:37 +0000] [433] [DEBUG] POST /api/v1/sketches/1/explore/
query_filter: {'from': 0, 'terminate_after': 40, 'size': 40, 'indices': [2], 'order': 'asc', 'chips': [], 'fields': [{'field': 'message', 'type': 'text'}]}
all_indices: ['1d9f6aca5d6c486aa9e42936cce0409e']
1. indices: [2] # <= Requested timeline in the frontend
all_timeline_ids: [1, 2]
1. timeline_ids: [1, 2] # <= One timeline too much since default to all_timelines_ids
2. indices: [2] # <= Still 2 since no "_all" in the list
3. indices: ['1d9f6aca5d6c486aa9e42936cce0409e'] # <= Translation to real search index happened via get_validated_indices(indices, sketch)
2. timeline_ids: [2] # <= Still the correct timeline_id after get_validated_indices(indices, sketch)
4. indices: ['1d9f6aca5d6c486aa9e42936cce0409e'] # <= Also still correct after utils.validate_indices(indices, self.datastore)
```

* Request via api client using "_all":
```
[2025-03-20 17:49:01 +0000] [433] [DEBUG] POST /api/v1/sketches/1/explore/
query_filter: {'from': 0, 'terminate_after': 40, 'size': 40, 'indices': ['_all'], 'order': 'asc', 'chips': [], 'fields': [{'field': 'message', 'type': 'text'}]}
all_indices: ['1d9f6aca5d6c486aa9e42936cce0409e']
1. indices: ['_all'] # <= Requested "_all" timelines via api client
all_timeline_ids: [1, 2]
1. timeline_ids: [1, 2]
2. indices: ['1d9f6aca5d6c486aa9e42936cce0409e'] # <= Gets replaced with the all_indices
3. indices: ['1d9f6aca5d6c486aa9e42936cce0409e']
2. timeline_ids: [] # <= empty after get_validated_indices(indices, sketch)!
4. indices: ['1d9f6aca5d6c486aa9e42936cce0409e']
```

So the easiest solution here is probably to replace `all_indices` with `all_timeline_ids`.
```suggestion
            indices = all_timeline_ids
```

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,23 @@ def post(self, sketch_id: int):
...
 
         # Make sure that the indices in the filter are part of the sketch.
-        # This will also remove any deleted timeline from the search result.
         indices, timeline_ids = get_validated_indices(indices, sketch)
-
+        # the timeline_ids seem to be [] all the time, make a check if it its empty
+        if not timeline_ids:
+            timeline_ids = all_timeline_ids
```
jkppr, line 170: If the change above is applied, this is not needed anymore.

```suggestion
```

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,23 @@ def post(self, sketch_id: int):
...
+        if not timeline_ids:
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "No valid timeline ids were found to perform the search on.",
+            )
```
jkppr, line 175: I think we can keep this. It does ensure an error in case we don't find a timeline id.

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,23 @@ def post(self, sketch_id: int):
 
         all_indices = list({t.searchindex.index_name for t in sketch.timelines})
         indices = query_filter.get("indices", all_indices)
+        all_timeline_ids = [t.id for t in sketch.timelines]
```
jkppr, line 159: This does not take into account if a timeline is ready to be searched, correct? So it would also return timeline_ids that are still in the processing status. This can lead to unexpected behaviour as well since we would return partial results for a timeline. We probably need something along the lines of
```
if timeline.get_status.status.lower() != "ready":
            continue
```

timesketch/api/v1/resources/explore.py
```diff
@@ -156,15 +156,23 @@ def post(self, sketch_id: int):
...
+        timeline_ids = query_filter.get("timelines", all_timeline_ids)
```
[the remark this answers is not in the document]
jaegeral: wel yeah we do not have that because we filter it on the frontend, but in theory we should actually do provide the timeline filter from the frontend to the backend to reduce the amount of data sent back and forth.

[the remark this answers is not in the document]
jkppr: it's already covered by `indices = query_filter.get("indices", all_indices)` (See comment below)

timesketch/api/v1/resources/explore.py
jaegeral: it does not, but the line later

```
        indices = utils.validate_indices(indices, self.datastore)
```

is checking that, so I think we are fine

jaegeral:
Ok I somehow went down a rabbit hole, I need to fix stuff first before it can be considered a good PR...

PR #3359, opened 2025-03-27
[Bandaid] introduce test case to test the issue

jaegeral:
TBD


jkppr (approved):
lgtm, thanks for the fix

PR #3364, opened 2025-03-28
Modify the github action workflows to pin the checkout version

jaegeral:
No more text


jkppr:
Is there a specific reason why you only pin the `actions/checkout` version but not the others?

PR #3371, opened 2025-04-02
Update deploy_timesketch.sh

Sh3b0:
Typo fix: prompt should be [y/N] instead of [Y/n] if the default action is not to start the containers.


jkppr (approved):
Thanks for the quick fix

PR #3400, opened 2025-05-08
[Workflows] Run unittests in paralell in github workflow

jaegeral:
Attempt to run unittests in paralell in github workflow

Description follows later


jkppr (approved):
General LGTM - let's try this.

PR #3402, opened 2025-05-09
[API Client] Robustness and Readability Enhancements

jaegeral:
## API Client Robustness and Readability Enhancements

This pull request introduces several improvements to the Timesketch API client, focusing on enhancing its robustness through retry mechanisms and improving code readability.

**Key Changes:**

*   **Enhanced Methods (in `client.py`):**
    *   **Retry Logic:** Implemented a robust retry mechanism with exponential backoff. This makes the creation of search indices and other calls more resilient to transient network issues or temporary server-side errors.
    *   **Docstring Improvement:** The docstring for several methods has been updated to be more comprehensive, detailing its functionality, arguments, return values, and the exceptions it might raise, including those related to the new retry logic.
*   **Enhanced `Sketch.create_timeline` Method (in `sketch.py`):**
    *   **Retry Logic:** Similar to `create_searchindex`, this method now includes a retry mechanism with exponential backoff to handle transient errors during timeline creation. It also performs more explicit validation of the API response to ensure the timeline was created successfully.

These changes aim to make the API client more reliable and the codebase easier to maintain.

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -773,38 +773,123 @@ class TimesketchApi:
 
     def create_searchindex(self, searchindex_name: str, opensearch_index_name: str):
         """Create a new SearchIndex.
+
+        This method attempts to create a new searchindex on the Timesketch server.
+        It implements a retry mechanism with exponential backoff if the initial
+        request fails due to network issues, API errors, or unexpected
+        response formats.
+
         Args:
             searchindex_name: Name for the searchindex.
             opensearch_index_name: The name of the index in opensearch.
+
         Returns:
             Instance of a SearchIndex object.
+
         Raises:
-            ValueError: If the SearchIndex fails to create.
+            RuntimeError: If the SearchIndex fails to create after all retries,
+                or if the API returns an unexpected response format.
+            requests.exceptions.RequestException: If a connection error persists
+                after all retries.
+            ValueError: If the API response cannot be JSON-decoded after all retries.
         """
         resource_url = f"{self.api_root}/searchindices/"
         form_data = {
             "searchindex_name": searchindex_name,
             "es_index_name": opensearch_index_name,
         }
-        response = self.session.post(resource_url, json=form_data)
+        last_exception = None
 
-        if response.status_code not in definitions.HTTP_STATUS_CODE_20X:
-            error.error_message(
-                response,
-                message="Error creating searchindex",
-                error=ValueError,
-            )
+        for attempt in range(self.DEFAULT_RETRY_COUNT):
+            try:
+                response = self.session.post(resource_url, json=form_data)
+                response_dict = error.get_response_json(
+                    response, logger
+                )  # Raises RuntimeError for non-20x, ValueError for JSON decode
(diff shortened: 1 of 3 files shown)


api_client/python/timesketch_api_client/client.py
```diff
@@ -365,102 +367,232 @@ def _create_session(
...
+        attempt = 0
         result = None
         while True:
-            retry_count += 1
-            response = self.session.get(resource_url, params=params)
+            attempt += 1
+            # If this is not the first attempt, wait before trying again
+            if attempt > 1:
+                # Simple fixed backoff before retrying
+                time.sleep(1)  # Wait for 1 second before the next attempt
```
jaegeral, line 407: happy to also make the exponential sleep here, but I was not sure if that is needed.

jkppr: Let's do an exponential backoff here if we already touch this code.

jkppr, line 407: Should we add a message here to let the user know that they ran into a backoff?

api_client/python/timesketch_api_client/client.py
```diff
@@ -365,102 +367,232 @@ def _create_session(
...
-                    "[{0:d}/{1:d}] Parsing the response for request '{2:s}'"
-                    "failed. Trying again...".format(
-                        retry_count, self.DEFAULT_RETRY_COUNT, resource_url
-                    )
+                    "[%d/%d] API error (RuntimeError) for request '%s' "
+                    "failed. Error: %s. Trying again...",
+                    attempt,
+                    self.DEFAULT_RETRY_COUNT,
+                    resource_url,
+                    e,
```
jkppr, line 427:
```suggestion
                    str(e),
```

I would recommend to send `e` through `str()` here to ensure it is compatible with `%s`. At least for me in the past, not doing this caused some issues before.

api_client/python/timesketch_api_client/client.py
```diff
@@ -365,102 +367,232 @@ def _create_session(
...
-                            resource_url, e
-                        )
-                    ) from e
+                if attempt >= self.DEFAULT_RETRY_COUNT:
+                    # Re-raise the original error after exhausting retries
+                    error_msg = (
+                        f"Error parsing response for request '{resource_url}'"
+                        f" - {e!s}"
+                    )
+                    raise ValueError(error_msg) from e
```
jkppr, line 436: What do we expect to change with a retry that would fix a "ValueError"?

jkppr (commented):
First round

api_client/python/timesketch_api_client/client.py
jaegeral: yep

api_client/python/timesketch_api_client/client.py
jaegeral: Good question, the value in question here is something from the server, that is expected to be JSON, but due to several reasons, including network error or temporary server issues, the server COULD return not valid JSON. So while in theory, that should never be the case, it could, does that make sense?

jkppr (approved):
lgtm, thanks

PR #3407, opened 2025-05-12
Fix TimelineChip failed mode

jkppr:
There was a bug where the a timeline in status "failed" would still show as processing all the time in the frontend even the status was correctly set to "fail".

This PR fixes this by adding the "fail"  status to the conditions for the `timelineAvailable` compute value. By setting this value true, the timeline gets out of the processing slot and into a final state.

Without the fix: 
image

With the fix:
image


jaegeral (approved):
LGTM

PR #3409, opened 2025-05-14
Efficient Bulk Export with Opensearch using PIT and Slicing

jkppr:
This pull request introduces a significant enhancement to the Opensearch datastore, enabling efficient bulk export of events using Opensearch's Point-In-Time (PIT) and Sliced Scroll capabilities.

### Key Changes:

**Opensearch Datastore Enhancements (`timesketch/lib/datastores/Opensearch.py`):**

*   **New `export_events_with_slicing` Method:**
    *   Provides a new public method to export large volumes of event data from specified Opensearch indices.
    *   Utilizes Opensearch Point-In-Time (PIT) to ensure a consistent view of the data across multiple paginated and sliced requests.
    *   Employs `search_after` and `slice` features for robust and performant pagination and parallelization.
    *   Yields individual event dictionaries, including `_id` and `_index`.
    *   Designed for efficient bulk export, minimizing load on the Opensearch cluster and improving export speed.
*   **New Private `_export_slice_worker` Method:**
    *   A threaded worker function responsible for processing a single slice of the export.
    *   Manages PIT creation and deletion for its assigned slice.
    *   Fetches documents using `search_after` within its slice.
    *   Puts fetched event data onto a shared queue for consumption by the main thread.
    *   Includes robust error handling, backpressure management for the queue, and logging.
*   **Configuration Options:**
    *   The `__init__` method of `OpensearchDataStore` now accepts `**kwargs` to allow overriding default Opensearch client parameters.
    *   New configuration parameters (retrieved from `current_app.config` or using sensible defaults) have been added to control the behavior of the sliced export:
        *   `Opensearch_SLICED_EXPORT_DEFAULT_PAGE_SIZE` (default: 10000)
        *   `Opensearch_SLICED_EXPORT_PIT_KEEP_ALIVE` (default: "5m")
        *   `Opensearch_SLICED_EXPORT_NUM_SLICES` (default: 4)
        *   `Opensearch_SLICED_EXPORT_REQUEST_TIMEOUT` (default: Opensearch client timeout + 20s)
        *   `Opensearch_SLICED_EXPORT_QUEUE_BUFFER_FACTOR` (default: 4, used to calculate internal queue size)
*   **Default Sort Criteria:**
    *   Added `_DEFAULT_PIT_SORT_CRITERIA = [{"_id": "asc"}]` for stable pagination in PIT exports if no specific sort order is provided.
*   **Imports:**
    *   Added necessary imports: `time`, `queue`, `threading`, and typing hints (`Generator`, `List`, `Any`).

### Benefits:

*   **Improved Export Performance:** Leverages Opensearch PIT and slicing for significantly faster and more reliable export of large datasets compared to traditional scroll APIs, especially on busy clusters.
*   **Data Consistency:** PIT ensures that the exported data represents a consistent snapshot in time, avoiding issues with data changing during a long export process.
*   **Reduced Cluster Load:** Slicing allows the export workload to be distributed, potentially reducing the impact on the Opensearch cluster.
*   **Enhanced Configurability:** Provides new configuration options to fine-tune export parameters based on specific needs and environment.

### How the Sliced Export Works:

1.  A Point-In-Time (PIT) ID is created for the target indices, providing a fixed view of the data.
2.  The export is divided into a configurable number of `slices`.
3.  A separate worker thread is spawned for each slice.
4.  Each worker uses its `slice_id`, the `PIT ID`, and `search_after` to retrieve documents for its portion of the data.
5.  Documents are fetched in pages and put onto a thread-safe queue.
6.  The main thread consumes documents from the queue and yields them to the caller.
7.  Workers manage the lifecycle of their PIT context (extending keep-alive) and delete the PIT upon completion or error.
8.  The system includes mechanisms for backpressure (if the queue fills up) and coordinated stopping of workers in case of errors.

This new export functionality will be particularly beneficial for use cases requiring the extraction of large numbers of events from Timesketch, such as for external analysis, archiving, or integration with other systems.


timesketch/lib/datastores/opensearch.py
```diff
@@ -1237,3 +1247,290 @@ def flush_queued_events(self, retry_count=0):
...
+        current_sort_criteria: List[Dict[str, Any]],
+        current_page_size: int,
+        current_pit_keep_alive: str,
+        output_queue: queue.Queue,
+        stop_event: threading.Event,
+        worker_request_timeout: int
+    ):
+        """
+        Worker function for a single slice in a PIT export.
+        Fetches documents for its assigned slice and puts them onto the output_queue.
```
jaegeral, line 1266: could we make the docstring a bit more descriptive on what is going on, when errors are dropped, how the parameters are used?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1237,3 +1247,290 @@ def flush_queued_events(self, retry_count=0):
...
+        if not indices_for_pit:
+            raise ValueError("indices_for_pit cannot be empty.")
+        if not 1 <= num_slices <= 1024:  # OpenSearch constraint for slices
+            raise ValueError("num_slices must be between 1 and 1024.")
+
+        effective_sort_criteria = sort_criteria or _DEFAULT_PIT_SORT_CRITERIA
+        worker_timeout = request_timeout_per_slice or (self.timeout + 20) # Use configured or default+buffer
+
+        # Adjust num_slices from app config if not provided or if a default is desired.
+        # Example: num_slices = num_slices or current_app.config.get('OPENSEARCH_PIT_NUM_SLICES', 5)
```
jaegeral, line 1433: shall we remove that if we do not use a Config variable here?

jaegeral (commented):
some minor nits for now

timesketch/lib/datastores/opensearch.py
jkppr: This value should be configurable for sure. The question is where we consume the config. 
I would probably opt for a solution that uses the value from the configuration as default but still can be overwritten by a caller if necessary. What do you think?

jaegeral: how about we give it a Default here and if a config value is present to use the config?

jkppr: I went with the following priority: Caller arg > timesketch.conf setting > default value

timesketch/lib/datastores/opensearch.py
jkppr: done

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                    break
+                except ConnectionTimeout:
+                    os_logger.warning(
+                        "[Slice %s/%s] Connection timeout during search. "
+                        "Stopping slice.",
+                        log_slice_id,
+                        num_slices,
+                    )
+                    break
+                except Exception as e:  # pylint: disable=broad-exception-caught
```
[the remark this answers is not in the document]
jaegeral: this can be ignored

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                "Sliced export: Signaling any remaining worker threads to stop."
+            )
+            stop_event.set()
+
+            for i, thread in enumerate(threads):
+                if thread.is_alive():
+                    os_logger.info(
+                        "Sliced export: Waiting for worker thread %s to join.", i + 1
+                    )
+                    thread.join(timeout=10)  # Give threads a chance to clean up PIT
```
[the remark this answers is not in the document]
jaegeral: @jkppr WDYT?

timesketch/lib/datastores/opensearch.py
```diff
@@ -114,7 +121,7 @@ class OpenSearchDataStore:
         "yellow"  # Minimum health status required ('yellow' or 'green')
     )
 
-    def __init__(self, host="127.0.0.1", port=9200):
+    def __init__(self, host="127.0.0.1", port=9200, **kwargs):
```
jaegeral, line 124: Since you introduce `kwargs` as parameter, could you explain in the docstring, what is expected and what the keyword arguments are expected / might be provided?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                    if stop_event.is_set():
+                        break
+                    if "_source" in hit:
+                        event_data_to_export = {
+                            **hit["_source"],
+                            "_id": hit.get("_id"),
+                            "_index": hit.get("_index"),
+                        }
+                        # Try to put item onto the queue with timeout
+                        # This loop handles backpressure if the queue is full.
```
jaegeral, line 1449: what does backpressure mean here?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                            try:
+                                output_queue.put(
+                                    event_data_to_export, timeout=1
+                                )  # Short timeout for responsiveness
+                                break  # Item put successfully
+                            except queue.Full:
+                                # Queue is full, wait a bit and retry.
+                                # stop_event is checked at the start of this inner loop.
+                                time.sleep(0.1)  # Avoid busy-waiting
+                            except (  # pylint: disable=broad-exception-caught
```
jaegeral, line 1460: is there a way to refactor this so the code does not have to be that much indented and readability is improved?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                        ),
+                        log_slice_id,
+                        num_slices,
+                    )
+                    stop_event.set()
+                    break
+
+        except NotFoundError:
+            os_logger.warning(
+                "[Slice %s/%s] PIT ID %s not found or expired, "
```
jaegeral, line 1511: can we somehow distinguish between expired and not found PITs?
And in a similar vain is it possible to see if the index was not found? It feels like this catch is pretty broad for some very different issues that would deserve their own handling

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+
+        This method is designed for efficient bulk export of large datasets.
+        It yields individual event dictionaries, including '_id' and '_index'.
+
+        The following parameters can be overridden by the caller. If not provided,
+        they will default to values from the Timesketch configuration or
+        sensible internal defaults:
+            - `page_size`: Defaults to `OPENSEARCH_SLICED_EXPORT_DEFAULT_PAGE_SIZE`
+            - `pit_keep_alive`: Defaults to `OPENSEARCH_SLICED_EXPORT_PIT_KEEP_ALIVE`
+            - `num_slices`: Defaults to `OPENSEARCH_SLICED_EXPORT_NUM_SLICES` or 4
```
jaegeral, line 1597: be explicit that if the config variable is not set it defaults to 4 (ambiguity)

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+            - `request_timeout_per_slice`: Defaults to
+              `OPENSEARCH_SLICED_EXPORT_REQUEST_TIMEOUT` or a calculated value
+              based on the general OpenSearch timeout +20s.
```
jaegeral, line 1600: so what is the final default value? 30 sec (indicated below)?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+            indices_for_pit: List of index names to query.
+            base_query_body: The base OpenSearch query body (must not include
+                             size, sort, pit, or slice clauses). Can include
+                             "_source" to specify fields to return.
+            sort_criteria: Optional. The sort criteria for search_after.
+                           If None, defaults to `[{"_id": "asc"}]`.
+                           Must include a unique tie-breaker field (e.g. _id)
+                           for stable results across pages.
+            page_size: Optional. Number of documents per request per slice.
+                       Overrides the global default (10_000) if provided.
```
jaegeral, line 1615: s/10_000/10000

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+            pit_keep_alive: Optional. Keep-alive duration for the Point-In-Time context
```
jaegeral, line 1616: since this is a string, maybe explain a bit more in what format it should be provided?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+            - `request_timeout_per_slice`: Defaults to
+              `OPENSEARCH_SLICED_EXPORT_REQUEST_TIMEOUT` or a calculated value
+              based on the general OpenSearch timeout +20s.
+
+        The internal queue size is determined by `page_size`, `num_slices`, and
+        the `OPENSEARCH_SLICED_EXPORT_QUEUE_BUFFER_FACTOR` configuration.
+
+        Args:
+            indices_for_pit: List of index names to query.
+            base_query_body: The base OpenSearch query body (must not include
```
jaegeral, line 1607: do you want to make the `must not` section bold to stand out to users?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                             size, sort, pit, or slice clauses). Can include
```
jaegeral, line 1608: and also can we add a check in the code in case someone passes the must_not keys??

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+        they will default to values from the Timesketch configuration or
+        sensible internal defaults:
+            - `page_size`: Defaults to `OPENSEARCH_SLICED_EXPORT_DEFAULT_PAGE_SIZE`
+            - `pit_keep_alive`: Defaults to `OPENSEARCH_SLICED_EXPORT_PIT_KEEP_ALIVE`
+            - `num_slices`: Defaults to `OPENSEARCH_SLICED_EXPORT_NUM_SLICES` or 4
+            - `request_timeout_per_slice`: Defaults to
+              `OPENSEARCH_SLICED_EXPORT_REQUEST_TIMEOUT` or a calculated value
+              based on the general OpenSearch timeout +20s.
+
+        The internal queue size is determined by `page_size`, `num_slices`, and
```
jaegeral, line 1602: is that an infirmation a dev / user needs to know in the public dosctring or more details in the implementation?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+        """
+        if not indices_for_pit:
+            raise ValueError("indices_for_pit cannot be empty.")
+
+        # Getting default values if not provided otherwise by the caller.
+        effective_sort_criteria = sort_criteria or _DEFAULT_PIT_SORT_CRITERIA
+        effective_page_size = (
+            page_size if page_size is not None else self.sliced_export_default_page_size
+        )
+        effective_pit_keep_alive = (
```
jaegeral, line 1640: Just a comment to this (and the following) While I understand, what the code does, it takes a bit to really process what those three lines do. Did you consider to write in if/else block instead?

e.g. 

```
if num_slices is None:
    effective_num_slices = self.sliced_export_num_slices_default
else:
    effective_num_slices = num_slices
```

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+        Raises:
+            ValueError: If indices_for_pit is empty or num_slices is invalid.
+            errors.DataStoreQueryError: For critical issues during the export.
+        """
+        if not indices_for_pit:
+            raise ValueError("indices_for_pit cannot be empty.")
+
+        # Getting default values if not provided otherwise by the caller.
+        effective_sort_criteria = sort_criteria or _DEFAULT_PIT_SORT_CRITERIA
+        effective_page_size = (
```
jaegeral, line 1637: also, can page:size be negative? if not, consider checking

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                        effective_sort_criteria,
+                        effective_page_size,
+                        effective_pit_keep_alive,
+                        results_queue,
+                        stop_event,
+                        effective_worker_timeout,
+                    ),
+                    # Allows main thread to exit even if workers hang
+                    # (though join is preferred)
+                    daemon=True,
```
jaegeral, line 1697: given you have a timeout in the .join, do you think the daemon is needed?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                )
+                threads.append(thread)
+                thread.start()
+
+            while active_workers > 0:
+                if stop_event.is_set() and results_queue.empty():
```
jaegeral, line 1703: do you also want to stop the while loop when `active_workers <= 0`

timesketch/lib/datastores/opensearch.py
```diff
@@ -1271,3 +1296,505 @@ def flush_queued_events(self, retry_count=0):
...
+                )
+            except Exception as e_s:  # pylint: disable=broad-exception-caught
+                os_logger.error(
+                    "[Slice %s/%s] Failed to put sentinel: %s",
+                    log_slice_id,
+                    num_slices,
+                    str(e_s),
+                )
+
+    def export_events_with_slicing(
```
jaegeral, line 1574: a general question, if this is used on a busy system, how would one see which slice log relates to which? Would it make sense to create uuid() in any way and add it to every log messa written in one export job?

timesketch/lib/datastores/opensearch.py
jkppr: Backpressure is a mechanism where a data consumer signals to a data producer to slow down or temporarily stop sending data. This happens when the consumer can't process data as fast as the producer is sending it, preventing the consumer from being overwhelmed. It's like a traffic control system for data flow, ensuring stability and preventing data loss.

So this loop will trigger if the queue is full and the worker needs to wait a bit before it can add new data. This ensures the queue does not grow out of RAM etc.

timesketch/lib/datastores/opensearch.py
jkppr: Let me try putting some logic into a helper function. Maybe that helps to make the code more readable.

timesketch/lib/datastores/opensearch.py
jkppr: Fair point. I updated the note on the default value.

timesketch/lib/datastores/opensearch.py
jkppr: This is relevant for tuning the settings for your instance. By changing the `num_slices` and `OPENSEARCH_SLICED_EXPORT_QUEUE_BUFFER_FACTOR` you can adjust how much RAM the queue uses and how many connections are opened to Opensearch.

timesketch/lib/datastores/opensearch.py
jkppr: Fair point, it takes a moment to get used reading this type of conditional expression. It is just shorter than `if/else` blocks. But I'm fine with both option. Do you have preferences?

timesketch/lib/datastores/opensearch.py
jkppr: What the page_size can and cannot be is basically handled by the opensearch API. If you set it to negative (or larger than your opensearch instance is configured to handle) the opensearch client will raise an error with the search and halt all exports.

timesketch/lib/datastores/opensearch.py
jkppr: I went with a `if/else` approach for now. It is indeed easier to read.

timesketch/lib/datastores/opensearch.py
jkppr: The idea here is a fallback mechanism to safeguard against indefinite hangs if the join timeout is insufficient or if threads fail to terminate properly.

timesketch/lib/datastores/opensearch.py
jkppr: Yes, if the `active_workers` get negative at any time (don't know how) the loop should also stop.

timesketch/lib/datastores/opensearch.py
jkppr: Each slice log (if enabled in the opensearch.py - by default those logs are not visible) is listing its slice ID in the front. e.g. `"[Slice %s/%s] Retrieved %s docs. Total for this slice so far: %s."`.
And if used with the celery system for example they will be tied to the celery execution ID.

timesketch/lib/datastores/opensearch.py
jkppr: Fair point. I have reworked the errors and made them more specific in their messaging.

timesketch/lib/datastores/opensearch.py
jaegeral: ACK

timesketch/lib/datastores/opensearch.py
jaegeral: no that sounds fine

timesketch/lib/datastores/opensearch.py
jaegeral: ah okay, yeah then its fine

jaegeral (approved):
LGTM but please fix the linter before merging

timesketch/lib/datastores/opensearch.py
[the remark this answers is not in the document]
jkppr: I made it configurable. In general it would not matter much if the PIT is not deleted within 10sec, since they are created with a 5min keep alive. Once they are not accessed anymore for 5min they will be removed by the opensearch cluster anyway.

jkppr:
Opened #3427 to track adding tests for the OpenSearchDataStore class.

PR #3414, opened 2025-05-14
Improve OpenSearch search method docstring and error logging

jaegeral:
## Improve OpenSearch search method docstring and error logging

This pull request addresses improvements to the `search` method within the OpenSearch datastore (`timesketch/lib/datastores/opensearch.py`).

Key changes include:

-   **Enhanced Docstring:** The docstring for the `search` method has been significantly improved for clarity and completeness. It now provides more detailed descriptions for each parameter, explicitly defines the return type (`Union[Dict, int]`), and includes a `Raises` section to document potential `ValueError` exceptions that can occur due to OpenSearch errors.
-   **Improved Error Logging:** Error handling for `opensearchpy.exceptions.RequestError` and `opensearchpy.exceptions.TransportError` has been enhanced. When these exceptions occur during a search, the error message logged now includes more context, such as the sketch ID, the indices being queried. This additional information will greatly assist in debugging search-related issues.


timesketch/lib/datastores/opensearch.py
```diff
@@ -590,26 +590,49 @@ def search(
...
-            indices: List of indices to query
-            count: Boolean indicating if we should only return result count
-            aggregations: Dict of OpenSearch aggregations
-            return_fields: List of fields to return
-            enable_scroll: If OpenSearch scroll API should be used
+            sketch_id: The ID of the sketch the search is performed within. Used
+                for building label filters.
+            indices: A list of OpenSearch index names to query.
+            query_string: The query string to search for (e.g., "hostname:evil.com").
+                Defaults to an empty string, which matches all documents.
```
jkppr, line 606: > ... which matches all documents
Is that true? I don't think this is how it works?

timesketch/lib/datastores/opensearch.py
```diff
@@ -705,8 +728,13 @@ def search(
...
                 cause = str(e)
-
-            es_logger.error("Unable to run search query: %s", cause, exc_info=True)
+            log_message = (
+                f"Unable to run search query. Error: {cause}. "
+                f"Sketch ID: {sketch_id}. Indices: {indices}. "
+                f"Query String: '{query_string}'. Query Filter: {query_filter}. "
+                f"Query DSL: {query_dsl}"
+            )
+            os_logger.error(log_message, exc_info=True)
```
jkppr, line 737: For strings that are loaded into logger output we should always use %s lazy evaluation. This means the string formatting is deferred until it's absolutely necessary (i.e., when the log level is enabled and the message needs to be output). This ensures a better performance.

timesketch/lib/datastores/opensearch.py
```diff
@@ -705,8 +728,13 @@ def search(
...
             raise ValueError(cause) from e
```
jkppr, line 738: Do we also want to improve the error message that is presented to the user in the red snackbar? This would need to go into the `ValueError` here.

jkppr (changes requested):

timesketch/lib/datastores/opensearch.py
jaegeral: I was thinking about that as well, happy to change and yeah I think that would help

jkppr: yes, please add at least an info that something in their query was not legit and the user should revisit their search query.

timesketch/lib/datastores/opensearch.py
jaegeral: ack

timesketch/lib/datastores/opensearch.py
jaegeral: done

timesketch/lib/datastores/opensearch.py
jaegeral: adjusted the docsgtring

jkppr (approved):

timesketch/lib/datastores/opensearch.py
```diff
@@ -706,8 +729,24 @@ def search(
             else:
                 cause = str(e)
 
-            es_logger.error("Unable to run search query: %s", cause, exc_info=True)
-            raise ValueError(cause) from e
+            os_logger.error(
+                "Unable to run search query. Error: %s. "
+                "Sketch ID: %s. Indices: %s. "
+                "Query String: '%s'. Query Filter: %s. "
```
jkppr, line 735: We probably shouldn't log the query string here since this will add potential case data into the logs.
Let's remove it here for now. Long term solution would probably a prometheus logging for those if we need debugging. Or add a separate `os_logger.debug` call that a TS admin can explicitly enable to debug issues.

jkppr (changes requested):

timesketch/lib/datastores/opensearch.py
jaegeral: SGTM

jkppr (approved):

PR #3226, opened 2024-11-21
Compose dev

jbaptperez:
**IMPORTANT: All Pull Requests should be connected to an issue, if you don't
have an issue, please start by creating an issue and link it to the PR.**

Please provide enough information so that others can review your pull request:

+ What existing problem does this PR solve?
  - Issue #3225 (Adapt the development Compose set-up to allow restricted company contexts).
+ What new feature is being introduced with this PR?
  - Add to the development set-up the possibility to customise image builds (among other) using a `.env` file,
  - Update of the Compose file syntax to the Compose specification,
  - Other minor enhancements related to the development Compose set-up
+ Overview of changes to existing functions if required.
  - The current behavior remain the same, but the enhancement of the development set-up leads to different paths to the Dockerfiles and simplified commands to manage the development container.

**Checks**

- [x] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [x] Documentation updated.

**Closing issues**

Closes #3225.

### .gitattributes and .dockerignore

Adds .gitattributes file to handle line endings, in particular when developing on Windows and building Linux images.
Adds .dockerignore file to limit the context transferred when building Docker images.

### New Compose set-up for development

- Located at `contrib/docker/dev`.
- **Does NOT replace the `docker/dev` set-up** for now.
- Uses a **latest Compose syntax**.
- Relies only on Docker Compose CLI, not both Docker Compose and Docker CLIs anymore.
- **Allows image builds in enterprise context**: Optional dependency source proxies can be set for Ubuntu, Python and Node.
- Uses **optimized multi-stage builds** to locally build images.
- Centralizes variables in an non-versioned `.env` file, with a `.env.template` model file.
- Uses a structured file hierarchy for Compose services.
- Uses per service environment files.
- **Uses only named volumes** for speed, portability and to store all data that are not configuration or source files (use `docker compose down -v` to reset).
- **Splits the main service** (container) into one per component (setup, celery-worker, gunicorn, frontend).
- Makes the frontend service compatible with the 3 frontends implementations (switch via an environment variable).
- **Dramatically optimizes the setup stage (dedicated service)** (shallow clones/fetches the Sigma repository, calls `tsctl` once to register all rules at once).
- Uses a single entry point for each Docker service.
- **Makes each service run on PID 1** to properly receive Docker signals.**
[...]

PR #3424, opened 2025-05-20
Developemnt fixes

jbaptperez:
Minor enhancements in documentation.

Extracted from the [compose-dev](https://github.com/google/timesketch/pull/3226) PR.


docs/developers/frontend-development.md
```diff
@@ -46,13 +46,15 @@ Follow the steps in the previous section to get dependencies installed and the c
 
 You need two shells:
 
-1. Start the main webserver (for serving the API etc) in the first shell:
+1/ Start the main webserver (for serving the API etc.) in the first shell:
```
jaegeral, line 49: Was there a specific reason to change from `1.` to `1/`? (in some other places as well)

jbaptperez: I did it to remove a Markdown warning in my IDE because each numbered line are considered belonging to an independent numbered list of one element.
Actually, the code snippets in-between break a normal list.

I just pushed a new version that can combine a proper numbered list with a code snippet by just indent the code snippet.

jaegeral: excellent, thx

jaegeral (approved):
LGTM, thank you for the contribution, those document cleanup look small, but are much appreciated and important, so thank you for this and also the quick reaction on the feedback, that makes review smooth and nice :-)

PR #3431, opened 2025-05-23
[tsctl] Archive improvements

### LINKED ISSUE #3090: tsctl archive sketches
With a larger Timesketch setup, over time there can be many searchindex instances opened on the opensearch cluster. It is possible to manually archive a sketch in the UI which will close the searchindex on the opensearch side, limiting the consumption of resources.

For an easy clean-up and resource management, I propose a feature in the `tsctl` admin tool that allows to easily archiving all sketches that are have not been touched for the last X days.

This can also be done with an external script and the API. However, having it integrated into the admin tool makes sense to me.

https://github.com/google/timesketch/blob/master/timesketch/tsctl.py

jaegeral:
## Summary of Changes

This pull request significantly enhances Timesketch's archiving capabilities by refining the core logic for managing sketches, timelines, and their OpenSearch indexes. It introduces more robust error handling and pre-checks during archival, ensuring data consistency and preventing premature closure of shared indexes. Additionally, the `tsctl` command-line tool is upgraded to provide administrators with better visibility into archived sketches, particularly those with potentially inconsistent OpenSearch index states, and improved options for listing sketches. The changes also include updated documentation and frontend optimizations for archived sketches.

### Highlights

* **Improved Sketch Archiving Logic**: The internal `_archive_sketch` method has been significantly refactored to correctly handle SearchIndexes shared across multiple timelines. It now only attempts to close an OpenSearch index if *all* associated timelines are marked as archived. Error handling for OpenSearch interactions during closure has also been improved, ensuring the database status of the SearchIndex accurately reflects whether the OpenSearch index was successfully closed or confirmed not found.
* **Enhanced `tsctl list-sketches` Command**: The `tsctl list-sketches` command gains two new options: `--archived` to list only archived sketches, and `--archived-with-open-indexes` to specifically identify archived sketches that still have associated SearchIndexes in a non-archived state (e.g., 'ready', 'processing', 'fail') or are actually open in OpenSearch. This helps administrators identify potential issues with the archival process. An `--include-deleted` option was also added.
* **Updated Archive API Documentation**: The docstring for the GET method of the sketch archive API endpoint (`/api/v1/sketches/<int:sketch_id>/archive/`) has been updated to clearly describe the structure of the JSON response, including details about the sketch's archival status, ID, name, and the archival status of its associated timelines.
* **Archiving Pre-checks**: The `_archive_sketch` method now includes a check to prevent archiving sketches that contain timelines in certain non-archivable states (`processing`, `fail`, `timeout`), providing a clearer error message and actionable advice to the user.
* **Frontend Optimizations for Archived Sketches**: The frontend (both `frontend-ng` and `frontend-v3`) now includes guard clauses to prevent unnecessary API calls for fetching tag and data type information when a sketch is in an archived state, improving performance and reducing errors.

<details>
<summary><b>Changelog</b></summary>

* **docs/guides/user/sketch-overview.md**
    * Expanded documentation for 'Archive' and 'Unarchive' sections, detailing conditions, processes, and troubleshooting steps.
    * Added explanations for when a sketch can/cannot be archived/unarchived and what happens during these operations.
* **timesketch/api/v1/resources/archive.py**
    * Updated imports to include `HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR` and `db_session`.
    * Enhanced docstring for `GET /api/v1/sketches/<int:sketch_id>/archive/` to describe JSON response payload and potential errors.
    * Refactored `_archive_sketch` method to include pre-checks for sketch status and labels preventing archival.
    * Implemented checks to abort archiving if any timeline is in a 'processing', 'fail', or 'timeout' state, providing user-friendly error messages.
    * Revised logic for closing OpenSearch indexes: an index is now only closed if *all* associated timelines (across all sketches) are archived.
    * Improved error handling during OpenSearch index closure, setting SearchIndex DB status to 'archived' on success, 'fail' if the OpenSearch index is not found, and leaving status unchanged on other errors to indicate inconsistencies.
    * Ensured database changes (sketch, timeline, and searchindex statuses) are committed only after successful OpenSearch operations.
    * Refactored `_unarchive_sketch` method to include pre-checks for sketch status (must be 'archived').
    * Added warnings for timelines in 'processing' or 'fail' states during unarchiving, advising resolution rather than aborting the process for backward compatibility.
    * Implemented logic to attempt opening all necessary OpenSearch indices during unarchiving, handling cases where an index is already open or not found.
    * Updated sketch, timeline, and searchindex statuses to 'ready' upon successful unarchiving.
* **timesketch/frontend-ng/src/store.js**
    * Added guard clauses to `updateTimelineTags` and `updateDataTypes` actions to prevent fetching data for archived sketches.
* **timesketch/frontend-v3/src/stores/app.js**
    * Added guard clauses to `updateTimelineTags` and `updateDataTypes` actions to prevent fetching data for archived sketches.
* **timesketch/tsctl.py**
    * Modified `list-sketches` command to include new options: `--archived` (list only archived sketches), `--archived-with-open-indexes` (find archived sketches with inconsistent OpenSearch index states), and `--include-deleted` (show deleted sketches).
    * Implemented mutual exclusivity check for `--archived` and `--archived-with-open-indexes` options.
    * Enhanced `list-sketches` output to clearly indicate sketch ID, name, and status, with improved formatting for names containing spaces.
    * Updated `searchindex_info` command to display the status of the SearchIndex and its associated timelines.
</details>

<details>
<summary><b>Activity</b></summary>

* The author, @jaegeral, initiated the pull request and requested summaries multiple times.
* Initial summary and review comments were provided by `gemini-code-assist[bot]`, highlighting key changes and suggesting improvements, particularly regarding the `open_index_statuses` in `tsctl` and the re-introduction of type hints.
* Extensive discussions occurred with reviewer `jkppr` over multiple review cycles, covering various aspects of the archiving and unarchiving logic.
* Key discussions included the handling of timelines in 'processing', 'fail', or 'timeout' states during archiving, leading to refined pre-checks and actionable error messages.
* The logic for closing shared OpenSearch indexes was critically reviewed and improved to ensure an index is only closed if all associated timelines are archived, preventing data unavailability for other sketches.
* Error handling during OpenSearch interactions was a significant focus, with `jkppr` raising concerns about inconsistent states and `jaegeral` implementing fixes to ensure database statuses accurately reflect OpenSearch index states (e.g., setting to 'fail' if an index is not found).
* The `tsctl list-sketches` command's new options were refined based on feedback, including the correct definition of 'open' index statuses and the addition of an `--include-deleted` flag.
* UI issues related to unarchiving were identified by `jkppr` and subsequently fixed by `jaegeral`, ensuring a smooth user experience.
* Logging and error messages were improved throughout the codebase to provide clearer information to administrators and users, with careful consideration for not logging sensitive data.
* Backward compatibility for unarchiving sketches with previously 'failed' or 'processing' timelines was discussed, leading to a compromise of logging warnings instead of aborting the unarchive process, with a plan for future enforcement.
* The order of operations in the archive/unarchive methods was adjusted to ensure data consistency, such as updating the sketch status as the final step.
</details>

```
[2025-08-22 13:15:14 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/timelines/20/analysis/
[2025-08-22 13:15:33 +0000] [9640] [DEBUG] POST /api/v1/sketches/12/archive/
[2025-08-22 13:15:33,158] timesketch.api_archive/INFO Attempting to close OpenSearch index: d1d5d0586253410cb71c68b7735c9110 (DB ID: 17) for sketch 12.
[2025-08-22 13:15:33,266] timesketch.api_archive/INFO Successfully closed OpenSearch index: d1d5d0586253410cb71c68b7735c9110. SearchIndex DB object (ID: 17) status updated to 'archived'.
[2025-08-22 13:15:33,281] timesketch.api_archive/INFO Sketch 12 status set to 'archived'.
[2025-08-22 13:15:33 +0000] [9641] [DEBUG] GET /api/v1/sketches/12/
[2025-08-22 13:15:33 +0000] [9640] [DEBUG] GET /api/v1/users/me/
[2025-08-22 13:15:33 +0000] [9640] [DEBUG] GET /api/v1/searchtemplates/
[2025-08-22 13:15:33 +0000] [9640] [DEBUG] GET /api/v1/sigmarules/
[2025-08-22 13:15:33 +0000] [9643] [DEBUG] GET /api/v1/sketches/12/timelines/19/analysis/
[2025-08-22 13:15:33 +0000] [9641] [DEBUG] GET /api/v1/intelligence/tagmetadata/
[2025-08-22 13:15:33 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/aggregation/
[2025-08-22 13:15:33 +0000] [9640] [DEBUG] GET /dist/empty-state.png
[2025-08-22 13:15:33 +0000] [9643] [DEBUG] GET /api/v1/sketches/12/timelines/20/analysis/
[2025-08-22 13:15:35 +0000] [9643] [DEBUG] POST /api/v1/sketches/12/archive/
[2025-08-22 13:15:35,594] timesketch.api_archive/INFO Unarchiving sketch 12 ('New Sketch From Importer CLI').
[2025-08-22 13:15:35,602] timesketch.api_archive/INFO Attempting to open OpenSearch index: d1d5d0586253410cb71c68b7735c9110 (DB ID: 17) for sketch 12.
[2025-08-22 13:15:35,826] timesketch.api_archive/INFO Successfully opened OpenSearch index: d1d5d0586253410cb71c68b7735c9110.
[2025-08-22 13:15:35,840] timesketch.api_archive/INFO Sketch 12 status set to 'ready'.
[2025-08-22 13:15:35,852] timesketch.api_archive/INFO Timeline 19 ('foo') status set to 'ready'.
[2025-08-22 13:15:35,861] timesketch.api_archive/INFO Timeline 20 ('foo2') status set to 'ready'.
[2025-08-22 13:15:35,873] timesketch.api_archive/INFO SearchIndex DB object d1d5d0586253410cb71c68b7735c9110 (ID: 17) status updated to 'ready'.
[2025-08-22 13:15:35,879] timesketch.api_archive/INFO Unarchiving of sketch 12 complete.
[2025-08-22 13:15:35 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/
[2025-08-22 13:15:36 +0000] [9640] [DEBUG] GET /api/v1/users/me/
[2025-08-22 13:15:36 +0000] [9641] [DEBUG] POST /api/v1/sketches/12/aggregation/explore/
[2025-08-22 13:15:36 +0000] [9643] [DEBUG] POST /api/v1/sketches/12/aggregation/explore/
[2025-08-22 13:15:36 +0000] [9640] [DEBUG] GET /api/v1/searchtemplates/
[2025-08-22 13:15:36 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/searchhistorytree/
[2025-08-22 13:15:36 +0000] [9641] [DEBUG] GET /api/v1/sigmarules/
[2025-08-22 13:15:36 +0000] [9643] [DEBUG] GET /api/v1/intelligence/tagmetadata/
[2025-08-22 13:15:36 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/timelines/19/analysis/
[2025-08-22 13:15:36 +0000] [9640] [DEBUG] GET /api/v1/sketches/12/aggregation/
[2025-08-22 13:15:36 +0000] [9643] [DEBUG] GET /api/v1/sketches/12/questions/
[2025-08-22 13:15:36 +0000] [9641] [DEBUG] GET /api/v1/intelligence/tagmetadata/
[2025-08-22 13:15:36 +0000] [9642] [DEBUG] GET /api/v1/sketches/12/timelines/20/analysis/
```

closes #3090 

#vc4a


timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+        """Archives a sketch. This involves:
+        1. Setting the sketch status to 'archived'.
+        2. Setting the status of its 'ready' timelines to 'archived'.
+        3. For each SearchIndex associated with these timelines:
+           If all timelines using that SearchIndex are now 'archived',
+           attempt to close the index in OpenSearch.
+           If the OpenSearch index is successfully closed (or confirmed not found),
+           set the SearchIndex's status in the database to 'archived'.
+           Otherwise, the SearchIndex status is left unchanged (e.g., 'ready'),
+           allowing tools like tsctl to identify potential issues.
```
jkppr: How is this handling timelines that are still in the status processing or fail?
Especially for fail, there might be a searchindex created but not used. So this one also needs to be closed!

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
         if sketch.get_status.status == "archived":
             abort(
                 HTTP_STATUS_CODE_BAD_REQUEST,
-                "Unable to archive a sketch that was already archived "
-                "(sketch: {:d})".format(sketch.id),
+                f"Sketch {sketch.id} is already archived.",
             )
 
+        # Check if any timeline is in a non-archivable state
+        non_archivable_timeline_statuses = ["processing", "new", "fail", "timeout"]
```
jkppr: What makes those status "non-archivable"?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+        for timeline in sketch.timelines:
+            timeline_status = timeline.get_status.status
+            if timeline_status in non_archivable_timeline_statuses:
```
jkppr: This will prevent archiving a sketch with e.g. failed timelines. So what is the expected solution? Delete them first?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+                    f"Cannot archive sketch {sketch.id}. Timeline '{timeline.name}' "
+                    f"(ID: {timeline.id}) is in '{timeline_status}' state, which "
+                    f"prevents archival. Please ensure all timelines are 'ready' or "
+                    f"'archived'.",
+                )
+
         labels_to_prevent_deletion = current_app.config.get(
-            "LABELS_TO_PREVENT_DELETION", []
+            "LABELS_TO_PREVENT_DELETION",
+            [],  # Assuming same config key for archival prevention
```
jkppr: 1. This re-formatting because of the comment seems unnecessary.
2. Archival is a reversible and non destructive process that saves resources. Why should we prevent it?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
         for label in labels_to_prevent_deletion:
             if sketch.has_label(label):
                 abort(
                     HTTP_STATUS_CODE_FORBIDDEN,
-                    "A sketch with the label {:s} cannot be " "archived.".format(label),
+                    f"A sketch with the label {label} cannot be archived.",
                 )
 
         sketch.set_status(status="archived")
+        logger.info("Sketch %s status set to 'archived'.", sketch.id)
```
jkppr: I know this was not part of your PR, but when we already do such big changes: Shouldn't this be the last step after all sub-components are taken care of?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+
+        # Process timelines of the sketch
+        for timeline_in_sketch in sketch.timelines:
+            if (
+                timeline_in_sketch.get_status.status == "ready"
+            ):  # Only archive 'ready' timelines
```
jkppr: Why?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+                timeline_in_sketch.set_status(status="archived")
+                logger.info(
+                    "Timeline %s (part of sketch %s) status set to 'archived'.",
+                    timeline_in_sketch.id,
+                    sketch.id,
+                )
```
jkppr: Is "info" the correct log level here? Isn't this more of a debug info?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+            if not search_index.timelines:
+                logger.warning(
+                    "SearchIndex %s (ID: %s) has no associated timelines in DB. "
+                    "Skipping OpenSearch closure check.",
+                    search_index.index_name,
+                    search_index.id,
+                )
+                all_associated_timelines_archived = (
+                    False  # Cannot confirm, so don't archive OS index
+                )
```
jkppr:
```suggestion
                # Cannot confirm, so don't archive OS index
                all_associated_timelines_archived = False  
```

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
-            if not all(
-                x.get_status.status == "archived" for x in search_index.timelines
-            ):
-                continue
-            search_index.set_status(status="archived")
-            indexes_to_close.append(search_index.index_name)
+        for search_index in search_indexes_to_evaluate:
+            # Check if all timelines associated with this SearchIndex are now archived
+            all_associated_timelines_archived = True
+            if not search_index.timelines:
```
jkppr: 1. I don't understand your logic here, can you please elaborate?
`search_index.timelines` returns a list of all timelines associated with the index. How is this a check for "archived"?

2. Since you already set the sketch to archived above, this will now have a dangling searchindex?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+                    logger.info(
+                        "Successfully closed OpenSearch index: %s."
+                        " SearchIndex DB object "
+                        "(ID: %s) status updated to 'archived'.",
+                        search_index.index_name,
+                        search_index.id,
+                    )
+                except opensearchpy.exceptions.NotFoundError:
+                    # If index not found in OpenSearch, it's effectively gone/closed.
+                    search_index.set_status(status="archived")
```
jkppr: Is archived the correct status here? It implies that it can be unarchived, but if the index does not exist on the cluster this would not work. Maybe "deleted" or "fail" is better?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +478,144 @@ def _unarchive_sketch(self, sketch: Sketch):
...
+                    # For any other error during close, log it and
+                    # DO NOT change SearchIndex status.
+                    logger.error(
+                        "Failed to close OpenSearch index: %s (DB ID: %s). Error: %s. "
+                        "SearchIndex DB status remains '%s'.",
+                        search_index.index_name,
+                        search_index.id,
+                        str(e),
+                        original_si_status,
+                    )
```
jkppr: If you end up here, the sketch is still marked as archived but the searchindex won't. Should this be something that is surfaced to the user?

timesketch/tsctl.py
```diff
@@ -233,15 +233,106 @@ def drop_db():
...
+
+    By default, this command lists all sketches that have not been deleted.
+
+    - If the --archived flag is provided, it will only list sketches
+      that have an 'archived' status.
+
+    - If the --archived-with-open-indexes flag is provided, it will list
+      archived sketches that have one or more associated searchindexes
+      with a status of 'new', 'ready', 'processing', 'fail', or 'timeout'.
+      This option takes precedence over --archived.
```
jkppr: Why not using something like this instead of relying on precedence etc.?
```
    if archived and archived_with_open_indexes:
        # You can customize this message
        raise click.UsageError(
            "The options --archived and --archived-with-open-indexes are mutually exclusive. "
            "Please use only one."
        )
```

timesketch/tsctl.py
```diff
@@ -233,15 +233,106 @@ def drop_db():
...
+    """
+    all_sketches = Sketch.query.all()
+    # SearchIndex statuses that indicate it's not properly closed/archived
+    open_index_statuses = ["new", "ready", "processing", "fail", "archived", "timeout"]
+
+    if archived_with_open_indexes:
+        open_statuses_str = ", ".join(open_index_statuses)
+        print(
+            f"Searching for archived sketches with 'open' ({open_statuses_str}) searchindexes..."  # pylint: disable=line-too-long
```
jkppr: Maybe clarify if you mean the opensearch index status or the postgres status.

timesketch/tsctl.py
```diff
@@ -233,15 +233,106 @@ def drop_db():
...
+        )
+        found_sketches_info = []
+
+        for sketch in all_sketches:
+            if sketch.get_status.status != "archived":
+                continue
```
jkppr, line 407: Why not apply this filter on the psdql level by adjusting your query?

timesketch/tsctl.py
```diff
@@ -233,15 +233,106 @@ def drop_db():
...
+                for detail in sk_info["details"]:
+                    print(detail)
+        return
+
+    # Handle default listing or --archived only
+    sketches_to_display = []
+    for sketch in all_sketches:
+        current_status = sketch.get_status.status
+        if current_status == "deleted":
             continue
```
jkppr: How about a flag that allows me to also see sketches marked as deleted?

timesketch/api/v1/resources/archive.py
jaegeral: one of the listed status

jkppr: But why are those status "non-archivable"?

jaegeral: change to only processing.

timesketch/api/v1/resources/archive.py
jaegeral: remove

timesketch/api/v1/resources/archive.py
jaegeral: reduced it to only still processing timelines

timesketch/api/v1/resources/archive.py
jaegeral: good point, will move it to the end

timesketch/api/v1/resources/archive.py
jaegeral: done

timesketch/api/v1/resources/archive.py
jaegeral: changed it to only cover `                timeline_in_sketch.get_status.status not in ("archived", "deleted")
`

timesketch/api/v1/resources/archive.py
jaegeral: oh that is a very good point, changed

timesketch/api/v1/resources/archive.py
jaegeral: I changed it so the sketch would not be archived at this point, but I will think about other implications

timesketch/tsctl.py
jaegeral: done

timesketch/tsctl.py
jaegeral: tbh it looked more easy here and uses the same code flow as in the rest of the tool / project, and doing it in a different way looked more complicated from a code perspective and more LOC with what looks like a small gain in performance, like that query should be fairly quick.

timesketch/tsctl.py
jaegeral: good point, included

timesketch/tsctl.py
jaegeral: done, implemented in logic

timesketch/api/v1/resources/archive.py
jaegeral: the second point I think is addressed since I now archive later.

jaegeral: You're right, that comment could be more precise. The block of code is indeed about determining if the SearchIndex can be safely archived (which usually means closing its underlying OpenSearch index). The if not search_index.timelines: part is a specific check for an orphaned SearchIndex, which is one condition that would prevent it from being considered safe to archive.

timesketch/api/v1/resources/archive.py
jaegeral: so there is code for `        non_archivable_timeline_statuses = ["processing"]`

`            if timeline_in_sketch.get_status.status not in ("archived", "deleted"):`  So, a timeline that was "fail" will be marked as "archived" in the database as part of this step.

Their SearchIndex is then processed. If the underlying OpenSearch index doesn't exist, the SearchIndex (DB object) status becomes "fail". If it exists and is closed, it becomes "archived". If it exists but closure fails, its status remains as it was.

Does that make sense?

jaegeral: but I also updated the docstring to hopefully explain it better

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +480,173 @@ def _unarchive_sketch(self, sketch: Sketch):
...
                 )
 
-        sketch.set_status(status="archived")
+        # Process timelines of the sketch
+        for timeline_in_sketch in sketch.timelines:
+            # Archive timelines that are not already archived, deleted,
+            # or actively processing.
+            # The "processing" state is already checked and aborted earlier.
+            if timeline_in_sketch.get_status.status not in ("archived", "deleted"):
+                timeline_in_sketch.set_status(status="archived")
```
jaegeral: this should not be done for all timelines, failed ones should remain failed

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +480,173 @@ def _unarchive_sketch(self, sketch: Sketch):
...
-                continue
-            search_index.set_status(status="archived")
-            indexes_to_close.append(search_index.index_name)
+        errors_occurred = False
+        error_details = []
+
+        for search_index in search_indexes_to_evaluate:
+            # Determine if this SearchIndex can be safely archived (i.e., its
+            # OpenSearch index closed). This requires all associated timelines
+            # to be archived.
```
jkppr: This will have an unexpected side-effect: If a timeline is in status "fail" and then moved to "archived" it will be moved to "ready" when you un-archive it. So I recommend to keep the timeline as "fail" and accept both status as ok to be archived.

jkppr: Or as discussed: Don't change the timeline status at all since it does not matter much for the sketch archived state. (Needs to see how it impacts un-archive process)

timesketch/tsctl.py
jkppr: Both works for me.

jkppr (changes requested):

timesketch/api/v1/resources/archive.py
jaegeral: done

timesketch/api/v1/resources/archive.py
jaegeral: ok done

jaegeral: done

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +482,142 @@ def _unarchive_sketch(self, sketch: Sketch):
...
-        #    Each timeline has only a single search index, however
-        #    each search index can be part of multiple timelines.
-        #    Only archive a search index if all of it's timelines
-        #    are archived.
-        indexes_to_close = []
-        for timeline in sketch.timelines:
-            if timeline.get_status.status != "ready":
-                continue
-            timeline.set_status(status="archived")
-            search_index = timeline.searchindex
```
jkppr, line 500: In theory a searchindex should only have timelines from one Sketch. However, our data model would still allow for someone to use the API to have a searchindex shared between sketches. Therefore it might be worth to keep this check to be on the safe side. What do you think?

timesketch/api/v1/resources/archive.py
```diff
@@ -462,58 +482,142 @@ def _unarchive_sketch(self, sketch: Sketch):
         return HTTP_STATUS_CODE_OK
 
     def _archive_sketch(self, sketch: Sketch):
```
jkppr, line 445: Maybe we can mention that this is a revertable/non-destructive action in the doc string?

timesketch/api/v1/resources/archive.py
jkppr, line 1: Attempting to unarchive a sketch form the UI results in unavailable search indices in my testing. The server logs throw this error:
```
[2025-08-20 12:25:25 +0000] [132] [ERROR] Error handling request /api/v1/sketches/1/aggregation/explore/
  File "/usr/local/lib/python3.10/dist-packages/gunicorn/workers/sync.py", line 134, in handle
    self.handle_request(listener, req, client, addr)
  File "/usr/local/lib/python3.10/dist-packages/gunicorn/workers/sync.py", line 177, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 1498, in __call__
    return self.wsgi_app(environ, start_response)
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 1476, in wsgi_app
    response = self.handle_exception(e)
  File "/usr/local/lib/python3.10/dist-packages/flask_restful/__init__.py", line 298, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 1473, in wsgi_app
    response = self.full_dispatch_request()
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 882, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/usr/local/lib/python3.10/dist-packages/flask_restful/__init__.py", line 298, in error_router
    return original_handler(e)
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 880, in full_dispatch_request
    rv = self.dispatch_request()
  File "/usr/local/lib/python3.10/dist-packages/flask/app.py", line 865, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
  File "/usr/local/lib/python3.10/dist-packages/flask_restful/__init__.py", line 489, in wrapper
    resp = resource(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/flask/views.py", line 110, in view
    return current_app.ensure_sync(self.dispatch_request)(**kwargs)  # type: ignore[no-any-return]
  File "/usr/local/lib/python3.10/dist-packages/flask_restful/__init__.py", line 604, in dispatch_request
    resp = meth(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/flask_login/utils.py", line 290, in decorated_view
    return current_app.ensure_sync(func)(*args, **kwargs)
  File "/usr/local/src/timesketch/timesketch/api/v1/resources/aggregation.py", line 518, in post
    result_obj = aggregator.run(**aggregator_parameters)
  File "/usr/local/src/timesketch/timesketch/lib/aggregators/bucket.py", line 135, in run
    response = self.opensearch_aggregation(aggregation_spec)
  File "/usr/local/src/timesketch/timesketch/lib/aggregators/interface.py", line 378, in opensearch_aggregation
    aggregation = self.opensearch.client.search(
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/client/utils.py", line 180, in _wrapped
    return func(*args, params=params, headers=headers, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/client/__init__.py", line 2359, in search
    return self.transport.perform_request(
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/transport.py", line 455, in perform_request
    raise e
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/transport.py", line 416, in perform_request
    status, headers_response, data = connection.perform_request(
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/connection/http_urllib3.py", line 308, in perform_request
    self._raise_error(
  File "/usr/local/lib/python3.10/dist-packages/opensearchpy/connection/base.py", line 315, in _raise_error
    raise HTTP_EXCEPTIONS.get(status_code, TransportError)(
opensearchpy.exceptions.RequestError: RequestError(400, 'index_closed_exception', 'closed')
```

timesketch/tsctl.py
```diff
@@ -341,15 +341,132 @@ def drop_db():
...
+    is_flag=True,
+    help="Show only archived sketches. Mutually exclusive with --archived-with-open-indexes.",  # pylint: disable=line-too-long
+)
+@click.option(
+    "--archived-with-open-indexes",
+    is_flag=True,
+    help="Show archived sketches that have at least one searchindex with status "
+    "'new', 'ready', 'processing', 'fail','archived' or 'timeout'. "
+    "Mutually exclusive with --archived.",
+)
```
jkppr, line 358: Would it make sense to also have a `--open-with-closed-indexes` option?

jkppr (changes requested):
All good with the tsctl changes. But for the archive API there seems to be an issue when unarchiving sketches from the UI.
When I tested it I ended up with an active sketch but archived searchindices.

Please take another look below.

timesketch/api/v1/resources/archive.py
jaegeral: oh there seems to be a bug, let me fix it, the searchindex was not ipened when doing unarchive.

jaegeral: should be fixed now, I can archive and unarchive sketches in the UI without a problem

timesketch/api/v1/resources/archive.py
jaegeral: yep

timesketch/tsctl.py
jaegeral: made a new issue: https://github.com/google/timesketch/issues/3506

timesketch/api/v1/resources/archive.py
```diff
@@ -425,95 +443,296 @@
...
+            timeline_status = timeline.get_status.status
+            if timeline_status in non_archivable_timeline_statuses:
                 abort(
-                    HTTP_STATUS_CODE_FORBIDDEN,
-                    "A sketch with the label {:s} cannot be " "archived.".format(label),
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    f"Cannot archive sketch {sketch.id}. Timeline '{timeline.name}' "
+                    f"(ID: {timeline.id}) is in '{timeline_status}' state, which "
+                    f"prevents archival.",
                 )
```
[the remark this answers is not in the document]
jaegeral: oh good point, will bring it back

timesketch/api/v1/resources/archive.py
```diff
@@ -425,95 +443,296 @@
...
         if sketch.get_status.status != "archived":
             abort(
                 HTTP_STATUS_CODE_BAD_REQUEST,
                 "Unable to unarchive a sketch that wasn't already archived "
-                "(sketch: {:d})".format(sketch.id),
+                f"(sketch: {sketch.id})",
+            )
+
+        # Check if any timeline is in a non-unarchivable state
+        non_unarchivable_timeline_statuses = ["processing"]
```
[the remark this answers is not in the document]
jaegeral: good point

timesketch/api/v1/resources/archive.py
```diff
@@ -481,39 +628,139 @@ def _archive_sketch(self, sketch: Sketch):
...
+        # Check if any timeline is in a non-archivable state
+        non_archivable_timeline_statuses = ["processing", "fail", "timeout"]
         for timeline in sketch.timelines:
-            if timeline.get_status.status != "ready":
-                continue
-            timeline.set_status(status="archived")
-            search_index = timeline.searchindex
+            timeline_status = timeline.get_status.status
+            if timeline_status in non_archivable_timeline_statuses:
+                abort(
```
jkppr: When I hit this error I have the problem that my existing working timelines are in the "archived" state but the rest of my sketch is not, rendering it un-usable.

How to reproduce:
1. Upload a working new timeline
2. Upload a broken timeline
3. Click archive in the UI
4. see red error bar
5. reload sketch, see one timeline in permanent loading state.

timesketch/api/v1/resources/archive.py
```diff
@@ -481,39 +628,139 @@ def _archive_sketch(self, sketch: Sketch):
...
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    f"Cannot archive sketch {sketch.id}. Timeline '{timeline.name}' "
+                    f"(ID: {timeline.id}) is in '{timeline_status}' state, which "
+                    f"prevents archival.",
```
jkppr: Can we make this error message more actionable for the user? E.g. add something about please delete timelines that have a failed status.

timesketch/api/v1/resources/archive.py
jaegeral: updated the error handling

timesketch/api/v1/resources/archive.py
jaegeral: done

timesketch/api/v1/resources/aggregation.py
```diff
@@ -528,6 +532,29 @@ def post(self, sketch_id: int):
...
+                        HTTP_STATUS_CODE_BAD_REQUEST,
+                        "Unable to run aggregation on a closed index."
+                        f"index: {indices_msg:s} and parameters:"
+                        f" {aggregator_parameters!s}",
+                    )
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    f"Unable to run the aggregation, with error: {exc!s}"
+                    f"index: {indices_msg:s} and parameters: {aggregator_parameters!s}",
+                )
```
jkppr, line 574: Since you are already adding logging here, why not also log when `if exc.error == "index_closed_exception"` is not true? It would tell us why aggregations fail in other situations.

timesketch/api/v1/resources/aggregation.py
```diff
@@ -562,10 +589,21 @@ def post(self, sketch_id: int):
...
+            except RequestError as e:
+                if e.error == "index_closed_exception":
+                    abort(
+                        HTTP_STATUS_CODE_BAD_REQUEST,
+                        "Unable to run aggregation on a closed index.",
+                    )
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    f"Unable to run the aggregation, with error: {e!s}",
+                )
```
jkppr, line 643: For the abort errors in the `AggregationExploreResource` you have added a error log. Would it make sense to add them here as well?

timesketch/api/v1/resources/archive.py
```diff
@@ -425,53 +444,181 @@ def _export_view(self, view, sketch, zip_file):
...
         if sketch.get_status.status != "archived":
             abort(
                 HTTP_STATUS_CODE_BAD_REQUEST,
                 "Unable to unarchive a sketch that wasn't already archived "
-                "(sketch: {:d})".format(sketch.id),
+                f"(sketch: {sketch.id})",
+            )
+
+        # Check if any timeline is in a non-unarchivable state
+        non_unarchivable_timeline_statuses = ["processing"]
```
jkppr: Two questions:
1. Since you set all timelines into the `archived` status now and also prevent the archiving if the timeline is in a processing status, what is the reason for this check?
2. The name is a bit confusing and sounds like a double negative. `statuses_preventing_unarchival` be more clear?

timesketch/api/v1/resources/archive.py
```diff
@@ -425,53 +444,181 @@ def _export_view(self, view, sketch, zip_file):
...
+            timeline_status = timeline.get_status.status
+            if timeline_status in non_unarchivable_timeline_statuses:
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    f"Cannot unarchive sketch {sketch.id}. Timeline "
+                    f"'{timeline.name}' "
+                    f"(ID: {timeline.id}) is in '{timeline_status}' state, which "
+                    f"prevents unarchival.",
+                )
+
```
jkppr: You're logging the other abort events, why not this one?

timesketch/api/v1/resources/archive.py
```diff
@@ -425,53 +444,181 @@ def _export_view(self, view, sketch, zip_file):
...
-        """Unarchives a sketch by opening up all indices and removing labels.
+        """Archives a sketch. This involves:
+
+        1. Setting the sketch status to 'archived'.
+        2. For each SearchIndex associated with the sketch:
+          the system attempts to close the corresponding OpenSearch index.
+           - If the OpenSearch index is successfully closed, the SearchIndex's
+             database status is set to 'archived'.
+           - If the OpenSearch index is not found, it's considered effectively
+             closed, and the SearchIndex's database status is set to 'fail'.
```
jkppr: In context of your action in line 584 above, wouldn't this lead to those searchindex being set as ready in the db again when a sketch gets unarchived?

timesketch/api/v1/resources/archive.py
```diff
@@ -481,39 +628,153 @@ def _archive_sketch(self, sketch: Sketch):
         for label in labels_to_prevent_deletion:
             if sketch.has_label(label):
                 abort(
-                    HTTP_STATUS_CODE_FORBIDDEN,
-                    "A sketch with the label {:s} cannot be " "archived.".format(label),
+                    HTTP_STATUS_CODE_BAD_REQUEST,
```
jkppr: Is this really a BAD REQUEST if the sketch has a label that prevents archival? Wouldn't the old FORBIDDEN be a better match?

timesketch/api/v1/resources/archive.py
```diff
@@ -481,39 +628,153 @@ def _archive_sketch(self, sketch: Sketch):
...
+                    f"'{timeline_to_check.name}' (ID: {timeline_to_check.id}) is in "
+                    f"a non-archivable state: '{timeline_status}'."
+                )
+                suggestion = ""
+                if timeline_status in ["fail", "timeout"]:
+                    suggestion = " Please delete this timeline and try again."
+                elif timeline_status == "processing":
+                    suggestion = (
+                        " Please wait for the timeline to finish processing or "
+                        "delete it."
```
jkppr: iirc, you can't delete a processing timeline. So only way is to wait for it to finish. Which makes sense, since (as long as its not a bug) the search index is importing stuff during processing status.

timesketch/tsctl.py
```diff
@@ -341,15 +341,132 @@ def drop_db():
...
+    if archived:
+        output_type = "Archived sketches"
+    elif include_deleted:
+        output_type = "Sketches (including deleted, ready, and archived)"
+    else:  # Default
+        output_type = "Sketches (excluding deleted; i.e., ready and archived)"
+
+    print(f"{output_type}:")
+    for sketch in sketches_to_display:
+        print(f"{sketch.id} {sketch.name} (status: {sketch.get_status.status})")
```
jkppr:
```suggestion
        print(f"{sketch.id} '{sketch.name}' (status: {sketch.get_status.status})")
```

Sketch names can be odd sometimes especially when they have a space in them. Let's add something that makes it clear what is the full name and what is the rest.

Example why this helps:
```
# tsctl list-sketches --archived
Archived sketches:
3 try archive 2 (status: archived)

# tsctl list-sketches --archived
Archived sketches:
3 [try archive 2] (status: archived)

# tsctl list-sketches --archived
Archived sketches:
3 "try archive 2" (status: archived)
```

timesketch/tsctl.py
```diff
@@ -341,15 +341,132 @@ def drop_db():
...
+    """
+    all_sketches = Sketch.query.all()
+
+    if archived and archived_with_open_indexes:
+        raise click.UsageError(
+            "The options --archived and --archived-with-open-indexes "
+            "are mutually exclusive. Please use only one."
+        )
+
+    # SearchIndex statuses that indicate it's not properly closed/archived
```
jkppr, line 393: Am I missing anything or do you only check against inconsistency within the database and the real searchindex status in opensearch does not get checked?

jkppr (changes requested):
We are getting to the nitty parts :)

timesketch/api/v1/resources/archive.py
jaegeral: there might be instances where it is indef in preocessing state, that is why the recomendation to delete.

timesketch/api/v1/resources/aggregation.py
jaegeral: good point, added

timesketch/api/v1/resources/aggregation.py
jaegeral: yes

timesketch/api/v1/resources/archive.py
jaegeral: so this is just a safeguard that should hopefully never occur, but if it has timelines  in that status, it should fail, we could consider to reverse it and say if there is a timeline that has any status other than "Archived" --> fail. WDYT?

timesketch/api/v1/resources/archive.py
jaegeral: Good point, added

timesketch/api/v1/resources/archive.py
jaegeral: yeah, fine for me

timesketch/tsctl.py
jaegeral: Update of `tsctl`

tsctl list-sketches --archived-with-open-indexes
Searching for archived sketches with 'open' SearchIndex DB statuses (new, ready, processing, fail, timeout) OR indices that are actually open in OpenSearch...
No archived sketches with inconsistent index statuses found.

timesketch/api/v1/resources/archive.py
jaegeral: hm noe, this is that check: 
```
except opensearchpy.exceptions.NotFoundError:
                    # If index not found in OpenSearch, it's effectively gone/closed,
                    # but the intended 'close' operation failed because it wasn't there.
                    search_index.set_status(status="fail")
                    logger.warning(
                        "OpenSearch index: %s not found. "
                        " SearchIndex DB object (ID: %s) "
                        "status updated to 'fail'.",
                        search_index.index_name,
                        search_index.id,
                    )
                    errors_occurred = True
                    error_details.append(
                        f"OpenSearch index '{search_index.index_name}' not found. "
                        f"SearchIndex DB object (ID: {search_index.id}) "
                        "status updated to 'fail'."
                    )
```

timesketch/api/v1/resources/archive.py
```diff
@@ -425,53 +444,180 @@ def _export_view(self, view, sketch, zip_file):
...
         if sketch.get_status.status != "archived":
             abort(
                 HTTP_STATUS_CODE_BAD_REQUEST,
                 "Unable to unarchive a sketch that wasn't already archived "
-                "(sketch: {:d})".format(sketch.id),
+                f"(sketch: {sketch.id})",
             )
 
+        # Check if any timeline is in a state that prevents unarchiving.
+        statuses_preventing_unarchival = ["processing", "fail"]
```
jaegeral, line 676: Add only a warning in logs and create a Tsctl command to check for archived sketches that have failed or processing timelines in them.

docs/guides/user/sketch-overview.md
```diff
@@ -75,7 +75,26 @@ Delete the whole sketch. *Note:* this will not delete the Timelines.
 
 ### Archive
 
-Archive the whole sketch.
+Archiving a sketch is a way to preserve it in a read-only state while freeing up system resources. When a sketch is archived:
```
jkppr: But the sketch is not read-only either. It is archived. to read it you need to unarchive it first which makes it writeable as well.

timesketch/api/v1/resources/archive.py
jkppr: Discussed via chat.
Agreed-compromise:
* This check will add a breaking change for existing sketches, since their status was not updated during archive process in the past.
* Solution:
  1. Don't abort when this check fails.
  2. Add a warning only to request the admin to run some health checks in tsctl.
  3. Publish a tsctl clean-up script.
  
Follow-up comment: https://github.com/google/timesketch/pull/3431/files#r2332785506

Tracking issue: https://github.com/google/timesketch/issues/3516

timesketch/api/v1/resources/archive.py
jaegeral: done https://github.com/google/timesketch/pull/3517

timesketch/api/v1/resources/archive.py
jkppr: Yes, I was thinking about that situation as well. But then the user seeing the message can't delete it from the UI, since we prevent deletion of processing timelines. In that case, the user needs to reach out to an admin who needs to validate if the timeline is indeed in a processing loop caught and set it manually to fail before deletion.
Can we accommodate for that in our wording? e.g. here where it is user facing to have something that says: wait or contact admin

timesketch/api/v1/resources/archive.py
```diff
@@ -425,53 +444,177 @@ def _export_view(self, view, sketch, zip_file):
...
         sketch.set_status(status="ready")
+        logger.info("Sketch %s status set to 'ready'.", sketch.id)
 
-        indexes_to_open = []
+        # Set all timelines in the sketch to ready.
         for timeline in sketch.timelines:
             if timeline.get_status.status != "archived":
                 continue
             timeline.set_status(status="ready")
-            search_index = timeline.searchindex
```
jkppr: I think we might run into a logic error here. Consider this scenario:

1. During Archiving (`_archive_sketch`):

  * The process attempts to close the OpenSearch index for a timeline.
  * If the index is missing in OpenSearch, it raises a `NotFoundError`
  * The `except` block in `_archive_sketch` catches this, sets the SearchIndex status to `fail` in the database, and sets `errors_occurred` to `True`.
  * The code proceeds to set the Sketch status to archived and then commits all database changes.

2. During Unarchiving (`_unarchive_sketch`):

  * It collects SearchIndex objects to reopen. Since the problematic SearchIndex has a status of `fail` (not archived), it is correctly skipped and not added to the `search_indexes_to_open` set.
  * Assuming all other indices are found and opened successfully, `errors_occurred` remains `False`, and the function does not abort.
  * The logic then proceeds to open all timelines again, including the ones potentially associated with the failed search index, resulting in timelines that can't be used but look ready and will throw error 500 when searched.

Either we need to set timelines in status fail when we find their searchindex missing during the archival process (but timeline in fail will block archival according to your logic). Or we set them to fail when we unarchvie and notice the searchindex is gone or in a fail mode?

jkppr (changes requested):
Please have a look below.

docs/guides/user/sketch-overview.md
jaegeral: well technically the data is still available in OS, but I know what you mean, will adjust the wording

timesketch/api/v1/resources/archive.py
jaegeral: hm I think I can add the message we write to the logs to be adde to the user facing issue here

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,348 @@ def _export_view(self, view, sketch, zip_file):
...
+                error_msg = f"{base_error_msg}{suggestion}"
+                errors_occurred = True
+                abort(
+                    HTTP_STATUS_CODE_BAD_REQUEST,
+                    error_msg,
+                )
 
-        # TODO (kiddi): Move this to lib/datastores/opensearch.py.
-        if indexes_to_open:
+        for search_index in search_indexes_to_close:
```
jaegeral, line 575: Create another check if error_occured

timesketch/api/v1/resources/archive.py
jaegeral: I think it should be logical now O:-)

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                if timeline_status in ["fail", "timeout"]:
+                    suggestion = " Please delete this timeline and try again."
+                elif timeline_status == "processing":
+                    suggestion = (
+                        " Please wait for it to finish processing. If it seems to be "
+                        "stuck, contact your system administrator to resolve the "
+                        "issue."
+                    )
+                error_msg = f"{base_error_msg}{suggestion}"
+                logger.error(error_msg)
```
jkppr, line 508: Can we prevent that `timeline_to_check.name` gets posted to the backend log?
But we should include it for the `abort` message so the user knows what to do.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+
+        Returns:
+            An integer representing the HTTP status of the operation.
+            - HTTP_STATUS_CODE_OK if successful.
+
+        Raises:
+            HTTPException:
+                - If the sketch is not currently archived.
+                - If any timeline within the sketch is in a state that prevents
+                  unarchiving (e.g., 'processing').
```
jkppr: Can we update this to reflect the backwards compatibility we have discussed?

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                - If a critical error occurs while opening an OpenSearch index.
         """
-        if sketch.get_status.status == "archived":
+        logger.info("Unarchiving sketch %s ('%s').", sketch.id, sketch.name)
```
jkppr: Debatable: Should info logging contain potentially sensitive case data like `sketch.name`? We do try to avoid it for error and warnings. But info is something turned on intentionally by the admins?

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                if timeline_status == "fail":
+                    specific_advice = (
+                        "It is recommended to fix this timeline (e.g., by deleting it) "
+                        "because the sketch cannot be archived again in this state."
+                    )
+                elif timeline_status == "processing":
+                    specific_advice = (
+                        "Please wait for it to finish processing. If it seems to be "
+                        "stuck, contact your system administrator to resolve the issue."
+                    )
```
jkppr, line 695: If a timeline is still in processing state when being unarchived, it is most likely broken. So I would recommend to go with "Contact your admin to fix the status" only in this case. Waiting won't help here.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                    # only a processing timeline is stopping the flow
+                    errors_occurred = True
+
+                general_advice = " You can use 'tsctl find-inconsistent-archives' to find such sketches."  # pylint: disable=line-too-long
+                warning_msg = f"{base_warning_msg}{specific_advice}{general_advice}"
+                error_details.append(warning_msg)
+                logger.warning(warning_msg)
```
jkppr: This warning contains potential sensitive case data `timeline.name`.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+
+        if errors_occurred:
+            logger.error(
+                "Unarchiving sketch %s failed because one or more indices could not "
+                "be opened. Errors: %s",
+                sketch.id,
+                "; ".join(error_details),
```
jkppr, line 707: This `error_details` contains potential sensitive case details `timeline.name`.
Also: You already send each of the warning to the log above in line 699. So do you really need this again as error?

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                        "It is recommended to fix this timeline (e.g., by deleting it) "
+                        "because the sketch cannot be archived again in this state."
+                    )
+                elif timeline_status == "processing":
+                    specific_advice = (
+                        "Please wait for it to finish processing. If it seems to be "
+                        "stuck, contact your system administrator to resolve the issue."
+                    )
+                    # only a processing timeline is stopping the flow
+                    errors_occurred = True
```
jkppr: Why is a processing timeline stopping the unarchive flow? On an old instance there is a chance to run into this situation with an infinite processing timeline being archived. I recommend to handle this similar to what you do with `fail` above.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+                "Unarchiving sketch %s failed because one or more indices could not "
+                "be opened. Errors: %s",
+                sketch.id,
+                "; ".join(error_details),
+            )
+            abort(
+                HTTP_STATUS_CODE_INTERNAL_SERVER_ERROR,
+                "Failed to unarchive sketch. One or more OpenSearch indices could not "
+                f"be opened. Details: {'; '.join(error_details)}",
+            )
```
jkppr: For now in the context of #3518 I would argue that the status of a timeline should not be a blocking step in the unarchive process.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+
+        # 4. If all indices are open, update the database statuses.
+        sketch.set_status(status="ready")
```
jkppr: Shouldn't we do this after the timelines are handled? Otherwise we might run into some unexpected race condition where the sketch is "ready" but the timelines are not ready yet, so a reload of the sketch fails.

timesketch/api/v1/resources/archive.py
```diff
@@ -424,96 +442,364 @@ def _export_view(self, view, sketch, zip_file):
...
+
+        # Set all timelines in the sketch to ready.
+        for timeline in sketch.timelines:
+            if timeline.get_status.status != "archived":
+                continue
+            timeline.set_status(status="ready")
+            logger.info(
+                "Timeline %d ('%s') status set to 'ready'.",
+                timeline.id,
+                timeline.name,
```
jkppr: We should try to avoid logging `timeline.name` into the backend log.

jkppr (changes requested):
Some additional nits.

timesketch/api/v1/resources/archive.py
jaegeral: sure thing

timesketch/api/v1/resources/archive.py
jaegeral: no that should only be the id

timesketch/api/v1/resources/archive.py
jaegeral: ack

timesketch/api/v1/resources/archive.py
jaegeral: removed the timeline name above, I remove the one above, this one is enough

timesketch/api/v1/resources/archive.py
jaegeral: yes

timesketch/api/v1/resources/archive.py
jaegeral: ack

timesketch/api/v1/resources/archive.py
jaegeral: ack

timesketch/api/v1/resources/archive.py
jaegeral: easy enough, yes

timesketch/api/v1/resources/archive.py
jaegeral: ack

timesketch/api/v1/resources/archive.py
jaegeral: removed that comment

PR #3440, opened 2025-06-04
Various small changes

jaegeral:
### Highlights
* **Improved Logging**: Updates logging messages in two different areas (`opensearch.py` and `annotations.py`) to provide more context, such as including the sketch ID in OpenSearch errors and adding object type, ID, and sketch ID (where applicable) to warnings about objects with multiple statuses.
* **Model Status Handling**: Refines the `get_status` method in the annotations model to use `getattr` for safer access to the object ID and enhances the warning message when multiple statuses are found for an object.
* **TSCTL Command Cleanup**: Removes an unnecessary digit check for the search index ID in the `searchindex_info` command and corrects the example verification command printed after setting a search index status in `searchindex_status`.

### Changelog
* **timesketch/lib/datastores/opensearch.py**
  * Updated the logging format and added the sketch ID to the error message when an OpenSearch index is not found during a count operation.
* **timesketch/models/annotations.py**
  * Modified the `get_status` method to use `getattr` for accessing the object ID, defaulting to 'N/A' if not present.
  * Enhanced the warning message when an object has more than one status, including the object type name and detailed information (ID and Sketch ID if available).
  * Added comments explaining the logic for including Sketch ID in the log details.
* **timesketch/tsctl.py**
  * Removed the explicit check to ensure the `searchindex_id` is a digit in the `searchindex_info` function, as the type hint already indicates it should be an integer.
  * Corrected the example command printed after successfully setting a search index status to explicitly use the `--searchindex_id` flag.


timesketch/lib/datastores/opensearch.py
```diff
@@ -771,9 +772,9 @@ def search(
                 count_result = self.client.count(body=query_dsl, index=list(indices))
             except NotFoundError:
                 os_logger.error(
-                    "Unable to count due to an index not found: {:s}".format(
-                        ",".join(indices)
-                    )
+                    "Unable to count for sketch ID %s due to an index not found: %s",
+                    sketch_id,
+                    ",".join(indices),
```
jkppr:
```suggestion
                    "Unable to count for sketch ID [%s] due to an index not found: [%s]",
                    sketch_id,
                    ",".join(indices),
```

timesketch/lib/datastores/opensearch.py
```diff
@@ -825,8 +826,9 @@ def search(
                 cause = str(e)
 
             os_logger.error(
-                "Unable to run search query. Error: %s. "
-                "Sketch ID: %s. Indices: %s. ",
+                "Unable to run search query for user %s. Error: %s. "
+                "Sketch ID: %s. Indices: %s.",
+                current_user.username,
```
jkppr:
```suggestion
                "Unable to run search query for user [%s]. Error: %s. "
                "Sketch ID: [%s]. Indices: [%s].",
                current_user.username,
```

jkppr (approved):
lgtm, please accept the two small suggestions and fix linters.

PR #3445, opened 2025-06-06
Enable support for vue3 UI

jkppr:
This change will start adding the vue3 UI into the release builds. The default is still `frontend-ng` but this change will allow to switch to `frontend-v3` if configured.


berggren (approved):
LGTM

PR #3451, opened 2025-06-13
Allow batch editing of tags.

Annoraaq:
I pulled the dialog into its own component so it can be opened from different places (using different triggers).

Then I changed it such that it can handle an array of events.

### DIFF
--- a/timesketch/frontend-ng/src/components/Explore/EventTagDialog.vue
+++ b/timesketch/frontend-ng/src/components/Explore/EventTagDialog.vue
@@ -1,0 +1,190 @@
+<!--
+Copyright 2022 Google Inc. All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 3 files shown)


timesketch/frontend-ng/src/components/Explore/EventTagDialog.vue
```diff
@@ -0,0 +1,190 @@
+<!--
+Copyright 2022 Google Inc. All rights reserved.
```
jkppr:
```suggestion
Copyright 2025 Google Inc. All rights reserved.
```

jkppr (changes requested):
Thanks for adding this feature! This is a huge win.

One small nit I have already sent you via chat:
* Selecting multiple events with different tags and removing one tag that is not on all events results in a different tag to disappear until reloading.

Please check if you can fix that

Annoraaq:
> Thanks for adding this feature! This is a huge win.
> 
> One small nit I have already sent you via chat:
> 
> * Selecting multiple events with different tags and removing one tag that is not on all events results in a different tag to disappear until reloading.
> 
> Please check if you can fix that

Should be fixed. Thanks for testing thoroughly.

jkppr (approved):
Thanks for this amazing feature!

PR #3454, opened 2025-06-18
Introduce unified SearchGuideCard to help users start exploring

### LINKED ISSUE #2900: UI feedback: Improve initial search UX
When working on a new sketch the initial explore page has a lot of white space and no information on how to get started. The analyst has just their list of timelines and an empty search bar (see screenshot below).

image

While observing how analysts work with timesketch and from feedback, we noticed that many analysts start by running the `*` search to get all events for all timelines. This operation gets more and more expensive with larger timelines until it can even break (e.g. searching `*` on 50M events, does not work well and will most likely generate an error 500.)

The main task with this issue is to explore and track possible options to improve the start of an investigation.

Possible ideas that have been provided during feedback sessions:
* Show the omnibar in the empty space to allow for a quick start to search on specific data_types, tags or saved searches (screenshot below).
* Have some guidance displayed on how to start an efficient search
* Nudge the analyst to use the Investigate (DFIQ) feature to define what they ware looking to answer and then follow the approaches.

image

jkppr:
This PR revamps the user experience on the Explore page by introducing a new search guide, a welcome screen, and an improved "no results" page, all aimed at making the search functionality more intuitive and user-friendly.

The core of this change is the creation of new, reusable components to guide users in building search queries.

#### Key Changes:

*   **New Welcome Card (`ExploreWelcomeCard.vue`):**
    *   When a user first lands on the Explore page without a query, a new "Start Exploring" card is displayed.
    *   This card provides a list of common search query examples.
    *   It dynamically lists available tags, data types, and saved searches, giving users immediate, actionable starting points for their investigation.

*   **Improved "No Results" Page (`SearchNotFoundCard.vue`):**
    *   The "No search results" message has been replaced with a much more helpful and interactive card.
    *   It displays the user's query that returned no results and offers suggestions for improvement.
    *   Crucially, it includes the same search examples and dynamic lists (tags, data types, saved searches) as the welcome card, helping users quickly refine their failed query.

*   **On-Demand Search Help (`SearchHelpCard.vue`):**
    *   A new help icon (`mdi-help-circle-outline`) has been added to the main search bar in the Explore view.
    *   Clicking this icon opens a dialog containing search syntax examples, providing on-demand assistance without navigating away from the page.

*   **Component Refactoring for Reusability:**
    *   A central, reusable `SearchGuideCard.vue` component was created. It serves as the foundation for the Welcome, No Results, and Search Help cards, ensuring a consistent look and feel.
    *   The logic for displaying `DataTypes` and `SavedSearches` has been extracted from the left panel components into their own dedicated list components (`DataTypesList.vue`, `SavedSearchesList.vue`). This allows them to be reused in both the left panel and the new guide cards, reducing code duplication.

These changes collectively lower the barrier to entry for new users, make the powerful search syntax more discoverable, and provide a more helpful and consistent UI throughout the exploration workflow.

### Screenshots

**1. The New Search Help Dialog in Action:**
image

**2. The "No Results Found" State:**
image

**3. The new "Empty State":**
image

**Closing issues**

closes #2900


timesketch/frontend-ng/src/components/Explore/EmptyStateCard.vue
```diff
@@ -0,0 +1,292 @@
...
+    },
+  },
+  methods: {
+    emitSetQueryAndFilter(queryString) {
+      let eventData = {}
+      eventData.doSearch = true
+      eventData.queryString = queryString
+      EventBus.$emit('setQueryAndFilter', eventData)
+    },
+  },
```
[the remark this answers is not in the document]
jkppr: Those are commas.

timesketch/frontend-ng/src/components/LeftPanel/DataTypesList.vue
```diff
@@ -0,0 +1,110 @@
...
+  height: 28px;
+}
+
+::v-deep .v-data-footer__icons-before .v-icon,
+::v-deep .v-data-footer__icons-after .v-icon {
+  font-size: 1rem;
+}
+
+::v-deep .v-data-footer__pagination {
+  margin-left: 15px !important;
```
Annoraaq, line 107: nit: I would try to avoid !important in almost every case as it makes it really hard to override the impacted styles properly. It should be the last resort. If possible, I recommend to use more specific CSS rules instead.

timesketch/frontend-ng/src/components/Explore/EmptyStateCard.vue
```diff
@@ -0,0 +1,322 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+<template>
+  <v-card :outlined="!inDialog" :class="{ 'ma-4': !inDialog }">
+    <!-- Header Section -->
+    <div v-if="displayMode === 'InfoCard'" class="pa-4 pb-0">
```
Annoraaq: Optional opinionated design suggestion:

I see a couple of drawbacks of the "mode" approach.
The main problem is that the information about where the component will be used (context) is leaked into the component.
Concretely it needs to "know" whether it is used in a dialog or not and if it is an InfoCard, an EmptyState or a NotFound card. So whenever you add a new type this component has to change and needs potentially more case distinctions inside. Also all the methods and computed properties of all modes need to be in this component. This can be a source of errors and a burden to maintain.

An alternative would be to turn the dependency around:

- Instead of a case distinction about the mode inside EmptyStateCard you can create 3 more Components InfoCard, EmptyStateCard and NotFoundCard. These then use EmptyStateCard and provide it the header in a [slot](https://vuejs.org/guide/components/slots).
- nit: Instead of `inDialog` call the property something more generic (embedded/standalone or flat/outlined showBorder, ...). This then indicates what is the actual difference and not where it is used. It could be useful outside of a dialog in the future. So maybe everywhere where we don't need a border?

The main difference with this design is that the EmptyStateCard stays more generic and easier to maintain as it does not need to contain information about the context it is used in.

Annoraaq (approved):
Thanks for refactoring and cleaning up.

timesketch/frontend-ng/src/components/Explore/EmptyStateCard.vue
jkppr: I have changed the code to follow your advice, thanks.

PR #3468, opened 2025-07-12
Call validate_on_submit() on the login form.

### LINKED ISSUE #3467: Login form validate_on_submit not being called on form submissions
**Describe the bug**
At `timesketch/views/auth.py#L157` the form validation method "validate_on_submit" is not actually being called but is instead being checked as if it was an attribute of the class:
https://github.com/google/timesketch/blob/83fa0a74954de76af7c52bf131e4a691083e0114/timesketch/views/auth.py#L157

As far as I can tell this doesn't actually affect the username/password validation and so doesn't lead to a security issue. However, omitting the password parameter from the request entirely does lead to a 500 error which would be prevented by the form validation.

**To Reproduce**
This bug is visible in the code reference linked above.

**Expected behavior**
The validate_on_submit method should be called when the login form is submitted.

**Screenshots**
N/A

N/A

**Additional context**
N/A

jonathan-greig:
Fix for #3467 ensuring that validate_on_submit is properly called on the login form.

**Checks**
- [x] All tests succeed.
- [x] ~~Unit tests added.~~ (N/A)
- [x] ~~e2e tests added.~~ (N/A)
- [x] ~~Documentation updated.~~ (N/A)

**Closing issues**
closes #3467

### DIFF
--- a/timesketch/views/auth.py
+++ b/timesketch/views/auth.py
@@ -152,9 +152,8 @@ def login():
             db_session.commit()
 
     # Login form POST
-    # pylint: disable=using-constant-test
     form = UsernamePasswordForm()
-    if form.validate_on_submit:
+    if form.validate_on_submit():
         user = User.query.filter_by(username=form.username.data).first()
         if user:
             if user.check_password(plaintext=form.password.data):


jaegeral:
Uh excellent catch, thx!

jaegeral (approved):

PR #3470, opened 2025-07-14
[API] Enhancing the `sketchlist` functionality by introducing a new 'all' scope

### LINKED ISSUE #2117: Add an `all` scope to the get_sketches endpoint
Getting all sketches available to a user using the Python API. I can either get `user` sketches or `shared` sketches.

Increase the scope of the `get_sketches` API call, adding a `all` scope that gets both `user` and `shared` sketches.

Looping over all sketch IDs. Not tractable.

jaegeral:
## Summary of Changes

This pull request primarily focuses on enhancing the `sketchlist` functionality by introducing a new 'all' scope to the API, allowing users to retrieve all accessible sketches. It also refines the existing sketch listing logic to better handle archived sketches and significantly improves the API documentation for clarity. Additionally, the E2E test environment is updated to support multi-user testing.

### Highlights

* **New API Scope: 'all'**: I've introduced a new `all` scope to the `list_sketches` API endpoint and the corresponding API client documentation. This new scope allows users to retrieve all sketches they have access to, encompassing both sketches they own and those shared with them.
* **Improved Sketch Listing Logic**: The `list_sketches` API method has been enhanced to correctly handle the `include_archived` parameter across various existing scopes (`user`, `shared`, `archived`, `search`), ensuring that archived sketches are included or excluded as intended by the user's query.
* **API Documentation Enhancement**: The docstring for the `list_sketches` API endpoint (`GET /api/v1/sketches/`) has been significantly expanded to provide comprehensive details on all available query parameters, their possible values, and their effects, including the new `all` scope.
* **E2E Test Setup for Multiple Users**: The end-to-end (E2E) test environment setup has been updated to provision a second user, which will facilitate testing scenarios involving multiple users and shared resources, likely in support of the new 'all' and clarified 'shared' scopes.
* **Minor Error Message Refinements**: Several error messages returned by the API have been slightly refactored for improved clarity and conciseness, specifically for form validation failures and cases where sketch deletion or label actions are prevented.

Closes #2117

### DIFF
--- a/timesketch/lib/testlib.py
+++ b/timesketch/lib/testlib.py
@@ -696,7 +696,7 @@ class BaseTest(TestCase):
         """
         self.client.post(
             "/login/",
-            data=dict(username=username, password=password),
+            data={"username": username, "password": password},
             follow_redirects=True,
         )
 


timesketch/api/v1/resources/sketch.py
jkppr, line 1: Have you considered adding unit tests for the new and modified logic in `SketchListResource`? The existing tests in `timesketch/api/v1/resources_test.py` for this resource are quite minimal. Adding tests would make this PR even stronger.

timesketch/api/v1/resources/sketch.py
```diff
@@ -144,15 +168,32 @@ def get(self):
             else:
                 filtered_sketches = base_filter
         elif scope == "user":
-            filtered_sketches = base_filter.filter_by(user=current_user)
+            if include_archived:
+                filtered_sketches = base_filter_with_archived.filter_by(
+                    user=current_user
+                )
+            else:
+                filtered_sketches = base_filter.filter_by(user=current_user)
```
jkppr: Instead of repeating the `if include_archived` check for every scope, you could also check once before the scope checks and set the query method like this: `query = base_filter_with_archived if include_archived else base_filter`

And then you can just run `filtered_sketches = query.filter_by(user=current_user)` etc for each scope. This would make the code repeat itself less.

timesketch/api/v1/resources/sketch.py
```diff
@@ -84,8 +84,32 @@ def __init__(self):
...
+        Returns a list of sketches that the user has access to, filtered
+        and paginated according to the provided query parameters.
+
+        Query Parameters:
+            scope (str): Optional. Defines the scope of the sketches to return.
+                Can be one of:
+                - "user": (Default) Sketches owned by the current user.
+                - "shared": Sketches shared with the current user.
+                - "all": All sketches accessible by the user (owned and shared).
+                - "recent": Sketches recently accessed by the user.
```
jkppr, line 96: Optional, but maybe worth it, since you already touched the `SketchListResource` class in this PR:
* The `recent` scope bypasses the pagination logic passed via the form. We could use the chance to add in support for that as well instead of hardcoding it to last 10?
* The `recent` scope currently fetches the last 10 View objects and then, for each view, it fetches the associated Sketch and its Status. This requires basically N+1 database queries which can slow down the API endpoint. We can fetch all the "last activity" timestamps in a single, efficient query instead after you have the list of sketches from the paginator, and then use a dictionary lookup in the loop.

This could look something like this (untested):
```
from sqlalchemy.orm import joinedload, subqueryload
from timesketch.models.sketch import Sketch, View
[...]
if scope == "recent":
    views = (
        View.query.options(
            joinedload(View.sketch).subqueryload(Sketch.status)
        )
        .filter_by(user=current_user, name="")
        .order_by(View.updated_at.desc())
        .limit(10) # Or use pagination logic here
    )
    sketches = [
        view.sketch
        for view in views
        if view.sketch.get_status.status != "deleted"
    ]
    total_items = len(sketches)
[...]
```

timesketch/api/v1/resources/sketch.py
jaegeral: Good point

timesketch/api/v1/resources/sketch.py
jaegeral: let's keep that in a different issue, otherwise this PR becomes to complicated IMHO

timesketch/api/v1/resources/sketch.py
jaegeral: added some tests

timesketch/api/v1/resources/sketch.py
jaegeral: created https://github.com/google/timesketch/issues/3503

jaegeral (commented):
See updates WDYT?

jkppr (approved):
Thanks!

PR #3486, opened 2025-08-04
fix: Gemini styleguide

jaegeral:
Codehealth:

* Updating the `.gemini` Styleguide to better mimic the reality of the project and what we set up for `black`.
* Updating the `.gemini` Styleguide to include a check for Pull Requests to see if that helps increase the quality of that as well. If it gets to annoying, we can remove it again

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -461,7 +461,6 @@ confidence=HIGH,
 # --disable=W".
 disable=assignment-from-none,
         bad-inline-option,
-        broad-except,
         cyclic-import,
         deprecated-pragma,
         duplicate-code,


.pylintrc
```diff
@@ -461,6 +461,7 @@ confidence=HIGH,
 # --disable=W".
 disable=assignment-from-none,
         bad-inline-option,
+        broad-except,
```
berggren: why do we want to turn this off?

jaegeral: done as per side conversation, will disable it for `tsctl.py` but not the overall project.

PR #3501, opened 2025-08-15
DFIQv2 and Yeti Analyzer Refactor

jkppr:
This pull request introduces an overhaul of the Digital Forensics Investigative Questions (DFIQ) integration and refactors the Yeti analyzers to use the official `yeti-python` library.

#### Description

The primary goals of this change are to modernize our DFIQ implementation, improve its flexibility by integrating YETI as a data source, and enhance the robustness of our threat intelligence analyzers.

#### Key Changes

**1. DFIQv2 Implementation (`dfiq.py`, `scenarios.py`)**

*   **UUID as Primary Identifier:** All DFIQ objects (Scenarios, Facets, Questions) now use a `uuid` as their primary identifier for lookups, relationships, and sorting. The human-readable `dfiq_id` (e.g., "S1001") is retained for backwards compatibility in graph building and as a secondary identifier.
*   **YETI Integration:** The system can now fetch DFIQ templates directly from a connected YETI instance. A configuration flag (`YETI_DFIQ_ENABLED`) controls this behavior.
*   **Unified DFIQ Loading:** A new function `load_dfiq_from_config` in `scenarios.py` handles loading from both the local filesystem (`DFIQ_PATH`) and YETI. It intelligently merges the two sources, with YETI data taking precedence in case of UUID conflicts.
*   **Graph & Data Model Update:** The internal DFIQ graph in `dfiq.py` is now built using UUIDs. The in-memory data model (`ScenarioTemplate`, `FacetTemplate`, etc.) has been updated to reflect parent-child relationships correctly.
*   **API Enhancements:**
    *   `POST /api/v1/sketches/<id>/scenarios/` now prioritizes `uuid` for template lookups, providing a more stable reference than `dfiq_id` or `name`.
    *   The API now correctly creates the full hierarchy of Scenarios, Facets, and Questions in the database from a chosen template.
*   **Legacy DFIQ Removal:** The old `scenarios.yaml`, `facets.yaml`, and `questions.yaml` files have been removed, along with the old DFIQ YAML files, in favor of the new structured directory format. A `README.md` is added to guide users.

**2. Yeti Analyzer Refactor (`yetiindicators.py`, `yetiindicators_test.py`)**

*   **Adoption of `yeti-python` Library:** All Yeti analyzers now use the official `YetiApi` client instead of direct `requests` calls.
*   **Simplified Authentication:** Manual session authentication and token handling have been removed. The `YetiApi` library now manages authentication and token refreshes transparently.
*   **Robust Initialization & Error Handling:** The `YetiBaseAnalyzer.__init__` method now handles authentication. If authentication fails, it raises a `RuntimeError`, causing the analyzer to fail fast with a clear error message instead of proceeding in an invalid state.
*   **Code Cleanup:** Replaced direct API endpoint calls with high-level library methods (e.g., `api.search_graph`, `api.search_bloom`), making the code cleaner and more maintainable.
*   **Updated Tests:** Unit tests have been refactored to mock the `YetiApi` class and its methods, aligning with the new implementation.


jkppr:
Note: This PR requires https://github.com/yeti-platform/yeti-python to push a new release first to work correctly.

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -366,7 +346,21 @@ def get_entities(self, type_selector: List[str]) -> Dict[str, dict]:
...
+                # pylint: disable=protected-access
+                url = f"{self.api._url_root}/api/v2/entities/search"
+                response_bytes = self.api.do_request(
+                    "POST", url, json_data={"query": query, "count": 0}
+                )
+                data = json.loads(response_bytes)
+            except yeti_errors.YetiApiError as e:
+                raise RuntimeError(
+                    f"Error {e.status_code} retrieving entities from Yeti: {e}"
+                ) from e
```
tomchop: Added in https://github.com/yeti-platform/yeti-python/pull/18

tomchop (commented):
Thanks for these changes! First pass on the analyzer, will take a look at the rest soon

timesketch/lib/analyzers/yetiindicators.py
jkppr: Once your PR is merged, are you going to release a new yeti-python version?

jkppr:
The failing linter will be fixed once yeti-python 2.0.9 is published.

timesketch/lib/analyzers/yetiindicators.py
jkppr: Alright, I have updated the section to use the new method in the yeti-python lib.

timesketch/frontend-ng/src/components/Scenarios/QuestionCard.vue
```diff
@@ -557,9 +547,6 @@ export default {
       }
       return DOMPurify.sanitize(marked(markdown))
     },
-    getDfiqQuestionUrl(id) {
-      return 'https://dfiq.org/questions/' + id + '/'
-    },
```
tomchop, line 562: Can we add a link to Yeti here if it's enabled?

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -78,46 +86,37 @@ def __init__(self, index_name, sketch_id, timeline_id=None):
...
+            raise RuntimeError("The 'yeti-python' library is not installed.")
+
         root = current_app.config.get("YETI_API_ROOT")
+        if not root:
+            raise RuntimeError("YETI_API_ROOT is not configured.")
+
         if root.endswith("/"):
             root = root[:-1]
-        self.yeti_api_root = root
+        yeti_api_root = root
```
tomchop: redundant assignment? why not do `yeti_api_root = current_app.config.get` in L95 directly?

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -78,46 +86,37 @@ def __init__(self, index_name, sketch_id, timeline_id=None):
...
-        self._yeti_session = requests.Session()
+        yeti_api_key = current_app.config.get("YETI_API_KEY")
         tls_cert = current_app.config.get("YETI_TLS_CERTIFICATE")
-        if tls_cert and self.yeti_web_root.startswith("https://"):
-            self._yeti_session.verify = tls_cert
 
-        self._intelligence_refs = set()
-        self._intelligence_attribute = {"data": []}
+        if not all([yeti_api_root, yeti_api_key]):
+            raise RuntimeError("YETI_API_ROOT or YETI_API_KEY is not configured.")
```
tomchop: You're already checking for YETI_API_ROOT in L97.

timesketch/lib/analyzers/yetiindicators.py
```diff
@@ -138,8 +137,9 @@ def mark_event(
             msg = f'Observable match: "{indicator["value"]}" (ID: {indicator["id"]})\n'
         for neighbor in neighbors:
             tags.add(slugify(neighbor["name"]))
-            emoji_name = TYPE_TO_EMOJI[neighbor["type"]]
-            event.add_emojis([emojis.get_emoji(emoji_name)])
+            emoji_name = TYPE_TO_EMOJI.get(neighbor["type"])
+            if emoji_name:
+                event.add_emojis([emojis.get_emoji(emoji_name)])
```
tomchop: I'd rather raise an exception (or store an error somewhere, if possible) if the analyzer is using an unsupported emoji

timesketch/lib/dfiq.py
```diff
@@ -221,151 +255,176 @@ def questions(self):
             A list of Question objects.
         """
         return sorted(
-            [c for c in self.components.values() if isinstance(c, Question)],
-            key=lambda x: x.id,
+            [c for c in self.components.values() if isinstance(c, QuestionTemplate)],
+            key=lambda x: x.uuid,
         )
 
+    def get_by_id(self, dfiq_id: str):
```
tomchop, line 328: Would be great to have return types here as well

timesketch/lib/dfiq.py
```diff
@@ -221,151 +255,176 @@ def questions(self):
...
+
     @staticmethod
     def _convert_yaml_object_to_dfiq_component(yaml_object):
-        """Converts a YAML object to a DFIQ component.
+        """Converts a YAML object to a DFIQ component."""
+        component_uuid = yaml_object.get("uuid")
+        if not component_uuid:
+            component_uuid = str(uuid.uuid4())
+            logger.warning(
+                "DFIQ object '%s' ('%s') is missing a UUID. "
```
tomchop, line 385: DFIQ objects with no UUIDs are invalid. It's safe to assume that throughout the rest of the code.

If you're worried about backwards compatibility with existing DFIQ implementations, I suggest asking users to add UUIDs to their existing DFIQ objects.

timesketch/lib/dfiq.py
```diff
@@ -221,151 +255,176 @@ def questions(self):
...
-    def _load_dfiq_items_from_yaml(self):
-        """Loads DFIQ items from YAML files.
+            dfiq_object = self._convert_yaml_object_to_dfiq_component(
+                component_from_yaml
+            )
+            if dfiq_object:
+                component_dict[dfiq_object.uuid] = dfiq_object
+                # Only add to the map if an ID exists.
+                if dfiq_object.id:
+                    id_to_uuid_map[dfiq_object.id] = dfiq_object.uuid
```
tomchop, line 467: Why not store the object directly, so you can avoid a lookup in the future?

timesketch/lib/dfiq.py
```diff
@@ -117,49 +124,74 @@ def search_templates(self):
...
+            dfiq_id=dfiq_id,
+            description=description,
+            tags=tags,
+            parent_ids=None,
+        )
+        self.facets = []
+        self.questions = []
 
 
 class DFIQ:
```
tomchop: `DFIQCatalog` would be a better name. I was confused and thought this was a DFIQBase class from which you would make Questions, Scenarios, etc.

timesketch/lib/dfiq.py
```diff
@@ -221,151 +255,176 @@ def questions(self):
...
-        )
-        components.update(
-            self._load_yaml_files_by_type("Approach", self.yaml_data_path)
-        )
-        return components
+        return component_dict, id_to_uuid_map
 
-    def _build_graph(self, components):
-        """Builds a graph of DFIQ components.
+    def _load_dfiq_items_from_yaml(self):
```
tomchop: This function should be named something more explicit (e.g. `load_dfiq_from_yaml_files`).

Why not reuse from_yaml_list as well?

timesketch/lib/dfiq.py
```diff
@@ -221,151 +255,176 @@ def questions(self):
...
+                    if isinstance(child_obj, FacetTemplate):
+                        content.facets.append(child_uuid)
+                    elif isinstance(child_obj, QuestionTemplate):
+                        content.questions.append(child_uuid)
+
+            elif isinstance(content, FacetTemplate):
+                for child_uuid in children_uuids:
+                    child_obj = self.get_by_uuid(child_uuid)
+                    if isinstance(child_obj, QuestionTemplate):
+                        content.questions.append(child_uuid)
```
tomchop, line 558: This will hide structural errors in your DFIQ catalog, ie. Facets can only have Questions as children, any other item should be invalid.

timesketch/frontend-ng/src/components/Scenarios/QuestionCard.vue
jkppr: An import can be combined from file based dfiq data and yeti data. Do we have any option to distinguish between them? If not we would risk linking to a UUID that does not exist in yeti.

How would the link need to look like for yeti? Right now timesketch only knows the API endpoint. So e.g. in a docker deployment that would be `http://yeti-api:8000`.

tomchop: Ah yeah, I'm assuming you're not storing the yeti internal ID (numeric) of the DFIQ object. Altho, maybe that's a good way to distinguish the ones coming from yeti VS the ones loaded from files (they won't have one)

jkppr: Can I get the URL of the instance from the API? (Because e.g. in a docker deployment the API root is not equal to the URL a user needs to be linked to.

timesketch/lib/dfiq.py
jkppr: With the `id_to_uuid_map` I have a lookup that follows the `ID -> UUID -> Object` pattern (lookup in dicts is cheap). It establishes the UUID as the single identifier for component objects. This allows me in the future to just remove this initial `ID -> UUID` lookup step when we got rid of all the human readable IDs in DFIQ. Once that support is not needed anymore, we would be left with just the `UUID -> Object` lookup.

Given that the `get_by_id` method is not really a performance critical path, I would keep it this way until we remove the old IDs.

timesketch/lib/dfiq.py
jkppr: Good point. I renamed the function and introduced a `_populate` helper method that provides the logic used in both import methods.

timesketch/lib/dfiq.py
jkppr: Good point. I added a warning if the structure is not correct.

timesketch/frontend-ng/src/components/Scenarios/QuestionCard.vue
jkppr: As discussed, I have added a `YETI_WEB_ROOT` into the timesketch config that can be set in case API Root and web root are different.

timesketch/lib/dfiq.py
tomchop: Gotcha!

PR #3508, opened 2025-09-01
Enhance Scenario and Question API Client Functionality

jkppr:
This PR introduces enhancements to the Python API client, focusing on providing interactive capabilities for managing Scenarios and Questions within a sketch. It adds methods for modifying states, updating properties, and linking evidence.

## Description

The primary goal of this change is to empower users to programmatically interact with and update the investigative state of Scenarios and Questions via the API client. Previously, these objects were largely read-only. This update introduces setters, new methods, and supporting backend changes to enable a more dynamic workflow.

Key additions include:

  * **Modifying Scenarios and Questions**: Users can now change names, descriptions, status, and priority directly from the client.
  * **Managing Conclusions**: Functionality to add and list conclusions for investigative questions.
  * **Linking Events as Facts**: A new convenience method to link specific events to a conclusion, formally marking them as supporting facts.
  * **Performance Improvements**: Properties on the `Scenario` and `Question` objects are now cached after the initial API call to reduce redundant lookups.

-----

### **API Client (`timesketch_api_client`)**

  * **`Scenario` Class:**

      * Added a setter for `display_name` (e.g., `scenario.display_name = "New Name"`).
      * New method `set_status(status)` to update the scenario's status.
      * New method `list_facets()` to retrieve all facets for a scenario.

  * **`Question` Class:**

      * Added setters for `name`, `display_name`, and `description`.
      * New methods `set_status(status)` and `set_priority(priority)`.
      * New methods `add_conclusion(conclusion_text)` and `list_conclusions()` to manage conclusions.
      * A private `_update()` helper was added to streamline POST requests for updating the question object.

  * **`Sketch` Class:**

      * New high-level method `link_event_to_conclusion(events, conclusion_id)` to formally associate events with a conclusion by applying the `__ts_fact` label.
      * The `label_events()` method was enhanced to support `conclusion_id` and a `remove` flag, which provides the underlying mechanism for the new linking feature.

### **Backend API (`timesketch/api/v1/`)**

  * **`EventResource`**: The resource now correctly includes the `_id` and `_index` of an event in the `objects` part of the response, making it easier for the client to reference specific events.
  * **`QuestionListResource`**: Added logic to automatically generate a UUID for new questions if one is not provided, improving data consistency.
  * **`QuestionConclusionListResource`**: Fixed a bug by changing `filter_by` to `query.filter_by` for retrieving conclusions.

### **Testing**

  * Comprehensive unit tests have been added in `scenario_test.py` and `sketch_test.py` to validate all new client-side functionality.
  * The mocking library (`test_lib.py`) has been updated with new mock responses to support the test cases for resource modification and new endpoints.

### DIFF
--- a/api_client/python/timesketch_api_client/scenario.py
+++ b/api_client/python/timesketch_api_client/scenario.py
@@ -59,6 +59,57 @@ class Scenario(resource.BaseResource):
         return scenario["objects"][0]["name"]
 
     @property
+    def display_name(self):
+        """Property that returns the scenario display name."""
+        scenario = self.lazyload_data(refresh_cache=True)
+        return scenario["objects"][0]["display_name"]
+
+    @display_name.setter
+    def display_name(self, value):
+        """Sets the display name of the scenario."""
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": value}
+        response = self.api.session.post(resource_url, json=data)
+        _ = error.check_return_status(response, logger)
+        self.lazyload_data(refresh_cache=True)
+
+    def set_status(self, status):
+        """Sets the status of the scenario.
+
+        Args:
+            status (str): The new status for the scenario.
+
+        Returns:
+            bool: True if the status was set successfully.
+        """
+        resource_url = (
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"scenarios/{self.id}/status/"
+        )
+        data = {"status": status}
+        response = self.api.session.post(resource_url, json=data)
+        status = error.check_return_status(response, logger)
+        self.lazyload_data(refresh_cache=True)
+        return status
+
+    def list_facets(self):
+        """Lists all facets for the scenario.
+
+        Returns:
+            A list of dictionaries, each representing a facet.
+        """
+        resource_url = (
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"scenarios/{self.id}/facets/"
+        )
+        response = self.api.session.get(resource_url)
+        response_json = error.get_response_json(response, logger)
+        objects = response_json.get("objects", [])
+        if objects and isinstance(objects[0], list):
+            return objects[0]
+        return objects
+
+    @property
     def dfiq_identifier(self):
         """Property that returns the dfiq identifier.
 
@@ -107,6 +158,13 @@ class Question(resource.BaseResource):
             api=api, resource_uri=f"sketches/{self.sketch_id}/questions/{self.id}/"
         )
 
+    def _update(self, data):
+        """Internal helper to update the question."""
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        response = self.api.session.post(resource_url, json=data)
+        _ = error.check_return_status(response, logger)
+        self.lazyload_data(refresh_cache=True)
+
     @property
     def name(self):
         """Property that returns the question name.
@@ -116,6 +174,64 @@ class Question(resource.BaseResource):
         """
         question = self.lazyload_data()
         return question["objects"][0]["name"]
+
+    @name.setter
+    def name(self, value):
+        """Sets the name of the question."""
+        self._update({"name": value})
+
+    @property
+    def display_name(self):
+        """Property that returns the question display name."""
+        question = self.lazyload_data()
+        return question["objects"][0]["display_name"]
+
+    @display_name.setter
+    def display_name(self, value):
+        """Sets the display name of the question."""
+        self._update({"display_name": value})
+
+    def set_status(self, status):
+        """Sets the status of the question."""
+        self._update({"status": status})
+
+    def set_priority(self, priority):
+        """Sets the priority of the question."""
+        self._update({"priority": priority})
+
+    def list_conclusions(self):
+        """Lists all conclusions for the question.
+
+        Returns:
+            A list of dictionaries, each representing a conclusion.
+        """
+        resource_url = (
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"questions/{self.id}/conclusions/"
+        )
+        response = self.api.session.get(resource_url)
+        response_json = error.get_response_json(response, logger)
+        objects = response_json.get("objects", [])
+        if objects and isinstance(objects[0], list):
+            return objects[0]
+        return objects
+
+    def add_conclusion(self, conclusion_text):
+        """Adds a conclusion to the question for the current user.
+
+        Args:
+            conclusion_text (str): The text of the conclusion.
+
+        Returns:
+            A dictionary with the API response.
+        """
+        resource_url = (
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"questions/{self.id}/conclusions/"
+        )
+        data = {"conclusionText": conclusion_text}
+        response = self.api.session.post(resource_url, json=data)
+        return error.get_response_json(response, logger)
 
     @property
     def dfiq_identifier(self):
@@ -136,6 +252,11 @@ class Question(resource.BaseResource):
         """
         question = self.lazyload_data()
         return question["objects"][0]["description"]
+
+    @description.setter
+    def description(self, value):
+        """Sets the description of the question."""
+        self._update({"description": value})
 
     @property
     def approaches(self):
--- a/api_client/python/timesketch_api_client/scenario.py
+++ b/api_client/python/timesketch_api_client/scenario.py
@@ -44,6 +44,8 @@ class Scenario(resource.BaseResource):
         self.uuid = uuid
         self.api = api
         self.sketch_id = sketch_id
+        self._name = None
+        self._display_name = None
         super().__init__(
             api=api, resource_uri=f"sketches/{self.sketch_id}/scenarios/{self.id}/"
         )
@@ -55,14 +57,20 @@ class Scenario(resource.BaseResource):
         Returns:
             Scenario name as string.
         """
+        if self._name:
+            return self._name
         scenario = self.lazyload_data()
-        return scenario["objects"][0]["name"]
+        self._name = scenario["objects"][0]["name"]
+        return self._name
 
     @property
     def display_name(self):
         """Property that returns the scenario display name."""
-        scenario = self.lazyload_data(refresh_cache=True)
-        return scenario["objects"][0]["display_name"]
+        if self._display_name:
+            return self._display_name
+        scenario = self.lazyload_data()
+        self._display_name = scenario["objects"][0]["display_name"]
+        return self._display_name
 
     @display_name.setter
     def display_name(self, value):
@@ -70,8 +78,10 @@ class Scenario(resource.BaseResource):
         resource_url = f"{self.api.api_root}/{self.resource_uri}"
         data = {"scenario_name": value}
         response = self.api.session.post(resource_url, json=data)
-        _ = error.check_return_status(response, logger)
-        self.lazyload_data(refresh_cache=True)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._display_name = updated_object["display_name"]
+        self._name = updated_object["name"]
 
     def set_status(self, status):
         """Sets the status of the scenario.
@@ -154,6 +164,9 @@ class Question(resource.BaseResource):
         self.uuid = uuid
         self.api = api
         self.sketch_id = sketch_id
+        self._name = None
+        self._display_name = None
+        self._description = None
         super().__init__(
             api=api, resource_uri=f"sketches/{self.sketch_id}/questions/{self.id}/"
         )
@@ -162,8 +175,11 @@ class Question(resource.BaseResource):
         """Internal helper to update the question."""
         resource_url = f"{self.api.api_root}/{self.resource_uri}"
         response = self.api.session.post(resource_url, json=data)
-        _ = error.check_return_status(response, logger)
-        self.lazyload_data(refresh_cache=True)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._name = updated_object["name"]
+        self._display_name = updated_object["display_name"]
+        self._description = updated_object["description"]
 
     @property
     def name(self):
@@ -172,8 +188,11 @@ class Question(resource.BaseResource):
         Returns:
             Question name as string.
         """
+        if self._name:
+            return self._name
         question = self.lazyload_data()
-        return question["objects"][0]["name"]
+        self._name = question["objects"][0]["name"]
+        return self._name
 
     @name.setter
     def name(self, value):
@@ -183,8 +202,11 @@ class Question(resource.BaseResource):
     @property
     def display_name(self):
         """Property that returns the question display name."""
+        if self._display_name:
+            return self._display_name
         question = self.lazyload_data()
-        return question["objects"][0]["display_name"]
+        self._display_name = question["objects"][0]["display_name"]
+        return self._display_name
 
     @display_name.setter
     def display_name(self, value):
@@ -250,8 +272,11 @@ class Question(resource.BaseResource):
         Returns:
             Question description as string.
         """
+        if self._description:
+            return self._description
         question = self.lazyload_data()
-        return question["objects"][0]["description"]
+        self._description = question["objects"][0]["description"]
+        return self._description
 
     @description.setter
     def description(self, value):
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -1408,12 +1408,15 @@ class Sketch(resource.BaseResource):
         response = self.api.session.get(resource_url_base + resource_url_params)
         return error.get_response_json(response, logger)
 
-    def label_events(self, events, label_name):
+    def label_events(self, events, label_name, remove=False, conclusion_id=None):
         """Labels one or more events with label_name.
 
         Args:
             events: Array of JSON objects representing events.
             label_name: String to label the event with.
+            remove (bool): If true, the label will be removed instead of added.
+            conclusion_id (int): Optional. ID of a conclusion to link the
+                event to.
 
         Returns:
             Dictionary with query results.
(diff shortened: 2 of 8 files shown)


api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+    @display_name.setter
+    def display_name(self, value):
+        """Sets the display name of the scenario."""
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": value}
+        response = self.api.session.post(resource_url, json=data)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._display_name = updated_object["display_name"]
+        self._name = updated_object["name"]
```
tomchop: I don't know if this is a convention across the API client, but having an HTTP request when the user (of the API) is just assigning a value sounds very surprising. 

Can we change this to something like below, `set_display_name`?

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1408,12 +1408,15 @@ def get_event(self, event_id, index_id):
...
-    def label_events(self, events, label_name):
+    def label_events(self, events, label_name, remove=False, conclusion_id=None):
         """Labels one or more events with label_name.
 
         Args:
             events: Array of JSON objects representing events.
             label_name: String to label the event with.
+            remove (bool): If true, the label will be removed instead of added.
+            conclusion_id (int): Optional. ID of a conclusion to link the
+                event to.
```
tomchop, line 1414: It's a little awkward to call "label_events" if I want to link a conclusion to them.

api_client/python/timesketch_api_client/scenario.py
jkppr: This is the `@display_name.setter`. Reading happens via the property in line 68 above. For setting the value, we need to update it in the database, so a HTTP request is required.

api_client/python/timesketch_api_client/sketch.py
jkppr: This is something happening via the event label endpoint on the backend since linking involves both: Linking & adding a label so the frontend knows this is linked. (similar to stars).

Since it is indeed awkward I have also added the `link_event_to_conclusion` function below. This allows for users not to think about what happens under the hood when linking events.

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        self._name = scenario["objects"][0]["name"]
+        return self._name
+
+    @property
+    def display_name(self):
+        """Property that returns the scenario display name."""
+        if self._display_name:
+            return self._display_name
+        scenario = self.lazyload_data()
+        self._display_name = scenario["objects"][0]["display_name"]
```
jaegeral: would it make sense to add something in case there is no display_name?

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        return self._display_name
+
+    @display_name.setter
+    def display_name(self, value):
+        """Sets the display name of the scenario."""
```
jaegeral: Can you make the docstring a bit more verbose, e.g.

```
Updates the scenario's display name.

Sends a POST request to the API to update the scenario's display name,
then refreshes the local object's internal state with the updated data.

Args:
    value (str): The new display name to be set.
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": value}
+        response = self.api.session.post(resource_url, json=data)
+        self._data = error.get_response_json(response, logger)
```
jaegeral: If the http request is failing would you still update it locally?

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+    def display_name(self):
+        """Property that returns the scenario display name."""
+        if self._display_name:
+            return self._display_name
+        scenario = self.lazyload_data()
+        self._display_name = scenario["objects"][0]["display_name"]
+        return self._display_name
+
+    @display_name.setter
+    def display_name(self, value):
```
jaegeral: Can we do
```
def display_name(self, value: str) -> None:
``` 

instead?

api_client/python/timesketch_api_client/scenario.py
jaegeral: Why is the name here updated? Isn't that method about the display name?

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        """Sets the display name of the scenario."""
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": value}
+        response = self.api.session.post(resource_url, json=data)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._display_name = updated_object["display_name"]
+        self._name = updated_object["name"]
+
+    def set_status(self, status):
```
jaegeral: please use type hints here as well?

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        """Sets the status of the scenario.
+
+        Args:
+            status (str): The new status for the scenario.
```
jaegeral: I assume the status of a scenario has one of fixed set of values to choose from, I would suggest we enforce that in the setter so we do not end with scenarios in `status: weird` situations.

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": value}
+        response = self.api.session.post(resource_url, json=data)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._display_name = updated_object["display_name"]
+        self._name = updated_object["name"]
+
+    def set_status(self, status):
+        """Sets the status of the scenario.
```
jaegeral: I think it warrents a comment in docstring to say that it updates the value in the database and does a lazy_load.

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+        response = self.api.session.post(resource_url, json=data)
+        status = error.check_return_status(response, logger)
+        self.lazyload_data(refresh_cache=True)
+        return status
+
+    def list_facets(self):
+        """Lists all facets for the scenario.
+
+        Returns:
+            A list of dictionaries, each representing a facet.
```
jaegeral: ```    list[dict]: A list of dictionaries, where each dictionary represents a facet.```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"scenarios/{self.id}/status/"
+        )
+        data = {"status": status}
+        response = self.api.session.post(resource_url, json=data)
+        status = error.check_return_status(response, logger)
+        self.lazyload_data(refresh_cache=True)
+        return status
+
+    def list_facets(self):
```
jaegeral:
```suggestion
    def list_facets(self) -> List[Dict[str, Any]]:
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -55,8 +58,67 @@ def name(self):
...
+            A list of dictionaries, each representing a facet.
+        """
+        resource_url = (
+            f"{self.api.api_root}/sketches/{self.sketch_id}/"
+            f"scenarios/{self.id}/facets/"
+        )
+        response = self.api.session.get(resource_url)
+        response_json = error.get_response_json(response, logger)
+        objects = response_json.get("objects", [])
+        if objects and isinstance(objects[0], list):
```
jaegeral, line 165:
```suggestion
        # Handle the case where the API returns a nested list
        if objects and isinstance(objects[0], list):
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
         self.sketch_id = sketch_id
+        self._name = None
+        self._display_name = None
+        self._description = None
+        self._data = None
         super().__init__(
             api=api, resource_uri=f"sketches/{self.sketch_id}/questions/{self.id}/"
         )
 
+    def _update(self, data):
```
jaegeral:
```suggestion
    def _update(self, data: Dict[str, Any]) -> None:
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+        """Internal helper to update the question."""
```
jaegeral:
```suggestion
        """Internal helper to update the question with new data.

        This method sends a POST request to the API with the provided data
        and updates the internal state of the question object based on the
        API's response.

        Args:
                data (dict): A dictionary containing the updated question data.
                     Expected keys include 'name', 'display_name', and 'description'.
        """
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        response = self.api.session.post(resource_url, json=data)
+        self._data = error.get_response_json(response, logger)
+        updated_object = self._data["objects"][0]
+        self._name = updated_object["name"]
```
jaegeral:
```suggestion
        self._name = updated_object.get("name", self._name)
        self._display_name = updated_object.get("display_name", self._display_name)
        self._description = updated_object.get("description", self._description)
```

Have a safety net in case the values are not present in the `update_object`

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+        self._display_name = updated_object["display_name"]
+        self._description = updated_object["description"]
+
+    def add_attribute(self, name, value, ontology=None, description=None):
```
jaegeral:
```suggestion
    def add_attribute(self, name: str, value: str, ontology: Optional[str] = None, description: Optional[str] = None) -> None:
```
Make sure the optional aspect outlined in the docstring is reflected in the type.

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+        """Adds an attribute to the question.
```
jaegeral:
```suggestion
        """Adds a single attribute to the question.

         This method constructs a new attribute and updates the question on the API
         by calling the internal `_update` helper method.

```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+            ontology (str): Optional ontology for the attribute.
+            description (str): Optional description for the attribute.
+        """
+        attribute = {
+            "name": name,
+            "value": value,
+            "ontology": ontology,
+            "description": description,
+        }
+        self._update({"attributes": [attribute]})
```
jaegeral, line 318: I think here is a logic flaw, you are not adding it but replace all attributes existing before with the to be added one?

roughly:
```
current_attributes = self._data.get("attributes", [])
current_attributes.append(new_attribute)
self._update({"attributes": current_attributes})
```

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -103,19 +165,119 @@ def __init__(self, sketch_id, question_id, uuid, api):
...
+
     @property
     def name(self):
```
jaegeral: Same as above, and I assume for the others as well, check docstring, typing

jaegeral (changes requested):
A first pass with some comments

api_client/python/timesketch_api_client/scenario.py
tomchop: No I know, I was just referring to the fact that this was "hidden" in a setter. Explicitly calling set_display_name (or something like that) would make it much more intuitive that an HTTP request is about to happen. But as I said, maybe that's a (IMHO, not great) convention across the API client

api_client/python/timesketch_api_client/scenario.py
jkppr: Good catch. I'll add a fallback to using `name` which should be a required field.

api_client/python/timesketch_api_client/scenario.py
jkppr: To my understanding of the API client code, the error handling is done in the `error.get_response_json()` function. If the HTTP request failed, this error function would raise an error and by this does not change the display_name locally but throw the error.

Am I missing something here?

api_client/python/timesketch_api_client/scenario.py
jkppr: I'll refactor the function as recommended by tomchop earlier. But then I'll also ensure to add proper doc strings and typing.

api_client/python/timesketch_api_client/scenario.py
jkppr: No, this is handled on the back-end correctly. The way the [add_attribute](https://github.com/google/timesketch/blob/master/timesketch/models/annotations.py#L413) mixin method is designed is that it will append attributes to the existing list.

api_client/python/timesketch_api_client/scenario.py
jkppr: Good catch, this is not necessary.

api_client/python/timesketch_api_client/scenario.py
jkppr: Good catch. We don't use status or scenarios yet. So I'll remove the set_status for scenarios from the API client.

Once we add status for Scenarios we should do the validation of valid status values on the API Endpoint though and not in the client.

api_client/python/timesketch_api_client/scenario.py
jkppr: I see what you mean. Makes sense, I'll refactor accordingly.

api_client/python/timesketch_api_client/scenario.py
jkppr: I have removed this function for setting status with scenarios since we have no concept of this yet.

api_client/python/timesketch_api_client/scenario.py
jaegeral: ah okay got it

api_client/python/timesketch_api_client/scenario.py
jkppr: Changed by refactoring.

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -26,140 +27,385 @@
...
-    def name(self):
+    def name(self) -> str:
         """Property that returns the scenario name.
 
         Returns:
             Scenario name as string.
         """
+        if self._name:
+            return self._name
+        scenario = self.lazyload_data()
```
tomchop: Maybe for a future PR - can we change lazyload_data to automatically populate all attributes it can? It seems that for each "new" attribute we're trying to read from, we'd do the same exact API call that would return the same exact results.

tomchop: And, maybe, just always lazy_load the data as soon as the object is created? Are API calls that expensive?

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -26,140 +27,385 @@
...
+        resource_url = f"{self.api.api_root}/{self.resource_uri}"
+        data = {"scenario_name": display_name}
+        response = self.api.session.post(resource_url, json=data)
+
+        try:
+            response_json = error.get_response_json(response, logger)
+            self._data = response_json
+            updated_object = self._data["objects"][0]
+            self._display_name = updated_object.get("display_name")
+        except (RuntimeError, ValueError) as e:
```
tomchop, line 144: You can probably restrain the scope of the except statement here to just around `get_response_json` (and maybe include the `post`?)

api_client/python/timesketch_api_client/scenario.py
```diff
@@ -26,140 +27,385 @@
...
-    def description(self):
+    def description(self) -> str:
         """Property that returns the scenario description.
 
         Returns:
             Description as string.
         """
         scenario = self.lazyload_data()
-        return scenario["objects"][0]["description"]
+        return scenario["objects"][0].get("description", "")
```
tomchop: You're never storing the result of the lazyload_response in an attribute (unlike in `def name()`)

api_client/python/timesketch_api_client/scenario.py
jkppr: You raise some valid points here. 
Let me start with the second topic: Can't we lazyloads during object creation. I would argue it is not "Lazy loading" anymore if we do this? Doesn't "lazy" mean only when needed? Also I would see a risk for e.g. when Sketch.list_scenarios is called and 100 objects get created and for all it will pull already all data. 
So for now, I would prefer to keep it as is and aligned with the current concept of the API client.


But for your first point, I think at least for the scenarios I can already improve this. But I would like to not refactor the whole API client BaseResources here.

api_client/python/timesketch_api_client/scenario.py
jkppr: Good catch. Fixed it.

PR #3552, opened 2025-10-08
feat: introduce a log_pretext in log_analyzer.py

jaegeral:
Tracing and debugging the llm `log_analyzer`is kind of hard since the interaction with the LLM provide can take really long, sessions from various timelines and sketches could and will overlap. If an error occurs it is really hard to determine what caused the issue.

This PR introduces a uniq id that is created for each execution and then added to every logline moving forward. So going back in the logs it can be re-assembled what happened and what sketch the execution is tied to.

### DIFF
--- a/timesketch/lib/llms/features/log_analyzer.py
+++ b/timesketch/lib/llms/features/log_analyzer.py
@@ -14,6 +14,7 @@
 """LogAnalyzer feature for automated log analysis using LLMs via an external service."""
 import json
 import logging
+import uuid
 from typing import Any, Dict, Optional, Generator
 
 from timesketch.models import db_session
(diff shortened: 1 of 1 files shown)


timesketch/lib/llms/features/log_analyzer.py
```diff
@@ -104,8 +105,10 @@ def execute(  # pylint: disable=unused-argument
             LogAnalysisError: If the log stream preparation fails.
             Exception: For unexpected errors during execution.
         """
+        session_id = uuid.uuid4().hex[:8]  # create a uuid to track a session
```
tomchop, line 108: Why not use the entire UUID here?

tomchop (approved):
one nit, otherwise LGTM

timesketch/lib/llms/features/log_analyzer.py
jaegeral: my thought was that the entire one would be very long and just having a piece is enough to trace it

PR #3562, opened 2025-10-31
feat: log secGemini response to /tmp/ if timesketch app mode is set to DEBUG

jaegeral:
If Timesketch is set to DEBUG, log the responses of secgemini to a file in `/tmp/`. This drastically increases ability to debug and troubleshoot interactions with the LLM backend.


itsmvd (approved):

jkppr (approved):
Had to fix some minor logic bugs but now it works.
* Streams results now directly into files to save on memory.
* Records all messages during the conversation and not just the final one.
* Stores the messages as jsonl for easier processing.

PR #3581, opened 2025-11-10
feat: API client: refactor retry logic

jaegeral:
There use to be retry logic in some (but not all) methods that fetch or post data to the API in the api_client. In some cases that was okay but not good across the board for flaky errors. This PR is an attempt to make a retry logic across the whole client.

## Summary of Changes

This pull request enhances the robustness and consistency of the Timesketch API client by centralizing and improving its retry logic. It introduces a better retry mechanism that handles both transient network errors and specific application-level response issues, making the client more resilient to flaky API behavior. The changes also provide greater flexibility by allowing users to configure retry parameters, ensuring a more reliable interaction with the Timesketch API.

### Highlights

* **Centralized Retry Logic**: The API client's retry mechanism has been refactored to use `urllib3.Retry` with a custom `VerboseRetry` class, centralizing error handling for network-level issues (e.g., 5xx status codes) and providing more informative `MaxRetryError` messages.
* **Configurable Retry Parameters**: The `TimesketchApi` constructor now accepts `retry_count` and `backoff_factor` arguments, allowing users to customize the retry behavior for HTTP requests.
* **Application-Level Retry Handling**: A new private helper method, `_send_request_with_retry`, has been introduced to handle application-level retries for cases where the API returns a 200 OK status but with an empty, falsy, or invalid JSON payload, which `urllib3.Retry` does not cover.
* **Consolidated API Method Calls**: Several API methods, including `fetch_resource_data`, `create_sketch`, `create_user`, `create_searchindex`, and `create_sigmarule`, have been updated to leverage the new `_send_request_with_retry` for consistent and robust error handling.
* **Improved Error Specificity**: Error handling for unexpected API response formats in methods like `create_sketch` and `create_searchindex` has been updated to raise `ValueError` instead of the more general `RuntimeError`.

<details>
<summary><b>Changelog</b></summary>

* **api_client/python/timesketch_api_client/client.py**
    * Imported `HTTPAdapter`, `MaxRetryError`, and `Retry` from `requests.adapters` and `urllib3.util.retry` respectively.
    * Introduced `VerboseRetry` class for enhanced retry logging and exception messages.
    * Modified `TimesketchApi` class to include `retry_count` and `backoff_factor` attributes, configurable via the constructor.
    * Updated `_create_session` to configure the `requests.Session` with `HTTPAdapter` and the `VerboseRetry` strategy, including 'POST' in `allowed_methods`.
    * Added `_send_request_with_retry` private method to encapsulate HTTP requests and handle application-level retries for falsy or invalid JSON responses.
    * Refactored `fetch_resource_data`, `create_sketch`, `create_user`, `create_searchindex`, and `create_sigmarule` to utilize `_send_request_with_retry`.
    * Changed `DEFAULT_RETRY_COUNT` from 5 to 4 (representing 1 initial attempt + 4 retries).
    * Updated docstrings for `TimesketchApi` and related methods to reflect new retry behavior and configurable parameters.
    * Changed exception type from `RuntimeError` to `ValueError` for unexpected API response formats in `create_sketch` and `create_searchindex`.
* **api_client/python/timesketch_api_client/client_test.py**
    * Added `TimesketchApiRetryTest` class to verify that the `requests.Session` is correctly configured with the `HTTPAdapter` and `VerboseRetry` parameters.
* **api_client/python/timesketch_api_client/retry_test.py**
    * Added a new file containing unit tests for the `VerboseRetry` class, covering warning logging on retryable errors and the format of final `MaxRetryError` exceptions with and without response bodies.
* **api_client/python/timesketch_api_client/test_lib.py**
    * Added `auth_text_data` for mocking authentication responses.
    * Modified `MockSession` to include a `mount` method and an updated `request` method to support testing the new retry logic.
* **end_to_end_tests/client_test.py**
    * Updated expected exception type from `RuntimeError` to `NotFoundError` in `test_delete_sketch` to align with more specific error handling for resource not found scenarios.
</details>

This PR was supported by Gemini

### DIFF
--- a/api_client/python/timesketch_api_client/client.py
+++ b/api_client/python/timesketch_api_client/client.py
@@ -62,9 +62,9 @@ class TimesketchApi:
 
                         session: Authenticated HTTP session with retry logic configured.
 
-                        _retry_count: The number of retries configured for the session.
+                        retry_count: The number of retries configured for the session.
 
-                        _backoff_factor: The backoff factor used for retries."""
+                        backoff_factor: The backoff factor used for retries."""
 
     DEFAULT_OAUTH_SCOPE = [
         "https://www.googleapis.com/auth/userinfo.email",


api_client/python/timesketch_api_client/client.py
```diff
@@ -396,83 +409,19 @@ def fetch_resource_data(self, resource_uri, params=None):
...
+            return result
+
+        # If we reach here, it means get_response_json did not raise an error
+        # but returned a falsy result (e.g., empty dict/list). This should
+        # be treated as an error after all retries by VerboseRetry.
+        error_msg = (
+            f"Unable to fetch JSON resource data for request: '{resource_url}'"
+            f" - Response: '{result!s}'"
+        )
+        raise RuntimeError(error_msg)
```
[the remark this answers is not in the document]
jaegeral: hm that is a valid point, WDYT @jkppr  I am happy to make the retry logic for http code 200 again in (but more centrally)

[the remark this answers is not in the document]
jkppr: Yes makes sense. 

The new session-level `urllib3.Retry` takes care of all transport-level and transient 5xx errors. And if you add back a manual re-try loop that can be used inside key methods like `fetch_resource_data`, `create_sketch`, etc. that only checks for application-level errors that are not network dependent, we should have everything covered.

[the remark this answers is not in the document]
jaegeral: ok, tried to make that change now.

api_client/python/timesketch_api_client/client.py
```diff
@@ -349,6 +357,17 @@ def _create_session(
 
         session = requests.Session()
 
+        # Configure retry logic
+        retry_strategy = VerboseRetry(
+            total=self.DEFAULT_RETRY_COUNT,
+            backoff_factor=0.5,
+            status_forcelist=[500, 502, 503, 504],
+            allowed_methods=["HEAD", "GET", "PUT", "DELETE", "OPTIONS", "POST"],
```
jkppr, line 393: If a POST request to create a sketch succeeds but the connection drops before the client receives the response, the retry mechanism will send the exact same request again, potentially creating a duplicate sketch. 
Is this something we want to just accept for now and observe how it behaves with our deployment?

api_client/python/timesketch_api_client/client.py
```diff
@@ -366,121 +385,99 @@ def _create_session(
...
-        Retries occur for network connection errors, API errors (non-20x status codes),
-        JSON decoding errors, or if the API returns a successful (20x) response
-        with a "falsy" JSON payload (e.g., null, empty list/dictionary).
+        This is the core private helper for all API requests. It wraps the
+        session requests and adds a manual retry loop to handle cases where
+        the API returns a 200 OK status but with an empty or invalid JSON
+        payload.
 
+        For GET requests, it is preferred to use the `fetch_resource_data`
+        wrapper method for clarity and convenience.
```
jkppr: What is the clarity and convenience here?

api_client/python/timesketch_api_client/client.py
```diff
@@ -66,8 +73,9 @@ class TimesketchApi:
     DEFAULT_OAUTH_LOCALHOST_URL = "http://localhost"
     DEFAULT_OAUTH_API_CALLBACK = "/login/api_callback/"
 
-    # Default retry count for operations that attempt a retry.
-    DEFAULT_RETRY_COUNT = 5
+    # Default total attempts (initial request + retries) for operations.
+    # A value of 4 means 1 initial attempt + 4 retries = 5 total attempts.
+    DEFAULT_RETRY_COUNT = 4
 
     def __init__(
```
jkppr, line 86: Should we make the retry value configurable via the init? So default to 4 retries but if someone likes 0 or 100 they can choose themselves depending on their needs and deployment.

api_client/python/timesketch_api_client/client.py
```diff
@@ -366,121 +385,99 @@ def _create_session(
...
-                logger.info(
-                    "Waiting %.1fs before next attempt for request '%s'.",
-                    backoff_time,
-                    resource_url,
-                )
-                time.sleep(backoff_time)
+        resource_url = f"{self.api_root}/{resource_uri}"
 
+        last_exception = None
+        for attempt in range(3):
```
jkppr: Should we make this configurable as well? Set via the init method?

api_client/python/timesketch_api_client/client.py
```diff
@@ -366,121 +385,99 @@ def _create_session(
...
                 )
-            except ValueError as e:
-                if attempt >= self.DEFAULT_RETRY_COUNT:
-                    # Re-raise the original error after exhausting retries
-                    error_msg = (
-                        f"Error parsing response for request '{resource_url}'"
-                        f" - {e!s}"
-                    )
-                    raise ValueError(error_msg) from e
+                time.sleep(1)
```
jkppr: No increasing backoff here? 
I'm fine with going with static for now, and add increasing backoff later if needed.

api_client/python/timesketch_api_client/client.py
```diff
@@ -491,117 +488,51 @@ def create_sketch(self, name, description=None):
...
+            sketch_id = objects[0]["id"]
+            return self.get_sketch(sketch_id)
+
+        # If we reach here, it means get_response_json did not raise an error
+        # but returned an unexpected 'objects' format or it was empty.
+        error_message_detail = (
+            "API for sketch creation returned an unexpected 'objects' "
+            f"format or it was empty. Response: {response_dict!s}"
         )
+        raise RuntimeError(error_message_detail)
```
jkppr: Is RuntimeError here the correct choice? It seems a bit unspecific. Maybe `ValueError` since none of the answered value format matched? Or a custom `errors.ApiResponseError`?

api_client/python/timesketch_api_client/client.py
```diff
@@ -792,107 +719,40 @@ def create_searchindex(self, searchindex_name: str, opensearch_index_name: str):
...
+            searchindex_id = objects[0]["id"]
+            return index.SearchIndex(searchindex_id, api=self)
+
+        # If we reach here, it means get_response_json did not raise an error
+        # but returned an unexpected 'objects' format or it was empty.
+        error_message_detail = (
+            "API for searchindex creation returned an unexpected 'objects' "
+            f"format or it was empty. Response: {response_dict!s}"
         )
+        raise RuntimeError(error_message_detail)
```
jkppr: Same as with the other RuntimeError.

api_client/python/timesketch_api_client/client.py
```diff
@@ -349,6 +357,17 @@ def _create_session(
 
         session = requests.Session()
 
+        # Configure retry logic
+        retry_strategy = VerboseRetry(
+            total=self.DEFAULT_RETRY_COUNT,
+            backoff_factor=0.5,
```
jkppr: Should this be made configurable via the init method so users can choose to overwrite the value?

api_client/python/timesketch_api_client/client.py
jaegeral: I think that is a risk we would accept, fixingt that would require some kind of uniq identifier to be sent with every request and then check on the serverside if it was already sent or doing a get before each post request if the thing that is going to be POSTed is already there.

api_client/python/timesketch_api_client/client.py
jaegeral: because it was the method that was there before and was used all over the codebase

jaegeral: I updated the docstring

api_client/python/timesketch_api_client/client.py
jaegeral: sure thing

api_client/python/timesketch_api_client/client.py
jaegeral: hm `ApiResponseError` seems to be general. Will change it to a valueerror

api_client/python/timesketch_api_client/client.py
jaegeral: easy enough

api_client/python/timesketch_api_client/client.py
jaegeral: yup can do

api_client/python/timesketch_api_client/client.py
```diff
@@ -49,10 +52,19 @@
...
+
+                    Attributes:
+
+                        api_root: The full URL to the server API endpoint.
+
+                        session: Authenticated HTTP session with retry logic configured.
+
+                        _retry_count: The number of retries configured for the session.
+
+                        _backoff_factor: The backoff factor used for retries."""
```
jkppr: The arguments are not private are they? 

```suggestion
                        retry_count: The number of retries configured for the session.

                        backoff_factor: The backoff factor used for retries."""
```

api_client/python/timesketch_api_client/client.py
```diff
@@ -79,6 +93,8 @@ def __init__(
         client_secret="",
         auth_mode="userpass",
         create_session=True,
+        retry_count=None,
```
jkppr: Why not default to the default value right away?

```suggestion
        retry_count=DEFAULT_RETRY_COUNT,
```

PR #3587, opened 2025-11-12
initial attempt to fix some delete workflows

jaegeral:
This pull request enhances the system's deletion workflows by adding new end-to-end tests and refining existing logic. It specifically addresses the proper handling of soft-deleted sketches during hard deletion and ensures that search indexes for failed timelines are correctly archived upon deletion.

### DIFF
--- a/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
+++ b/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
@@ -204,7 +204,11 @@ class SecGeminiLogAnalyzer(interface.LLMProvider):
     def generate_stream_from_logs(
         self,
         log_events_generator: Iterable[Dict[str, Any]],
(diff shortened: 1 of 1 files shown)


timesketch/api/v1/resources/timeline.py
```diff
@@ -576,7 +576,7 @@ def delete(self, sketch_id: int, timeline_id: int):
             if timeline_.id != timeline_id:
                 # There are more than a single timeline using this index_name,
                 # we can't close it (unless this timeline is archived).
```
jkppr, line 588: Should we update the comment accordingly to reflect the tates that will prevent index closure?

jkppr (approved):
lgtm, small comment below

timesketch/api/v1/resources/timeline.py
jaegeral: done, I adjusted the docstring and the comment in here to reflect it

PR #3597, opened 2025-11-14
Fix back-to-back SecGemini investigations

gpavlidi:
There are 2 main contributions here:

1. Dev quality-of-life changes (override SecGemini URLs for local dev, ability to access docker host's local servers etc)
2. Early termination from SecGemini's streaming to address the back-to-back investigations issue

### DIFF
--- a/docker/dev/docker-compose.yml
+++ b/docker/dev/docker-compose.yml
@@ -30,9 +30,6 @@ services:
     restart: always
     volumes:
       - ../../:/usr/local/src/timesketch/
-    # Make docker host accessible for local development.
-    extra_hosts:
-      - host.docker.internal:host-gateway
 
   opensearch:
     container_name: opensearch
--- a/docs/developers/getting-started.md
+++ b/docs/developers/getting-started.md
@@ -146,6 +146,32 @@ Point your browser to `http://localhost:5001/` to access the new frontend UI.
 All changes to the `timesketch/frontend-ng/` path will be automatically build
 and loaded in the new frontend.
 
+### Accessing services on your host machine
+
+In some development scenarios, you might need the Timesketch container to
+communicate with a service running directly on your host machine (the "docker
+host"). A common use case is when you are running local LLM (Large Language
+Model) services (e.g., Ollama) that need access to hardware like GPUs which may
+not be available to the container.
+
+To enable this, you need to make the docker host accessible from within the
+Timesketch container. You can achieve this by adding an `extra_hosts` entry to
+the `timesketch` service in your `docker/dev/docker-compose.yml` file:
+
+```yaml
+services:
+  timesketch:
+    ...
+    # Make docker host accessible for local development.
+    extra_hosts:
+      - host.docker.internal:host-gateway
+```
+
+After adding this and restarting your container, you can configure Timesketch to
+connect to services on your host using `http://host.docker.internal:<PORT>`. For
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -54,3 +54,6 @@ vagrant/*.log
 
 # Exclude debugging profiles
 profiles/
+
+# Exclude timesketch.dev.conf
+data/timesketch.dev.conf
```
jkppr: Was this file created/renamed manually? If yes, I would opt for not submitting it to the upstream .gitignore. If this is the result of some automated process during the development workflow by Timesketch, it's a different story.

docker/dev/docker-compose.yml
```diff
@@ -30,6 +30,9 @@ services:
     restart: always
     volumes:
       - ../../:/usr/local/src/timesketch/
+    # Make docker host accessible for local development.
+    extra_hosts:
+      - host.docker.internal:host-gateway
```
jkppr: What is your use-case here?
This seems like a specific change relevant to your personal local dev environment and not something everyone needs?

data/timesketch.conf
```diff
@@ -419,7 +419,10 @@ LLM_PROVIDER_CONFIGS = {
         'secgemini_log_analyzer_agent': {
             'logs_processor_api_url': '',
             'api_key': '',
-            'model': 'logs_analysis_agent-1.1'
+            'model': 'logs_analysis_agent-1.1',
+            'base_url': '',
+            'wss_url': '',
+            'agents_config': {},
```
itsmvd, line 438: Can we add a doc-string above to explain what agents_config does / what can be specified?

itsmvd (approved):

.gitignore
gpavlidi: Yeah I can see your point, will push an update undoing this change

docker/dev/docker-compose.yml
gpavlidi: This is the standard way to be able to access your docker host from the containers and I think it's applicable to multiple dev usecases. This specific one use case is facilitating local development with different LLM services that are not provided as docker containers. It allows you to provide urls like {'base_url': 'http://host.docker.internal:8000'} in timesketch.conf for 'secgemini_log_analyzer_agent' configuration as an example. 

I can see this being useful for other LLM providers like ollama, for example if your docker host has a GPU that is not accessible in a container and you want Timesketch to forward requests to your docker host.

data/timesketch.conf
gpavlidi: Yes good point. Will add an example config to set the max investigation rounds
{'logs_analysis_loop_agent':{'max_iterations': 10}}

docker/dev/docker-compose.yml
jkppr: Thanks for the explanation. With Timesketch we operate the compose files as bare minimum for the general use-case and if someone wants to do some special use-case situations they need to adapt these changes locally for them. Otherwise it will get messy supporting all the different use-cases.

Hence I would prefer not to add this here and rather add a small section in the [developer docs](https://github.com/google/timesketch/blob/master/docs/developers/getting-started.md) explaining whats needed if you want to use a local service running on your host. Could you add a small section there?

jkppr: I took care of it 👍

PR #3612, opened 2025-11-25
Add a custom prompt field to the AI investigation view

itsmvd:
Passing the custom prompt as kwargs through the analyzer framework to the log analysis agent

image

### DIFF
--- a/.github/workflows/build_frontend-ng.yml
+++ b/.github/workflows/build_frontend-ng.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
       - name: Set up Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version: '20'
 
--- a/.github/workflows/build_frontend-v3.yml
+++ b/.github/workflows/build_frontend-v3.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
 
       - name: Set up Node.js
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v6
         with:
           node-version: '20'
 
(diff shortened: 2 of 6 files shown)


timesketch/api/v1/resources/settings.py
```diff
@@ -19,6 +19,9 @@
 from flask_restful import Resource
 from flask_login import login_required
 from timesketch.lib.llms.providers import manager as llm_manager
+from timesketch.lib.llms.providers.secgemini_log_analyzer_agent import (
+    DEFAULT_PROMPT as LOG_ANALYZER_DEFAULT_PROMPT,
+)
```
jkppr: Instead of importing this from the log_analyzer file, let's make the DEFAULT_PROMPT configurable either in a config file in the `data` folder or as part of the `timesketch.conf`. This would then also allow admins to set a different default prompt than ours.

timesketch/api/v1/resources_test.py
```diff
@@ -31,6 +31,9 @@
 from timesketch.models.sketch import InvestigativeQuestionApproach
 from timesketch.models.sketch import Facet
 from timesketch.api.v1.resources import ResourceMixin
+from timesketch.lib.llms.providers.secgemini_log_analyzer_agent import (
+    DEFAULT_PROMPT as LOG_ANALYZER_DEFAULT_PROMPT,
+)
```
jkppr: Same as  in the setting.py, let's move this into the `timesketch.conf`.

timesketch/api/v1/resources/settings.py
jkppr: Done

timesketch/frontend-v3/src/utils/RestApiClient.js
```diff
@@ -437,13 +437,20 @@ export default {
   getAnalyzers(sketchId) {
     return RestApiClient.get("/sketches/" + sketchId + "/analyzer/");
   },
-  runAnalyzers(sketchId, timelineIds, analyzers, forceRun = false) {
+  runAnalyzers(sketchId, timelineIds, analyzers, forceRun = false, customPrompt = null) {
```
jkppr: Let's not call it "customPrompt" here.

jkppr: Done

PR #3619, opened 2025-12-05
feat(tsctl): Add `sync-groups-from-json` command for bulk user management

jkppr:
This PR introduces a new CLI command, `sync-groups-from-json`, to `tsctl.py`. This command enables administrators to synchronize user group memberships based on an input JSON configuration file. It allows for the automation of user and group management, facilitating integration with external identity providers or bulk onboarding workflows.

**Key Changes:**
*   **New Command:** Added `sync-groups-from-json <filepath>` to the `tsctl` CLI.
*   **JSON Import:** Accepts a JSON file structured as a dictionary where keys are group names and values are lists of usernames.
*   **Automated Provisioning:**
    *   Creates groups if they do not exist in the database.
    *   Creates users if they do not exist, assigning them a secure, randomly generated password.
*   **Membership Synchronization:**
    *   Adds users to groups defined in the JSON.
    *   Removes users from groups if they are not listed in the JSON for that specific group.
*   **Performance:** Implements pre-fetching of existing users and groups to prevent N+1 database queries.
*   **Safety:** Includes a `--dry-run` flag to simulate changes and log intended actions without committing to the database.

**Usage Example:**

1. Create a `groups.json` file:
   ```json
   {
     "analysts": ["alice@timesketch.org", "bobtimesketch.org"],
     "administrators": ["alice@timesketch.org"]
   }
   ```

2. Run in dry-run mode to inspect changes:
   ```bash
   tsctl sync-groups-from-json groups.json --dry-run
   ```

3. Commit changes to the database:
   ```bash
   tsctl sync-groups-from-json groups.json
   ```

**Implementation Details:**
*   Uses `secrets` and `string` for cryptographically strong random password generation for new accounts.
*   Uses `db_session.flush()` to ensure new objects have IDs available for relationship mapping within the transaction.
*   Validates input file existence and JSON structure before processing.
*   Logs warnings for database groups that are not present in the input file (orphaned groups are skipped, not deleted).
*   Added e2e tests.
*   Updated documentation.

### DIFF
--- a/docs/guides/admin/admin-cli.md
+++ b/docs/guides/admin/admin-cli.md
@@ -247,7 +247,7 @@ tsctl add-group-member analysts --username john
 tsctl remove-group-member analysts --username john
 ```
 
-#### `sync-group-memberships`
+#### `sync-groups-from-json`
 
 Synchronizes user group memberships from a JSON file. This command will create,
 add, and remove users from groups to match the state defined in the JSON file.
@@ -284,10 +284,10 @@ is a list of usernames to be in that group.
 **Example Usage:**
 ```shell
 # Perform a dry run to see what changes would be made
-tsctl sync-group-memberships /tmp/groups.json --dry-run
+tsctl sync-groups-from-json /tmp/groups.json --dry-run
 
 # Apply the changes to the database
-tsctl sync-group-memberships /tmp/groups.json
+tsctl sync-groups-from-json /tmp/groups.json
 ```
 
 ---


timesketch/tsctl.py
```diff
@@ -3438,3 +3440,148 @@ def import_db(filepath, yes):
                 sqlalchemy.text("SET session_replication_role = 'origin'")
             )
         click.echo("Database import finished.")
+
+
+@cli.command(name="sync-group-memberships")
```
jaegeral: hm I wonder if we could give it a better name, when reading sync group membership, I would not expect it to also create users / groups but only to see if an existing user is in a existing group if not add it.

timesketch/tsctl.py
```diff
@@ -3438,3 +3440,148 @@ def import_db(filepath, yes):
...
+    """Synchronize user group memberships from a JSON file.
+
+    The JSON file should be a dictionary where keys are group names and values
+    are lists of usernames (email addresses).
+
+    The script will:
+    1. Create groups that don't exist.
+    2. Create users that don't exist (with a random password).
+    3. Add users to groups defined in the JSON.
+    4. Remove users from groups if they are NOT in the JSON list for that group.
```
jaegeral, line 3460: should we call out that users not existing would be not be removed

timesketch/tsctl.py
```diff
@@ -3438,3 +3440,148 @@ def import_db(filepath, yes):
...
+            click.echo(f" -> {g}")
+
+    if not dry_run:
+        click.echo("Committing changes to database...")
+        try:
+            db_session.commit()
+            click.echo("Sync complete.")
+        except Exception as e: # pylint: disable=broad-except
+            click.echo(f"Error: Failed to commit changes: {e}")
+            db_session.rollback()
```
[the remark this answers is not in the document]
jaegeral: It would be fine for me to use the broad exception

timesketch/tsctl.py
jkppr: My thought was: It synchronizes the internal group state with the one provided in the input file. Hence: `sync-group-memberships`. But I can see the issue with the `memberships` part.

So maybe we can name it simply `sync-groups`?

timesketch/tsctl.py
```diff
@@ -3438,3 +3440,148 @@ def import_db(filepath, yes):
...
             )
         click.echo("Database import finished.")
+
+
+@cli.command(name="sync-group-memberships")
+@click.argument("filepath")
+@click.option(
+    "--dry-run", is_flag=True, help="Calculate changes/logs without committing."
+)
+def sync_group_memberships(filepath, dry_run):
```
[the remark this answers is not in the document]
jkppr: This is a specific e2e task for the tsctl. I'll keep it in one function for now.

timesketch/tsctl.py
jkppr: We settled on `sync-groups-from-json`

PR #3623, opened 2025-12-05
Update secgemini_log_analyzer_agent.py

babirous:
Update the prompt passed to secgemini


jkppr:
@itsmvd any objections for the proposed prompt?

PR #3626, opened 2025-12-11
chg: [importer] reuse existing sketch if name is given

### LINKED ISSUE #3625: timesketch_importer reuse existing sketch by name
Currently, with timesketch_importer.py there are two ways to define the sketch:
1. reuse existing sketch by giving sketch_id
2. create new sketch by giving sketch_name 

Reuse existing sketch by giving sketch_name

**Additional context**
A pull request this minor change is coming in a moment

cvandeplas:
- timesketch_importer.py recreates new sketches when a sketch_name is given, instead of reusing the existing sketch 

- reuse of the existing sketch when a name is given, if the sketch does not exist, create a new one with that name (original behaviour) 

Added one function to get the `get_sketch_by_name()` plus a minor change try/except to try getting the existing sketch.

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

closes #3625


importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
...
+        KeyError: If a sketch with the given name is not found.
+
+    Returns:
+        A sketch object.
+    """
+    sketches = ts_client.list_sketches()  # may return Sketch objects or dicts
+    for s in sketches:
+        # Sketch object
+        if hasattr(s, "name") and s.name.lower() == sketch_name.lower():
+            return s
```
jaegeral: this would only return the first sketch with a given name, but a sketch name is not unique, so how would you deal with that situation?

cvandeplas: A sketch name is indeed not unique (which is a bit confusing sometimes in terms of usability, although I understand it's a feature). 
The proposed approach is pragmatic: just take the first one.

This small change fits into a bigger change I'm working on: adapting the timesketch_importer.py to also allow monitoring files on the filesystem to feed into timesketch. 
Comparable to a splunk universal forwarder it will keep track of the files and the position in the files.
The idea is that we can then feed data into timesketch based on a hostname/device serial number (sketch_name), and timeline_name the time of the forensic capture.

Happy to discuss live by chat on Slack

importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
...
+        # dict representation
+        if isinstance(s, dict) and s.get("name", "").lower() == sketch_name.lower():
+            sketch_id = s.get("id")
+            if sketch_id and hasattr(ts_client, "get_sketch"):
+                return ts_client.get_sketch(sketch_id)
+            return s
```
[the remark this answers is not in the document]
jkppr: @cvandeplas can you please take a look at this?

The specific issue is that if the code reaches the `return s` line, it returns a `dict`. However, the calling function (`main`) expects a `Sketch` object and immediately accesses `my_sketch.id`. If a `dict` is returned, this will raise an `AttributeError` and crash the tool.

It should be updated to ensure the `get_sketch_by_name` only ever returns a `Sketch` object. If the dictionary representation is missing an ID, we should `continue` the loop rather than returning the raw dictionary. Or raise an error if this is something you think should be flagged.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
...
+    Args:
+        ts_client: An instance of timesketch_api_client.TimesketchApi (or similar).
+        sketch_name: The name of the sketch to find.
+
+    Raises:
+        KeyError: If a sketch with the given name is not found.
+
+    Returns:
+        A sketch object.
+    """
```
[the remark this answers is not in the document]
jkppr: @cvandeplas please add the types for arguments as suggested.

importer_client/python/tools/timesketch_importer.py
jkppr: We have discussed this in the Timesketch team. While we understand the goal of automation, relying on the "first" result is risky because the API's return order isn't guaranteed to be deterministic. This could lead to data being silently fragmented across different sketches if duplicates exist.

To merge this, we need logic to handle naming conflicts safely:

1.  **Default Behavior:** It must be transparent. If multiple sketches have the same name, the tool should **error out** (in non-interactive mode) or **prompt the user** to select the correct ID (in interactive mode). We cannot silently choose one.
2.  **Automation Support:** Since you need this for the monitoring feature, please add a flag (e.g., `--sketch-strategy=latest` or `--ignore-sketch-duplicates`) that explicitly tells the tool how to resolve the conflict (e.g., picking the one with the highest ID).

This ensures that ignoring a conflict is a conscious choice by the user, not an implicit side effect.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
...
+    sketches = ts_client.list_sketches()  # may return Sketch objects or dicts
+    for s in sketches:
+        # Sketch object
+        if hasattr(s, "name") and s.name.lower() == sketch_name.lower():
+            return s
+        # dict representation
+        if isinstance(s, dict) and s.get("name", "").lower() == sketch_name.lower():
+            sketch_id = s.get("id")
+            if sketch_id and hasattr(ts_client, "get_sketch"):
```
jkppr: What's the background for checking `hasattr(ts_client, "get_sketch")` here?

The ts_client argument is typed as `client.TimesketchApi`. If the API client instance does not have the `get_sketch` method, the library is fundamentally broken or the wrong object type was passed (which static analysis should catch).

importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
...
+    Returns:
+        A sketch object.
+    """
+    sketches = ts_client.list_sketches()  # may return Sketch objects or dicts
+    for s in sketches:
+        # Sketch object
+        if hasattr(s, "name") and s.name.lower() == sketch_name.lower():
+            return s
+        # dict representation
+        if isinstance(s, dict) and s.get("name", "").lower() == sketch_name.lower():
```
jkppr: `s.get("name", "").lower() == sketch_name.lower()` this makes the it even more important to handle naming conflicts correctly since the logic would now consider "Case-64d7280f" and "case-64D7280F" the same. While this makes sense in the context of usability, it needs solid error and edge case handling.

importer_client/python/tools/timesketch_importer.py
```diff
@@ -59,6 +59,35 @@ def configure_logger_default():
         handler.setFormatter(logger_formatter)
 
 
+def get_sketch_by_name(
```
jkppr: Architectural comment

So looking up a sketch by name is a generic operation that would be useful for other integrations as well, not just the importer. (like the existing `get_sketch`)

Instead of keeping this logic isolated in `timesketch_importer.py`, could you move this method into the core `TimesketchApi` class in `api_client/python/timesketch_api_client/client.py` [here](https://github.com/google/timesketch/blob/master/api_client/python/timesketch_api_client/client.py#L630)?

This would ensure the method can be used in other tools as well and keeps the "business logic" of how we handle sketches in one place. It also let's the importer focus on what it can do best: Import Data.

You can add it right next to the existing `get_sketch(self, sketch_id)` method. It should handle all the duplicate name issues etc. that we discussed in this PR there as well.

Then, in `timesketch_importer.py`, you can simply call `my_sketch = ts_client.get_sketch_by_name(sketch_name)`.

cvandeplas:
Thank you very much for the thinking and lots of provided input. 
I already started to work on this but it will not be wrapped up in a few days (there are a few other higher prio's in the TODO stack). It'd be great if we can keep this PR open a bit longer in the meantime.  I'll convert it to draft.

jkppr:
Hi @cvandeplas , how is this PR going?
Since it's been some time since the last update, I propose we close this PR and you can open a new one once the solution can be reviewed again. 

I'm also happy to discuss any design decisions via the slack channel or in the discussion section.

cvandeplas:
It's a bit busy, still on the TODO list. I'll close the PR, and will fall back to the input you gave here and to the discussions via the slack channel. Thanks again for the support !

PR #3630, opened 2025-12-11
chg: [importer] file monitoring

### LINKED ISSUE #3629: timesketch_importer monitoring files and folders
A feature we're missing is the ability to continuously monitor for new files and folders, and updated files and ingest them into Timesketch, comparable to what Splunk universal forwarder does. 

The ability to monitor a folder, and all subfolders for new files, or changed files and ingest the data into timesketch.
The sketch_name and timeline_name may need to be dynamic, for example based on the folder path of the files.

- [timesketch-importer.sh](https://github.com/google/timesketch/blob/master/contrib/timesketch-importer.sh) is not an option as 
  - it only imports new files, when they are fully there (file close), 
  - it doesn't support monitoring changes in files (new entries added). 
  - It also relies on filesystem features (inotify) which may not be available with NFS or SMB
  - and requires files to be available on the timesketch instance itself

A draft PR will be raised in a moment to request input on a proposed implementation.

cvandeplas:
Implements feature request #3629 with the intent to collect your feedback and improve the code until we reach something that can be merged.

## What new feature is being introduced with this PR?
- ingesting files continuously into timesketch, new files + updates in the files

## Overview of changes:
Addition of command line parameters:
 - `monitor`: to enable the monitoring feature
 - `monitor_state_file` : optional state tracking file, as default is homedir.
 - `sketch_name_path_pos` : the position in the path that should be used to find the sketch name
 - `timeline_name_path_pos`: same, for the timeline name.

The sketch and timeline name path position can be used outside of the file monitoring feature. 
 
Implemented full monitoring flow
- Added WatchdogDirectoryMonitor class (polling via PollingObserverVFS to be filesystem agnostic).
- Added _WatchdogEventHandler to enqueue filesystem events.
- Worker thread + queue to process files asynchronously.
- Initial scan (_initial_scan) to enqueue existing files at startup.
- Per-file state tracking persisted to JSON (load/save).

Potential improvements: (later)
- Multiple worker threads to improve performance. But considering we need to be careful not to overload timesketch, I'm cautious to do this immediately

**Checks**

- [ ] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.

**Closing issues**

closes #3629


jaegeral:
Not reviewing yet, but I wonder if it make sense for you to just fork the import_client into https://github.com/google/timesketch/tree/master/contrib ?

cvandeplas:
I considered making a separate tool, but then realized that most of the code of the timesketch_importer is to support multiple authentication methods and such generic features. So to me it felt logic to keep it together.

jkppr:
I've added some guidance to the issue: https://github.com/google/timesketch/issues/3629#issuecomment-3670733364

cvandeplas:
Thanks again for the input. As mentioned: It's on my (our) TODO list.
I also noticed an issue in expired sessions with long-term ingestion, so I'd also need to review how the authenticated session is handled.

cvandeplas:
As with #3626 , 
It's a bit busy, still on the TODO list. I'll close the PR, and will fall back to the input you gave here and to the discussions via the slack channel. Thanks again for the support !

Issue #3327, opened 2025-03-07
Add a 403 error page


jkppr:
Accessing a Sketch I have no access to results in the odd situation that I end up on the "empty state" page and get the 403 snackbar. I think it would be better to have a dedicated 403 error page, since the empty state page could indicate I can upload timelines to the sketch, which I cant.

![Image](https://github.com/user-attachments/assets/3d34abd8-da0d-4e32-bbe7-ac2d5f7eadbf)

jkppr:
Thanks @JohannesLks for adding this to the frontend-ng UI. As discussed, it would be amazing if you could also add the SketchAccessDenied logic to the new https://github.com/google/timesketch/tree/master/timesketch/frontend-v3 UI.

JohannesLks:
@jkppr Work in Progress :)

PR #3651, opened 2026-01-02
feat: Add dedicated 403 Access Denied page for sketches

JohannesLks:
When accessing a sketch without permission, users now see a dedicated Access Denied page instead of the misleading empty state page.

Changes:
- Add AccessDenied.vue component with lock icon and home button
- Add sketchAccessDenied state and mutation to store
- Detect 403 errors in updateSketch action
- Suppress 403 snackbar for sketch endpoints (dedicated page handles it)
- Conditionally render AccessDenied in Sketch.vue

Fixes: https://github.com/google/timesketch/issues/3327

**Checks**

- [X] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [ ] Documentation updated.


timesketch/frontend-ng/src/components/AccessDenied.vue
```diff
@@ -0,0 +1,42 @@
...
+limitations under the License.
+-->
+
+<template>
+  <v-container fill-height fluid>
+    <v-row align="center" justify="center">
+      <v-sheet class="pa-4" style="background: transparent">
+        <div class="text-center">
+          <v-icon size="80" color="grey">mdi-lock</v-icon>
+          <div style="font-size: 2em" class="mb-3 mt-3">Access Denied</div>
```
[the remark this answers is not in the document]
jkppr: Yeah, I don't mind. Let's skip this one.

timesketch/frontend-ng/src/utils/RestApiClient.js
```diff
@@ -51,6 +51,11 @@ RestApiClient.interceptors.response.use(
       return Promise.reject(error)
     }
 
+    // Skip snackbar for 403 errors on sketch endpoints - we show a dedicated Access Denied page instead
+    if (status === 403 && url.includes('/sketches/')) {
+      return Promise.reject(error)
+    }
+
```
jkppr: The logic to suppress the snackbar is too broad and will likely hide legitimate permission errors for actions taken inside a sketch.

This can also happen if a user has read access to a sketch but tries to perform a write action (e.g., Save a View, Delete a Timeline, Star an Event) and the backend denies it (403), this interceptor will catch it because those endpoints also contain /sketches/.
Since many components rely on the global `errorSnackBar` event (via this interceptor) to inform the user of failure, the user will click a button and nothing will happen, no error, no success.

Therefore I would opt to keep the errorSnackBar and just accept that the user gets the info twice rather than risk missing access denied errors in other situations. 

```suggestion

```

timesketch/frontend-ng/src/components/AccessDenied.vue
```diff
@@ -0,0 +1,42 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+-->
+
+<template>
+  <v-container fill-height fluid>
```
jkppr: Adding a height here so the message is placed on the same height as the empty sketch state message.

```suggestion
  <v-container fill-height fluid style="height: 70vh">
```

timesketch/frontend-ng/src/components/AccessDenied.vue
```diff
@@ -0,0 +1,42 @@
...
+    <v-row align="center" justify="center">
+      <v-sheet class="pa-4" color="transparent">
+        <div class="text-center">
+          <v-icon size="80" color="grey">mdi-lock</v-icon>
```
jkppr: Increase the icon size and add a lighter level of grey to align it with the empty state messaging optically. 

```suggestion
          <v-icon size="100" color="grey lighten-1" class="mb-3">mdi-lock</v-icon>
```

timesketch/frontend-ng/src/components/SketchAccessDenied.vue
jkppr, line 1: Can you rename this file to `SketchAccessDenied.vue` please?

jkppr:
@JohannesLks would you be open to also add this to the new frontend-v3 UI which is still under development? (Can be in a separate PR).

timesketch/frontend-ng/src/utils/RestApiClient.js
JohannesLks: Thanks, I initially wanted to clean this up, but I understand the concern and dont want to risk hiding legitimate errors

JohannesLks:
@jkppr 

> @JohannesLks would you be open to also add this to the new frontend-v3 UI which is still under development? (Can be in a separate PR).

Sure

timesketch/frontend-ng/src/views/Sketch.vue
```diff
@@ -378,6 +381,7 @@ import TsTimelinesTable from '../components/LeftPanel/TimelinesTable.vue'
 import TsQuestionCard from '../components/Scenarios/QuestionCard.vue'
 import TsSettingsDialog from '../components/SettingsDialog.vue'
 import TsInvestigation from '../components/LeftPanel/Investigation.vue'
+import TsAccessDenied from '../components/AccessDenied.vue'
```
jkppr:
```suggestion
import TsAccessDenied from '../components/SketchAccessDenied.vue'
```

jkppr (approved):

PR #3663, opened 2026-01-07
Fix HTTP 413 errors via multipart upload and dynamic chunking in ImportStreamer

jkppr:
This PR updates the `timesketch-import-client` to resolve persistent `HTTP 413 Payload Too Large` errors when uploading large datasets (CSV, JSONL). This issue arose due to the recent enforcement of `MAX_FORM_MEMORY_SIZE` (default 200MB) in Werkzeug/Flask, combined with the data expansion caused by URL-encoding large JSON payloads.

**Key Changes:**

1.  **Switch to `multipart/form-data`:**
    *   Changed the transport method in `_upload_data_frame` and `_upload_data_buffer` from `application/x-www-form-urlencoded` to `multipart/form-data`.
    *   **Reasoning:** URL-encoding JSON data increases payload size by approximately 2x-3x. Multipart allows sending the JSON string as a raw file part, eliminating encoding overhead and ensuring the on-wire size matches the actual data size.

2.  **Dynamic Recursive Chunking:**
    *   Implemented logic to pre-calculate the byte size of the JSON payload before transmission.
    *   If a chunk exceeds the configured server limit, the client now recursively splits the batch in half until it fits within the limit.
    *   This replaces the reliance on a static row count (default 50,000), which was unreliable for datasets containing large individual events.

3.  **Configurable Payload Limits:**
    *   Added a `--max-payload-size` argument to the CLI (`timesketch_importer`) and a corresponding setter in `ImportStreamer`.
    *   Defaults to 200MB (matching Werkzeug defaults) but allows administrators to align the client with custom server configurations.

4.  **Optimization & Cleanup:**
    *   Removed legacy `time.sleep(2)` calls in the upload loop. Retries and backoff are already handled robustly by the `urllib3` `HTTPAdapter` in the API client session.
    * Adding the default `MAX_FORM_MEMORY_SIZE` value of 200Mb to the app.py to handle situations where the config is not updated.

**Impact:**
*   Eliminates false-positive 413 errors for chunks that are physically smaller than the server limit but inflated by encoding.
*   Improves upload performance by removing client-side encoding and server-side decoding CPU overhead.
*   Ensures large uploads are robust against memory limits regardless of individual row sizes.

### DIFF
--- a/importer_client/python/timesketch_import_client/importer.py
+++ b/importer_client/python/timesketch_import_client/importer.py
@@ -1019,18 +1019,18 @@ class ImportStreamer(object):
         """Set the timestamp description field."""
         self._timestamp_desc = description
 
-    def set_max_payload_size(self, size_in_bytes):
+    def set_max_payload_size(self, size_in_bytes: int) -> None:
         """Set the maximum payload size allowed by the server.
 
         Args:
             size_in_bytes (int): The server limit (e.g. MAX_FORM_MEMORY_SIZE).
         """
+        if size_in_bytes <= 0:
+            raise ValueError(f"Payload size must be positive, got {size_in_bytes}")
         self._max_payload_size = size_in_bytes
-        # Recalculate safe limit
-        self._safe_payload_limit = self._max_payload_size - self.PAYLOAD_SAFETY_BUFFER
-        if self._safe_payload_limit <= 0:
-            # Fallback if someone sets a ridiculously low limit
-            self._safe_payload_limit = 1024 * 1024  # 1MB minimum
+        self._safe_payload_limit = max(
+            (self._max_payload_size - self.PAYLOAD_SAFETY_BUFFER), 1048576
+        )
 
     @property
     def state(self):


timesketch/app.py
```diff
@@ -115,6 +115,8 @@ def create_app(
     # This is needed to support large form uploads (e.g. from import client)
     if "MAX_FORM_MEMORY_SIZE" in app.config:
         app.request_class.max_form_memory_size = app.config["MAX_FORM_MEMORY_SIZE"]
+    else:
+        app.request_class.max_form_memory_size = 209715200
```
jaegeral:
```suggestion
    app.request_class.max_form_memory_size = app.config.get("MAX_FORM_MEMORY_SIZE", 209715200)
```

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -20,17 +21,16 @@
...
 import uuid
 
 import numpy
 import pandas
-
-from timesketch_api_client import timeline
 from timesketch_api_client import definitions
+from timesketch_api_client import timeline
 from timesketch_api_client.error import UnableToRunAnalyzer
 from timesketch_import_client import utils
```
jaegeral, line 32:
```suggestion

from typing import Union
```

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -906,6 +1019,19 @@ def set_timestamp_description(self, description):
         """Set the timestamp description field."""
         self._timestamp_desc = description
 
+    def set_max_payload_size(self, size_in_bytes):
```
jaegeral:
```suggestion
    def set_max_payload_size(self, size_in_bytes: int) -> None:
```

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -906,6 +1019,19 @@ def set_timestamp_description(self, description):
...
+        """Set the maximum payload size allowed by the server.
+
+        Args:
+            size_in_bytes (int): The server limit (e.g. MAX_FORM_MEMORY_SIZE).
+        """
+        self._max_payload_size = size_in_bytes
```
jaegeral, line 1030:
```suggestion
        if size_in_bytes <= 0:
            raise ValueError(f"Payload size must be positive, got {size_in_bytes}")
        self._max_payload_size = size_in_bytes
```

in case someone provides a negative number

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -906,6 +1019,19 @@ def set_timestamp_description(self, description):
...
+        # Recalculate safe limit
+        self._safe_payload_limit = self._max_payload_size - self.PAYLOAD_SAFETY_BUFFER
+        if self._safe_payload_limit <= 0:
+            # Fallback if someone sets a ridiculously low limit
+            self._safe_payload_limit = 1024 * 1024  # 1MB minimum
```
jaegeral:
```suggestion
        self._safe_payload_limit = max(calculated_limit, 1048576)
```

will pick the larger one of the two and with the check above we already have the negative covered. WDYT?

importer_client/python/timesketch_import_client/importer.py
```diff
@@ -289,42 +297,93 @@ def _reset(self):
...
+        # 1. Determine which data to use (Instance buffer or recursive argument)
+        if data_lines is None:
+            data_lines = self._data_lines
+
+        if not data_lines:
             return None
 
+        # 2. Serialize to check size
+        # We construct the JSON string here based on the local variable
+        events_string = "\n".join([json.dumps(x) for x in data_lines])
```
jaegeral, line 326: Is there a faster way to estimate the size? because if I am understanding it right, this is a recursive call, so you dump 100 % in the first call, then you split it in half eventually, to parse the first 50 % that you already parsed again and the second 50 % again etc etc?

jkppr: It's a tradeoff I was willing to accept. The reliability and speed we get via the other optimizations should equal this out (also since the recursion should only trigger rarely).

importer_client/python/timesketch_import_client/importer.py
jkppr: Where is this used?

jaegeral: I introduced typing later down below, that is why I would suggest it here

PR #3666, opened 2026-01-09
fix(e2e): test messages for importing file delay

jaegeral:
Improving the text and retry logic when importing a file into a timeline for e2e tests.

This improvement contains the filenames for log messages, but since this is only for tests it should be.


end_to_end_tests/interface.py
```diff
@@ -107,22 +107,32 @@ def import_timeline(self, filename, index_name=None, sketch=None):
         retry_count = 0
         while True:
             if retry_count >= max_retries:
-                raise TimeoutError
+                raise RuntimeError(
```
jkppr, line 110: Why change this from TimeoutError to RuntimeError? If the request took too many retries isn't this kind of a timeout?

jaegeral: Yeah I was on the edge for it since the connection did not really time out but something during runtime errored out. Happy to change it back...

PR #3671, opened 2026-01-09
feat: Add dialog for CSV export limit in Explore view

jkppr:
The Timesketch Explore View allows to download events as CSV. However, due to the way OpenSearch pagination works (10k events per page) triggering this download for more than 10k events has proven to be unreliable and brittle. Therefore I propose an UX improvement to limit the UI download of CSV to max 10k events and for views with more than 10k to point the user to the CLI client documentation that can be used to download all events.

- Checks if total hits > 10000 before exporting.
- Shows a dialog pointing to CLI client if limit is exceeded.
- Added unit tests for EventList export logic.

image

### DIFF
--- a/timesketch/api/v1/resources/upload.py
+++ b/timesketch/api/v1/resources/upload.py
@@ -97,7 +97,11 @@ class UploadFileResource(resources.ResourceMixin, Resource):
         if data_label in ("csv", "json", "jsonl"):
             data_label = "csv_jsonl"
 
-        indices = [t.searchindex for t in sketch.active_timelines]
+        indices = (
+            t.searchindex
+            for t in sketch.timelines
+            if t.get_status.status not in ("deleted", "archived")
+        )
         for index in indices:
             if index.has_label(data_label) and sketch.has_permission(
                 permission="write", user=current_user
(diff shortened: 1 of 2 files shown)


timesketch/frontend-ng/src/components/Explore/EventList.vue
```diff
@@ -970,6 +991,10 @@ export default {
         })
     },
     exportSearchResult: function () {
+      if (this.totalHits > 10000) {
```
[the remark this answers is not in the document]
jkppr: In this case I consider the magic number as okay.

timesketch/frontend-ng/src/components/Explore/EventList.test.js
```diff
@@ -0,0 +1,143 @@
...
+          es_total_count_complete: 0
+        },
+        objects: []
+      }
+    })
+
+    // Call exportSearchResult
+    await wrapper.vm.exportSearchResult()
+
+    expect(ApiClient.exportSearchResult).toHaveBeenCalled()
```
Annoraaq, line 109: Maybe we can explicitly assert here that the dialog is **not** shown?

Annoraaq (approved):
LGTM

timesketch/frontend-ng/src/components/Explore/EventList.test.js
jkppr: Good point. Added it.

PR #3690, opened 2026-01-28
feat: better logging in `timesketch/lib/tasks.py:set_timeline_status`

jaegeral:
Updated `_set_timeline_status` in `tasks.py` to include the `sketch.id` in all log messages (successes, status updates, and index refresh failures). This provides much-needed context when troubleshooting background processing issues.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,8 +25,6 @@ import glob
 import os
 import sys
 
-import pkg_resources
-
 from setuptools import find_packages
 from setuptools import setup
 
@@ -52,13 +50,11 @@ def parse_requirements_from_file(path):
       str: package resource requirement.
     """
     with open(path, "r", encoding="utf-8") as file_object:
-        file_contents = file_object.read()
-    for req in pkg_resources.parse_requirements(file_contents):
-        try:
-            requirement = str(req.req)
-        except AttributeError:
-            requirement = str(req)
-        yield requirement
+        for line in file_object:
+            line = line.split("#", 1)[0].strip()
+            if not line:
+                continue
+            yield line
 
 
 timesketch_description = (


timesketch/lib/tasks.py
```diff
@@ -275,20 +280,42 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
...
     datastore = OpenSearchDataStore()
     # Retry refreshing the index a few times if it fails.
+    index_name = timeline.searchindex.index_name
     for i in range(5):
         try:
-            datastore.client.indices.refresh(index=timeline.searchindex.index_name)
+            datastore.client.indices.refresh(index=index_name)
             break  # Success
-        except NotFoundError:
+        except Exception as e:  # pylint: disable=broad-except
```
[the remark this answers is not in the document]
jaegeral: no

timesketch/lib/tasks.py
```diff
@@ -275,20 +280,42 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
     db_session.add(timeline)
     db_session.commit()
 
+    # Defensive check for orphaned timelines to prevent worker crashes.
+    sketch_id = timeline.sketch.id if timeline.sketch else 0
```
jkppr, line 283: How is this a "defensive check for orphaned timelines"? 

Isn't this just getting a sketch_id from the timeline so you can print it in the logs?

timesketch/lib/tasks.py
```diff
@@ -275,20 +280,42 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
...
+
+    logger.info(
```
jkppr: Are you sure this should be informational logging? Wouldn't this spam our logs with each change of a status? This feels like something that should be debugging level to me.

timesketch/lib/tasks.py
```diff
@@ -275,20 +280,42 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
...
+                    "Unable to refresh index: %s in sketch (ID: %d). "
+                    "Gave up after 5 attempts. Error: %s",
+                    index_name,
+                    sketch_id,
+                    str(e),
+                    exc_info=True,
                 )
             else:
+                # Show error message for attempts 0-4 only if debug is enabled
+                logger.debug(
```
jkppr, line 312: why not include the sketch I'd here too?

jkppr (commented):
please see below

timesketch/lib/tasks.py
jaegeral: yes

timesketch/lib/tasks.py
jaegeral: yes

timesketch/lib/tasks.py
jaegeral: that is true, let me change it

PR #3699, opened 2026-02-09
draft: refactoring of event annotation and e2e tests

jaegeral:
This is a draft for now to work on the annotate events class


jaegeral:
Fwiw this is something I might work in the future again, for now it can remain in draft.

PR #3702, opened 2026-02-09
Fix chunked upload inconsistent file size by using random access write

jkppr:
The current upload implementation used `open(..., "ab")` for writing uploaded chunks. In "ab" (append binary) mode, `seek()` operations are ignored, and all writes are appended to the end of the file. This caused issues when:
1. Chunks were retried (duplicate data appended).
2. Chunks arrived out-of-order (data appended instead of placed at correct offset).

This resulted in "inconsistent file size" errors and corrupted files.

This change switches to `os.open(..., os.O_RDWR | os.O_CREAT, 0o600)` and `os.fdopen(..., "rb+")`, which allows proper seeking and overwriting of data at the specified `chunk_byte_offset`. This makes chunk uploads idempotent and robust against out-of-order delivery.

Added regression tests in `timesketch/api/v1/resources_test.py`.

### DIFF
--- a/docker/dev/build/docker-entrypoint.sh
+++ b/docker/dev/build/docker-entrypoint.sh
@@ -75,13 +75,6 @@ if [ "$1" = 'timesketch' ]; then
   # Add web user
   tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}"
 
-  # Add Sigma rules
-  git clone https://github.com/SigmaHQ/sigma /usr/local/src/sigma
-  # for each line in sigma_rules.txt execute the command
-  for line in $(cat sigma_rules.txt); do
-    tsctl import-sigma-rules $line
-  done
-
   # Wrap up things
   echo "Timesketch development server is ready!"
 


timesketch/api/v1/resources/upload.py
```diff
@@ -477,9 +478,15 @@ def _upload_file(
...
         try:
-            with open(file_path, "ab") as fh:
-                fh.seek(chunk_byte_offset)
-                fh.write(file_storage.read())
+            fd = os.open(file_path, os.O_RDWR | os.O_CREAT, 0o600)
+            try:
+                with os.fdopen(fd, "rb+") as fh:
+                    fh.seek(chunk_byte_offset)
+                    fh.write(file_storage.read())
+            except Exception as e:  # pylint: disable=broad-exception-caught
```
jaegeral, line 494: would it make sense to check in the error and if it fails to attempt to use the old way of doing it?

jkppr: I wouldn't recommend that here. The old way (ab mode) was actually the cause of the bug because it ignores seek and forces an append to the end.

If the new method fails and we fallback to the old one, we'd just be silently writing data to the wrong offset (corrupting the file) instead of throwing an error. I think it's safer to fail hard here than to risk saving bad data.

jaegeral: ack, lets ship it

PR #1987, opened 2021-10-16
Import client: Support Plaso event filters

joachimmetz:

PR #3710, opened 2026-02-11
Feature: Add support for Plaso event filters in importer and backend

### LINKED ISSUE #1931: Support psort filters when ingesting into Timesketch
Some hosts produce very large plaso data sets. As an example, a domain controller produced nearly 18 million parsed events when processed with log2timeline. Often I don't want to ingest all of those events. Many of them are from a timeframe that is irrelevent. Also, the majority of parsed events from the DC are Windows event logs. The way log2timeline parses Windows event logs results in duplicate events: one for `Creation Time` and one for `Last Modification Time`. I understand where these come from, but 99.99% of the time I only care about the Creation Time events. So, I have a psort filter I can use that will output just the Windows event log Creation events and also filter to a time range of interest. This works great for psort output to CSV. However, I don't know of a way to do an equivalent filter when importing into Timesketch. It would be great to be able to use psort filters with `timesketch_importer`. 

Here is an example of a psort filter that will narrow down those 18 million events to under 2 million. This is what I'd like to replicate with `timesketch_importer` or a similar option.

```
psort.py --output-time-zone 'UTC' -o dynamic -w dc-triage.csv dc-triage.plaso "(((parser == 'winevtx') and (timestamp_desc == 'Creation Time')) or (parser != 'winevtx')) and ( date > datetime('2020-02-01T00:00:00'))"
```

The only other option I'm aware of is to use `timesketch_importer` or the webui to upload all the data and then go back and delete unwanted documents from the Elasticsearch index. For example, this will delete documents prior to 2021-02-01 and will delete any `winevtx` events with the `timestamp_desc` containing "Modification":

```
curl -XPOST --header 'Content-Type: application/json' localhost:9200/plaso-dc-triage-index/_delete_by_query -d '{"query": {"range": {"datetime": {"time_zone": "+00:00","lt": "2021-02-01T00:00:00"}}}}'
```

```
curl -XPOST --header 'Content-Type: application/json' localhost:9200/plaso-dc-triage-index/_delete_by_query -d '{"query" : {"bool" : {"must": [{"match": {"parser": "winevtx"}}, {"match": {"timestamp_desc": "Modification"}}]}}}'
```

**Additional context**
This issue is tangentially related to [Plaso issue #3813](https://github.com/log2timeline/plaso/issues/3813)

jkppr:
(This PR is a new attempt on #1931 and based on prior work by @joachimmetz in #1987 )

This PR introduces the ability to filter Plaso events during import via the `timesketch_importer` client. Users can now provide a standard Plaso filter string (e.g., `'data_type is "fs:stat"'`) which is propagated through the API and applied during the `psort.py` processing phase on the worker.

**Changes:**

* **Importer Client:**
  * Added `--plaso-event-filter` flag to the CLI tool.
  * Updated `ImportStreamer` to accept the filter and include it in the upload metadata payload.
  * Added validation to warn if a filter is provided for non-Plaso file types.

* **API:**
  * Updated `UploadFileResource` to extract the `plaso_event_filter` from the request form and pass it to the ingestion pipeline.

* **Backend Tasks:**
  * Updated `build_index_pipeline` and `run_plaso` signatures to accept the filter argument.

* **Security Hardening:** Refactored the `psort.py` command construction in `run_plaso`. The command list now uses the `--` delimiter to strictly separate flags from positional arguments (the file path and filter). This prevents argument injection attacks and resolves parsing issues where filters or file paths could be misinterpreted as flags.

closes #1931


jkppr:
> Generally fine, I wonder if it would be a nice thing to also mention that in `docs`?

Yes! Good point. Let me update the docs with this as well.

PR #3721, opened 2026-02-13
Fix & refactor dev docker-entrypoint.sh script

### LINKED ISSUE #3709: Hot reload timesketch.conf in development environment & avoid relogins
Having to `docker compose restart` everytime you make a change in timesketch.conf when developing new features etc. introduces a lot of toil. You have to relogin, restart celery, web, frontend etc.

I propose that in the development environment we hot reload timesketch.conf.

This would involve the following actions:

* Symlink /etc/timesketch/timesketch.conf to the mounted host file 
* Use Gunicorn's hot reload, update `tsdev.sh` to use `--reload-extra-file` pointing to the config which would trigger a worker restart whenever the config changes
*  Use environment variables for dynamic configuration (using app.config.from_prefixed_env()). By moving "infrastructure" settings, like the SECRET_KEY and database/OpenSearch connection strings into environment variables. Because we'd now be using symlinks, the container would normally use sed to overwrite values in the host's timesketch.conf. We can set a persistent SECRET_KEY (stored in the environment or a local .env file) so that the key is no longer randomly regenerated on every boot. This stops the "forced logout" every time the container is restarted. Plus we'd also keep a clean git state, cause we avoid timesketch.conf from being written to on every restart.

All of this only applies to the development environment.

jkppr:
* Refactor dev docker-entrypoint.sh to support a local persistent key per dev env preventing frequent re-authentication after each restart.
* Adding the local key to the gitignore file to prevent leakages in PRs
* Fix some format mismatches with sed commands on timesketch.conf
* Moving configs that don't need change to links instead of copies

closes #3709

### DIFF
--- a/docker/dev/build/docker-entrypoint.sh
+++ b/docker/dev/build/docker-entrypoint.sh
@@ -52,7 +52,6 @@ setup_config() {
   fi
 
   # Set up the OpenSearch connection
-  # FIX: Changed OPENSEARCH_ADDRESS to OPENSEARCH_HOST to match the if-check and docker-compose
   if [ "$OPENSEARCH_HOST" ] && [ "$OPENSEARCH_PORT" ]; then
     sed -i 's#OPENSEARCH_HOSTS = \[{"host": "opensearch", "port": 9200}\]#OPENSEARCH_HOSTS = [{"host": "'$OPENSEARCH_HOST'", "port": '$OPENSEARCH_PORT'}]#' /etc/timesketch/timesketch.conf
   fi
@@ -92,7 +91,7 @@ if [ "$1" = 'timesketch' ]; then
   setup_config
 
   # Add web user
-  tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}" || true
+  tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}"
 
   # Wrap up things
   echo "Timesketch development server is ready!"


docker/dev/build/docker-entrypoint.sh
```diff
@@ -23,42 +14,58 @@ setup_config() {
...
   ln -sf /usr/local/src/timesketch/data/llm_summarize /etc/timesketch/
+  ln -sf /usr/local/src/timesketch/data/regex_features.yaml /etc/timesketch/regex_features.yaml
+  ln -sf /usr/local/src/timesketch/data/winevt_features.yaml /etc/timesketch/winevt_features.yaml
+  ln -sf /usr/local/src/timesketch/data/tags.yaml /etc/timesketch/tags.yaml
+  ln -sf /usr/local/src/timesketch/data/intelligence_tag_metadata.yaml /etc/timesketch/intelligence_tag_metadata.yaml
+  ln -sf /usr/local/src/timesketch/data/plaso.mappings /etc/timesketch/plaso.mappings
+  ln -sf /usr/local/src/timesketch/data/generic.mappings /etc/timesketch/generic.mappings
+  ln -sf /usr/local/src/timesketch/data/ontology.yaml /etc/timesketch/ontology.yaml
+  ln -sf /usr/local/src/timesketch/data/data_finder.yaml /etc/timesketch/data_finder.yaml
+  ln -sf /usr/local/src/timesketch/data/bigquery_matcher.yaml /etc/timesketch/bigquery_matcher.yaml
```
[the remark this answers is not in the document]
jaegeral: I really like that

docker/dev/build/docker-entrypoint.sh
```diff
@@ -5,60 +5,75 @@ setup_config() {
...
-  if [ $POSTGRES_USER ] && [ $POSTGRES_PASSWORD ] && [ $POSTGRES_ADDRESS ] && [ $POSTGRES_PORT ]; then
+  if [ "$POSTGRES_USER" ] && [ "$POSTGRES_PASSWORD" ] && [ "$POSTGRES_ADDRESS" ] && [ "$POSTGRES_PORT" ]; then
     sed -i 's#postgresql://<USERNAME>:<PASSWORD>@localhost#postgresql://'$POSTGRES_USER':'$POSTGRES_PASSWORD'@'$POSTGRES_ADDRESS':'$POSTGRES_PORT'#' /etc/timesketch/timesketch.conf
   fi
 
   # Set up the OpenSearch connection
-  if [ $OPENSEARCH_HOST ] && [ $OPENSEARCH_PORT ]; then
-    sed -i 's#OPENSEARCH_HOST = \x27127.0.0.1\x27#OPENSEARCH_HOST = \x27'$OPENSEARCH_HOST'\x27#' /etc/timesketch/timesketch.conf
-    sed -i 's#OPENSEARCH_PORT = 9200#OPENSEARCH_PORT = '$OPENSEARCH_PORT'#' /etc/timesketch/timesketch.conf
+  # FIX: Changed OPENSEARCH_ADDRESS to OPENSEARCH_HOST to match the if-check and docker-compose
```
jaegeral: Can we rephrase that comment?

docker/dev/build/docker-entrypoint.sh
```diff
@@ -77,7 +92,7 @@ if [ "$1" = 'timesketch' ]; then
   setup_config
 
   # Add web user
-  tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}"
+  tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}" || true
```
jaegeral: what does the ```true``` change? I think just supressing errors is not ideal.

If the concern is that the user might already exist, we could do something like:


```suggestion
  # Check if the user exists by searching the user list
  if tsctl list-users | grep -q "${TIMESKETCH_USER}"; then
      echo "User ${TIMESKETCH_USER} already exists. Skipping creation."
  else
      echo "Creating user ${TIMESKETCH_USER}..."
      tsctl create-user --password "${TIMESKETCH_USER}" "${TIMESKETCH_USER}"
  fi
```

docker/dev/README.md
```diff
@@ -1,121 +1,100 @@
...
-### Non-interactive
-
-Running the following as a script after `docker compose up -d` will bring up the development environment in the background for you.
+OR
 
 ```bash
-export CONTAINER_ID="$(docker container list -f name=timesketch-dev -q)"
-docker exec $CONTAINER_ID celery -A timesketch.lib.tasks worker --loglevel info
-docker exec $CONTAINER_ID gunicorn --reload -b 0.0.0.0:5000 --log-file - --timeout 120 timesketch.wsgi:application
+bash utils/tsdev.sh web
```
jaegeral, line 57: Hm, I know what your intention with this is, but I would suggest we leave the current commands in, maybe mention tsdev.sh as another option (as tilt).

jaegeral: ah nevermind, I saw you did that below

jkppr: Correct, I just added them as additional option.

docker/dev/build/docker-entrypoint.sh
jkppr: I removed the true. Timesketch does not care if a user already exists and will just overwrite the PW. Does not matter for dev / dev

PR #3731, opened 2026-02-19
fix: Handling Large String Fields

jaegeral:
**Graceful Handling of Large Fields (>32KB)**
   * Problem: OpenSearch has a hard limit of 32,766 bytes for keyword fields. When importing data with
     massive strings (like the ~38KB FeatureVector field), OpenSearch would reject the entire event with
     a max_bytes_length_exceeded_exception.
   * Solution: Updated data/generic.mappings to include a dynamic_template with ignore_above.
       * Large strings are still indexed as `text` (keeping them searchable).
       * The `keyword` sub-field (used for charts/sorting) is skipped for values exceeding 32k
         characters, preventing the import from crashing.

New End-to-End Test
   * Addition: Added test_large_field_upload_jsonl to end_to_end_tests/upload_test.py.
   * Verification: This test programmatically generates a JSONL file with a 38,000-character field,
     uploads it, and asserts that the timeline is successfully indexed and the data remains searchable.
   `test_field_length_limit_searchability`:
       * Under Limit (32,760 chars): Verifies that fields just below the 32,766-byte threshold are fully
         searchable via exact matching (TestField:"value") and text search.
       * Over Limit (32,770 chars): Verifies that fields just above the threshold are successfully
         imported (no document rejection) and remain searchable via text/wildcard queries
         (TestField:prefix*), even though the .keyword sub-field is skipped by OpenSearch.

   `test_extreme_field_length_upload`:
       * Extreme Size (100KB): Verifies system stability when handling very large "blob" fields. It
         ensures that 100KB fields are imported without data corruption and are searchable by other
         metadata (like the message field).

  Documentation Updates
   * User Guide: Updated docs/guides/user/import-from-json-csv.md with a new section explaining the 32KB
     limit and how Timesketch handles it.

### DIFF
--- a/docs/guides/user/import-from-json-csv.md
+++ b/docs/guides/user/import-from-json-csv.md
@@ -55,14 +55,17 @@ Similar to a CSV file, you can upload the JSONL file even if it does not have th
 
 ## Handling Large String Fields
 
-OpenSearch has a hard limit of **32,766 bytes** for "keyword" fields. Keyword fields are used for exact matching, sorting, and aggregations (charts).
+OpenSearch has a hard limit of **32766 bytes** for "keyword" fields. Keyword fields are used for exact matching, sorting, aggregations (charts) and regex searches.
 
 If an imported file contains a field that exceeds this limit (e.g., a very long feature vector or a massive blob of text), OpenSearch would normally reject the entire event with a `max_bytes_length_exceeded_exception`.
 
-Timesketch handles this by using a `dynamic_template` with an `ignore_above: 256` setting. This means:
-*   Fields longer than 256 characters are still indexed as **text** (and are therefore searchable).
+Timesketch handles this by using a `dynamic_template` with an `ignore_above: 32766` setting. This means:
+*   Fields larger than 32766 bytes are still indexed as **text** type (and are therefore searchable).
 *   However, the **keyword** sub-field is not created for those specific long values.
-*   **Impact:** You can search for the content of these large fields, but you cannot use them in aggregations (charts) or sort by them if they exceed the limit.
+*   **Impact:**
+    *   You can search for the content of these large fields based on the **text** type.
+    *   You cannot use them in aggregations (charts) or sort by them if they exceed the limit.
+    *   You cannot search the keyword specific field type anymore (e.g. `message.keyword:/.*<regex>.*/`)
 
 ## Upload the file to Timesketch
 


end_to_end_tests/upload_test.py
```diff
@@ -404,5 +406,55 @@ def test_plaso_import(self):
...
+                "message": unique_message,
+                "data_type": "test:large_field",
+                "LargeField": large_field_value,
+                "id": f"test-uuid-{rand}",
+            }
+            file_object.write(json.dumps(line_data) + "\n")
+
+        # Import the timeline
+        self.import_timeline(file_path, sketch=sketch)
+        os.remove(file_path)
```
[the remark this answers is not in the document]
jaegeral: done

data/generic.mappings
```diff
@@ -1,5 +1,21 @@
...
+  "dynamic_templates": [
+    {
+      "strings": {
+        "match_mapping_type": "string",
+        "mapping": {
+          "type": "text",
+          "fields": {
+            "keyword": {
+              "type": "keyword",
+              "ignore_above": 256
```
jkppr: This will lead to unexpected results with analysts since some fields can be searched with wildcards and others not. (TS does some behind the scenes adjustments on what field it searches, see the data store lib).

I recommend either setting this to its max possible (hard upper limit) by default or leave it out. 

In general the idea with the mappings file is to allow every admin to use a setting that works for them.

jaegeral: okidoki, I will check a little more, lets discuss it next week together

jaegeral: and I will add some more test cases for it

jaegeral:
ok tests passed, lets review next week

docs/guides/user/import-from-json-csv.md
```diff
@@ -53,6 +53,17 @@ Unlike JSON files, imports in JSONL format can be streamed from disk, making the
 
 Similar to a CSV file, you can upload the JSONL file even if it does not have the required headers. However, the uploaded file should have those headers whose semantic meaning is correlated to the ones of `message`, `datetime` and `timestamp_desc`.
 
+## Handling Large String Fields
+
+OpenSearch has a hard limit of **32,766 bytes** for "keyword" fields. Keyword fields are used for exact matching, sorting, and aggregations (charts).
```
jkppr: Can we link to a source in the OpenSearch docs here?

docs/guides/user/import-from-json-csv.md
```diff
@@ -53,6 +53,17 @@ Unlike JSON files, imports in JSONL format can be streamed from disk, making the
...
+
+If an imported file contains a field that exceeds this limit (e.g., a very long feature vector or a massive blob of text), OpenSearch would normally reject the entire event with a `max_bytes_length_exceeded_exception`.
+
+Timesketch handles this by using a `dynamic_template` with an `ignore_above: 256` setting. This means:
```
jkppr:
```suggestion
Timesketch handles this by using a `dynamic_template` with an `ignore_above: 32766` setting. This means:
```

Also below, please update to match the value in the config used. 

Maybe also add a line about why we use a default that is the max value. (To support solid regex & substring search on the full text)

docs/guides/user/import-from-json-csv.md
```diff
@@ -53,6 +53,17 @@ Unlike JSON files, imports in JSONL format can be streamed from disk, making the
...
+*   Fields longer than 256 characters are still indexed as **text** (and are therefore searchable).
+*   However, the **keyword** sub-field is not created for those specific long values.
+*   **Impact:** You can search for the content of these large fields, but you cannot use them in aggregations (charts) or sort by them if they exceed the limit.
```
jkppr: Also can't use the `field.keyword:*<searchterm>*` method for substring searches or `field.keyword:/.*<regex>.*/`.

jkppr (approved):
LGTM but please fix the docs as mentioned in the comments below.

docs/guides/user/import-from-json-csv.md
jkppr: Looks like there are none...

PR #3738, opened 2026-02-21
Deprecate store_aggregation and update docs

p-nowodzinski:
Deprecate store_aggregation() function

Added DeprecationWarning to store_aggregation() and updated notebook demo to remove a reference to calling store_aggregation()

Changed "You can also store your own aggregations using gd_sketch.store_aggregation function" to "You can also store your own aggregations using gd_sketch.save function".

Guides users to use Aggregation.save() instead


jkppr:
@jaegeral can you take a look into this?

notebooks/colab-timesketch-demo.ipynb
jaegeral, line 1: I think it is okay to not touch the notebook.

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1348,15 +1349,21 @@ def store_aggregation(
         Returns:
           A stored aggregation object or None if not stored.
         """
-        if self.is_archived():
-            raise RuntimeError("Unable to store an aggregator on an archived sketch.")
+        warnings.warn(
```
jaegeral: What did we actually gain if you raise a warning instead of a logger message? I think we would be better of to remove the method all together. Or remove the rest of the code.

jaegeral (commented):
PTAL

jaegeral:
Hey @p-nowodzinski , any thoughts on the PR? I might lean towards just removing the method tbh

p-nowodzinski:
Hi, apologies for the delayed response and thanks for the feedback.

That makes sense as I initially suggested deprecation in case you preferred to keep the function, but as it doesn’t appear to be used elsewhere, removing it would help simplify the codebase. I'm happy to proceed with removing store_aggregation(). Let me know your preference and I’ll update the PR accordingly.

jaegeral:
Yeah remove it

PR #3785, opened 2026-03-26
dependency: upgrade PyJWT to 2.12.0

jkppr:
PyJWT versions <= 2.11.0 do not validate the 'crit' (Critical) Header Parameter as defined in RFC 7515 §4.1.11. This violates the requirement that a JWS must be rejected if it contains extensions listed in 'crit' that are not understood by the recipient.

This change:
- Upgrades PyJWT to 2.12.0 in requirements.txt.
- Updates `timesketch/lib/google_auth.py` to catch `InvalidTokenError` during unverified header parsing. PyJWT 2.12.0 now raises this exception if the 'crit' header contains unsupported extensions.
- Adds a unit test in `google_auth_test.py` to verify that tokens with unknown critical extensions are correctly rejected.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -20,7 +20,7 @@ numpy==1.26.4
 scipy>=1.10.0
 oauthlib==3.3.1
 pandas==2.2.3
-PyJWT==2.4.0
+PyJWT==2.12.0
 python_dateutil==2.9.0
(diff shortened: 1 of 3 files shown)


jaegeral (approved):
ack

PR #3795, opened 2026-05-04
feat: Initial contribution for opentelemetry

jaegeral:
This Pull Request introduces comprehensive OpenTelemetry (OTel) instrumentation to Timesketch, enabling
  distributed tracing across the web API, background workers, and analyzers. The implementation is
  architecturally aligned with OpenRelik (https://github.com/openrelik/openrelik-common/blob/main/openrelik_common/telemetry.py)

1. Framework-Wide Tracing
   * Flask (API): Automatically captures spans for all incoming HTTP requests, including route patterns,
     status codes, and user agents.
   * Celery (Workers): Instruments task producers and workers, ensuring trace context is propagated
     across process boundaries.
   * Base Analyzer: Integrated OTel into the common analyzer interface, meaning every analyzer (Sigma,
     Sketch, etc.) now automatically reports sketch_id, analyzer_name, and execution status.

2. Local Dev Experience
   * Integrated Stack: Added a local Jaeger UI and an OTel Collector gateway to both docker-compose.yml
     and the Tiltfile.
   * Visualization: Traces can be viewed locally at http://localhost:16686/jaeger.
   
How to Verify
   1. Start the dev environment (docker-compose up or tilt up).
   2. Install dependencies: docker exec timesketch-dev pip install -r requirements.txt.
   3. Trigger any API call or Analyzer.
   4. View the trace waterfall in the local Jaeger UI.

### DIFF
--- a/contrib/Tiltfile
+++ b/contrib/Tiltfile
@@ -10,6 +10,8 @@ dc_resource("opensearch", labels=["docker"])
 dc_resource("postgres", labels=["docker"])
 dc_resource("redis", labels=["docker"])
 dc_resource("prometheus", labels=["docker"])
+dc_resource("otel-collector", labels=["docker"])
+dc_resource("jaeger", labels=["docker"], links=["http://localhost:16686/jaeger"])
 
 # --- 2. Managed Dev Processes ---
 backend_deps = [
--- a/docker/dev/docker-compose-telemetry.yml
+++ b/docker/dev/docker-compose-telemetry.yml
@@ -1,0 +1,26 @@
+services:
+  otel-collector:
+    container_name: otel-collector
+    image: otel/opentelemetry-collector:0.100.0
+    command: ["--config=/etc/otel-collector-config.yaml"]
+    volumes:
+      - ./otel-collector-config.yaml:/etc/otel-collector-config.yaml
+    ports:
+      - "127.0.0.1:4317:4317"
+      - "127.0.0.1:4318:4318"
+    depends_on:
+      - jaeger
+
+  jaeger:
+    container_name: jaeger
+    image: jaegertracing/all-in-one:latest
+    restart: always
+    ports:
+      - "127.0.0.1:16686:16686"
+    environment:
+      - COLLECTOR_OTLP_ENABLED=true
+
+networks:
+  default:
+    name: timesketch-network
+    external: true
--- a/docker/dev/docker-compose.yml
+++ b/docker/dev/docker-compose.yml
@@ -29,9 +29,10 @@ services:
       - CHOKIDAR_USEPOLLING=true
       - PROMETHEUS_MULTIPROC_DIR=/tmp/
       - ENABLE_STRUCTURED_LOGGING=true
-      - TIMESKETCH_OTEL_MODE=otlp-grpc
-      - TIMESKETCH_OTLP_GRPC_ENDPOINT=otel-collector:4317
-      - TIMESKETCH_OTLP_INSECURE=true
+      # Telemetry settings (Uncomment to enable locally)
+      # - TIMESKETCH_OTEL_MODE=otlp-grpc
+      # - TIMESKETCH_OTLP_GRPC_ENDPOINT=otel-collector:4317
+      # - TIMESKETCH_OTLP_INSECURE=true
     restart: always
     volumes:
       - ../../:/usr/local/src/timesketch/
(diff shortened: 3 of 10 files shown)


contrib/Tiltfile
```diff
@@ -10,6 +10,8 @@ dc_resource("opensearch", labels=["docker"])
 dc_resource("postgres", labels=["docker"])
 dc_resource("redis", labels=["docker"])
 dc_resource("prometheus", labels=["docker"])
+dc_resource("otel-collector", labels=["docker"])
+dc_resource("jaeger", labels=["docker"], links=["http://localhost:16686/jaeger"])
```
jaegeral, line 14: I am 50 / 50 on letting the otel-collector and jaeger in there as well as in the docker file, happy to hear feedback.

timesketch/lib/telemetry.py
```diff
@@ -0,0 +1,181 @@
...
+            endpoint=endpoint, insecure=insecure
+        )
+    elif otel_mode == "otlp-http":
+        endpoint = os.environ.get(
+            "TIMESKETCH_OTLP_HTTP_ENDPOINT", "http://localhost:4318/v1/traces"
+        )
+        trace_exporter = http_exporter.OTLPSpanExporter(endpoint=endpoint)
+    elif otel_mode == "otlp-cloud-trace":
+        # Direct OTLP to GCP Cloud Trace
+        # See: go/ct:writing-spans-from-borg-1p-howto
```
rgayon: Not sure if the go/ link is relevant

Did you test this cloud-trace enabled version? If not, maybe don't include it here?

jaegeral: not tested yet, removed the go link

contrib/Tiltfile
jaegeral: now that it is a seperate docker file it should be fine

rgayon (approved):

timesketch/lib/telemetry.py
jaegeral: ok tested now

rgayon (approved):

docker/dev/docker-compose-telemetry.yml
jkppr, line 1: Why a separate compose file and not adding it to the existing one?

docker/dev/docker-compose.yml
```diff
@@ -28,7 +28,11 @@ services:
       - TIMESKETCH_PASSWORD=dev
       - CHOKIDAR_USEPOLLING=true
       - PROMETHEUS_MULTIPROC_DIR=/tmp/
-      # - ENABLE_STRUCTURED_LOGGING=true
+      - ENABLE_STRUCTURED_LOGGING=true
```
jkppr, line 31: Why enable structured logging by default? That is usually only needed for GKE setups and not local dev?

docker/dev/otel-collector-config.yaml
jkppr, line 1: Is this a config only for dev setups or also for deployments? If the later, should we move it to data/ instead?

timesketch/lib/analyzers/interface.py
```diff
@@ -1147,9 +1148,17 @@ def run_wrapper(self, analysis_id):
...
+            telemetry.add_attribute_to_current_span("timeline_id", self.timeline_id)
+            telemetry.add_event_to_current_span(f"Starting analyzer: {self.name}")
+
             result = self.run()
             analysis.set_status("DONE")
-        except Exception:  # pylint: disable=broad-except
+
+            telemetry.add_attribute_to_current_span("status", "success")
+            telemetry.add_event_to_current_span(f"Analyzer {self.name} completed")
+        except Exception as e:  # pylint: disable=broad-except
```
jkppr, line 1161: So when telemetry stuff fails, the full analysis is marked as failed. Should this be distinguished?

timesketch/lib/analyzers/interface.py
jkppr, line 1: Why do we need the specific changes in this file? Isn't this part of the celery framework level traces already?

If this needs to be included on this level, what about import tasks that also run in celery? (see tasks.py)

docker/dev/docker-compose-telemetry.yml
jaegeral: @rgayon had a point that if you want to run bare minimum dev environment, you do not need telemetry stuff and or if you want to ship to Cloud telemetry, you do not need it either.

timesketch/lib/analyzers/interface.py
jaegeral: Not all of it is already part of celery, some info is specific and useful e.g. the sketch id. For other things like import tasks, I would do it follow up PRs, I just wanted to get started somewhere to keep it small.

timesketch/lib/analyzers/interface.py
jaegeral: obviously, telemetry stuff would never fail (/ironie) - will look into it, I will create an issue for it

docker/dev/docker-compose-telemetry.yml
jkppr: We have the same for notebooks: https://github.com/google/timesketch/blob/master/docker/dev/docker-compose.yml#L102

We just commented those our for now. Default is: No traces for the beginning and who wants it just removes the comments.

timesketch/lib/analyzers/interface.py
jkppr: Can you look into how much more or less work it is to add the missing information to the celery task context instead of adding custom telemetry to the analyzers?

docker/dev/docker-compose.yml
jaegeral: disabled in the follow up PR

docker/dev/otel-collector-config.yaml
jaegeral: https://github.com/google/timesketch/pull/3797/changes#r3193694689

docker/dev/docker-compose-telemetry.yml
jaegeral: https://github.com/google/timesketch/issues/3800

timesketch/lib/analyzers/interface.py
jaegeral: we do not have the information about analyzer tasks specifics on the celery level, IMHO those 13 lines or so here for analyzers and potentially a few on the importer side are worth it. Of course we get some basic for every celery job, but not really detailed ones. But I would suggest to move that conversation offline and get some exp first with how useful it is for analyzers before making any decisions.

PR #3797, opened 2026-05-05
feat(otel): Otel draft (likely obsolete)

jaegeral:
This Pull Request implements Phase 2 of the OpenTelemetry instrumentation for Timesketch. The focus of this phase is extending distributed tracing into the data layer (OpenSearch and PostgreSQL)

### Data Layer Visibility
- **OpenSearch:** Implemented manual instrumentation for `search()` and `count()` methods. This captures targeted indices and internal processing latency (`took_ms`) without exposing the query content.

Replaced by: https://github.com/google/timesketch/pull/3839


timesketch/lib/datastores/opensearch.py
```diff
@@ -837,151 +838,178 @@ def search(
...
-            scroll_timeout = "1m"  # Default to 1 minute scroll timeout
-
-        # Exit early if we have no indices to query
-        if not indices:
-            return {"hits": {"hits": [], "total": 0}, "took": 0}
-
-        # Make sure that the list of index names is uniq.
-        indices = list(set(indices))
+        tracer = telemetry.get_tracer(__name__)
+        with tracer.start_as_current_span("opensearch.search") as span:
```
rgayon: Does the code still run as expected if there is no tracing enbled?

docker/dev/docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
       - TIMESKETCH_PASSWORD=dev
       - CHOKIDAR_USEPOLLING=true
       - PROMETHEUS_MULTIPROC_DIR=/tmp/
-      - ENABLE_STRUCTURED_LOGGING=true
+      # - ENABLE_STRUCTURED_LOGGING=true
```
jaegeral, line 31: https://github.com/google/timesketch/pull/3795#discussion_r3193211593

data/otel-collector-config.yaml
jaegeral, line 1: https://github.com/google/timesketch/pull/3795#discussion_r3193217480

timesketch/lib/datastores/opensearch.py
```diff
@@ -27,6 +27,9 @@
 from typing import Generator, List, Dict, Optional, Any, Union
 
 from dateutil import parser, relativedelta
+from flask import abort
```
jaegeral, line 30: just changing the order of imports to make lint happy

jaegeral:
Had `timesketch/models/acl.py:324:4: W9016: "permission, user" missing in parameter type documentation (missing-type-doc)`

docs/OpenTelemetry.md
```diff
@@ -33,6 +30,7 @@ Telemetry is controlled entirely via environment variables.
 | `TIMESKETCH_OTLP_HTTP_ENDPOINT` | OTLP collector endpoint (HTTP). | `http://jaeger:4318/v1/traces` |
 | `TIMESKETCH_OTLP_INSECURE` | Use insecure (non-TLS) connection. | `true` (default for dev) |
 | `TIMESKETCH_ENV` | Environment identifier. | `production`, `development` |
+| `ENABLE_STRUCTURED_LOGGING` | Enable JSON logging with trace context. | `true`, `false` |
```
jkppr: Is that true? Does this affect the traces too? Or is this only for the general logging we do?

timesketch/lib/datastores/opensearch.py
jkppr, line 1: Are there any opentelemetry modules we can enable on our Opensearch cluster to get better insights from the other end too?

timesketch/lib/telemetry.py
```diff
@@ -31,13 +32,141 @@
...
+        "secret",
+        "key",
+        "session",
+        "cookie",
+        "auth",
+        "credential",
+    ]
+
+    # Attributes that are explicitly exempt from PII redaction (e.g. analyst identity)
+    EXEMPT_PII_ATTRIBUTES = {"user.name", "user.id", "timesketch.user_id"}
```
jkppr: It looks like this is never used. Should it?

timesketch/lib/telemetry.py
```diff
@@ -31,13 +32,141 @@
...
+                # 1. If it's a credential key, redact the whole thing
+                if is_credential_key:
+                    redacted_keys.append(key)
+                    # pylint: disable=protected-access
+                    span._attributes[key] = "[REDACTED]"
+                    continue
+
+                # 2. Check for sensitive keywords in the value
+                lower_val = value.lower()
+                if any(keyword in lower_val for keyword in SENSITIVE_KEYWORDS):
```
jkppr: Since the `SENSITIVE_KEYWORDS` contain pretty general stuff like `key` or `auth`, this will scrub data a bit too eagerly don't you think? e.g., searching for "windows_auth" or an error saying "index key_logs missing" will be redacted

Maybe limit value-based redaction to regex word boundaries or only run value-checks against suspicious keys rather than unconditionally across all string attributes. WDYT?

timesketch/lib/telemetry.py
```diff
@@ -152,7 +295,69 @@ def instrument_flask_app(app, **kwargs):
...
+def get_tracer(name: str):
+    """Returns a tracer instance.
+
+    Args:
+        name (str): The name of the tracer.
+
+    Returns:
+        opentelemetry.trace.Tracer: A tracer instance.
+    """
+    return trace.get_tracer(name)
```
jkppr, line 392: If I'm not mistaken this will lead to application crashes if `HAS_OTEL=False`. Either we need to return a dump tracker or ensure with other methods that the `with` blocks in the core application (e.g. ACL.py) don't crash.

timesketch/lib/telemetry.py
```diff
@@ -31,13 +32,141 @@
...
+                lower_key = key.lower()
+                is_credential_key = any(
+                    keyword in lower_key for keyword in SENSITIVE_KEYWORDS
+                )
+
+                # 1. If it's a credential key, redact the whole thing
+                if is_credential_key:
+                    redacted_keys.append(key)
+                    # pylint: disable=protected-access
+                    span._attributes[key] = "[REDACTED]"
```
jkppr: Why do we need to mutate a protected member here?

timesketch/models/acl.py
```diff
@@ -324,28 +325,36 @@ def has_permission(self, user, permission):
...
             permission.
         """
-        public_ace = self.is_public
-        if public_ace and permission == "read":
-            return public_ace
-        if isinstance(permission, bytes):
-            permission = codecs.decode(permission, "utf-8")
-        return self._get_ace(permission=permission, user=user)
+        tracer = telemetry.get_tracer(__name__)
+        with tracer.start_as_current_span("acl.has_permission") as span:
```
jkppr: If what your PR message says is true (one request can generate 40+ calls), doesn't this wrapper now create 40+ tiny spans per request which will result in a bloat of traces and load on the GCP or Jaeger end?
Do we need this level of granularity or could we also aggregate AACL checks into a single parent span?

Also: How does this impact memory usage of the Timesketch services? I imagine the trace tree per request with multiple users is rather large?

timesketch/lib/telemetry_test.py
```diff
@@ -0,0 +1,76 @@
...
+        exported_span = spans[0]
+        attrs = exported_span.attributes
+
+        # Assertions
+        self.assertEqual(attrs["password"], "[REDACTED]")
+        self.assertEqual(attrs["api_token"], "[REDACTED]")
+        self.assertEqual(attrs["custom_field"], "[REDACTED]")
+
+        # Verify analyst identity is NOT redacted
+        self.assertEqual(attrs["user.name"], "analyst@google.com")
```
jkppr, line 68: Wait, this is not redacted because it does not contain any of the keywords used on your list. So wouldn't this test be more accurate when you check for `auth-analyst@google.com` instead?

And we should not use `@google.com` here. Maybe use `@example.com` instead.

timesketch/lib/telemetry.py
```diff
@@ -31,13 +32,141 @@
...
+        "session",
+        "cookie",
+        "auth",
+        "credential",
+    ]
+
+    # Attributes that are explicitly exempt from PII redaction (e.g. analyst identity)
+    EXEMPT_PII_ATTRIBUTES = {"user.name", "user.id", "timesketch.user_id"}
+
+    class SensitiveDataScrubber(SpanProcessor):
```
jkppr: How does this impact performance? If this scrubs every string doing a sub-string search for a list of keywords to redact, won't this impact the API and worker performance of Timesketch?

Would it be sufficient to work only on a list of known risk attribute keys?

jkppr (changes requested):
General: 
Your PR says `Safe Telemetry Calls: Implemented a @safe_telemetry_call decorator ...` but the whole diff does not contain this decorator at all. Can you please ensure the PR message reflects what the implementation actually contains?

docs/OpenTelemetry.md
jaegeral: I will remove it from the markdown file, it is a general thing, not related to otel.

timesketch/lib/telemetry.py
jaegeral: let me see, I had a better idea yesterday evening on how to avoid any search related things to spill.

jaegeral:
I think I will need to split up this PR. To messy to review

timesketch/models/acl.py
jaegeral: yes it would create that many, but that is also true if we just enable sqlalchemy tracing :-/

docker/dev/docker-compose.yml
jaegeral: this will go to https://github.com/google/timesketch/pull/3805

data/otel-collector-config.yaml
jaegeral: https://github.com/google/timesketch/pull/3805

timesketch/lib/datastores/opensearch.py
jaegeral: unfortunately not

PR #3812, opened 2026-05-26
feat: add ExploreWildcard API endpoint and CLI support for raw wildcard queries

jaegeral:
Implementing the initial skeleton and routing for a new backend function `ExploreWildcardResource` and supporting Timesketch CLI client tools. This creates the architectural foundation for subsequent raw OpenSearch wildcard query work in small, reviewable increments.

### DIFF
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -20,7 +20,7 @@ import os
 import json
 import time
 import logging
-from typing import Dict, Generator, List, Optional
+from typing import Dict, Generator, List, Optional, Union
 
 import pandas
 
@@ -1100,6 +1100,40 @@ class Sketch(resource.BaseResource):
             return search_obj.to_pandas()
 
         return search_obj.to_dict()
+
+    def explore_wildcard(
+        self,
+        query_string: str,
+        fields: Union[str, List[str]] = "message",
+        limit: Optional[int] = None,
+    ) -> Dict[str, Union[Dict, List]]:
+        """Explore the sketch with raw wildcard queries (Skeleton endpoint).
+
+        Args:
+            query_string: String representation of the raw wildcard query (e.g. '*evil*').
+            fields: Comma-separated list or single field target to search (default: 'message').
+            limit: Optional integer representing maximum entries to return.
+
+        Returns:
+            A dictionary containing the parsed search result schema (metadata and objects).
+
+        Raises:
+            RuntimeError: If the sketch is currently archived.
+        """
+        # TODO: Align explore_wildcard signature and behavior with standard explore()
+        # by supporting as_pandas=False, as_object=False, and dynamic return types.
+        if self.is_archived():
+            raise RuntimeError("Unable to query an archived sketch.")
+
+        resource_url = f"{self.api.api_root}/sketches/{self.id}/explore_wildcard/"
+        form_data = {
+            "query": query_string,
+            "fields": fields,
+        }
+        if limit:
+            form_data["filter"] = {"size": limit}
+        response = self.api.session.post(resource_url, json=form_data)
+        return error.get_response_json(response, logger)
 
     def list_available_analyzers(self):
         """Returns a list of available analyzers."""
--- a/api_client/python/timesketch_api_client/sketch.py
+++ b/api_client/python/timesketch_api_client/sketch.py
@@ -1104,16 +1104,13 @@ class Sketch(resource.BaseResource):
     def explore_wildcard(
         self,
         query_string: str,
-        fields: Union[str, List[str]] = "message",
         limit: Optional[int] = None,
     ) -> Dict[str, Union[Dict, List]]:
         """Explore the sketch with raw wildcard queries (Skeleton endpoint).
 
         Args:
             query_string: String representation of the raw wildcard query
-                (e.g. '*evil*').
-            fields: Comma-separated list or single field target to search
-                (default: 'message').
+                (e.g. '*evil*' or 'message:*evil*').
             limit: Optional integer representing maximum entries to return.
 
         Returns:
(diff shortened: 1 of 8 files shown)


api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1101,6 +1101,43 @@ def explore(
...
+        # by supporting as_pandas=False, as_object=False, and dynamic return types.
+        if self.is_archived():
+            raise RuntimeError("Unable to query an archived sketch.")
+
+        resource_url = f"{self.api.api_root}/sketches/{self.id}/explore_wildcard/"
+        form_data = {
+            "query": query_string,
+            "fields": fields,
+        }
+        if limit:
```
jkppr: Minor nit: If a caller ever passes `limit=0` (which, although unlikely, is a valid integer), `if limit` evaluates to `False`. So maybe use `if limit is not None:` instead?

timesketch/api/v1/resources/explore.py
```diff
@@ -643,3 +643,113 @@ def get(self, sketch_id):
...
 
         return jsonify(schema)
+
+
+class ExploreWildcardResource(resources.ResourceMixin, Resource):
+    """Explore resources for running wildcard queries on the datastore.
+
+    This API endpoint allows exact-match wildcard searching on targeted document
+    fields (e.g. message, xml_string), bypassing standard Lucene string query
+    parser tokenization.
```
jkppr, line 653: Only on targeted document fields or also across all fields?

timesketch/api/v1/resources/explore.py
```diff
@@ -643,3 +643,113 @@ def get(self, sketch_id):
...
+        """Handles POST request to the resource.
+
+        Handler for /api/v1/sketches/:sketch_id/explore_wildcard/
+
+        Args:
+            sketch_id: Integer primary key for a sketch database model.
+
+        Input JSON parameters (request body):
+            query (str): Raw wildcard search expression (e.g. '*evil_term*').
+            fields (str): Comma-separated list of targeted fields (default: 'message').
```
jkppr: This is not how it works from the frontend. We can't provide a separate list of fields to use, since this would require us to do some string extraction on the frontend which I would prefer not to do. So the frontend will only send a `query` which can be `*search term*` or `message:searchterm` for example.

Additional nit: if you already accept JSON Input parameters, why use a comma separated str and not a list?

timesketch/api/v1/resources/explore.py
```diff
@@ -643,3 +643,113 @@ def get(self, sketch_id):
...
+
+        Returns:
+            A Flask Response (JSON format) containing the skeleton search response,
+            including an empty hits list.
+
+        Raises:
+            HTTPStatus.NOT_FOUND (404): If the sketch cannot be found.
+            HTTPStatus.FORBIDDEN (403): If current user lacks read
+                permissions on sketch.
+            HTTPStatus.BAD_REQUEST (400): If the sketch is currently archived.
```
jkppr, line 679: Is it a bad request if the sketch is archived? The requester did nothing wrong.

timesketch/api/v1/resources/explore.py
```diff
@@ -643,3 +643,113 @@ def get(self, sketch_id):
...
+            abort(
+                HTTP_STATUS_CODE_BAD_REQUEST,
+                "No active timelines with valid search indices found in this sketch.",
+            )
+
+        # TODO: In the actual search execution phase (Iteration 2), we should verify
+        # that the target fields in these indices support non-tokenized exact
+        # wildcard matching (i.e. they are mapped as 'keyword' type or possess a
+        # '.keyword' multi-field). If none of the requested fields are suitable
+        # for exact wildcard matching, we should raise a 400 Bad Request error.
```
jkppr: Critical nit: We want to work with the `wildcard` field type only here! Do not work with keyword or text field types for this method. If no wildcard field type exists (so no `message.wildcard`, we should not expose this method or return this as info).

timesketch/api/v1/resources/explore.py
```diff
@@ -643,3 +643,113 @@ def get(self, sketch_id):
...
+        # just accept the query and fields parameters but do not build the raw
+        # OpenSearch search query yet.
+        request_data = request.json or {}
+        query_string = request_data.get("query", "")
+
+        fields_raw = request_data.get("fields", "message")
+        if isinstance(fields_raw, str):
+            fields_list = [f.strip() for f in fields_raw.split(",") if f.strip()]
+        elif isinstance(fields_raw, list):
+            fields_list = [str(f).strip() for f in fields_raw if str(f).strip()]
```
jkppr: You can define in the API definition and form validation what format you accept. I would recommend to go with list only and skip the str extraction.

cli_client/python/timesketch_cli_client/commands/search.py
```diff
@@ -366,3 +366,54 @@ def describe_saved_search(ctx: click.Context, search_id: int):
...
+    "-q",
+    required=True,
+    help="Search query in raw wildcard format (e.g. *evil*)",
+)
+@click.option(
+    "--fields",
+    "-f",
+    default="message",
+    help="Comma-separated string of fields to match (default: message)",
+)
```
jkppr: This is something that works with a CLI but won't fly with the UI. The fields will be included in the query for most situations. e.g. searching for `message:*searchterm*` which also make the bool logic easier.

jkppr (approved):
LGTM, but please address the comments below first.

api_client/python/timesketch_api_client/sketch.py
jaegeral: good point

timesketch/api/v1/resources/explore.py
jaegeral: will keep it for now, we might change the error code later

timesketch/api/v1/resources/explore.py
jaegeral: Changed already

timesketch/api/v1/resources/explore.py
jaegeral: addressed

PR #3813, opened 2026-05-26
feat: implement wildcard search support for explore API.

jaegeral:
This pull request implements exact-match wildcard search functionality in Timesketch. 

Another change is include adding a `--compare` flag to the CLI client wildcard search command, verifying active wildcard mappings in the OpenSearch datastore, and introducing end-to-end tests.

### DIFF
--- a/data/generic.mappings
+++ b/data/generic.mappings
@@ -25,6 +25,6 @@
     },
     "timesketch_label": {
       "type": "nested"
-     }
+    }
   }
 }
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -809,18 +809,31 @@ class ExploreWildcardResource(resources.ResourceMixin, Resource):
             tl_colors[timeline.searchindex.index_name] = timeline.color
             tl_names[timeline.searchindex.index_name] = timeline.name
 
-        # Parse timesketch labels for response hits
-        for event in result["hits"]["hits"]:
-            event["selected"] = False
-            event["_source"]["label"] = []
-            try:
-                for label in event["_source"]["timesketch_label"]:
-                    if sketch.id != label["sketch_id"]:
-                        continue
-                    event["_source"]["label"].append(label["name"])
-                del event["_source"]["timesketch_label"]
-            except KeyError:
-                pass
+        # Parse timesketch labels for response hits safely
+        if isinstance(result, dict):
+            hits_dict = result.get("hits")
+            if isinstance(hits_dict, dict):
+                hits_list = hits_dict.get("hits")
+                if isinstance(hits_list, list):
+                    for event in hits_list:
+                        if not isinstance(event, dict):
+                            continue
+                        event["selected"] = False
+
+                        _source = event.get("_source")
+                        if isinstance(_source, dict):
+                            _source["label"] = []
+                            timesketch_label = _source.get("timesketch_label")
+                            if isinstance(timesketch_label, list):
+                                for label in timesketch_label:
+                                    if isinstance(label, dict):
+                                        if sketch.id != label.get("sketch_id"):
+                                            continue
+                                        _source["label"].append(label.get("name"))
+                                try:
+                                    del _source["timesketch_label"]
+                                except KeyError:
+                                    pass
 
         # Resolve total hits count
         es_total_count = result["hits"]["total"]
--- a/timesketch/api/v1/resources/explore.py
+++ b/timesketch/api/v1/resources/explore.py
@@ -653,96 +653,7 @@ class ExploreWildcardResource(resources.ResourceMixin, Resource):
     parser tokenization.
     """
 
-    def _verify_wildcard_mappings(
-        self, indices: list[str], fields_list: list[str]
-    ) -> dict[str, str]:
-        """Verifies that all targeted fields possess active wildcard mappings.
 
-        Args:
-            indices (list): List of active index names to search.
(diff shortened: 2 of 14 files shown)


timesketch/api/v1/resources/explore.py
```diff
@@ -734,29 +802,169 @@ def post(self, sketch_id: int):
...
-        # Return a valid standard skeleton search JSON structure
+        # Extract limit parameter safely from filter settings
+        limit = request_data.get("filter", {}).get("size")
+        if limit is not None:
+            try:
+                limit = int(limit)
+            except (ValueError, TypeError):
+                limit = 40
+        else:
+            limit = 40
```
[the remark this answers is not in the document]
jaegeral: done

timesketch/api/v1/resources/explore.py
```diff
@@ -734,29 +734,127 @@ def post(self, sketch_id: int):
...
+                                    for exact_path in field_paths.values()
+                                ],
+                                "minimum_should_match": 1,
+                            }
+                        }
+                    ],
+                    "filter": [],
+                }
+            },
+            "size": min(limit, 10000) if limit is not None else 40,
```
jkppr: This seems to be a duplicated check. You check the limit is not None in line 755 already. So here it should be fine to just do `"size": min(limit, 10000)`. Or you skip the part in 755ff and keep this one. Your call.

timesketch/tsctl.py
```diff
@@ -677,6 +686,100 @@ def list_sketches(
...
+    for timeline in sketch.timelines:
+        if not timeline.searchindex:
+            continue
+        si = timeline.searchindex
+        print(f"Timeline: '{timeline.name}' (ID: {timeline.id})")
+        print(f"SearchIndex DB: '{si.name}' | Physical Index: '{si.index_name}'")
+
+        # Retrieve mappings from OpenSearch
+        try:
+            mappings_data = datastore.client.indices.get_mapping(index=si.index_name)
```
jkppr: Two Nits: 
1. Most timelines in a sketch share the searchindex. (e.g. all Plaso files have one searchindex). So collecting a set of indices first before calling the mapping data makes this more efficient.
2. If a sketch has 50 timelines, this makes 50 synchronous requests. OpenSearch supports fetching multiple index mappings at once. Consider batching the physical index names into a single comma-separated string or list (e.g., datastore.client.indices.get_mapping(index=all_index_names)), executing the request once, and then looping through your results in memory.

timesketch/api/v1/resources/explore.py
```diff
@@ -734,29 +734,127 @@ def post(self, sketch_id: int):
...
+                                        if sketch.id != label.get("sketch_id"):
+                                            continue
+                                        _source["label"].append(label.get("name"))
+                                try:
+                                    del _source["timesketch_label"]
+                                except KeyError:
+                                    pass
+
+        # Resolve total hits count
+        es_total_count = result["hits"]["total"]
```
jkppr: This is prone to a potential KeyError if the results has no entries with those names

timesketch/api/v1/resources/explore.py
```diff
@@ -734,29 +734,127 @@ def post(self, sketch_id: int):
...
+                            if isinstance(timesketch_label, list):
+                                for label in timesketch_label:
+                                    if isinstance(label, dict):
+                                        if sketch.id != label.get("sketch_id"):
+                                            continue
+                                        _source["label"].append(label.get("name"))
+                                try:
+                                    del _source["timesketch_label"]
+                                except KeyError:
+                                    pass
```
jkppr: Why not use `_source.pop("timesketch_label", None)`? That saves the try / except.

data/generic.mappings
jkppr, line 1: Is this needed? The mapping was already adjusted in https://github.com/google/timesketch/pull/3814/changes#diff-a64a56a7fdc6bb0a9c9a2acf4a1ad49daf0283abf2b8aa7627e5a286b966ad6a

timesketch/lib/testlib.py
```diff
@@ -161,10 +161,27 @@ def search(
...
-        return {}
+        index_param = kwargs.get("index", "test")
+        if isinstance(index_param, str):
+            indices = [index_param]
+        elif isinstance(index_param, list):
+            indices = index_param
+        else:
+            indices = ["test"]
+
+        properties = {
```
jkppr, line 177: Want to include some `keyword` types too?

timesketch/api/v1/resources/explore.py
```diff
@@ -734,29 +734,127 @@ def post(self, sketch_id: int):
...
+                "bool": {
+                    "must": [
+                        {
+                            "bool": {
+                                "should": [
+                                    {
+                                        "wildcard": {
+                                            exact_path: {
+                                                "value": wildcard_query,
+                                                "case_insensitive": True,
```
jkppr, line 780: I'm not sure this will work. But we will need to test it.

jkppr (approved):
LGTM, But please take a look at the nits below.

timesketch/api/v1/resources/explore.py
jaegeral: oh yeah good point, will remove

timesketch/api/v1/resources/explore.py
jaegeral: classic, will fix

data/generic.mappings
jaegeral: that is only formatting by black

PR #3816, opened 2026-05-27
Support Ubuntu 26.04 base image & major dependency updates

jkppr:
This pull request introduces support for Ubuntu 26.04 as the base image for Timesketch. It includes necessary adjustments to Docker configurations, dependency versions, and internal logic to ensure compatibility with newer system libraries and updated versions of core tools like Pandas and OpenSearch. The changes maintain backward compatibility while preparing the infrastructure for future releases.

### Changes

* **Base Image Upgrade**: Updated Docker configurations to support Ubuntu 26.04 as the base image for both development and production environments.
* **Dependency Updates**: Refreshed core requirements including Pandas 3.x, NumPy 2.x, and updated OpenSearch to version 2.19.5.
* **Pandas 3.x Compatibility**: Updated CSV upload logic to handle Pandas 3.x date resolution and added filtering for out-of-range dates.
* **Plaso Integration**: Improved Plaso task execution by resolving paths via shutil and ensuring correct Python environment usage.


timesketch/lib/tasks.py
```diff
@@ -965,20 +966,32 @@ def run_plaso(
...
-    ]
+    if resolved_psort_path.endswith(".py"):
+        cmd = [
+            sys.executable,
+            resolved_psort_path,
+        ]
+    else:
+        cmd = [
+            resolved_psort_path,
+        ]
```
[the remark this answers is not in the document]
jkppr: I've tested it in the dev image and it works.

.github/workflows/e2e-tests.yml
```diff
@@ -9,22 +9,18 @@ on:
       - '*.md'
 
 jobs:
-  # Ubuntu 24.04
-  PyPi-plaso-stable-opensearch-v2-ubuntu2404:
-    runs-on: ubuntu-24.04
+  PyPi-plaso-stable-opensearch-v2-ubuntu2604:
+    runs-on: ubuntu-latest
```
jaegeral, line 13: can we instead pin that to 2604?

.github/workflows/e2e-tests.yml
```diff
@@ -33,22 +29,19 @@ jobs:
       - name: Dump docker logs on failure
         if: failure()
         uses: jwalton/gh-docker-logs@v2
-  PyPi-plaso-staging-opensearch-v2-ubuntu2404:
+  PyPi-plaso-staging-opensearch-v2-ubuntu2604:
     continue-on-error: true
-    runs-on: ubuntu-24.04
+    runs-on: ubuntu-latest
```
jaegeral, line 34: same here

.github/workflows/e2e-tests.yml
jkppr: No, GitHub does not have Ubuntu 26.04 runners yet.

.github/workflows/e2e-tests.yml
```diff
@@ -9,22 +9,18 @@ on:
...
 jobs:
-  # Ubuntu 24.04
-  PyPi-plaso-stable-opensearch-v2-ubuntu2404:
-    runs-on: ubuntu-24.04
+  PyPi-plaso-stable-opensearch-v2-ubuntu2604:
+    runs-on: ubuntu-latest
     timeout-minutes: 30
-    strategy:
-      matrix:
-        os: [ubuntu-24.04]
```
jaegeral, line 18: have you tested if we could also just add 26 to the matrix to spin up a parallel flow?

jkppr: You mean instead of 2 jobs have 1 job that has a matrix for the plaso tracks?

docker/e2e/Dockerfile
```diff
@@ -33,22 +33,28 @@ RUN \
...
     python3-setuptools \
     python3-venv \
     python3-wheel \
     tzdata \
     wget \
     && \
-    # 3. UPGRADE PLASO (Conditional)
+    # 3. INSTALL PSYCHOPG2 (Conditional)
+    # Ubuntu 24.04 has it in apt, 26.04 does not.
+    if [ "$UBUNTU_VERSION" = "24.04" ]; then \
```
jaegeral: could we instead an else tree here instead of hiding getting this further down below?

jaegeral (commented):
Smaller nits

docker/e2e/Dockerfile
jkppr: We can actually remove the 2404 completely if we go with it per default now. WDYT?

jaegeral:
Also maybe update the title so it is much clearer that it is not only ubuntu upgrade but also major dependencies.

.github/workflows/e2e-tests.yml
jkppr: keep as is for now.

.github/workflows/e2e-tests.yml
```diff
@@ -9,22 +9,18 @@ on:
...
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
-          fetch-depth: 0
+           fetch-depth: 0
```
jaegeral: why adding this whitespace?

jaegeral (approved):
small nit, otherwise LGTM

.github/workflows/e2e-tests.yml
jkppr:
```suggestion
          fetch-depth: 0
```

jkppr: fixed

PR #3817, opened 2026-05-27
obsolete: feat: add otel to wildcardsearch

jaegeral:
Add telemetry to WildCardSearch

This will be closed soon and replaced with a otel for opensearch explore in general now that we have refactored the explore class


jkppr:
Given the changes in https://github.com/google/timesketch/pull/3818 this needs probably some refactor.

PR #3818, opened 2026-05-28
Support Native Wildcard Fields Search

jkppr:
This pull request introduces a unified approach for real substring searches on OpenSearch datasets. Instead of using tokenized Lucene parsing on standard text fields, this feature maps and leverages native OpenSearch `wildcard` field types and subfields (e.g. `.wildcard` fields), offering improved exact-match wildcard searches.

### 1. OpenSearch Datastore & Query Builder Refactoring
- **Added `get_wildcard_fields()`**: Programmatically traverses OpenSearch index mapping property trees (including dot-notated nested maps and subfields) to identify wildcard-capable fields across all requested sketch indices.
- **Introduced `_build_wildcard_query_dsl()`**: 
  - Parses basic query strings into native OpenSearch `bool` DSL queries.
  - Dynamically tokenizes input text (taking into account quoted strings) and maps logical operators (`AND`, `OR`, `NOT`) into the respective `must`, `should`, and `must_not` clauses.
  - Routes global queries to a `multi_match` query across all active `*.wildcard` fields and field-specific queries to their exact `.wildcard` paths.
- **Refactored `search()` / `build_query()`**: Integrated `search_wildcard_fields` as a core boolean parameter to seamlessly compile queries with native wildcard queries when toggled.

### 2. API & Routing Upgrades
- **Standardized API Delegation**: Rewrote the `ExploreWildcardResource` skeleton API to act as a clean pass-through to the primary `ExploreResource` with `search_wildcard_fields = True` pre-injected. This avoids code duplication, simplifies routing logic, and guarantees standardized response payloads.
- **Form Handling & Client Support**:
  - Added the `search_wildcard_fields` BooleanField in `ExploreForm` to parse the parameter correctly via WTForms.
  - Updated the Python `timesketch_api_client` (`Search` & `Sketch`) and Timesketch CLI configurations to expose and persist the new `search_wildcard_fields` option.

### 3. Mapping Adjustments
- Standardized the `wildcard` subfield settings within `data/generic.mappings` and `data/plaso.mappings` to systematically index strings under a `.wildcard` sibling field.

### 4. Comprehensive Tests
- Rewrote and expanded coverage in `timesketch/lib/datastores/opensearch_test.py` to fully validate wildcard-field mapping traversal, global search compilation, field-specific search routing, and complex boolean operators routing.
- Refactored existing tests in `timesketch/api/v1/resources_test.py` to adapt to endpoint delegation and ensure robust functionality.


timesketch/api/v1/resources/explore.py
```diff
@@ -648,9 +657,9 @@ def get(self, sketch_id):
 class ExploreWildcardResource(resources.ResourceMixin, Resource):
```
jkppr, line 657: @jaegeral Given this endpoint was introduced with this feature and by now is just setting the flag and passing the job to the normal search endpoint, should we still keep this one? Do you see any benefit? Or should we refactor to use the `search_wildcard_fields` parameter only?

timesketch/lib/datastores/opensearch.py
```diff
@@ -779,119 +791,178 @@ def build_query(
...
+                }
+
+            if dsl_node:
+                # Append the node to the active boolean operator list
+                bool_query[current_operator].append(dsl_node)
+
+            # Reset to default AND after processing a term
+            current_operator = "must"
+
+        return bool_query
```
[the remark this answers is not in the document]
jkppr: > multi_match Wildcard Failure

This is wrong. Your info is outdated. It works like this.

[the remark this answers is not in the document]
jkppr: > Instead of writing a custom, fragile state-machine parser, leverage OpenSearch's native query_string query parser.

In theory a good idea, but wildcard field types are not supported in the query_string parser. So calls to e.g. `message.wildcard:*test*` will fail silently.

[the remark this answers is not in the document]
jkppr: > Broken OR Logic

Okay, that is correct. I'll fix it.

timesketch/api/v1/resources/explore.py
```diff
@@ -662,220 +671,21 @@ def post(self, sketch_id: int):
...
-            "ExploreWildcardResource: Sketch ID: %d, Query: %r, "
-            "Extracted Fields: %r, Extracted Wildcard Query: %r, "
-            "Resolved Paths: %r",
-            sketch_id,
-            query_string,
-            fields_list,
-            wildcard_query,
-            field_paths,
-        )
+        request.json["search_wildcard_fields"] = True
```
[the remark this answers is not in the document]
jkppr: The claim that mutating request.json raises an AttributeError: can't set attribute 'json' is incorrect under the active Flask/Werkzeug version.

Once Flask parses the incoming JSON request body via request.get_json(), the payload is cached as a standard, mutable Python dict under the request.json proxy property.

Executing request.json["search_wildcard_fields"] = True does not reassign the property itself; rather, it performs a standard key-value mutation directly on the cached dictionary. This behavior is fully supported, safe, and has been verified correct through integration testing in the active runtime container.

timesketch/lib/datastores/opensearch.py
```diff
@@ -779,119 +791,178 @@ def build_query(
...
-                        {},
-                    ).get("properties")
-
-                if not isinstance(properties, dict):
-                    supported_in_all_indices = False
-                    break
+            if wildcard_fields is None:
+                wildcard_fields = current_wildcard_fields
+            else:
+                wildcard_fields.update(current_wildcard_fields)
```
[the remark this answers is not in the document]
jkppr: This is intentional. If a sketch supports wildcard mapping all text type fields will also have a wildcard field. So there might be a slight chance of discrepancy but it is unlikely.

api_client/python/timesketch_api_client/search.py
```diff
@@ -937,9 +941,26 @@ def query_filter(self, query_filter):
...
         self.commit()
 
+    @property
+    def search_wildcard_fields(self):
+        """Return whether wildcard fields search mode is enabled."""
+        return self._search_wildcard_fields
+
+    @search_wildcard_fields.setter
+    def search_wildcard_fields(self, enabled):
+        """Enable or disable wildcard fields search mode."""
```
jaegeral: I think it would be useful to mention that the default here is `false`?

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1030,6 +1031,7 @@ def explore(
         max_entries=None,
         file_name="",
         as_object=False,
+        search_wildcard_fields=False,
```
jaegeral:
```suggestion
        search_wildcard_fields: bool = False,
```

Typing (yes it is not there for others, but does not hurt to start at some point)

api_client/python/timesketch_api_client/search.py
```diff
@@ -937,9 +941,26 @@ def query_filter(self, query_filter):
...
         self._extract_chips(query_filter)
         self.commit()
 
+    @property
+    def search_wildcard_fields(self):
+        """Return whether wildcard fields search mode is enabled."""
+        return self._search_wildcard_fields
+
+    @search_wildcard_fields.setter
+    def search_wildcard_fields(self, enabled):
```
jaegeral:
```suggestion
    def search_wildcard_fields(self, enabled: bool) -> None:
```

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -1106,22 +1115,19 @@ def explore_wildcard(
         query_string: str,
         limit: Optional[int] = None,
     ) -> Dict[str, Union[Dict, List]]:
-        """Explore the sketch with raw wildcard queries (Skeleton endpoint).
+        """Explore the sketch with raw wildcard queries (Deprecated).
+
+        This method is maintained for backward-compatibility. Newer scripts should
+        use explore(query_string, search_wildcard_fields=True) instead.
```
jaegeral:
```suggestion
        use explore(query_string, search_wildcard_fields=True) instead.
        TODO: Refactor / Remove this method
```

data/generic.mappings
```diff
@@ -1,30 +1,21 @@
...
-              "type": "wildcard"
+    "date_detection": false,
+    "dynamic_templates": [
+        {
+            "strings": {
+                "match_mapping_type": "string",
+                "mapping": {
+                    "type": "text",
+                    "fields": {
+                        "keyword": {"type": "keyword", "ignore_above": 256},
```
jaegeral, line 10: Do we need to document that somewhere to users when they have fields with longer fields?

timesketch/lib/forms.py
```diff
@@ -204,6 +204,9 @@ class ExploreForm(BaseForm):
     )
     scroll_id = StringField("Scroll ID", default="")
     file_name = StringField("Export to File")
+    search_wildcard_fields = BooleanField(
```
jaegeral: just an observation, I found `search_wildcard_fields` not intuitive to understand it being a bool, it could also contain the fields. What would you think about renaming it to `is_wildcard_search_enabled` or `enable_wildcard_search`?

jaegeral (commented):
PTAL

data/generic.mappings
jkppr: The mapping is supposed to be the TS admins individual responsibility since it highly depends on what data they ingest. But that said, yes adding documentation on this makes sense

timesketch/lib/forms.py
jkppr: Fair point. Let me think about it

timesketch/api/v1/resources/explore.py
jkppr: => #3820

timesketch/lib/forms.py
jkppr: Since this is a flag telling the backend what search mode the user intends to use, I propose `use_wildcard_fields`.

data/generic.mappings
jkppr: I've added documentation.

jaegeral (approved):
LGTM

PR #3822, opened 2026-06-04
api: enforce cross-sketch ownership on analyzer sessions, conclusions, and event relabeling

adilburaksen:
## Summary

Follow-up to #3821. Several more API handlers load a child object by the ID in the URL but never verify it belongs to the sketch in the URL — the same cross-sketch consistency check the sibling handlers (and #3821's aggregation fix) enforce. This lets a user act on objects belonging to a **different** sketch than the one they have access to.

## Gaps fixed

1. **`AnalyzerSessionResource.get`** (`analysis.py`) — read the analyzer session by `session_id` and returns it after checking only read permission on `<sketch_id>`, with no `analysis_session.sketch_id == sketch.id` check. A user with read access to their own sketch can pass an arbitrary `session_id` and read **another sketch's** analyzer session (its `result`, `log`, `parameters`, timeline). Added the not-found guard + cross-sketch check.

2. **`QuestionConclusionListResource` / `QuestionConclusionResource`** (`scenarios.py`, GET/POST/PUT/DELETE) — load `InvestigativeQuestion` by `question_id` with no `question.sketch_id == sketch.id` check, so conclusions could be read/created/edited/deleted against a question in another sketch. Added the check to all four handlers.

3. **`QuestionConclusionResource.put`** (`scenarios.py`) — was the only mutating handler missing `@login_required`. Added it (consistent with every sibling).

4. **`MarkEventsWithTimelineIdentifier.post`** (`event.py`) — the user-supplied `searchindex` was not tied to the validated `timeline`, so a write-capable user could relabel events on an arbitrary search index. Added `timeline.searchindex_id == searchindex.id` after the existing timeline-belongs-to-sketch check.

All checks mirror the existing pattern (`HTTP_STATUS_CODE_NOT_FOUND`, same f-string style). No behavior change for legitimate same-sketch requests.

## Testing

`py_compile` and `black --check` pass on all three files; changes are additive (53 insertions, 0 deletions).


jaegeral:
Fwiw, there is also: https://github.com/google/timesketch/pull/3777

adilburaksen:
Thanks @jaegeral — yes, #3777 (by @mohammadmseet-hue, opened in March) covers the same cross-sketch ownership class and predates this, so credit there for the overlap:

- The **aggregation** fix in #3821 overlaps #3777's `POST /sketches/<id>/aggregation/<aid>/` — #3777 got there first, so that one is effectively a duplicate.
- The **conclusions** `GET`/`POST` handlers here overlap #3777's `questions/<qid>/conclusions/` coverage.

This PR also closes endpoints that #3777 does **not** touch, which I'd suggest keeping:

- **`AnalyzerSessionResource.get`** (`analysis.py`) — reads an analyzer session by `session_id` with no `analysis_session.sketch_id == sketch.id` check. This is distinct from #3777's `timelines/<tid>/analysis/` (timeline_id) fix in the same file.
- **`MarkEventsWithTimelineIdentifier.post`** (`event.py`) — the user-supplied `searchindex` is not tied to the validated `timeline`, allowing event relabeling on an arbitrary search index. #3777 doesn't touch `event.py`.
- **`QuestionConclusionResource` `PUT`/`DELETE`** plus the missing `@login_required` on `PUT` — #3777 only covers conclusions `GET`/`POST`.

Happy to rebase this down to just the non-overlapping handlers (analyzer-session `get`, event relabel, conclusions `PUT`/`DELETE` + `login_required`) and let #3777 own the aggregation and conclusions `GET`/`POST` coverage, if that's cleaner for review. Whatever you prefer.

jkppr:
Given we have multiple PRs addressing similar ACL issues, I'll do some cherry picking. I'll remove the analysis.py changes in this PR but merge the events.py and scenario.py from this PR.

jkppr (approved):
Thanks for your contribution

adilburaksen:
While looking at the event relabeling path here, I found one more sibling of the same cross-sketch ownership class in `event.py` that none of the related PRs (#3777 / #3821 / #3822 / #3823) cover:

`EventTaggingResource.post` (the `__ts_fact` branch) loads an `InvestigativeQuestionConclusion` by a request-body `conclusion_id` and links it to the event without checking the conclusion belongs to the sketch in the URL:

```python
if not conclusion:
    abort(HTTP_STATUS_CODE_BAD_REQUEST, "Conclusion ID is required to add a fact.")
# (no check that the conclusion belongs to this sketch)
if not form.remove.data:
    event.conclusions.append(conclusion)
if form.remove.data:
    event.conclusions.remove(conclusion)
```

So a user with write access to sketch A can attach/detach a fact linkage against an `InvestigativeQuestionConclusion` that belongs to a different sketch B. The fix mirrors the ownership check already used in `scenarios.py`:

```python
if not conclusion.investigativequestion.sketch.id == sketch.id:
    abort(HTTP_STATUS_CODE_NOT_FOUND, "No conclusion found with this ID.")
```

Happy to push it into this PR's `event.py` changes, or you can fold it into the consolidation — whichever is cleaner for the cherry-pick.

jkppr:
@adilburaksen Thanks for flagging those additional paths. Please feel free to open a new PR based on latest master.

adilburaksen:
Opened #3844 for the `event.py` `__ts_fact` sibling path, based on latest master, as suggested. Thanks @jkppr.

PR #3823, opened 2026-06-04
api: scope analyzer result and session lookups to the requested sketch

evilgensec:
### What

`AnalysisResource.get` (`GET /api/v1/sketches/<sketch_id>/timelines/<timeline_id>/analysis/`) and `AnalyzerSessionResource.get` (`GET /api/v1/sketches/<sketch_id>/analyzer/sessions/<session_id>/`) authorize the caller against the sketch in the URL (`Sketch.get_with_acl` + `has_permission(current_user, "read")`), then load the result objects by a *different* ID:

- `Timeline.get_by_id(timeline_id)` → `Analysis.query.filter_by(timeline=...)`
- `AnalysisSession.get_by_id(session_id)`

Neither verifies that the loaded timeline/session belongs to the authorized sketch. A user who can read any sketch (for example one they own) can pass that sketch's `sketch_id` to satisfy the ACL check together with another sketch's `timeline_id` / `session_id` (sequential integers) and read that sketch's analyzer results and sessions — verdicts, logs, parameters, timeline name, and analyst identity — without having access to it.

### Fix

Add the sketch-scope check the sibling sub-object handlers already use. `AggregationResource.get` does `if aggregation.sketch_id != sketch.id: abort(404)`; `DataSourceResource`, `GraphResource`, and `EventResource` likewise scope the child to the sketch. The two handlers now return 404 when the timeline or session is from a different sketch. `AnalyzerSessionResource.get` also gains the missing existence check (it could previously call `to_json(None)`).

`Timeline` and `AnalysisSession` both carry a `sketch_id` column, so each check is a direct comparison.


jkppr (approved):
Thanks for your contribution

PR #3824, opened 2026-06-07
Fix imported search template visibility

### LINKED ISSUE #3385: Search templates not showing in UI
Search Templates are not showing in the UI and I am getting TypeErrors from VueJS.
I have tried it with multiple versions and the current version I have tried it with is: 20250408

After having imported the following example search template:

`- id: "1234"
  short_name: "Failed Logons"
  display_name: "Failed Logon Attempts"
  description: "Detects failed logon attempts"
  query_string: |
    event_id: 4625
  query_filter:
    field: event_id
    value: 4625
  query_dsl:
    bool:
      must:
        - match:
            event_id: 4625
  tags:
    - "authentication"
    - "security"
`
Using the command:
tsctl import-search-templates _/directory_

I can see them in the _/api/v1/searchtemplates_ and the number in the UI also increases when I upload new search templates. 

Image
With the following complete code:

Image

Have I done something wrong or is this a js problem?

**Screenshots**

**Additional context**

kiwigitops:
## Summary
- grant public read access to search templates imported from YAML so server-admin imports remain visible through the ACL-filtered API
- add a regression that imports a template through `tsctl import-search-templates` and reads it back from `/api/v1/searchtemplates/`

Closes #3385.

## Testing
- `python -m pytest timesketch/api/v1/resources_test.py::SearchTemplateResourceTest -q`
- `python -m py_compile timesketch\tsctl.py timesketch\api\v1\resources_test.py`
- `git diff HEAD~1..HEAD --check`


timesketch/api/v1/resources_test.py
```diff
@@ -710,6 +710,44 @@ def test_invalid_searchtemplate(self):
         response = self.client.get("/api/v1/searchtemplates/2/")
         self.assert404(response)
 
+    def test_imported_searchtemplate_resource(self):
+        """Imported search templates are visible through the API."""
+        from timesketch.tsctl import cli
```
jkppr: This test works, but is a bit of an anti-pattern when it comes to our test design. It uses the tsctl CLI inside the API endpoint tests. This is something we should move to the [e2e/tsctl_test.py](https://github.com/google/timesketch/blob/master/end_to_end_tests/tsctl_test.py)

jkppr (changes requested):

timesketch/api/v1/resources_test.py
jkppr: I've moved the test.

jkppr (approved):
Thanks for your contribution.

PR #3826, opened 2026-06-08
feat(otel): e2e otel tests

jaegeral:
* add a very simple otel e2e test
* Add a otel profile to the docker config of e2e tests.
* This also adds a new Ci workflow with enabled telemetry. Opted for an additional workflow instead of adjusting the old one to be a little flexible, might merge the workflows in the future.


docker/e2e/docker-compose.yml
```diff
@@ -15,6 +15,8 @@ services:
       - TIMESKETCH_ADMIN_USER_PASSWORD=admin
       - TIMESKETCH_USER2=test2
       - TIMESKETCH_PASSWORD2=test2
+      - TIMESKETCH_OTEL_MODE=otlp-grpc
+      - TIMESKETCH_OTLP_GRPC_ENDPOINT=otel-collector:4317
```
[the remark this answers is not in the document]
jaegeral: check it out now

end_to_end_tests/telemetry_test.py
```diff
@@ -0,0 +1,61 @@
...
+        response.raise_for_status()
+        traces_data = response.json()
+
+        # 5. Assert that at least one trace has been received by Jaeger
+        traces = traces_data.get("data", [])
+        self.assertGreater(
+            len(traces),
+            0,
+            "No telemetry traces were found in Jaeger for service 'timesketch'",
+        )
```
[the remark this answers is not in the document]
jaegeral: done

end_to_end_tests/telemetry_test.py
```diff
@@ -0,0 +1,62 @@
...
+    def test_telemetry_connectivity(self):
+        """Verify that OpenTelemetry spans are successfully exported to Jaeger."""
+        # 1. Check if Jaeger API is reachable. If not, skip gracefully.
+        jaeger_api_url = "http://jaeger:16686/api"
+        try:
+            response = requests.get(f"{jaeger_api_url}/services", timeout=5)
+            response.raise_for_status()
+        except requests.exceptions.RequestException:
+            print("Jaeger is not reachable. Telemetry profile is not active. Skipping.")
+            return
```
jkppr: Doesn't this exception result in a successful CI test if jaeger to connect? E.g. when the jaeger container crashes silently on startup this would not raise but return and silently skip the tests? Is this intentional?

Given jaeger must be reachable for telemetry to work, we should probably allow it to raise a failure.

end_to_end_tests/telemetry_test.py
```diff
@@ -0,0 +1,62 @@
...
+
+        # 2. Trigger a simple API request to generate some telemetry
+        self.api.list_sketches()
+
+        # 3. Wait for the asynchronously buffered spans to flush to the collector
+        time.sleep(5)
```
jkppr: What if flushing takes 6 seconds? 
Wouldn't this be more resilient when we poll for a trace and drop out early if found and e.g. fail after 30sec?

end_to_end_tests/tools/run_end_to_end_tests.sh
```diff
@@ -14,10 +14,16 @@ DEFAULT_OPENSEARCH_VERSION=2.19.5
 # Container ID for the web server
 export CONTAINER_ID="$(sudo -E docker container list -f name=e2e_timesketch -q)"
 
+# Configure telemetry if requested
+COMPOSE_PROFILES=""
+if [ "$TELEMETRY" = "true" ]; then
+  export TIMESKETCH_OTEL_MODE=otlp-grpc
+  COMPOSE_PROFILES="--profile telemetry"
+fi
```
jkppr: iirc docker compose supports env variables for profiles. If you use this, you can also keep `sudo -E docker compose -f ./docker/e2e/docker-compose.yml up -d` in line 20.

```suggestion
if [ "$TELEMETRY" = "true" ]; then
  export TIMESKETCH_OTEL_MODE=otlp-grpc
  export COMPOSE_PROFILES=telemetry
fi
```

.github/workflows/e2e-tests.yml
```diff
@@ -50,3 +50,24 @@ jobs:
       - name: Dump docker logs on failure
         if: failure()
         uses: jwalton/gh-docker-logs@v2
+  PyPi-plaso-stable-opensearch-v2-ubuntu2604-telemetry:
```
jkppr, line 56: Why create a new extra runner job for this that will also take 22min to finish while basically testing all the same stuff over again?

Can't we enable this for one of the other two runners instead to include it there? 

And if you need to run a dedicated runner, can we limit it on the telemetry_test?

jkppr (commented):
Please see comments below

end_to_end_tests/telemetry_test.py
jaegeral: oh good point, let me add some checks

end_to_end_tests/telemetry_test.py
jaegeral: yep

.github/workflows/e2e-tests.yml
jaegeral: done it is now limiting the tests

end_to_end_tests/tools/run_end_to_end_tests.sh
jaegeral: done

end_to_end_tests/tools/run_end_to_end_tests.sh
```diff
@@ -14,9 +14,14 @@ DEFAULT_OPENSEARCH_VERSION=2.19.5
 # Container ID for the web server
 export CONTAINER_ID="$(sudo -E docker container list -f name=e2e_timesketch -q)"
 
+# Configure telemetry if requested
+if [ "$TELEMETRY" = "true" ]; then
+  export TIMESKETCH_OTEL_MODE=otlp-grpc
+  export COMPOSE_PROFILES=telemetry
+fi
```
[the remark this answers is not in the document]
jaegeral: well this is only for workflows in github, so not really a problem

.github/workflows/e2e-tests.yml
```diff
@@ -53,3 +53,24 @@ jobs:
...
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
jkppr, line 58: Optional nit: Since we now run only the telemetry here, we can probably reduce the timeout.

jkppr (approved):
lgtm

.github/workflows/e2e-tests.yml
jaegeral: ah I do not think it would bother to keep it.

PR #3828, opened 2026-06-09
fix(cli-client): output_format usage in cli client

jaegeral:
Fix the output_format usage in cli client.
It was not coherent named (`output` vs `output_format`) 

And added some tests

```
root@ff572fe90e27:/# timesketch config set output text
root@ff572fe90e27:/# timesketch config set output-format text
root@ff572fe90e27:/# timesketch config get output-format
text
root@ff572fe90e27:/# timesketch config get output
text
root@ff572fe90e27:/# timesketch config set output-format foo
Unsupported format. Choose between text csv tabular json
```


cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -121,13 +121,13 @@ def output_format(self):
...
-            self.config_assistant.set_config("output", output_format)
+            self.config_assistant.set_config("output_format", output_format)
             self.config_assistant.save_config()
             return output_format
         try:
-            output_format = self.config_assistant.get_config("output")
+            output_format = self.config_assistant.get_config("output_format")
         except KeyError:  # in case value does not exist in the config file
-            self.config_assistant.set_config("output", DEFAULT_OUTPUT_FORMAT)
+            self.config_assistant.set_config("output_format", DEFAULT_OUTPUT_FORMAT)
```
jkppr: Is this change backwards compatible? E.g. when someone has set their default output format in their .timesketchrc file manually using the old `output` key?

end_to_end_tests/cli_client_e2e_test.py
```diff
@@ -156,6 +157,72 @@ def test_cli_sketch_export(self):
...
+        rc_path = os.path.join(home_path, ".timesketchrc")
+        token_path = os.path.join(home_path, ".timesketch.token")
+
+        rc_existed = os.path.exists(rc_path)
+        if rc_existed:
+            os.rename(rc_path, f"{rc_path}.bak")
+
+        token_existed = os.path.exists(token_path)
+        if token_existed:
+            os.rename(token_path, f"{token_path}.bak")
```
jkppr: If I get this right, you park the existing timesketch configs in a .bak file and restore them on finally? 

But if the tests fail with e.g. a SIGKILL in a CI runner, finally is not called so the tmp files now persists. Is this a problem?

if yes, could we allow to define the config to use via a cli flag? `--config` ?

cli_client/python/timesketch_cli_client/commands/config.py
```diff
@@ -30,25 +30,50 @@ def set_group():
...
 @set_group.command("sketch")
 @click.argument("sketch_id")
 @click.pass_context
-def set_sketch(ctx, sketch_id):
+def set_sketch(ctx: click.Context, sketch_id: str) -> None:
     """Set the active sketch.
 
     Args:
         ctx: Click CLI context object.
         sketch_id: ID of the sketch to save to config.
```
jkppr:
```suggestion
        sketch_id: ID of the sketch to save to config (pass an empty string to clear).
```

jkppr (approved):
LGTM, just some minor nits below

end_to_end_tests/cli_client_e2e_test.py
jaegeral: yep

end_to_end_tests/cli_client_e2e_test.py
```diff
@@ -195,34 +188,23 @@ def test_cli_integration(self):
...
+            self.assertions.assertEqual(result.exit_code, 0, f"Failed: {result.output}")
 
             # Test config get sketch
-            result = self.runner.invoke(cli, ["config", "get", "sketch"])
-            self.assertions.assertEqual(result.exit_code, 0)
+            result = self.runner.invoke(cli, ["--config", rc_path, "config", "get", "sketch"])
+            self.assertions.assertEqual(result.exit_code, 0, f"Failed: {result.output}")
             self.assertions.assertEqual(result.output.strip(), "42")
 
-        finally:
```
jaegeral: now only using a temp file instead of deleting the stuff in finally

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -142,8 +142,14 @@ def output_format(self):
     required=False,
     help="Set output format [json, text, tabular, csv] (overrides global setting).",
 )
+@click.option(
```
jaegeral, line 145: adding a new config file parameter as per Janosch feedback

jkppr (approved):
Looks good, thanks

PR #3830, opened 2026-06-09
draft(e2e): reduce time for e2e tests

jaegeral:
Improving the e2e tests:
* By using cached images
* introducing timing information on every e2e test so we can see over time if tests take longer.

from Gemini:

The Optimization: Since GitHub Actions runs on clean VMs with no local cache, standard docker builds pull and compile everything sequentially from scratch on every commit. This change hooks Buildx into the GitHub Actions cache (type=gha), saving intermediate layers.
The Result: If dependencies or Dockerfiles do not change, image building drops from 5+ minutes to seconds, loading the pre-built image directly into Docker Compose.
So it helps if you run commits on a PR.


docker/e2e/docker-compose.yml
```diff
@@ -1,5 +1,6 @@
 services:
   timesketch:
+    image: e2e-timesketch:latest
```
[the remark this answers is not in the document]
jaegeral: good point

.github/workflows/e2e-tests.yml
```diff
@@ -15,13 +15,32 @@ jobs:
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
-          fetch-depth: 0
+          fetch-depth: 1
```
jkppr, line 18: This change is good, we should keep it. But it saves only about 7 seconds I think.

.github/workflows/e2e-tests.yml
jkppr, line 1: Does the e2e test also work if we mark it as read only on the repository? This lowers the risk of running 3rd packages in the CI even more.

.github/workflows/e2e-tests.yml
```diff
@@ -15,13 +15,32 @@ jobs:
...
+          build-args: |
+            UBUNTU_VERSION=26.04
+            PLASO_PPA_TRACK=stable
+      - name: Cache Docker images
+        uses: actions/cache@0057852bfaa89a56745cba8c7296529d2fc39830  # v4.2.0
+        with:
+          path: /var/lib/docker
+          key: ${{ runner.os }}-docker-${{ hashFiles('docker/e2e/docker-compose.yml') }}
+          restore-keys: |
+            ${{ runner.os }}-docker-
```
jkppr, line 39: This buildx step seems not to save much time at all. In that case we can also skip it if you ask me.

end_to_end_tests/interface.py
jkppr, line 1: I like those! Let's keep the time infos 👍

jkppr (changes requested):
> If dependencies or Dockerfiles do not change, image building drops from 5+ minutes to seconds, loading the pre-built image directly into Docker Compose.

I don't think this correct. Checking past runs of our e2e test logs in GitHub the build of the docker container takes around ~2 min in total. And with this PR the build via buildX also takes about 2min + 30sec to run docker compose. 

So the 5min claim seems a bit over optimistic and not matching the latest version of the e2e tests.

See comments below.

.github/workflows/e2e-tests.yml
jaegeral: yeah it is not that much

PR #3834, opened 2026-06-10
Upgrade some frontend-ng dependencies

jkppr:
This PR updates the dependencies for `shell-quote`, `uuid`, `postcss` and `serialize-janascript`.

+ A small fix for an in-efficient regex section.


timesketch/frontend-ng/src/utils/ThreatIntelMetadata.js
```diff
@@ -1,5 +1,5 @@
 const IOCTypes = [
-    { regex: /^(\/[\S]+)+$/i, type: 'fs_path', humanReadable: 'Filesystem path' },
```
jaegeral, line 2: is that an intentional change in this PR?

jkppr: Yes, It fixes an inefficient regex part.

jaegeral (approved):
LGTM

PR #3837, opened 2026-06-10
Fix: Path and Redirect issues in the python API

jkppr:
This PR improves our upload path and page redirects to make them even more resilient against attacks. (And also makes the CodeQL happy)

### DIFF
--- a/timesketch/views/auth.py
+++ b/timesketch/views/auth.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 """This module implements HTTP request handlers for the user views."""
 
+import logging
 from typing import Optional
 from urllib.parse import urlparse
 
@@ -52,6 +53,8 @@ from timesketch.models import db_session
 from timesketch.models.user import Group
 from timesketch.models.user import User
 
+logger = logging.getLogger("timesketch.views.auth")
+
 # Register flask blueprint
 auth_views = Blueprint("user_views", __name__)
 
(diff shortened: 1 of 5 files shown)


timesketch/views/auth.py
```diff
@@ -60,6 +62,34 @@
 ]
 
 
+def is_safe_url(target):
```
jaegeral: Typing

timesketch/views/auth.py
```diff
@@ -60,6 +62,34 @@
...
+    # Remove control characters that browsers might strip
+    sanitized_target = "".join(c for c in target if c not in "\t\r\n\x0b\x0c")
+    # Replace backslashes with forward slashes to emulate browser behavior
+    sanitized_target = sanitized_target.replace("\\", "/")
+    # Must start with a single '/' and not '//'
+    if not sanitized_target.startswith("/") or sanitized_target.startswith("//"):
+        return False
+    try:
+        parsed = urlparse(sanitized_target)
+    except ValueError:
```
jaegeral: except ValueError as e:

return the e content so that a caller knows why it failed?

timesketch/views/auth_test.py
```diff
@@ -68,13 +68,34 @@ def test_login_invalid_next_url(self, mock_logger):
             "//example.com",
             "/\\example.com",
             "http://example.com",
+            "///example.com",
+            "/\t/example.com",
+            "/\n/example.com",
+            "/\r/example.com",
+            "/\x0b/example.com",
+            "/\x0c/example.com",
+            "javascript:alert(1)",
```
jaegeral, line 77: how about local file paths?

timesketch/views/spa.py
```diff
@@ -31,7 +32,13 @@ def redirect_view(sketch_id, view_id):
     Returns:
         Redirect to new URL scheme.
     """
-    return redirect(f"/sketch/{sketch_id:d}/explore?view={view_id:d}")
+    return redirect(
+        url_for(
+            "spa_views.overview",
```
jaegeral, line 41: Not familiar with that syntax, can you write a comment on what this means where it is redirected to?

jaegeral (commented):
PTAL

timesketch/views/spa.py
jkppr: I'll update the docstring for the method to explain this in more detail. TL;DR: Redirect from legacy `/sketch/<id>/explore/view/<id>/` to new `/sketch/<id>/explore?view=<id>`

timesketch/views/auth_test.py
jkppr: Sure, we can test that too. I'll add one.

timesketch/views/auth.py
jkppr: This would not make sense or improve the UX of the caller. The method is a pure bool return method. if anything is not okay about the redirect path we will fallback to root page and be good with it. The user does not benefit from any trace here since there are no manual redirect parameters in Timesketch.

jkppr: Based on offline discussion we'll add logging in False situations.

jaegeral (approved):
LGTM nice one

PR #3838, opened 2026-06-10
ci: Restrict default GITHUB_TOKEN permissions to read-only across all workflows

jkppr:
This PR adds an explicit `permissions: { contents: read }` block to all 6 GitHub Actions workflows in `.github/workflows/`. 

This addresses the GitHub CodeQL warning `actions/missing-workflow-permissions` by enforcing the principle of least privilege, ensuring that the default `GITHUB_TOKEN` generated for workflow runs does not inherit write permissions.

#### **Why is this safe?**
None of the workflows rely on the default `GITHUB_TOKEN` for write operations:
* **Build/Deploy Workflows** (`build_frontend-ng.yml`, `build_frontend-v3.yml`, and `documentation.yml`): These workflows perform write operations (pushing back to the repository or deploying to GitHub Pages), but they are configured to authenticate via an SSH Deploy Key (`secrets.DEPLOY_KEY`) rather than using the default `GITHUB_TOKEN`.
* **Testing/Linting Workflows** (`e2e-tests.yml`, `linters.yml`, and `unit-tests.yml`): These workflows only perform read operations (checking out the code, running tests, and running linters) and have no need for write permissions.

By explicitly declaring `contents: read`, we document the exact permissions required by these workflows and safeguard them against future default permission changes in GitHub.

### DIFF
--- a/.github/workflows/build_frontend-ng.yml
+++ b/.github/workflows/build_frontend-ng.yml
@@ -9,6 +9,9 @@ on:
       - '!timesketch/frontend-ng/dist/**' # Exclude changes in the dist folder
   workflow_dispatch:
 
+permissions:
+  contents: read
+
 jobs:
   build-frontend-ng:
     timeout-minutes: 30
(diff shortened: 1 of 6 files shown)


jaegeral (approved):
LGTM

Issue #3801, opened 2026-05-08
Otel Plan


jaegeral:
Here is a small plan on how / what to do with OTEL

- [ ] Enable OTEL for SQLAlchemy - Important: Do not log the query itself as it could countain sensitive data https://github.com/google/timesketch/pull/3840
- [x] Enable OTEL for OpenSearch --> https://github.com/google/timesketch/pull/3839
       - [ ] streamed searches https://github.com/google/timesketch/pull/3839#discussion_r3388994911
- [ ] Enable OTEL for LLM Model module --> https://github.com/google/timesketch/pull/3798/
- [x] Check that if OTEL is not configured, it would not crash -- https://github.com/google/timesketch/pull/3826

https://github.com/google/timesketch/pull/3795

PR #3839, opened 2026-06-10
feat(otel): Add basic OpenSearch Telemetry

jaegeral:
This pull request introduces OpenSearch search telemetry instrumentation by adding an instrument_search decorator in `timesketch/lib/telemetry.py` and applying it to the OpenSearch datastore's search method.

* Time taken for a search
* sketch-id

Screenshot 2026-06-10 at 15 07 49

### DIFF
--- a/end_to_end_tests/telemetry_test.py
+++ b/end_to_end_tests/telemetry_test.py
@@ -39,7 +39,7 @@ class TelemetryTest(interface.BaseEndToEndTest):
             except requests.exceptions.RequestException as e:
                 last_error = e
                 time.sleep(1)
-        
+
         self.assertions.fail(
             f"Jaeger is not reachable at {jaeger_api_url}: {last_error}"
         )
@@ -111,7 +111,9 @@ class TelemetryTest(interface.BaseEndToEndTest):
                         tags = {
                             t.get("key"): t.get("value") for t in span.get("tags", [])
                         }
-                        if "db.opensearch.took_ms" in tags and str(tags.get("timesketch.sketch_id")) == str(self.sketch.id):
+                        if "db.opensearch.took_ms" in tags and str(
+                            tags.get("timesketch.sketch_id")
+                        ) == str(self.sketch.id):
                             found_took_ms = True
                             break
                     if found_took_ms:
--- a/end_to_end_tests/telemetry_test.py
+++ b/end_to_end_tests/telemetry_test.py
@@ -27,27 +27,31 @@ class TelemetryTest(interface.BaseEndToEndTest):
 
     NAME = "telemetry_test"
 
-    def setup(self):
-        """Import a test timeline to ensure search works."""
-        self.import_timeline("evtx_direct.csv")
-
-    def test_telemetry_connectivity(self):
-        """Verify that OpenTelemetry spans are successfully exported to Jaeger."""
-        # 1. Check if Jaeger API is reachable.
+    def _wait_for_jaeger(self):
+        """Wait for Jaeger API to become reachable."""
         jaeger_api_url = "http://jaeger:16686/api"
         last_error = None
         for _ in range(30):
             try:
                 response = requests.get(f"{jaeger_api_url}/services", timeout=5)
                 response.raise_for_status()
-                break
+                return
             except requests.exceptions.RequestException as e:
                 last_error = e
                 time.sleep(1)
-        else:
-            self.assertions.fail(
-                f"Jaeger is not reachable at {jaeger_api_url}: {last_error}"
-            )
+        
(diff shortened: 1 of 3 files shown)


timesketch/lib/telemetry.py
```diff
@@ -37,17 +37,67 @@
...
-except (ImportError, ModuleNotFoundError):
+except (ImportError, ModuleNotFoundError) as e:
     HAS_OTEL = False
     trace = None
     StatusCode = None
     INVALID_SPAN = None
+    logger = logging.getLogger("timesketch.telemetry")
+    logger.warning(
+        "OpenTelemetry is not installed. Some features may crash. Error: %s", e
+    )
```
jkppr: "Some features may crash"

isn't this a bit dramatic? Wasn't the code designed to be fully backwards compatible and if no OpenTelemetry is installed it should not crash at all?

Also, wouldn't a info be enough?

end_to_end_tests/telemetry_test.py
```diff
@@ -76,6 +80,55 @@ def test_telemetry_connectivity(self):
         )
         print("Telemetry infrastructure E2E connectivity verified successfully!")
 
+    def test_opensearch_telemetry(self):
+        """Verify that OpenSearch telemetry spans are exported."""
+        jaeger_api_url = "http://jaeger:16686/api"
```
jkppr, line 89: iirc the test runner executes test methods alphabetically. So `test_opensearch_telemetry` will run before `test_telemetry_connectivity`. Currently, both tests will need to wait for Jaeger to become active during container startup, resulting in duplicated polling logic.

Why not throw it in a private `_wait_for_jaeger()` method and call it in the beginning of each method? (Or maybe the setup?)

end_to_end_tests/telemetry_test.py
```diff
@@ -76,6 +80,55 @@ def test_telemetry_connectivity(self):
...
+                traces_data = response.json()
+                traces = traces_data.get("data", [])
+
+                # Check if we found the took_ms tag in any span
+                for trace in traces:
+                    for span in trace.get("spans", []):
+                        tags = {
+                            t.get("key"): t.get("value") for t in span.get("tags", [])
+                        }
+                        if "db.opensearch.took_ms" in tags:
```
jkppr: This wouls also trigger for other searches executed in the same time? Should we ensure the span is for the exact sketch_id maybe?

timesketch/lib/datastores/opensearch.py
```diff
@@ -1105,6 +1106,7 @@ def is_operand_like_right(t):
         return {"must": [final_node], "must_not": [], "filter": []}
 
     # pylint: disable=too-many-arguments
+    @telemetry.instrument_search
```
jkppr, line 1109: Way better solution! 👍

timesketch/lib/datastores/opensearch.py
jkppr, line 1: What about scrolled searches? E.g. `search_stream()` does not go through `search()`. Maybe worth to also decorate that one?

timesketch/lib/telemetry.py
```diff
@@ -37,17 +37,67 @@
     from opentelemetry.sdk.trace.export import BatchSpanProcessor
 
     HAS_OTEL = True
-except (ImportError, ModuleNotFoundError):
+except (ImportError, ModuleNotFoundError) as e:
     HAS_OTEL = False
     trace = None
     StatusCode = None
     INVALID_SPAN = None
```
jkppr: Dead code

```suggestion
```

jkppr (commented):
next round

timesketch/lib/telemetry.py
jaegeral: yep, will adjust, and yeah it is as much backwards compatible as possible :)

end_to_end_tests/telemetry_test.py
jaegeral: yep

end_to_end_tests/telemetry_test.py
jaegeral: easy thing

timesketch/lib/datastores/opensearch.py
jaegeral: hm one at a time, I would like to keep this simple for now, and do a follow up, I added a note in the otel plan https://github.com/google/timesketch/issues/3801

timesketch/lib/telemetry.py
jaegeral: done

jkppr (approved):
thanks for adding this telemetry

PR #3840, opened 2026-06-10
feat(otel): Add SQLALchemy basic telemetry

jaegeral:
This pull request introduces SQLAlchemy telemetry instrumentation using OpenTelemetry, adding the necessary dependency, instrumenting the database engine in both the web application and Celery workers, and adding end-to-end tests.

Part of https://github.com/google/timesketch/issues/3801


timesketch/lib/telemetry.py
```diff
@@ -229,6 +248,23 @@ def instrument_flask_app(app, **kwargs):
...
+    Args:
+        engine (sqlalchemy.engine.Engine): The SQLAlchemy engine (ignored in favor
+            of global instrumentation).
+        **kwargs: Additional arguments passed to SQLAlchemyInstrumentor().instrument().
+    """
+    if not is_enabled():
+        return
+    instrumentor = SQLAlchemyInstrumentor()
+    if not instrumentor.is_instrumented_by_opentelemetry:
+        instrumentor.instrument(**kwargs)
```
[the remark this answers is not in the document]
jaegeral: done already

end_to_end_tests/telemetry_test.py
```diff
@@ -135,6 +136,109 @@ def test_opensearch_telemetry(self):
...
+
+    def test_celery_telemetry_upload(self):
+        """Verify that telemetry is generated during a file upload (Celery task)."""
+        # 1. Trigger an upload which runs a Celery task
+        rand = uuid.uuid4().hex
+        sketch = self.api.create_sketch(name=f"test_telemetry_upload_{rand}")
+        file_path = (
+            "/usr/local/src/timesketch/end_to_end_tests/test_data/sigma_events.jsonl"
+        )
+        self.import_timeline(file_path, sketch=sketch)
```
jkppr: Doesn't the `import_timeline` function automatically prepend the path already? This should work just as `self.import_timeline("sigma_events.jsonl", sketch=sketch)` too. (it does in other tests at least)

end_to_end_tests/telemetry_test.py
```diff
@@ -135,6 +136,109 @@ def test_opensearch_telemetry(self):
...
+
+                            for val in tags.values():
+                                if secret_term in str(val):
+                                    found_secret_leak = True
+
+                if found_db_system:
+                    break
+            except requests.exceptions.RequestException:
+                pass
+            time.sleep(1)
```
jkppr, line 194: iirc the Jaeger container retains traces across the entire test suite, so `test_sqlalchemy_telemetry` and `test_celery_telemetry_upload` will find older database traces from previous tests (like `test_telemetry_connectivity`) on their very first poll. This causes the loop to break immediately (`found_db_system = True`) without actually waiting for or validating the newly generated traces.
or do I miss something?

jkppr (changes requested):
Please see nits below

end_to_end_tests/telemetry_test.py
jaegeral: oh true, let me add a start time to the jaeger check to only get traces after the test started

jkppr (approved):
Thanks for the update

PR #3857, opened 2026-06-19
Fix LLM multiprocessing serialization error (cannot pickle '_thread.lock')

jkppr:
This PR resolves a blocking serialization error (`TypeError: cannot pickle '_thread.lock' object`) encountered when executing LLM features in Timesketch. 

#### Root Cause
In a recent update, the LLM providers were unified to use the modern `google-genai` SDK. The `GoogleGenAI` provider instantiates `genai.Client` in its `__init__`, which holds active connection pools and thread locks. When `LLMResource._execute_llm_call` spawned a child process (to enforce the LLM timeout) and passed the `llm_provider` instance as an argument, Python's `multiprocessing` library failed to serialize the provider object because of the internal thread locks.

Additionally, instantiating the provider inside the child process using `LLMManager.create_provider` fails because `current_app` (Flask context) is not available in the spawned child process (especially under `forkserver` start method).

#### Solution (Pass Primitives & Instantiate in Child)
To resolve this without leaking serialization complexity or breaking future features, we avoid passing any complex object instances across the process boundary:

1.  **Serialization Payloads:**
    *   Added `to_dict()` to `LLMProvider` which extracts the provider class and its config dictionary.
    *   Added `get_execution_context()` to `LLMFeatureInterface` which extracts the feature name and response schema.
2.  **API Resource Layer:**
    *   Updated `LLMResource._execute_llm_call` to extract these payloads and pass them to the subprocess helper.
    *   Updated the module-level helper `_get_content_with_timeout_helper` to receive these payloads and dynamically instantiate the provider class with its config inside the child process. This avoids pickling the instance and resolves the lack of Flask context in the child process.
3.  **Future-Proofing:** 
    *   This approach is robust against future changes; if developers add non-picklable attributes (like DB sessions) to `Feature` classes (e.g., `LogAnalyzer`), it will not break the multiprocessing boundary because the `Feature` instance is never pickled.

#### Testing
*   Added `test_execute_llm_call_real` integration test to `resources_test.py` using a picklable dummy feature to verify the end-to-end multiprocessing flow works correctly under `forkserver` mode.
*   Formatting (`black`) and Linting (`pylint`) have been run and verified clean on all modified files.


timesketch/api/v1/resources/llm.py
jaegeral: should we have a timeout here as well?
```suggestion
                process.join(timeout=self._LLM_TIMEOUT_WAIT_SECONDS)
```

jaegeral (approved):
A very small nit / suggestion, otherwise LGTM

timesketch/api/v1/resources/llm.py
jkppr: Good point, but 30sec for waiting on SIGTERM to finish is a bit long. I'll add a 2 sec timeout instead and kill the process if it's still alive to prevent zombies.

PR #3860, opened 2026-06-22
fix(delete): Fix soft delete crash when sketch was soft_deleted

jaegeral:
This pull request resolves an issue where attempting to delete a soft-deleted sketch via the CLI fails. 

It handles exceptions raised when lazy-loading timeline info on soft-deleted sketches, defaulting these values to 'N/A'.

Fwiw, we made the decision back when we implemented deletion, that only admins can see soft-deleted sketches (also in the sketch-list with `include_deleted=true` .  And currently we do not have a flag for `include_deleted` in the pull single sketch endpoint (which would be only usable by admins).


cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -348,31 +349,72 @@ def delete_sketch(ctx: click.Context, force_delete: bool) -> None:
...
+    if is_archived:
         click.echo("Error Sketch is archived")
         ctx.exit(1)
 
+    try:
+        sketch_name = sketch.name
+        sketch_desc = sketch.description
+        sketch_status = sketch.status
+        sketch_labels = sketch.labels
+        timelines = sketch.list_timelines()
```
jkppr, line 356: If I'm not mistaken, the API ([_get_sketch_for_admin](https://github.com/google/timesketch/blob/master/timesketch/api/v1/resources/sketch.py#L343)) returns `timelines: []`. This means for admins, the dry-run here will show no timelines, but `--force_delete` will still permanently delete them from the DB.

That seems a bit risky because the admin won't see what they are actually deleting. Or is this intentional?

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -348,31 +349,72 @@ def delete_sketch(ctx: click.Context, force_delete: bool) -> None:
...
+    timelines = []
+
+    try:
+        is_archived = sketch.is_archived()
+    except NotFoundError as e:  # pylint: disable=unused-variable
+        click.echo(
+            f"Warning: Sketch {sketch.id} appears to be soft-deleted or inaccessible."
+        )
+        if not force_delete:
+            click.echo("If you want to permanently delete it, use --force_delete")
```
jkppr: Isn't this guidance a bit misleading? We only enter this `except NotFoundError` block if the sketch doesn't exist at all, or if the user is a non-admin (since admins can check archived status on soft-deleted sketches without triggering a 404).

In both cases, telling them to try `--force_delete` will just lead to another failure (either because they aren't an admin, or because the sketch genuinely doesn't exist). Maybe we should just report it as not found or inaccessible without suggesting `--force_delete`?

jkppr (changes requested):
Fwiw, catching `NotFoundError` in the CLI feels a bit like a band-aid for API limitations. If we instead update the server-side API (e.g., `_get_sketch_for_admin`) to return basic timeline metadata (just names/IDs from DB, skipping OpenSearch calls) for soft-deleted sketches when accessed by admins, we could avoid these try-catch blocks in the CLI entirely.

WDYT?

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: oh yeah it might be good to at least mention the id and status of the timeline, we do not need to expose the name and description for admins.

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: yep let me fix it

jaegeral: done PTAL

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: DOne ptal, I now return redacted info and I also added another permission check in the get_sketch_for_admin just to be on the safe side

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -328,31 +329,70 @@ def delete_sketch(ctx: click.Context, force_delete: bool) -> None:
...
         ctx.exit(1)
 
+    try:
+        sketch_name = sketch.name
+        sketch_desc = sketch.description
+        sketch_status = sketch.status
+        sketch_labels = sketch.labels
+        timelines = sketch.list_timelines()
+    except NotFoundError as e:  # pylint: disable=unused-variable
+        pass
```
jkppr:
```suggestion
    except NotFoundError:
        pass
```

if you don't use the `e` then you can just drop it. No need to disable lint anymore.

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -328,31 +329,70 @@ def delete_sketch(ctx: click.Context, force_delete: bool) -> None:
...
-    for timeline in sketch.list_timelines():
+    for timeline in timelines:
+        timeline_desc = "N/A"
+        timeline_status = "N/A"
+        try:
+            # timeline.description and timeline.status lazy-load from the API.
+            timeline_desc = timeline.description
+            timeline_status = timeline.status
+        except NotFoundError as e:  # pylint: disable=unused-variable
+            pass
```
jkppr:
```suggestion
        except NotFoundError:
            pass
```

Same as the other one.

jkppr (approved):
2 minor nits, otherwise lgtm

PR #3861, opened 2026-06-22
feat(telemetry): Add search metadata attributes to OpenTelemetry spans`

jkppr:
This PR expands OpenSearch telemetry in Timesketch by logging key search configuration parameters and result metadata to OpenTelemetry traces. This allows operations and dev teams to correlate search performance and modes.

#### 1. Added Telemetry Attributes
* **`timesketch.search.use_wildcard_fields`** (Boolean): Tracks whether wildcard field mode was enabled.
* **`timesketch.search.is_count`** (Boolean): Tracks if the query was a count-only execution.
* **`timesketch.search.enable_scroll`** (Boolean): Tracks if pagination scroll mode was enabled.
* **`timesketch.search.is_dsl`** (Boolean): Tracks if a custom OpenSearch DSL query was passed (from programmatic API clients or analyzers).
* **`timesketch.search.indices_count`** (Integer): The number of timelines/indices queried simultaneously.
* **`timesketch.search.page_size` / `page_offset`** (Integer): Limit and offset boundaries from the search request's pagination filters.
* **`timesketch.search.total_hits`** (Integer): Total matches found in OpenSearch.
* **`timesketch.search.returned_hits`** (Integer): Actual records returned in this page.

#### 2. Technical Architecture & Safety
* **Data Privacy Guarantee:** To prevent leaking sensitive search keywords, IOCs, or PII into telemetry, the actual query string (`query_string`) and raw query DSL (`query_dsl`) are **never** logged to the spans.
* **Non-Blocking Telemetry:** Post-execution telemetry parsing is isolated in its own `try-except` block to ensure telemetry extraction errors **never** propagate or crash a user's search query.
* **Cached Signature Extraction:** The function signature of `search()` is parsed and cached using Python's `inspect.signature` **once** at import-time to avoid dynamic argument-binding CPU overhead during query execution.

#### 3. How has this been tested?
* **Unit Tests:** Added a new test suite under `timesketch/lib/telemetry_test.py` verifying helper behaviors, mock tracing, and spec mock validation.
* **E2E Tests:** Updated `end_to_end_tests/telemetry_test.py` to execute normal vs wildcard searches and verified Jaeger traces receive the correct span attribute configurations.

### DIFF
--- a/timesketch/lib/telemetry_test.py
+++ b/timesketch/lib/telemetry_test.py
@@ -1,0 +1,117 @@
+# Copyright 2026 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for telemetry."""
+
+# pylint: disable=protected-access,unused-argument
+
+from unittest import mock
+
+try:
+    from opentelemetry import trace as otel_trace
+except ImportError:
+    otel_trace = None
+from timesketch.lib import telemetry
+from timesketch.lib.testlib import BaseTest
+
+
+class TelemetryTest(BaseTest):
+    """Tests for telemetry functions."""
+
+    def test_extract_total_hits(self):
+        """Test _extract_total_hits helper function."""
+        # Test int value
+        self.assertEqual(telemetry._extract_total_hits(123), 123)
+
+        # Test invalid types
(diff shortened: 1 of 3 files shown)


timesketch/lib/telemetry_test.py
```diff
@@ -0,0 +1,152 @@
...
+        mock_span.reset_mock()
+        res_count = dummy_search(7, ["index1"], "query", {}, True, None, True, False)
+        self.assertEqual(res_count, 42)
+
+        mock_span.set_attribute.assert_any_call("timesketch.sketch_id", 7)
+        mock_span.set_attribute.assert_any_call(
+            "timesketch.search.use_wildcard_fields", False
+        )
+        mock_span.set_attribute.assert_any_call("timesketch.search.is_count", True)
+        mock_span.set_attribute("timesketch.search.enable_scroll", True)
```
jaegeral: shouldn't you also use assert_any_call here? Because set_attribute will always run? Or is that intended and you only want to set it for the test purpose?

jaegeral (approved):
One small question, otherwise LGTM

timesketch/lib/telemetry_test.py
jkppr: Good catch, yes this should be an assert_any_call.

PR #3864, opened 2026-06-24
Fix race condition in `_set_timeline_status` on new imports

jkppr:
This PR addresses a race condition that occurs during new timeline imports (specifically CSV/JSONL imports).

**Problem:**
When a new import is initiated, the task updates the datasource status to `"processing"`, which in turn calls `_set_timeline_status`. Previously, `_set_timeline_status` unconditionally attempted to refresh the corresponding OpenSearch index. However, for CSV/JSONL imports, the index is created *after* this `"processing"` status is set. This resulted in a guaranteed `NotFoundError` (404) and a 4-second delay due to the retry loop (5 attempts with 1-second sleep) before the import could proceed.

**Solution:**
Wrapped the index refresh logic in `_set_timeline_status` with a check to only execute when the status is `"ready"`. The index is now only refreshed when the timeline is fully populated and ready for analysis, avoiding unnecessary errors and delays during the initial processing phase.

### Testing
*   **Unit Tests:** Ran `pytest timesketch/lib/` inside the dev container (302 passed, 108 skipped).
*   **End-to-End Tests:** Ran the full E2E test suite inside the container (80/80 tests passed, including `import_plaso_filter_test` and `status_race_test`).

### DIFF
--- a/timesketch/lib/tasks.py
+++ b/timesketch/lib/tasks.py
@@ -291,33 +291,34 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
     )
 
     # Refresh the index so it is searchable for the analyzers right away.
-    datastore = OpenSearchDataStore()
-    # Retry refreshing the index a few times if it fails.
-    index_name = timeline.searchindex.index_name
-    for i in range(5):
-        try:
-            datastore.client.indices.refresh(index=index_name)
-            break  # Success
-        except Exception as e:  # pylint: disable=broad-except
-            if i == 4:  # Last attempt
-                logger.error(
-                    "Unable to refresh index: %s in sketch (ID: %d). "
-                    "Gave up after 5 attempts. Error: %s",
-                    index_name,
-                    sketch_id,
-                    str(e),
-                    exc_info=True,
-                )
-            else:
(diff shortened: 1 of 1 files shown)


timesketch/lib/tasks.py
```diff
@@ -291,33 +291,34 @@ def _set_timeline_status(timeline_id: int, status: Optional[str] = None):
...
-                logger.debug(
-                    "Attempt %d to refresh index %s in sketch (ID: %d)failed: %s",
-                    i + 1,
-                    index_name,
-                    sketch_id,
-                    str(e),
-                )
-                time.sleep(1)  # Wait a second before retrying
+    if status == "ready":
+        datastore = OpenSearchDataStore()
```
jaegeral, line 300:
```suggestion
        if not timeline.searchindex:
            logger.error(
                "Timeline (ID: %d) has no associated search index.", timeline_id
            )
            return
        datastore = OpenSearchDataStore()
```

jaegeral: suggestion based on the gemini comment below

jaegeral (changes requested):
otherwise LGTM

PR #3866, opened 2026-06-25
fix(cli): add config_section and imports

jaegeral:
Fixing the cli-client to add `config_section` and improving various imports.

I've updated the ignored-argument-names configuration inside .pylintrc by appending |^ctx$ to it to ignore docstring checks in `cli.py`


cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -77,7 +81,9 @@ def __init__(
                 sys.exit(1)
 
         self.config_assistant = timesketch_config.ConfigAssistant()
-        self.config_assistant.load_config_file(conf_file, load_cli_config=True)
+        self.config_assistant.load_config_file(
+            conf_file, section=config_section, load_cli_config=True
```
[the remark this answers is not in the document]
jaegeral: done

cli_client/python/timesketch_cli_client/commands/events.py
```diff
@@ -101,7 +108,14 @@ def annotate(ctx, timeline_id, event_id, tag, comment):
 )
 @click.pass_context
 def add_event(ctx, message, date, attributes, timestamp_desc):
-    """Add an event to the sketch."""
+    """Add an event to the sketch.
+    Args:
```
jkppr, line 124: Missing blank line

```suggestion
    """Add an event to the sketch.

    Args:
```

cli_client/python/timesketch_cli_client/commands/config.py
```diff
@@ -37,8 +37,8 @@ def get_config_parameter(ctx: click.Context, name: str) -> None:
     """Get the value of a configuration parameter.
 
     Args:
-        ctx: Click CLI context object.
-        name: Name of the configuration parameter to get.
+        ctx (click.Context): Click CLI context object.
```
jkppr: More of a general question: We already have typing in the method. Do we really need them in the doc string too then? Feels a bit too much tbh

cli_client/python/timesketch_cli_client/commands/events.py
```diff
@@ -153,15 +167,7 @@ def tag_mod(
...
         tag (Optional[str]): The tag to remove from the event, or a comma-separated
-                             list of tags. If None, lists the event's tags.
-
-    Errors:
-        * If the specified timeline or event does not exist.
-        * If the specified event does not exist.
-
-    Outputs:
-        Text: If a tag is specified, a message indicating the tag was removed.
-              If no tag is specified, the event's current tags are printed.
```
jkppr, line 164: Why remove this? Are the errors not part of this method anymore?

cli_client/python/timesketch_cli_client/commands/events.py
```diff
@@ -40,6 +40,13 @@ def annotate(ctx, timeline_id, event_id, tag, comment):
 
     If neither tag or comment are specified, the command will return the
     current annotations for the event.
+
+    Args:
```
jkppr, line 50: Since the CLI uses click, doesn't this all get provided when called via the `-h` or `--help` flag? Is this intentional?

e.g. If a user runs `timesketch events add-event --help` in their terminal, the output will now literally print the `Args:` section. The user will be told to provide `ctx (click.Context)`, which is an internal framework object they cannot even pass via the CLI.

cli_client/python/timesketch_cli_client/commands/search_test.py
```diff
@@ -68,7 +68,10 @@ def test_describe_non_existent_saved_search(self):
 
     @mock.patch("timesketch_cli_client.commands.search.search.Search")
     def test_search_describe_flag(self, MockSearchClient):
-        """Test the 'search --describe' command."""
+        """Test the 'search --describe' command.
+        Args:
+            MockSearchClient (mock.MagicMock): MockSearchClient parameter.
```
jkppr: For this and the other args added in the test files: Does this really add any development info that is relevant? 

Also iirc this will be printed as the test name currently executed when you run pytest in verbose mode. This seems a bit excessive?

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -77,7 +84,9 @@ def __init__(
...
```
jkppr, line 119: This binds this specific version of `timesketch-cli-client` to a version of `timesketch-api-client` where `load_config_file` and `get_client` both explicitly accept the new `section=` parameter. Is this compatible with older API clients? Or will this fail?

jkppr (changes requested):
First round of review

cli_client/python/timesketch_cli_client/cli.py
jaegeral: hm good point, let me put it in a try catch block in case the parameter is not there to fallback

cli_client/python/timesketch_cli_client/commands/events.py
jaegeral: yeah let me remove the ctx argument from docstring

cli_client/python/timesketch_cli_client/commands/config.py
jaegeral: will address

cli_client/python/timesketch_cli_client/commands/events.py
jaegeral: by accident

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -395,7 +383,9 @@ def create_story(ctx: click.Context, title: str) -> None:
 @sketch_group.command("list-stories", help="List all stories in the sketch.")
 @click.pass_context
 def list_stories(ctx: click.Context):
-    """List all stories in the sketch."""
+    """List all stories in the sketch.
+    Args:
```
jkppr: This seems not needed anymore if there are no args listed.

```suggestion
```

cli_client/python/timesketch_cli_client/commands/sketch.py
jkppr, line 1: Minor format nit: Check the Args sections again! You have removed the `ctx` lines but kept the empty Args and also second lines like `client and output format.` which are now out of context....

jkppr (changes requested):

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: done, I have also re-checked (by hand ;) ) all the other commands

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -52,32 +56,71 @@ def __init__(
...
+                        config_path=conf_file,
+                        config_section=config_section,
+                        load_cli_config=True,
+                    )
+                else:
+                    if config_section != DEFAULT_CONFIG_SECTION:
+                        click.echo(
+                            "WARNING: The installed timesketch-api-client does not "
+                            "support custom config sections. Falling back to default."
+                        )
```
jkppr, line 94: Falling back to the default config section if the requested one isn't supported seems a bit risky to me. If a user explicitly asks for a specific section (like `prod`) and it silently falls back to default (which might be `dev`), they could run commands on the wrong server.

I suggest we should fail and exit with an error here instead of warning and continuing.

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -52,32 +56,71 @@ def __init__(
...
         self.output_format_from_flag = output_format_from_flag
 
         if not api_client:
             try:
-                # TODO: Consider other config sections here as well.
-                self.api = timesketch_config.get_client(
-                    config_path=conf_file, load_cli_config=True
-                )
+                try:
+                    sig = inspect.signature(timesketch_config.get_client)
```
jkppr: Using `inspect.signature` to check for dependency features at runtime feels a bit fragile. Since both the CLI and API clients live in this same repo, can we just bump the required version of `timesketch-api-client `in `cli_client/python/setup.py` instead? That way we can assume the feature is present and clean up this check.

jkppr (changes requested):

cli_client/python/timesketch_cli_client/cli.py
jaegeral: done

cli_client/python/timesketch_cli_client/cli.py
jaegeral: done

.pylintrc
```diff
@@ -673,7 +673,7 @@ dummy-variables-rgx=_+$|(_[a-zA-Z0-9_]*[a-zA-Z0-9]+?$)|dummy|^ignored_|^unused_
 
 # Argument names that match this expression will be ignored. Default to name
 # with leading underscore.
-ignored-argument-names=_.*|^ignored_|^unused_
+ignored-argument-names=_.*|^ignored_|^unused_|^ctx$
```
jkppr, line 676: If Click context is not needed in a command, isn't it enough to remove `@click.pass_context`, rename the parameter to `_ctx`, or explicitly disable `pylint` locally on the function with `# pylint: disable=unused-argument`?

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -148,8 +158,16 @@ def output_format(self):
...
     help="Path to the config file.",
 )
+@click.option(
+    "--config-section",
+    "config_section",
+    default=DEFAULT_CONFIG_SECTION,
+    show_default=True,
+    required=False,
+    help="The config section to use.",
+)
```
jkppr, line 198: Adding `--config-section` is nice. A quick thought on tests & error handling:

1. Could we add a `CliRunner` test in `config_test.py` to verify passing `--config-section` via actual CLI command invocation?
2. How does this behave if a user passes a non-existent section name? We might want to catch that cleanly so it gives a friendly error message instead of falling back or throwing a traceback.

jkppr (approved):
General LGTM. Just 2 small nits below:

.pylintrc
jaegeral: It is needed, but I do not want to have it mentioned in docstring as a param

jkppr: Ah, I remember. Yeah, ignore my comment then.

cli_client/python/timesketch_cli_client/cli.py
jaegeral: good point

.pylintrc
jaegeral: as discussed offline, we had this conversation before my leave

cli_client/python/timesketch_cli_client/cli.py
```diff
@@ -33,42 +38,73 @@
...
         self.api = api_client
         self.sketch_from_flag = sketch_from_flag
         self.output_format_from_flag = output_format_from_flag
 
         if not api_client:
+            # Check if the config file exists and contains the requested section.
+            actual_conf_file = conf_file
+            if not actual_conf_file:
+                home_path = os.path.expanduser("~")
+                actual_conf_file = os.path.join(home_path, ".timesketchrc")
```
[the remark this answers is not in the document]
jaegeral: done

tomchop (approved):

PR #3868, opened 2026-06-25
reuse existing sketch if name is given

### LINKED ISSUE #3625: timesketch_importer reuse existing sketch by name

cvandeplas:
## What existing problem does this PR solve?

`timesketch_importer.py` recreates new sketches when a `sketch_name` is given, instead of reusing the existing sketch.

## What new feature is being introduced with this PR?

Reuse of an existing sketch when a name is given. If the sketch does not exist, create a new one with that name (original behaviour).

This is a reworked version of #3626, addressing all review feedback from @jkppr and @jaegeral.

## Changes

- **`api_client/python/timesketch_api_client/client.py`**: Added `get_sketches_by_name()` method to the `TimesketchApi` class, returning a list of matching sketches. Conflict resolution (duplicates) is the caller'\''s responsibility.
- **`api_client/python/timesketch_api_client/sketch.py`**: Added `created_at` and `creator` properties to the `Sketch` class.
- **`importer_client/python/tools/timesketch_importer.py`**: Uses the new API method and adds a `--sketch-strategy` flag (`ask`/`newest`/`oldest`) to handle duplicate sketch names.
- **`docs/guides/user/upload-data.md`**: Documented the new `--sketch-name` and `--sketch-strategy` flags.

## Design decisions

- **Sketch lookup lives in the API client, not the importer.** Per review feedback, `get_sketches_by_name()` is a generic operation useful beyond the importer, so it belongs in `TimesketchApi`.
- **Returns a list, not a single sketch.** Sketch names are not unique. The method returns all matches and raises `KeyError` if none are found. Callers decide how to resolve duplicates.
- **Case-sensitive matching.** Exact name match avoids accidental collisions (e.g. "Case-A" vs "case-a").
- **`--sketch-strategy` flag for automation.** In interactive mode (default `ask`), the user is prompted to select. For automated pipelines, `newest` or `oldest` can be passed to avoid interactive prompts.

## Checks

- [x] All tests succeed
- [x] Unit tests added (`get_sketches_by_name`, `created_at`, `creator`, duplicate handling, case sensitivity)
- [ ] e2e tests added
- [x] Documentation updated

## Closing issues

closes #3625


api_client/python/timesketch_api_client/client.py
```diff
@@ -688,6 +688,31 @@ def get_sketch(self, sketch_id):
...
+
+        Raises:
+            KeyError: If no sketch with the specified name is found.
+
+        Returns:
+            list[sketch.Sketch]: A list of sketch objects.
+        """
+        sketches = [
+            sketch_obj
+            for sketch_obj in self.list_sketches()
```
jkppr: Calling `list_sketches()` without arguments pulls all sketches from the server and filters them in memory. If a user has thousands of sketches, this is going to be super slow and trigger a ton of API requests.

Can you update `list_sketches` to accept `search_query` and use the backend's `scope="search"`. That way the server does the heavy lifting, and we just do exact matching on a few candidates in python.

jkppr: By default, `list_sketches()` uses `scope="user"`, meaning we only search for sketches owned by the current user. If someone is trying to import data into a sketch that was shared with them, they won't find it.

If you switch this to use server-side search (`scope="search"`), it will automatically look at all accessible sketches (owned and shared).

api_client/python/timesketch_api_client/sketch.py
```diff
@@ -161,6 +161,42 @@ def last_activity(self):
...
         return meta.get("last_activity", "")
 
+    @property
+    def created_at(self):
+        """Property that returns sketch creation time.
+
+        Returns:
+            str: Sketch creation time as string.
+        """
+        data = self.lazyload_data(refresh_cache=True)
```
jkppr: Why are you using `refresh_cache=True` here? Unlike ACLs or attributes, `created_at` and `creator` are static and will never change after the sketch is created.

Forcing a refresh here triggers redundant API calls every time we access these properties (like when sorting duplicates or listing them in the importer).

importer_client/python/tools/timesketch_importer.py
```diff
@@ -636,10 +652,51 @@ def main(args=None):
...
+                    print("Multiple sketches found with the name '{0:s}':".format(sketch_name))
+                    for s in sketches:
+                        print(" - [{0:d}] created_at: {1:s}, by {2:s}".format(
+                            s.id, s.created_at, s.creator))
+
+                    selected_option = cli_input.ask_question(
+                        "Select the sketch to use by entering the corresponding number",
+                        input_type=int,
+                        default=sketches[0].id,
+                    )
```
jkppr: You are passing `input_type=int` directly to `ask_question`. If the user makes a typo and enters a non-integer (like `'abc'` or just hits enter), this is going to raise a `ValueError` and crash the importer. Can you wrap this in a try-except to handle typos gracefully and re-prompt.

jkppr (changes requested):
Thanks for the contribution and following up on the earlier PR. This is the first round of review.

jkppr:
Please also ensure pylint & black are happy. You can run this in your dev setup:
```
docker exec -it -w /usr/local/src/timesketch timesketch-dev black <files you changed>

docker exec -it -w /usr/local/src/timesketch timesketch-dev pylint --rcfile=.pylintrc <files you changed>
```

importer_client/python/tools/timesketch_importer.py
cvandeplas: The ask_question call is now wrapped in a while loop with a try/except ValueError. If the user enters something that can't be converted to an integer (typo, empty string, etc.), it prints a helpful message and re-prompts instead of crashing.

api_client/python/timesketch_api_client/client.py
cvandeplas: Excellent point. The list_sketches function is extended with an optional search_query, with scope selection. Considering search_query does partial matching (behind the scenes) on sketch name and description, a final filter is applied here to ensure only matching sketch names remain.

cvandeplas: Excellent point, I've covered this with the resolution to above comment.

api_client/python/timesketch_api_client/sketch.py
cvandeplas: I used the same code-logic that was applied in the surrounding functions like `last_activity()` , `my_acl()`. They also `refresh_cache=True` .  
Your point that the values will not change over time is valid, I therefore removed it in the `created_at()` and `creator()` functions.

cvandeplas:
Thanks @jkppr for the input. I think I've incorporated it in this update.

cvandeplas:
linting fixed

PR #3869, opened 2026-06-27
Update SecGemini provider to match sec-gemini v3 SDK APIs

samomi-oss:
**IMPORTANT: All Pull Requests should be connected to an issue, if you don't
have an issue, please start by creating an issue and link it to the PR.**

**Checks**

- [x] All tests succeed.
- [ ] Unit tests added.
- [ ] e2e tests added.
- [x] Documentation updated.

**Closing issues**

Put `closes #XXXX` in your comment to auto-close the issue that your PR fixes
(if such).

### DIFF
--- a/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
+++ b/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
@@ -191,8 +191,10 @@ class SecGeminiLogAnalyzer(interface.LLMProvider):
                                         desc = f.get("description", "")
                                         relevance = f.get("relevance", "")
                                         combined_conclusion = (
-                                            f"{CONCLUSION_LOG_RECORDS_HEADER}{desc}\n\n<br>\n\n"
-                                            f"{CONCLUSION_RELEVANCE_HEADER}{relevance}"
+                                            f"{CONCLUSION_LOG_RECORDS_HEADER}"
+                                            f"{desc}\n\n<br>\n\n"
+                                            f"{CONCLUSION_RELEVANCE_HEADER}"
+                                            f"{relevance}"
                                         )
                                         annotations = [
                                             {
--- a/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
+++ b/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
@@ -17,7 +17,6 @@ import json
 import logging
 import os
 import asyncio
-import inspect
 import pathlib
 import tempfile
 from datetime import datetime
@@ -31,11 +30,16 @@ from timesketch.lib.llms.providers import manager
 has_required_deps = True
 try:
     from sec_gemini import SecGemini
-    from sec_gemini.models.enums import MessageType
 except ImportError:
     has_required_deps = False
 
 logger = logging.getLogger(__name__)
+
+
+CONCLUSION_LOG_RECORDS_HEADER = "**What the selected log records are:**\n"
+CONCLUSION_RELEVANCE_HEADER = (
+    "**How the selected log records are related to the finding:**\n"
+)
 
 
 class SecGeminiLogAnalyzer(interface.LLMProvider):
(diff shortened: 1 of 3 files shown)


timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -114,105 +105,119 @@ async def _run_async_stream(self, log_path, prompt):
...
+                    msg_type = response.get("message_type")
+                    source = response.get("source")
+                    content = response.get("content", "")
+
+                    if msg_type == "MESSAGE_TYPE_RESPONSE":
+                        start_marker = "\n## Investigation Findings\n\n"
+                        if start_marker in content:
+                            json_str = content.split(start_marker)[1].strip()
+                            try:
+                                findings_list = json.loads(json_str)
```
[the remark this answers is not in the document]
samomi-oss: secgemini agent won't do that

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -114,105 +105,119 @@ async def _run_async_stream(self, log_path, prompt):
...
-        self._session.upload_and_attach_logs(
-            log_path, custom_fields_mapping=self.custom_fields_mapping
-        )
-        if hasattr(self._session, "logs_table") and hasattr(
-            self._session.logs_table, "blake2s"
-        ):
-            self.table_hash = self._session.logs_table.blake2s
+        try:
+            # Create session via the sessions manager
+            self._session = await self.sg_client.sessions.create()
```
jkppr, line 109: This omits the session_params we had in the past (e.g. model & logging).
Does the v3 SDK `sessions.create()` method support these parameters? If they are no longer supported at session creation, can they be passed via meta in `self._session.prompt(prompt, meta=...)`?

If they are indeed no longer supported, they should be removed from __init__ and the provider's configuration definition to avoid misleading users.

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -114,105 +105,119 @@ async def _run_async_stream(self, log_path, prompt):
...
-        self.session_id = self._session.id
-        # TODO: Could we check if the API key has logging enabled and if not ERR
-        logger.info("Started new SecGemini session: '%s'", self._session.id)
-        self._session.upload_and_attach_logs(
-            log_path, custom_fields_mapping=self.custom_fields_mapping
-        )
-        if hasattr(self._session, "logs_table") and hasattr(
-            self._session.logs_table, "blake2s"
-        ):
-            self.table_hash = self._session.logs_table.blake2s
```
jkppr, line 142: Since you remove this here, please make sure this is nor referenced in other parts of this file. e.g. here: https://github.com/google/timesketch/blob/master/timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py#L315

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -114,105 +105,119 @@ async def _run_async_stream(self, log_path, prompt):
...
+                                            "annotations": annotations,
+                                        })
+                                    json_str = json.dumps({"findings": translated_findings})
+                            except Exception as e:
+                                logger.error(
+                                    "Failed to translate SecGemini findings JSON: %s",
+                                    e,
+                                    exc_info=True,
+                                )
+                            yield json_str
```
jkppr, line 224: Is it expected to still yield json_str here even when json.loads fails? Wouldn't this contain a raw untranslated list then and will fail when attempting to parse and call it later?

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -70,19 +70,10 @@ def __init__(self, config: dict, **kwargs: Any):
         if self.server_url:
             os.environ["SEC_GEMINI_LOGS_PROCESSOR_API_URL"] = self.server_url
 
-        self.base_url = self.config.get("base_url")
-        self.wss_url = self.config.get("wss_url")
```
jkppr, line 74: The removal of those values will break compatibility with older versions, is that intentional?

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -114,105 +105,119 @@ async def _run_async_stream(self, log_path, prompt):
...
-        ):
-            self.table_hash = self._session.logs_table.blake2s
+        try:
+            # Create session via the sessions manager
+            self._session = await self.sg_client.sessions.create()
+            self.session_id = self._session.id
+            logger.info("Started new SecGemini session: '%s'", self.session_id)
+
+            # Upload logs file using the new files.upload API
+            await self._session.files.upload(str(log_path), content_type="text/plain")
```
jkppr, line 114: The old upload supported a custom mapping (e.g. `datetime` to `timesketch`). Is this still supported and needed? Or does this happen on the SecGemini side automatically now?

jkppr (changes requested):
Please take a look at the comments below and take care of the complaining linter & formatter.

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
jkppr: If those settings are not supported anymore, can you also update the default config accordingly please: https://github.com/google/timesketch/blob/master/data/timesketch.conf#L462

jkppr (approved):
Thanks heaps for the contribution 👍

PR #3874, opened 2026-07-01
Transition Gunicorn WSGI Workers to Threaded Model (gthread)

jkppr:
This PR transitions the Gunicorn WSGI server from `sync` workers to the threaded `gthread` worker class to prevent CPU-bound request serialization (e.g., timeline exports, large searches, high QPS) from blocking the server.

### Changes
*   **Gunicorn Config (`timesketch/gunicorn.conf.py`):** Added dynamic configuration for `worker_class`, `threads`, and `workers` via environment variables (`WSGI_WORKER_CLASS`, `NUM_WSGI_THREADS`, `NUM_WSGI_WORKERS`). Implemented safe parsing to handle empty/missing environment variables without crashing Gunicorn on startup.
*   **Release Config (`docker/release/`):** Defaulted release container entrypoint to `gthread` (4 threads). Exposed these variables in `config.env` and mapped them in `docker-compose.yml`.
*   **E2E Config (`docker/e2e/`):** Defaulted E2E container entrypoint to `gthread` (2 threads) so tests match the production model.
*   **Dev Config (`docker/dev/`, `utils/tsdev.sh`, `contrib/Tiltfile`):** Defaulted dev environment to `gthread` (2 threads, 2 workers) via `docker-compose.yml`. Removed hardcoded `--workers 4` arguments from `tsdev.sh` and `Tiltfile` to let the configuration file control allocation.

### Verification
*   All E2E tests passed successfully (80/80).
*   Verified via container logs and active processes (`ps -ef`) that Gunicorn initializes with `gthread` and the correct thread/worker counts.

### DIFF
--- a/timesketch/gunicorn.conf.py
+++ b/timesketch/gunicorn.conf.py
@@ -68,9 +68,9 @@ def get_debug_status():
 DEBUG = get_debug_status()
 
 # Worker Configuration
-worker_class = os.environ.get("WSGI_WORKER_CLASS") or "sync"
+worker_class = os.environ.get("WSGI_WORKER_CLASS") or "gthread"
 threads_val = os.environ.get("NUM_WSGI_THREADS")
-threads = int(threads_val) if threads_val and threads_val.isdigit() else 1
+threads = int(threads_val) if threads_val and threads_val.isdigit() else 4
 workers_val = os.environ.get("NUM_WSGI_WORKERS")
 workers = int(workers_val) if workers_val and workers_val.isdigit() else 1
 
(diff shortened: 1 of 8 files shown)


timesketch/gunicorn.conf.py
```diff
@@ -67,6 +67,13 @@ def get_debug_status():
 
 DEBUG = get_debug_status()
 
+# Worker Configuration
+worker_class = os.environ.get("WSGI_WORKER_CLASS") or "sync"
```
tomchop: Don't you want the default to be `gthread`? it's what you're setting in docker/dev/docker-compose.yml

tomchop (approved):

timesketch/gunicorn.conf.py
jkppr: This is the absolute fallback option if someone manages to spin up TS without any of the other configs applied. But I guess it's a fair point to change this to gthreads too.

PR #3875, opened 2026-07-02
adding byot functionality to secgemini investigations

samomi-oss:
This PR adds the local BYOT log search functionality to the SecGemini log analyzer agent in Timesketch.

PR #3878, opened 2026-07-09
Implement Sec-Gemini BYOT Log Analysis Support

jkppr:
This PR introduces support for local Bring Your Own Tool (BYOT) log analysis utilized by the Sec-Gemini agent, optimizes search aggregations, and implements transparent sketch access control (auto-sharing) with automatic cleanup.

The BYOT client is run in a separate lightweight docker container from the `contrib/secgemini_byot/` folder. It takes care of the reverse tunnel to the Sec-Gemini hub and implements the compatible Timesketch LogStore class that exposes the necessary search tools for Sec-Gemini to investigate a Sketch.

### 1. Client-side BYOT Connection Tunnel (`contrib/secgemini_byot/`)
* **Connection Tunnel:** Created a Docker-containerized client-side connection tunnel (`byot_client.py`) that establishes a secure outbound reverse websocket connection to the Sec-Gemini Hub.
* **Local MCP Server:** Uses the `FastMCP` framework to expose standard search and type-describe tools executing queries directly against the local Timesketch API (using the Python `timesketch-api-client`), bypassing any requirement to upload raw log files to the cloud.
* **Documentation:** Added a detailed `README.md` with build/run guidelines, Docker-Compose network configurations, and troubleshooting tips.

### 2. Date Histogram Aggregator Optimization (`timesketch_logstore.py`)
* **Dynamic Bounds Querying:** Optimized log type discovery in the BYOT client. Instead of executing OpenSearch `date_histogram` queries over a default 130-year span (which blocks queues for large sketches), the store now runs lightweight queries to discover the earliest and latest log events per data type.
* **Mid-night Buffer Align:** Applies a `+/- 1 day` buffer aligned to midnight (`00:00:00`) to dynamically scope histogram boundaries.

### 3. Sketch Auto-Sharing & Temporary Collaborator Access
* **ACL Resolution Wrappers:** Implemented `grant_permission_by_username()` and `revoke_permission_by_username()` on the SQL model `AccessControlMixin` (`acl.py`) with domain-stripping fallback to manage sketch permissions using LDAP/SSO strings.
* **Transparent sharing:** Integrated auto-sharing into the Sec-Gemini provider. When BYOT is active and `byot_username` is configured, the sketch is automatically shared (read-only) with the bot user at task startup.
* **Early Abort:** Validates that the configured bot user exists at startup, failing early with a descriptive config error instead of throwing a generic connection error during streaming.
* **Lifecycle Cleanup:** Implemented a generator-level `finally` block to automatically revoke the granted sketch permission once the streaming session completes, errors out, or is cancelled, preventing bot accounts from accumulating permanent sketch access.

### 4. Core Bugfix: Group Revocation Leak (`acl.py`)
* **Bug:** In `revoke_permission()`, the `_get_ace()` call was checking group membership by default. If a user had no direct sketch ACE but belonged to an allowed group, the query returned the group's ACE, causing the loop to delete the group entry—meaning **the entire group lost access to the sketch**.
* **Fix:** Updated the `revoke_permission()` call to specify `check_group=False`, targeting only the user's direct Access Control Entries.

### 5. Tests
* Added comprehensive coverage in `acl_test.py`:
  * Verifies username-based grant/revoke, duplicate grant protections, and email domain stripping fallback.
  * Verifies that revoking a user who only holds group-level permissions does not delete the group ACE (`test_revoke_user_with_group_permission`).

### DIFF
--- a/contrib/secgemini_byot/byot_client.py
+++ b/contrib/secgemini_byot/byot_client.py
@@ -1,0 +1,130 @@
+# Copyright 2026 Google Inc. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Entrypoint script for the client-side SecGemini BYOT container."""
+
+import asyncio
+import inspect
+import logging
+import os
+import sys
+
+from fastmcp import FastMCP
+from mcp.types import ToolAnnotations
+from sec_gemini.byot import ByotService
+from sec_gemini.logs_mcp.backends.sqlite import multi_db_tools
+from timesketch_api_client import client as ts_client
+
+from timesketch_logstore import DynamicLogStoreMap
+
+# Setup logging to stdout
+logging.basicConfig(
+    level=logging.INFO,
+    format="%(asctime)s [%(levelname)s] %(name)s: %(message)s",
+    stream=sys.stdout,
+)
+logger = logging.getLogger("timesketch.byot.client")
+
+
+def get_required_env(name: str) -> str:
+    val = os.getenv(name)
+    if not val:
+        logger.error("Missing required environment variable: %s", name)
+        sys.exit(1)
+    return val
+
+
+async def main():
+    """Main execution function for the BYOT client container."""
+    # Read configuration from environment
+    timesketch_url = get_required_env("TIMESKETCH_URL")
+    username = get_required_env("TIMESKETCH_USERNAME")
+    password = get_required_env("TIMESKETCH_PASSWORD")
+    secgemini_api_key = get_required_env("SEC_GEMINI_API_KEY")
+    tunnel_name = os.getenv("TUNNEL_NAME", "byot-sec-gemini-bot")
+
+    logger.info("Initializing Timesketch API client for %s...", timesketch_url)
+    try:
+        api = ts_client.TimesketchApi(
+            host_uri=timesketch_url,
+            username=username,
+            password=password,
+        )
+        # Attempt to fetch list of sketches to verify connection/credentials
+        sketches = list(api.list_sketches(scope="all"))
+        logger.info(
+            "Successfully authenticated. User has access to %d sketches.",
+            len(sketches),
+        )
+    except Exception:  # pylint: disable=broad-exception-caught
+        logger.exception("Failed to authenticate to Timesketch API")
+        sys.exit(1)
+
+    logger.info("Configuring LogStore map...")
+    # Monkeypatch the SQLite multi_db_tools LOG_STORES map with our API
+    # client-backed version.
+    multi_db_tools.LOG_STORES = DynamicLogStoreMap(api)
+
+    # Initialize FastMCP server
+    logger.info("Setting up FastMCP server: %s", tunnel_name)
+    mcp = FastMCP(tunnel_name)
+
+    # Register the tools from multi_db_tools
+    mcp.tool(
+        description=inspect.getdoc(multi_db_tools.describe_available_logs),
+        annotations=ToolAnnotations(readOnlyHint=True),
+    )(multi_db_tools.describe_available_logs)
+
+    mcp.tool(
+        description=inspect.getdoc(multi_db_tools.search_logs),
+        annotations=ToolAnnotations(readOnlyHint=True),
+    )(multi_db_tools.search_logs)
+
+    secgemini_host = os.getenv("SEC_GEMINI_HOST")
+
+    # Initialize and start BYOT Tunnel Service
+    logger.info("Starting BYOT reverse tunnel connection to SecGemini...")
+    byot_kwargs = {
+        "api_key": secgemini_api_key,
+        "name": tunnel_name,
+        "max_retries": -1,
+    }
+    if secgemini_host:
+        logger.info("Using custom SecGemini Hub host: %s", secgemini_host)
+        byot_kwargs["hub_url"] = secgemini_host
+
+    byot = ByotService(**byot_kwargs)
+
+    try:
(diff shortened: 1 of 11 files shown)


timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -275,6 +313,88 @@ def generate_stream_from_logs(
...
+            finally:
+                if granted_by_automation:
+                    try:
+                        sketch.revoke_permission_by_username("read", self.byot_username)
+                        logger.info(
+                            "Automatically revoked read permission on sketch "
+                            "%d from bot user %s",
+                            sketch.id,
+                            self.byot_username,
+                        )
```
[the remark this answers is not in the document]
jkppr: Not an issue since we block further executions of the sec-gemini analysis once the first started.

jkppr:
@samomi-oss could you please take a look at this implementation and test if it sufficiently replaces the proposal you had in #3875 ?

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -275,6 +313,91 @@ def generate_stream_from_logs(
...
+        if not self.upload_logs_to_secgemini:
+            if not sketch:
+                raise ValueError("Local log analysis (BYOT) requires a sketch object.")
+
+            granted_by_automation = False
+            if self.byot_username:
+                try:
+                    success = sketch.grant_permission_by_username(
+                        "read", self.byot_username
+                    )
```
samomi-oss, line 334: what if two users run AI investigation at the same time on the same sketch?

sketch.grant_permission_by_username returns False for the 2nd user as bot's permission is already granted by the 1st user.

contrib/secgemini_byot/byot_client.py
```diff
@@ -0,0 +1,133 @@
...
+            len(sketches),
+        )
+    except Exception:  # pylint: disable=broad-exception-caught
+        logger.exception("Failed to authenticate to Timesketch API")
+        sys.exit(1)
+
+    logger.info("Configuring LogStore map...")
+    # Monkeypatch the SQLite multi_db_tools LOG_STORES map with our API
+    # client-backed version.
+    multi_db_tools.LOG_STORES = DynamicLogStoreMap(api)
```
samomi-oss, line 79: When multiple analysts trigger investigations simultaneously on different sketches, the bot user is granted access to all of them at once. A hallucinating agent or a prompt-injected session from Analyst A could query the logs of Analyst B's sketch by simply passing the other sketch ID. The bot user will authorize it because it currently has access to both.

contrib/secgemini_byot/byot_client.py
```diff
@@ -0,0 +1,133 @@
...
+
+    byot = ByotService(**byot_kwargs)
+
+    try:
+        await byot.start(tools=[mcp])
+        logger.info("Waiting for tunnel connection to establish...")
+        await byot.wait_connected()
+        logger.info("Tunnel connected successfully! Listening for tool calls...")
+
+        # Keep running until terminated
```
samomi-oss: This might have some reliability issues if for some reason the connection gets distrubted and does not recover.

samomi-oss: consider having one tunnel per sketch

contrib/secgemini_byot/byot_client.py
samomi-oss: consider having one tunnel per sketch

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
```diff
@@ -144,21 +153,52 @@ async def _run_async_stream(self, log_path, prompt):
...
+                        if key in ("sec-gemini-enabled-byots", "config.mode", "mode"):
+                            continue
                         if isinstance(val, bool):
                             meta[key] = "true" if val else "false"
                         else:
                             meta[key] = str(val)
 
-                await self._session.prompt(prompt, meta=meta)
+                if not self.upload_logs_to_secgemini:
+                    meta["sec-gemini-enabled-byots"] = self.byot_tunnel_name
```
samomi-oss, line 167: This limits an investigation to only one tunnel when there exists multiple byot tunnels.
In your case, you have only one tunnel (with multi_db). In my PR, I had one tunnel (using single_db) per investigation and that's why I was using this. Anyway, in your case it does not matter to use it or not

timesketch/lib/llms/providers/secgemini_log_analyzer_agent.py
jkppr: In theory the problem happens with the removal of the user when the 1st run ends and the second suddenly has no access anymore. During initialization the 2nd run would just notice the sketch is already shared and just proceed.

While it's technically possible to run the investigation in parallel, at least in the UI we prevent this path and once the first session is started a second can't be triggered. 

Therefore I'll acknowledge the risk, but will keep it as is to not over-optimize before any real problem happens.

contrib/secgemini_byot/byot_client.py
jkppr: That’s technically true. If two sessions run simultaneously, `sec-gemini-bot` briefly holds read permissions on both sketches in the DB at once.

In practice, pulling cross-sketch data would require a pretty specific alignment: simultaneous execution plus either a numeric hallucination or a prompt injection inside Sketch A that targets the exact `ticket_id` currently active in Sketch B.

To my understanding, using one tunnel per sketch does not mitigate this risk, since API calls can be made via any tunnel as a tunnel is not linked to a specific sketch.

For the initial version of the BYOT implementation we can accept the risk in my opinion. Timesketch is used in trusted environments only and it requires a specific timing + malicious employee to trigger the abuse.

If strict multi-tenant isolation becomes a requirement down the road, we need to look into user specific authentications, or enforcing session-to-sketch scoping at the tool or gateway level. For prompt injection and hallucination protection I will need defer to the Sec-Gemini side.

contrib/secgemini_byot/byot_client.py
jkppr: Good point. I've added a health check and will utilize the restart option of the container to re-try.

contrib/secgemini_byot/byot_client.py
samomi-oss: > a tunnel is not linked to a specific sketch.

We can set up a byot tunnel for one specific sketch. We can discuss more later for future upgrades

PR #3888, opened 2026-07-30
feat(explore): add PPL and SQL as search modes

### LINKED ISSUE #3887: Support PPL and SQL queries within Timesketch

A search with Lucene (query search/wildcard) provides the list of events which contain the keyword. It does not allow analysts to obtain aggregated results or to perform extraction or transformation as part of the investigation process to get more precise results. Analysts who need those answers today can query OpenSearch directly, which loses the sketch scoping, the ACL check and the timeline filter along the way — and the results can't be pivoted back into the timeline.

Expose PPL and SQL from the Opensearch plugin, as additional search modes in the Explore view, scoped to the sketch's own indices.

- PPL and SQL join Query String and Wildcard in the existing search-mode selector, so there's one place to choose how a search is written.
- Results come back as a table, and any cell can be pivoted into the events behind it, handing the language back to Lucene.
- Queries are read-only, and every query is scoped to the sketch's indices server-side; a query naming an index outside the sketch is refused.
- The modes appear only when the cluster can serve them, in the same way wildcard search is offered only where the mappings support it.

- *Aggregations API.* Timesketch already has one, but it answers a fixed set of shapes chosen ahead of time. The cases above are ad hoc, and an analyst can't express a join or a multi-stage pipeline through it.
- *Querying OpenSearch directly outside Timesketch.* This bypasses the ACL and sketch scoping, and the results live outside the case.
- *A separate query console, distinct from the search bar.* Rejected because the language and the search mode would be two controls that can disagree.

**Additional context**

I have a working implementation and will open a draft PR against it for early feedback. Note it requires OpenSearch 3.7.0 or later, for reasons described in the PR.

stsawyer1:
closes #3887

## What existing problem does this PR solve?

Lucene (query search/wildcard) returns the list of events which contain the search query. But, it does not allow analysts to perform extraction, transformation, or aggregation as part of the investigation process. An analyst who needs that today leaves Timesketch and queries OpenSearch directly, losing sketch scoping, the ACL check, and the timeline filter. OpenSearch already ships a plugin that can answer those questions; this exposes it to a sketch.

## What new feature is being introduced?

PPL and SQL as two additional search modes with dedicated API routes, mirroring how `explore_wildcard` is separated from `explore`, so the language is fixed by the route and can never be steered by the request body:

    /sketches/<id>/explore/ppl/          /sketches/<id>/explore/sql/
    /sketches/<id>/explore/ppl/explain/  /sketches/<id>/explore/sql/explain/
    /sketches/<id>/explore/ppl/export/   /sketches/<id>/explore/sql/export/

On query execution, the sketch's own indices are injected as its source, and a query naming an index outside the sketch is refused rather than rewritten, so a rewrite bug can't silently widen a read.

In the frontend, PPL and SQL have been added to the search mode selector. Selecting these modes swaps the Lucene input for a monospace editor and returns a table instead of an event list. Queries are linted in the browser first, catching the query that succeeds while quietly meaning something else, but never blocking a run. In the results, a cell can be pivoted into the events behind it, handing the language back to Lucene and carrying the time range across. 

## Requires OpenSearch 3.7.0

Scoping injects a filter ahead of the rest of the query, and the behavior that depends on settled in 3.7.0, when Calcite became the default PPL engine. Older clusters accept the same queries but differ in how a filter ahead of a stats stage is pushed down. Rather than scope correctly on some versions and subtly not on others, the capability probe refuses below 3.7.0 and the UI drops the modes—the same way it drops wildcard on a sketch without wildcard mappings. Sketch metadata reports `supports_direct_query` alongside `supports_wildcard`, cached for five minutes, so an upgraded cluster lights the feature up without a restart.

`docker/dev/docker-compose.yml` already runs 3.7.0. `docker/release/config.env` and the e2e harness pin 2.19.5, so a stock release deployment won't see the feature until the operator upgrades.

## Overview of changes to existing functions

Three are trivial: route registration, the `supports_direct_query` metadata field, and a call in the app factory that builds the shared client at startup.

The substantive one is `timesketch/lib/datastores/opensearch.py`. The connection setup that `OpenSearchDataStore.__init__` builds inline moves into a new `build_opensearch_client()` so the direct-query package derives the cluster from the same configuration rather than parsing `OPENSEARCH_HOSTS` twice. That body is existing code moved, not rewritten. Two things are new: the builder sends nothing to the cluster, so a client can be built at startup without a cluster that is still coming up blocking the app; and it sizes the urllib3 pool to what one process can have in flight (`OPENSEARCH_POOL_MAXSIZE`, default 20), where the opensearch-py default of one keep-alive connection per node makes any two overlapping requests open and discard a throwaway. That second one helps all OpenSearch traffic.

If you'd rather not touch the datastore, the package can build its own client at the cost of duplicating the config parsing, and the pooling fix can be split out.

## Why I think this can go in the main branch

- **No new dependencies** and no new HTTP plumbing—cluster access goes through opensearch-py's existing `plugins.ppl` and `plugins.sql` namespaces, so queries fail over between nodes like every other OpenSearch call. These plugins have identical methods and parameters in the pinned 2.8.0 version of opensearch-py through current (3.2.0).
- **Self-contained**: one backend package, three Vue components, one lint utility, plus the four files above.
- **Fails safe**: if OpenSearch version is too low or without the plugin, the UI never offers the modes
- **Robust test suite**, weighted toward scoping and read-only enforcement.

**Checks**

- [x] All tests succeed.
- [x] Unit tests added.
- [x] e2e tests added. The e2e dockerfile defaults to 2.19.5, so this needs `OPENSEARCH_VERSION` raised for that job.
- [x] Documentation updated. Added a section to `docs/guides/user/search-query-guide.md`

### DIFF
--- a/data/timesketch.conf
+++ b/data/timesketch.conf
@@ -39,6 +39,10 @@ OPENSEARCH_SSL = False
 OPENSEARCH_VERIFY_CERTS = True
 OPENSEARCH_CA_CERTS = None
 OPENSEARCH_TIMEOUT = 10
+# Connections kept alive per node. Needs to cover the requests a single
+# process can have in flight at once: the web server's threads, or a
+# worker's Celery concurrency.
+OPENSEARCH_POOL_MAXSIZE = 20
 OPENSEARCH_FLUSH_INTERVAL = 5000
 OPENSEARCH_FLUSH_BYTE_SIZE = 52428800
 OPENSEARCH_INDEX_WAIT_TIMEOUT = 10
--- a/docs/guides/user/search-query-guide.md
+++ b/docs/guides/user/search-query-guide.md
@@ -265,3 +265,218 @@ Here are some common searches:
 ## Common questions
 
 There is a frequent question around Windows Event logs and how they are represented in Timesketch when imported from Plaso. For that we recommend reading up on [Common misconception about Windows EventLogs](https://osdfir.blogspot.com/2021/10/common-misconceptions-about-windows.html)
+
+## PPL and SQL
+
+Query String and Wildcard search return events.
+[PPL](https://docs.opensearch.org/latest/sql-and-ppl/ppl/index/) and
+[SQL](https://docs.opensearch.org/latest/sql-and-ppl/sql/index/) are sent to
+OpenSearch directly, so they can also aggregate: count events per host, group
+events by `data_type`, or rank field values by how often they occur. Use them
+for questions about totals and distributions, and use Query String when you
+want to read the events themselves.
+
+Both languages need OpenSearch 3.7.0 or later with the SQL plugin installed. If
+`PPL` and `SQL` are missing from the search mode menu, the cluster does not
+provide them. Timesketch re-checks the cluster periodically, so the modes appear
+on their own after an upgrade.
+
+### Selecting a search mode
+
+The button on the left of the query bar selects the language:
+
+| Mode  | Language                              |
+| ----- | ------------------------------------- |
+| `QS`  | Query String (Lucene), the default    |
+| `WC`  | [Wildcard](#wildcard-search-mode)     |
+| `PPL` | OpenSearch Piped Processing Language  |
+| `SQL` | OpenSearch SQL                        |
+
+### Indexes and query scope
+
+PPL and SQL run their searches directly against the OpenSearch indexes that hold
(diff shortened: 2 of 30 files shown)


jaegeral:
Thx for the Pull request, this is a quite a lot of code, we need to discuss this internally how to approach it and ensure maintenance from our side.

PR #3898, opened 2026-08-24
fix(cli_client): fix JSONL export double newline and Pyrefly type error

jaegeral:
When exporting sketch annotations to JSONL (export-only-with-annotations --output-format jsonl),
df.to_json(orient="records", lines=True) already appends a newline (\n) to every record, including
the last one.

Doing final_df.to_json(...) + "\n" caused two issues:

1. Malformed JSONL: It added an extra trailing newline (\n\n), creating an empty record at the end
of the file that causes strict JSONL / NDJSON parsers (e.g. jq, BigQuery) to fail.
2. Pyrefly Build Failure: Because DataFrame.to_json() is typed as Optional[str] in Pandas stubs,
binary concatenation with Literal['\n'] fails static type checking in Google3 Blaze builds.

Changes:

• Removed redundant + "\n" from sketch.py when writing JSONL output.
* Direct write to a temp file by Pandas.
• Updated sketch_test.py to assert against an explicit JSON Lines string fixture rather than a
circular to_json() + "\n" call. So this way it is clearer, what is expected and is not calling the same method, potentially causing the same double n.

According to: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_json.html
```
linesbool, default False
If ‘orient’ is ‘records’ write out line-delimited json format. 
```

Fixing:

```
timesketch_cli_client/commands/sketch.py:552:21
    |
552 |                       final_df.to_json(orient="records", lines=True, date_format="iso")
    |  _____________________^----------------------------------------------------------------
    | |                     |
    | |                     has type `None`
553 | |                     + "\n"
    | |_______________________---^
    |                         |
    |                         has type `Literal['\n']`
```

### DIFF
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -546,14 +546,10 @@ def export_only_with_annotations(
         # Write the final DataFrame to the file
         with open(filename, "w", encoding="utf-8") as fh:
             if output_format == "csv":
-                fh.write(
-                    final_df.to_csv(index=False, header=True, lineterminator="\n")
-                )
+                fh.write(final_df.to_csv(index=False, header=True, lineterminator="\n"))
             elif output_format == "jsonl":
                 fh.write(
-                    final_df.to_json(
-                        orient="records", lines=True, date_format="iso"
-                    )
+                    final_df.to_json(orient="records", lines=True, date_format="iso")
                 )
 
         end_time = time.time()
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -549,14 +549,14 @@ def export_only_with_annotations(
                 csv_data = final_df.to_csv(
                     index=False, header=True, lineterminator="\n"
                 )
-                if csv_data:
+                if csv_data is not None:
                     fh.write(csv_data)
             elif output_format == "jsonl":
                 json_data = final_df.to_json(
                     orient="records", lines=True, date_format="iso"
                 )
-                if json_data:
-                    fh.write(f"{json_data}\n")
+                if json_data is not None:
+                    fh.write(json_data)
 
         end_time = time.time()
         click.echo(
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -544,8 +544,8 @@ def export_only_with_annotations(
 
         click.echo(f"Writing {exported_count} events to file...")
 
-        # Write the final DataFrame to a temporary file, then move atomically
-        # this is to avoid accidental data deletion in an exception case.
+        # Write the final DataFrame to a temporary file, then move atomically.
+        # This is to avoid accidental data deletion in an exception case.
         temp_filename = f"{filename}.tmp"
         try:
             if output_format == "csv":
@@ -565,7 +565,8 @@ def export_only_with_annotations(
                     force_ascii=False,
                 )
             os.replace(temp_filename, filename)
-        except Exception:
+        except Exception as e:  # pylint: disable=broad-except:
+            click.echo(f"Error writing to temporary file: {e}", err=True)
             if os.path.exists(temp_filename):
                 try:
                     os.remove(temp_filename)
--- a/cli_client/python/timesketch_cli_client/commands/sketch.py
+++ b/cli_client/python/timesketch_cli_client/commands/sketch.py
@@ -13,9 +13,10 @@
 # limitations under the License.
 """Commands for sketches."""
 
-import time
-import os
 import json
+import os
+import tempfile
+import time
 from typing import Optional
 import click
 import pandas as pd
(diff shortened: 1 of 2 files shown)


cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -546,11 +546,14 @@ def export_only_with_annotations(
...
+                    final_df.to_csv(index=False, header=True, lineterminator="\n")
+                )
             elif output_format == "jsonl":
                 fh.write(
-                    final_df.to_json(orient="records", lines=True, date_format="iso")
-                    + "\n"
+                    final_df.to_json(
+                        orient="records", lines=True, date_format="iso"
+                    )
                 )
```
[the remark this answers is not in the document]
jaegeral: but we have: lines bool, default False
If ‘orient’ is ‘records’ write out line-delimited json format. Will throw ValueError if incorrect ‘orient’ since others are not list-like.

jaegeral:
But that would still not fix the py error, so we could either:
```
with open(filename, "w", encoding="utf-8") as fh:
        if output_format == "csv":
            csv_data = final_df.to_csv(index=False, header=True, lineterminator="\n")
            if csv_data is not None:
                fh.write(csv_data)
        elif output_format == "jsonl":
            json_data = final_df.to_json(orient="records", lines=True, date_format="iso")
            if json_data is not None:
                fh.write(json_data)
```

So having checks if json_data and csv_data is not None:
or we could do:

```
if output_format == "csv":
        final_df.to_csv(
            filename,
            index=False,
            header=True,
            lineterminator="\n",
            encoding="utf-8",
        )
    elif output_format == "jsonl":
        final_df.to_json(
            filename,
            orient="records",
            lines=True,
            date_format="iso",
        )
```

Meaning we would directly write to file, but that would mean to adjust the test cases for unit tests to have mock file system.

jaegeral:
Result of offline discussion, we will use pandas native function, which might require a bit more code, but is the better fix

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -543,15 +544,39 @@ def export_only_with_annotations(
 
         click.echo(f"Writing {exported_count} events to file...")
 
-        # Write the final DataFrame to the file
-        with open(filename, "w", encoding="utf-8") as fh:
+        # Write the final DataFrame to a temporary file, then move atomically.
+        # This is to avoid accidental data deletion in an exception case.
+        temp_filename = f"{filename}.tmp"
```
jkppr: Can we use `tempfile.NamedTemporaryFile` instead to not have predictable and collision prone tmp files?
(https://docs.python.org/3/library/tempfile.html)

Clean-up can then happen in a `finally` block after the `try`.

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -543,15 +544,39 @@ def export_only_with_annotations(
...
-                    + "\n"
+                final_df.to_json(
+                    temp_filename,
+                    orient="records",
+                    lines=True,
+                    date_format="iso",
+                    force_ascii=False,
                 )
+            os.replace(temp_filename, filename)
+        except Exception as e:  # pylint: disable=broad-except:
```
jkppr:
```suggestion
        except Exception as e:  # pylint: disable=broad-except
```

cli_client/python/timesketch_cli_client/commands/sketch.py
```diff
@@ -543,15 +544,39 @@ def export_only_with_annotations(
...
             elif output_format == "jsonl":
-                fh.write(
-                    final_df.to_json(orient="records", lines=True, date_format="iso")
-                    + "\n"
+                final_df.to_json(
+                    temp_filename,
+                    orient="records",
+                    lines=True,
+                    date_format="iso",
+                    force_ascii=False,
```
jkppr, line 572: Why do we need this?

cli_client/python/timesketch_cli_client/commands/sketch_test.py
```diff
@@ -189,47 +184,39 @@ def test_export_only_with_annotations_jsonl_with_limit(
...
+            self.assertIn("Applying limit of 2 events.", result.output)
+            self.assertIn("Writing 2 events to file...", result.output)
+            self.assertIn(
+                "Export finished: 2 unique annotated events written.", result.output
+            )
+
+            with open("output.jsonl", "r", encoding="utf-8") as f:
+                written_content = f.read()
+
+            expected_jsonl = (
```
jkppr, line 211: For the assertion, can we generate `expected_jsonl` dynamically from `expected_df` instead of hardcoding the raw string?

Similar how its done with `expected_df.to_csv` above.

jkppr (changes requested):

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: ACK

cli_client/python/timesketch_cli_client/commands/sketch.py
jaegeral: This is to force pandas to not attempt to escape all strings to ascii, but e.g. let them in UTF8.

cli_client/python/timesketch_cli_client/commands/sketch_test.py
jaegeral: discussed offline, we let it that way, if it causes problems moving forward, we can revisit this.

jkppr (approved):
Please fix failing linters. Otherwise LGTM

PR #3899, opened 2026-08-25
Fix non-SecGemini log analyzer reports 🤖🤖🤖

### LINKED ISSUE #3890: LLM log analyzer: UnboundLocalError on session_id makes every non-SecGemini provider report failure, even on success 🤖🤖🤖
### Summary

`llm_log_analyzer.py` binds `session_id` only inside a branch guarded by
`llm_provider.NAME == "secgemini_log_analyzer_agent"`, but uses it unconditionally a few
lines later. With any other provider the analyzer raises `UnboundLocalError`, which the
broad `except` converts into `result_status = "ERROR"`.

Because the DFIQ objects are committed **before** that point, the result is an analyzer run
that reports failure while having succeeded: the investigative questions and conclusions are
in the Investigation View, and the run says it errored.

This fires on every run with a non-SecGemini provider regardless of what the provider
returns, so it also masks other failure modes — the "did not identify any specific findings"
path is unreachable in practice.

### The code

[`timesketch/lib/analyzers/llm_log_analyzer.py`](https://github.com/google/timesketch/blob/master/timesketch/lib/analyzers/llm_log_analyzer.py#L140-L158)

```python
            # Add provider-specific details if available
            if llm_provider.NAME == "secgemini_log_analyzer_agent":
                session_id = getattr(llm_provider, "session_id", "N/A")     # only bound here
                ...

            full_response_text = result.get("full_response_text")
            if full_response_text:
                if self.output.result_status == "ERROR":
                    story_title = f"[err] Full AI Log Analysis Report - [{session_id}]"
                else:
                    story_title = f"Full AI Log Analysis Report - [{session_id}]"   # <- unbound
```

`full_response_text` is set on every return path of `LogAnalyzer.execute()`, so the guard at
that line is effectively always true.

### Observed

```
[timesketch.analyzers.dfiq.llm_log_analyzer/ERROR] An unexpected error occurred during
Log Analyzer execution: cannot access local variable 'session_id' where it is not
associated with a value

  File ".../timesketch/lib/analyzers/llm_log_analyzer.py", line 157, in run
    story_title = f"Full AI Log Analysis Report - [{session_id}]"
                                                    ^^^^^^^^^^
UnboundLocalError: cannot access local variable 'session_id' where it is not associated with a value
```

Analyzer result: `"result_status": "ERROR"`.
Database at the same moment: the `InvestigativeQuestion` and its
`InvestigativeQuestionConclusion` **exist**, created by `process_response` before the
exception.

### Reproduction

A minimal provider — no model or network needed. Save as
`timesketch/lib/llms/providers/repro_provider.py` and add
`from timesketch.lib.llms.providers import repro_provider` to
`timesketch/lib/llms/providers/__init__.py`.

```python
import json
from timesketch.lib.llms.providers import interface, manager


class ReproLogAnalyzer(interface.LLMProvider):
    NAME = "repro_log_analyzer"
    SUPPORTS_STREAMING = True

    def generate_stream_from_logs(self, log_events_generator, prompt=None, sketch=None):
        for _ in log_events_generator:
            pass
        yield json.dumps({
            "findings": [{
                "log_records": [{"record_id": self.config["record_id"]}],
                "annotations": [{
                    "investigative_question": "Repro question",
                    "conclusions": ["Repro conclusion"],
                    "priority": "critical",
                }],
            }],
            "report_summary": "repro",
        })

    def generate(self, prompt, response_schema=None):
        raise NotImplementedError()


manager.LLMManager.register_provider(ReproLogAnalyzer)
```

`timesketch.conf` — `record_id` must be the OpenSearch `_id` of an event in an active
timeline of the sketch:

```python
LLM_PROVIDER_CONFIGS = {
    "log_analyzer": {"repro_log_analyzer": {"record_id": "<an event _id>"}},
}
```

Restart the Celery worker (it reads the config at startup), then run the `llm_log_analyzer`
analyzer on the sketch.

**Result:** analyzer status `ERROR` with the traceback above, **and** one new question with
one conclusion visible in the Investigation View.

### Suggested fix

Hoist the binding so it is always defined:

```python
            session_id = getattr(llm_provider, "session_id", "N/A")

            # Add provider-specific details if available
            if llm_provider.NAME == "secgemini_log_analyzer_agent":
                table_hash = getattr(llm_provider, "table_hash", "N/A")
                ...
```

I am happy to send that as a PR if useful.

### Environment

Timesketch `master` @ `0c45e475`, `docker/dev` stack, OpenSearch 3.7.0, Python 3.14.4,
Ubuntu 26.04.

### Note

Found while building a provider for a local model. Prepared with AI assistance, hence the
🤖🤖🤖 per CONTRIBUTING.md.

dev-rehaann:
## Problem

Closes #3890.

The analyzer only bound `session_id` for the SecGemini provider, but used it whenever a full response story was created. Every non-SecGemini provider therefore raised `UnboundLocalError` after otherwise successful processing and reported an ERROR result.

## Change

Resolve `session_id` before the provider-specific block. Providers that expose a session ID keep using it; all others get the existing `N/A` fallback.

The new unit test uses a provider that deliberately has no `session_id` and verifies both a SUCCESS analyzer result and a full report story titled with `[N/A]`.

## Checks

- [x] Targeted unit test succeeds: 1 passed (2 framework setup skips)
- [x] Unit test added
- [x] Black check passes
- [x] Pylint passes
- [ ] E2E test not required for this isolated analyzer regression
- [ ] Documentation not affected

This pull request was prepared with an automated coding agent, as indicated by the required robot marker in the title.


jkppr (approved):
Thanks for your contribution 👍

PR #3903, opened 2026-08-25
Fix SecGemini LogStore signature compatibility

samomi-oss:
This PR adds the has_enrichment search parameter to TimesketchLogStore.search_logs to align with the upcoming sec-gemini SDK changes, preventing runtime TypeError crashes when the SDK is upgraded on client environments.


contrib/secgemini_byot/timesketch_logstore.py
```diff
@@ -313,7 +313,8 @@ async def search_logs(
         must_contain_all_of: Optional[Iterable[str]] = None,
         must_not_contain_any_of: Optional[Iterable[str]] = None,
         order_by: ls.Order = ls.Order.CHRONOLOGICAL,
-        exclude_log_type: Optional[Any] = None,  # pylint: disable=unused-argument
+        exclude_log_type: Optional[Any] = None,
+        has_enrichment: Optional[bool] = None,
```
jkppr, line 330: If the main goal is preventing `TypeError` whenever upstream SDK updates roll out, should we also add `**kwargs: Any` here (and pass/log it)? Otherwise, any time the SDK adds another parameter in the future, we'll be back playing whack-a-mole with runtime crashes.

contrib/secgemini_byot/timesketch_logstore.py
```diff
@@ -360,6 +365,19 @@ def _search_logs_sync(
         if log_type:
             terms.append(f'data_type:"{log_type}"')
 
+        if exclude_log_type:
```
jkppr, line 412: We should probably pass `exclude_log_type` / `et` through `_escape_opensearch_query()` just to be safe. If a log type ever contains quotes or Lucene special characters, it could throw a query parse error or mess with the query string.

jkppr (changes requested):
Please update also the docstrings to reflect on the new args in _search_logs_sync

contrib/secgemini_byot/timesketch_logstore.py
samomi-oss: done

contrib/secgemini_byot/timesketch_logstore.py
samomi-oss: done

samomi-oss:
> Please update also the docstrings to reflect on the new args in _search_logs_sync

done

jkppr (approved):

PR #3906, opened 2026-08-27
Change UPLOAD_ENABLED access to use get method

jaegeral:
Change UPLOAD_ENABLED access to use get method


jkppr (changes requested):
Please add more info what problem this is supposed to solve and why this change is needed and how it fixes it these problems.

PR #3910, opened 2026-09-15
draft: fix: ACL check

jaegeral:
Draft for checking ACL modification


jaegeral:
talked with Janosch, we will archive this PR, since we addressed the problem in #3911 .

PR #3915, opened 2026-09-22
Add configurable index prefix support

### LINKED ISSUE #3914: Configurable OpenSearch index prefix for shared cluster multi-tenancy
When running Timesketch in a shared OpenSearch / Elasticsearch cluster alongside other tools (such as Malcolm or network sensor log stores), all indices reside in the same global namespace. Timesketch currently creates its datastore indices using bare 32-character hexadecimal UUIDs (e.g. `a89933473b2a48948beee2c7e870209f`) or arbitrary user-defined names.

Because bare UUIDs do not share a common, predictable prefix, cluster administrators cannot apply granular OpenSearch Role-Based Access Control (RBAC) to restrict the Timesketch service account to its own indices (e.g. `timesketch-*`). Consequently, Timesketch must either be granted overly broad cluster-wide index permissions, or individual index permissions must be updated out-of-band as indices are created.

An optional configuration setting in `timesketch.conf`:
```python
OPENSEARCH_INDEX_PREFIX = "timesketch-"
```
defaulting to `""` to preserve existing upstream behavior.

When this prefix is configured:
1. Every OpenSearch index created and managed by Timesketch follows the format `<prefix><32 lowercase hex characters>` (e.g. `timesketch-a89933473b2a48948beee2c7e870209f`).
2. The datastore layer (`OpenSearchDataStore.create_index`) strictly validates and enforces this pattern (`^<prefix>[0-9a-f]{32}$`), rejecting unauthorized or non-canonical index creations.
3. The API and upload layer normalizes client-provided identifiers, allowing upload continuation to seamlessly match existing `SearchIndex` records whether a client returns the newly prefixed index name or a legacy bare UUID.
4. Upload chunk identifiers (`chunk_index_name`) remain bare UUIDs for file-assembly operations, decoupled from datastore index names.
5. Setting `OPENSEARCH_INDEX_PREFIX = ""` maintains 100% backwards compatibility with upstream Timesketch.

1. **Cluster-level administrative proxy / alias rules**: Difficult to maintain, prone to race conditions during index creation, and complicates direct client/worker connectivity.
3. **Granting Timesketch cluster-wide index wildcard permissions (`*`)**: Violates the principle of least privilege in multi-tenant or shared OpenSearch environments where other security datasets reside on the same cluster.

**Additional context**
Closed by https://github.com/google/timesketch/pull/3915

je4yfc:
Adds an optional `OPENSEARCH_INDEX_PREFIX` setting that gives Timesketch a distinct OpenSearch index namespace, for example:

```text
timesketch-<32 hex characters>
```

This allows deployments that share an OpenSearch cluster with other applications to restrict the Timesketch service account to an index pattern such as `timesketch-*` rather than granting access across unrelated namespaces.

The default remains:

```python
OPENSEARCH_INDEX_PREFIX = ""
```

When unset, existing index naming behavior is preserved.

Closes https://github.com/google/timesketch/issues/3914

### Index naming

Adds `timesketch/lib/index_name.py` with helpers to:

* validate 32-character hexadecimal index identifiers;
* validate configured index prefixes;
* canonicalize bare UUID-style identifiers into prefixed names;
* validate the final `<prefix><32 hex>` format.

Configured prefixes must be lowercase, begin with an alphanumeric character, and contain only `[a-z0-9_-]`. Invalid configuration is rejected during application startup and datastore initialization.

### Datastore enforcement

`OpenSearchDataStore.create_index()` now enforces the canonical namespace when a prefix is configured.

Normalization is performed before the datastore boundary; the datastore only accepts names already matching the configured namespace.

This also changes the generated `index_name` default from `uuid4().hex` in the function signature to generation at call time, ensuring separate calls receive separate identifiers.

### Index creation paths

Prefix-aware naming is applied to:

* file and data uploads;
* `SearchIndex` API creation;
* manual-event indices;
* timeline creation;
* the GCS importer.

Upload `chunk_index_name` values remain bare UUID file-assembly identifiers and are intentionally independent of OpenSearch index names.

Upload continuation also accepts either the canonical prefixed name returned by Timesketch or the corresponding bare UUID from legacy importer clients and resolves both to the same `SearchIndex`.

### Backwards compatibility

When `OPENSEARCH_INDEX_PREFIX` is empty:

* arbitrary names accepted by the existing `SearchIndex` API remain supported;
* generated datastore names remain bare UUID hex strings;
* existing upload validation behavior is preserved.

Namespace restrictions apply only when a prefix is explicitly configured.

## Testing

Adds coverage for:

* prefix syntax and type validation;
* bare, prefixed, uppercase and malformed index names;
* canonicalization and generated-name uniqueness;
* rejection of invalid and falsy non-string configuration values;
* datastore namespace enforcement;
* empty-prefix backwards compatibility;
* prefixed `SearchIndex` API creation;
* upload continuation using both prefixed and legacy bare UUID identifiers.


jaegeral:
Not going into much review here, just some thoughs:
* e2e are broken, please fix
* linters are also not ready
* How do you deal with already existing indexes from systems that were setup before that change?

I am generally not sure if we want to advocate for a system that shares the OS cluster with other systems. Also there are already APIs to give Indexes a set name, so you could have your importing flow done via API scripts and give it whatever name you want.

je4yfc:
> Not going into much review here, just some thoughs:
> 
>     * e2e are broken, please fix
> 
>     * linters are also not ready
> 
>     * How do you deal with already existing indexes from systems that were setup before that change?
> 
> 
> I am generally not sure if we want to advocate for a system that shares the OS cluster with other systems. Also there are already APIs to give Indexes a set name, so you could have your importing flow done via API scripts and give it whatever name you want.

Thanks for the feedback. The E2E and linter failures should be addressed now; both are clean locally. I'll fix it if I'm still wrong. 

On existing deployments the current behaviour is unchanged. When OPENSEARCH_INDEX_PREFIX is unset the existing indexes continue to work as they do today. Enabling a prefix is an explicit opt-in configuration change; I have not made automatic migration of existing unprefixed indexes part of this PR. If you would like that expansion or want prefixes to be default, for a future major release cycle, I can add that, but it should be a separate PR. 

I also want to clarify that this is not proposing that Timesketch should advocate shared OpenSearch clusters as the default. In environments where Timesketch and something like Malcolm are already sharing infrastructure, having a deterministic namespace is a fairly normal way to avoid collisions and make ownership explicit.  the goal here is not to change the recommended default architecture.  It is an optional capability. 

The existing APIs that allow callers to choose index names are useful, but they solve a different problem. A deployment-level prefix gives the datastore a consistent namespace. Additionally, for smaller environments, running separate OpenSearch infrastructure for every application can represent a meaningful compute and operational cost.

jaegeral:
I rather move the conversation in the bug

je4yfc:
I fixed the final linter bug on test_init_with_invalid_prefix. Had to change mock_client to _mock_client. Not sure how that slipped past local pylint, maybe I didn't read it correctly. I used the following to try and be clever and replicate the test: 

```
❯ git --no-pager diff origin/master --name-only --diff-filter=ACMR |  grep '\.py$' | xargs -r -n1 python -m pylint --rcfile=.pylintrc
(base) 
```

